PR #6, opened 2023-02-17
Add new _nodeps targets for test runs

cjeanner:
On some environment, we may not be able to install dependencies as part of the test run. Providing molecule_nodeps and pre_commit_nodeps target allows to ensure we can prepare the env in a step, and run the tests in a second step, without the need to get a working `sudo` or related commands.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,16 +13,24 @@ setup_molecule: setup_tests ## Setup molecule environment
 	bash scripts/setup_molecule
 
 .PHONY: molecule
-molecule: setup_molecule ## Run molecule tests
+molecule: setup_molecule molecule_nodeps ## Run molecule tests with dependencies install
+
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -13,16 +13,22 @@ setup_molecule: setup_tests ## Setup molecule environment
 	bash scripts/setup_molecule
 
 .PHONY: molecule
-molecule: setup_molecule ## Run molecule tests
+molecule: setup_molecule molecule_nodes ## Run molecule tests with dependencies install
+
+.PHONY: molecule_nodes
```
viroel, line 18: molecule_nodeps ?

cjeanner: good catch, thanks!

viroel:
thanks, LGTM

PR #7, opened 2023-02-20
Provides our own CI container

cjeanner:
In order to ensure we have all the needed pieces in the container for Prow, we provide our own container, with everything installed directly from the Containerfile.ci

In openshift-ci, the container may be built using one of the parameters pointing to the Containerfile directly, and we should be good.

The container did successfuly build locally, and running some "make" command was also working fine.

### DIFF
--- a/containerfiles/Containerfile.ci
+++ b/containerfiles/Containerfile.ci
@@ -1,0 +1,10 @@
+FROM registry.access.redhat.com/ubi9/ubi-init:latest
+ENV USE_VENV=no
+
+RUN dnf update -y && dnf install -y git python3-pip make gcc sudo && yum clean all
+
+COPY ../ /opt/sources
+WORKDIR /opt/sources
+RUN ls -l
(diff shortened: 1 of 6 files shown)


containerfiles/Containerfile.ci
```diff
@@ -0,0 +1,10 @@
+FROM registry.access.redhat.com/ubi9/ubi-init:latest
+ENV USE_VENV=no
+
+RUN dnf update -y && dnf install -y git python3-pip make gcc sudo && yum clean all
+
+COPY ../ /opt/sources
```
viroel, line 6: I think that should work in Prow, but we will only know after merging and testing.
In order to have this working, you will need to start from pipeline:src, which will have the code for you to COPY.

cjeanner: I've already prepared the update for patch against release - we'll see how it goes once we get this one in. A pity we can't Depends-On....

raukadah (approved):

PR #9, opened 2023-02-20
Better handling of the chmod

cjeanner:
Instead of a blind chmod -R 0775 on files and directories, let's associate the right mode to the right content type.

Also, in order to prevent pre-commit to crash due to files being edited, let's tell git to not care about file mode changes.


viroel:
lgtm

PR #10, opened 2023-02-20
Correct some tests paths

cjeanner:
Some of the calls to python scripts were still hard-coded to the venv, even if we don't use it.

This patch ensures everything is now taking the USE_VENV parameter.

### DIFF
--- a/scripts/run-local-test
+++ b/scripts/run-local-test
@@ -27,5 +27,12 @@ export ROLE_NAME="${ROLE_NAME:-$1}"
 
 # Run local test
 pushd "${PROJECT_DIR}ci_framework/roles/${ROLE_NAME}"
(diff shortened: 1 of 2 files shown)


scripts/run-local-test
```diff
@@ -27,5 +27,12 @@ export ROLE_NAME="${ROLE_NAME:-$1}"
 
 # Run local test
 pushd "${PROJECT_DIR}ci_framework/roles/${ROLE_NAME}"
-${HOME}/test-python/bin/molecule -c "${PROJECT_DIR}.config/molecule/config_podman.yaml" test
+case ${USE_ENV:-yes} in
```
viroel, line 30: USE_VENV ?

viroel (changes requested):

viroel (approved):

PR #12, opened 2023-02-21
Modify how molecule is running

cjeanner:
In order to run molecule on Prow, we need to get rid of the "podman" driver, and use "delegate" instead. In addition, we're using "localhost", since it's already running in a container, so we have to ensure the ansible_connection is set to "local" instead of "ssh".

This leads to a third configuration file for molecule.

The _skeleton_role_ molecule.yml is also updated accordingly, so that new roles will benefit of the defaults set in the config files.


Makefile
```diff
@@ -1,6 +1,16 @@
...
+BUILD_VENV_CTX ?= yes
+MOLECULE_CONFIG ?= ${MOLECULE_CONFIG:-.config/molecule/config_podman.yml}
+
+# target vars for generic operator install info 1: target name , 2: operator name
+define vars
+${1}: export ROLE_LIST=${ROLE_LIST}
+${1}: export USE_VENV=${USE_VENV}
+${1}: export BUILD_VENV_CTX=${BUILD_VENV_CTX}
+${1}: export MOLECULE_CONFIG=${MOLECULE_CONFIG}
+endef
```
raukadah, line 13: it is not called anywhere can we remove it?

raukadah (changes requested):

Makefile
cjeanner: the ROLE_LIST is consumed in target, and the USE_VENV + MOLECULE_CONFIG are used in scripts. As I understand the "export" thing here, it's supposed to expose the make "parameters" as environment variable that may be then consumed by sub-shells, such as scripts. So this is what I need.

raukadah: yup, correct, thanks!

raukadah (approved):
/lgtm

PR #13, opened 2023-02-21
Run molecule only against modified role(s)

cjeanner:
Using some trick, we may be able to know what role(s) were modified in the PR. This patch enables this automated filtering.

It also open a wide door for external reuse of the Makefile and scripts: if you have your own repository, you should be able to clone this one in some nearby path, and then run the following:
```Bash
make run_ctx_molecule \
        MOLECULE_CONFIG=.config/molecule/config_local.yml \
        ROLE_DIR=../my_repo/roles
```

Note: relative path must be used (for now).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,12 @@
 SHELL := /bin/bash
-ROLE_LIST ?= ./ci_framework/roles/*
+ROLE_DIR ?= ci_framework/roles/
 USE_VENV ?= ${USE_VENV:-yes}
 BUILD_VENV_CTX ?= yes
 MOLECULE_CONFIG ?= ${MOLECULE_CONFIG:-.config/molecule/config_podman.yml}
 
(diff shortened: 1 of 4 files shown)


raukadah:
Based on my local run
```
+++ readlink -f scripts/run_molecule
++ dirname /opt/sources/scripts/run_molecule
+ PROJECT_DIR=/opt/sources/scripts/../
+ ROLE_DIR=ci_framework/roles/
+ test -d ci_framework/roles/
+ pushd ci_framework/roles/
/opt/sources/ci_framework/roles /opt/sources
+ git diff --name-only 'HEAD^..HEAD' .
++ echo ci_framework/roles/
++ sed 's|\.\?\./||g'
+ local_roledir=ci_framework/roles/
++ git diff --name-only 'HEAD^..HEAD' .
+ popd
```
Looks good.

raukadah (approved):

PR #17, opened 2023-02-28
Usage/container

cjeanner:


cjeanner:
/hold

I just want to be 100% sure of the procedure shown in the README. It *should* work, but since we may get a direct feedback from the edm-ansible project... :)

Makefile
```diff
@@ -57,18 +65,23 @@ ci_ctx: ## Build CI container with buildah
...
+			-e BASEDIR=$(BASEDIR) \
+			cfwm:latest bash -c "make pre_commit_nodeps  BASEDIR=$(BASEDIR)" ; \
 	fi
 
 .PHONY: run_ctx_molecule
 run_ctx_molecule: ci_ctx ## Run molecule check in a container
 	if [ "x$(BUILD_VENV_CTX)" == 'xyes' ]; then \
 		podman run --rm -e MOLECULE_CONFIG=$(MOLECULE_CONFIG) cfwm:latest \
-			make molecule_nodeps MOLECULE_CONFIG=$(MOLECULE_CONFIG) ; \
+			bash -c "make molecule_nodeps MOLECULE_CONFIG=$(MOLECULE_CONFIG)" ; \
```
cescgina, line 82: Just curious, why do we need to run this target with `bash`?

cjeanner: apparently in order to pass the make parameters properly. I seems they weren't taken into account, which was pretty annoying. Maybe (probably) I did some weird things in the Makefile... ?

cescgina: Maybe it should be `${MOLECULE_CONFIG}` instead of `$(MOLECULE_CONFIG)`? I believe the parenthesis will evaluate the parameter in a subshell (https://unix.stackexchange.com/a/267661), which could be what's causing the problems? I don't know much about the particulars of this makefile,so I'm kinda shooting in the dark here

cjeanner: that difference you point seems to be more in a shell than Makefile. I'm also shooting in the dark. Thing is: it's (apparently) working better like this - we may iterate a bit in order to clean things as follow-up. There's a need to get that BASEDIR parameter/variable so that the container may be used by other projects without too much pain (there's already a usecase with edpm-ansible).
So here's a proposal:
- we get it in
- if anyone has some better way to pass down the params to the container, we make a new PR
I'm pretty sure there's a way to use env variables in the Makefile, but finding proper info isn't that easy - or my search skills are really bad lately :(. So the best way would be to find the right way to consume the `-e BASEDIR=$(BASEDIR)` directly so that we can just call `make pre_commit_nodeps` without anything more on the command line... I may find some time to iterate locally on that next week - but I have a "small" time constraint for the edpm-ansible integration and some schedule conflicts with my other tasks :(.

If you have some time to iterate and do some more testing on your side, while ensuring we get the feature (and doc) in before next Tuesday, I'm all for it :).

cescgina: If it works, definitely go ahead, when I first ask I did not notice that the bash command was part of `podman run`, so it makes more sense. By the way, how could I pull the ci-framework image locally to test?

cjeanner:
```Bash
podman run --rm -ti registry.ci.openshift.org/openstack-k8s-operators/ci-framework-image:latest bash
```

Or you can build it using `make ci_ctx` or run tests directly by running `make run_ctx_pre_commit` and/or `make run_ctx_molecule` :)

I'll amend the README once some more content exists.

Makefile
```diff
@@ -57,18 +65,23 @@ ci_ctx: ## Build CI container with buildah
 .PHONY: run_ctx_pre_commit
 run_ctx_pre_commit: ci_ctx ## Run pre-commit check in a container
 	if [ "x$(BUILD_VENV_CTX)" == 'xyes' ]; then \
-		podman run --rm cfwm:latest make pre_commit_nodeps ; \
+		podman run --rm \
+			-e BASEDIR=$(BASEDIR) \
+			cfwm:latest bash -c "make pre_commit_nodeps BASEDIR=$(BASEDIR)" ; \
```
viroel, line 70: i am confused about that image 
shouldn't be: ci-framework-image:latest ?

Makefile
```diff
@@ -57,18 +65,23 @@ ci_ctx: ## Build CI container with buildah
...
 	else \
-		podman run --rm --security-opt label=disable -v .:/opt/sources cfwm:latest make pre_commit_nodeps ; \
+		podman run --rm --security-opt label=disable \
+			-v .:/opt/sources \
+			-e BASEDIR=$(BASEDIR) \
+			cfwm:latest bash -c "make pre_commit_nodeps  BASEDIR=$(BASEDIR)" ; \
```
viroel, line 75: ditto

Makefile
cjeanner: not locally. Here we're using the locally built image from the `ci_ctx` target

cjeanner: (but maybe we may get the same names.... Would it be clearer?)

viroel: oh yeah, i see that now. I'm not sure if we want the same here too. lgtm in this way

PR #19, opened 2023-03-01
Modify how container is built

cjeanner:
In order to make Prow job faster, let's provide a simple root image with only the basic tools (git, make, etc).

In Prow, we have a step where we deploy the dependencies (calling `make setup_molecule`) against the pull request content, so there's no need to pre-deploy things at this point, especially since it's done when we build the root_image.

For the local testing though, we still need to get the appropriate tools and dependencies.
In order to still be able to consume a locally built container, a new Containerfile is created and used by the `make ci_ctx` target. The target will now build 2 images - the base one, listed as `localhost/cifwm-build:latest`, using Containerfile.ci, and `localhost/cfwm:latest` built against Containerfile.tests.

Since the name of the "loaded" image doesn't change, it's transparent to the user.


Sandeepyadav93:
/lgtm

Thanks Cedric!

PR #21, opened 2023-03-02
Ansible: import test_deps from edpm-ansible

cjeanner:
This helper will be used mostly in the molecule tests, allowing to set repositories and install some packages during the "prepare" step.


cjeanner:
/retest

An issue in the container lead to this issue - should be corrected on Prow side.

raukadah (approved):

PR #28, opened 2023-03-06
Don't call non-existing role

cjeanner:
For now, we'll work without the env_data role (must be imported from edpm-ansible project).

Roles are probably not needing this one for debugging. May reconsider later.

This PR has:


cjeanner:
recheck

PR #36, opened 2023-03-06
Add a var to not run cron  - copy_container

rlandy:
This PR adds a var to not run the CentOS 9
cron tasks in
ci_framework/roles/copy_container/tasks/main.yml
and just run the install.

By default the behaviour remain unchanged.
The var is set to true in the defaults.

This PR has:


dasm:
Default behavior stays unchanged.
/lgtm

PR #41, opened 2023-03-09
Implement ansible-test

cjeanner:
While molecule and pre-commit already provide a nice covering, ansible-test makes some more "ansible-centric" checks, and allows to leverage some more tests in a convenient way.

Note that ansible-test is expecting a proper collection, with code located at a specific place. This is why we have to copy (and not just link) the project directory content. We take advantage of the already deployed collections by just linking them in the test path, avoiding yet another collection deploy.

This PR has:

### DIFF
--- a/scripts/run_ansible_test
+++ b/scripts/run_ansible_test
@@ -1,0 +1,48 @@
+#!/usr/bin/env bash
+# Copyright Red Hat, Inc.
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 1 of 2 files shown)


scripts/run_ansible_test
```diff
@@ -0,0 +1,48 @@
...
+case ${USE_VENV} in
+    y|yes|true):
+        ansible_test="${HOME}/test-python/bin/ansible-test"
+        collection_path="${HOME}.ansible/collections/ansible_collections"
+        ;;
+esac
+
+mkdir -p ${HOME}/code/ansible_collections/cifmw/general
+cp -ra ${PROJECT_DIR}ci_framework/* ${HOME}/code/ansible_collections/cifmw/general/
+ln -s ${collection_path}/* ${HOME}/code/ansible_collections/
```
adrianfusco, line 39: In case we run local we need to manually cleanup the `ansible_collections` folder otherwise will give error. Not a problem if it will be used just in CI.

cjeanner: imho we should run in container using the new `run_ctx_ansible_test` target - that will start a container with the `--rm` flag in order to get a env as clean as possible for each run.

PR #45, opened 2023-03-10
Modify how molecule is launched

cjeanner:
Running from within a container is nice, but we then lack the molecule logs. So instead of running in the container, let's leverage the virtualenv and related so that we can properly run on the host.


cjeanner:
recheck

PR #48, opened 2023-03-14
Change Containerfile.ci image base

adrianfusco:
Change image from ubi to centos stream. At this way we can have access to some repositories and being able to install python3-libvirt and dependencies

This PR has:


cjeanner:
Only concern is about the image size change - do you happen to have a quick comparison?

PR #53, opened 2023-03-16
Added support for rhos-release in repo-setup role

raukadah:
rhos-release tool is used in downstream to lay down
RHEL base os repos. repo-setup tool will be used
to generate dlrn repos.
    
Integrating rhos-release with in repo-setup role
provides a single role to deal with all the repos
 in downstream.
    
This PR has:

### DIFF
--- a/ci_framework/roles/repo_setup/tasks/main.yml
+++ b/ci_framework/roles/repo_setup/tasks/main.yml
@@ -20,3 +20,6 @@
   ansible.builtin.include_tasks: configure.yml
 - name: Generate additional artifacts
   ansible.builtin.include_tasks: artifacts.yml
+- name: Generate downstream base os repos
+  ansible.builtin.include_tasks: rhos_release.yml
+  when: cifmw_repo_setup_enable_rhos_release | bool
--- a/ci_framework/roles/repo_setup/tasks/rhos_release.yml
+++ b/ci_framework/roles/repo_setup/tasks/rhos_release.yml
@@ -1,0 +1,21 @@
+---
+- name: Install RHOS Release tool
+  ansible.builtin.package:
+    name: "{{ cifmw_repo_setup_rhos_release_rpm }}"
+    state: present
+
+- name: Get rhos-release tool version
+  become: true
+  ansible.builtin.command: rhos-release --version
+  register: rr_version
+
+- name: Print rhos-release tool version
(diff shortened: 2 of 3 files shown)


ci_framework/roles/repo_setup/tasks/rhos_release.yml
```diff
@@ -0,0 +1,16 @@
+---
+- name: Install RHOS Release tool
+  ansible.builtin.package:
+    name: "{{ cifmw_repo_setup_rhos_release_rpm }}"
+    state: present
+
+- name: Print the rhos-release tool version
+  become: true
+  ansible.builtin.command: rhos-release --version
```
cjeanner, line 9: I'm not sure this will actually be shown, unless you have a debug task for it.. ?

ci_framework/roles/repo_setup/tasks/main.yml
```diff
@@ -20,3 +20,6 @@
   ansible.builtin.include_tasks: configure.yml
 - name: Generate additional artifacts
```
cjeanner, line 21: I think we shouldn't generate those artifacts if we get rhos-release. Maybe renaming "artifactss.yml" to "dlrn.yml" and include it only if `not cifmw_repo_setup_enable_rhos_release | bool` ?

Also, we probably don't want to install repo-setup if we're with rhos-release... ?

cjeanner (changes requested):

ci_framework/roles/repo_setup/tasks/main.yml
raukadah: So here is the workflow
1. for upstream, repo-setup tool will generate centos repos and dlrn repos in a directory.
2. for downstream, rhos-release tool will generate rhel base os repos and repo-setup tool will generate dlrn repos only for rhel (with the same command). 

So we need to install both rhos-release and repo-setup tool together.
artifactss.yml playbook is generate the dlrn md5 hash in both cases (upstream and downstream). So may be we should not rename it.

cjeanner: hmmm ok. So it may be good to refactor this a bit in order to get all the generation of content in the artifacts.yml - this can be a follow-up, we can probably go as-is.

cjeanner (approved):

PR #55, opened 2023-03-16
Implement hooks in ci-framework

cjeanner:
Hooks are needed in order to allow users to side-load content in the deployment process.

The `deploy-edpm.yml` playbook has now a first structure, allowing to understand better how hooks may be leveraged.

Of course, some pre/post hooks are kind of "the same", since they are following each others without any other tasks in-between, but getting a clean name allows to know a bit more about the actual purpose of the hook in question.

The USAGE.md file has been updated in order to show a bit more the potentials for hooks.

This PR has:

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -63,6 +63,11 @@ pre_infra:
       source: foo.yml
     - name: My glorious CRD
       type: crd
+      host: https://my.openshift.cluster
+      username: foo
+      password: bar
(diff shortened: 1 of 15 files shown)


cjeanner:
As draft - I need to figure out how to test the CRD side-loading. Probably calling the `rhol_crc` role in order to get a small CRC.. or microshift once it's ready...

PR #60, opened 2023-03-17
Adds initial implementation of operator_build role

viroel:
This patch adds initial implementation of a role to build OpenStack operators.
The role can build multiple operators, when provided their 'name' and 'src' code directory.
When testing from a pull request, in order to properly build the meta-operator, the user needs to provide the PR owner and SHA, to be used on meta-operator go.mod.

This PR has:


ci_framework/roles/operator_build/tasks/build.yml
```diff
@@ -0,0 +1,119 @@
...
+    name: docker.io/library/golang:1.19
+
+- name: Set operator image names and tags
+  ansible.builtin.set_fact:
+    operator_img_tag: "{{ pr_sha }}"
+    operator_registry_prefix: "{{ cifmw_operator_build_push_registry }}/{{ cifmw_operator_build_push_org }}/{{ operator.name }}"
+    cacheable: true
+
+- name: Call manifests
+  community.general.make:
```
cjeanner, line 68: care to call `ci_make` instead? That one will create reproducer scripts in `{{ cifmw_basedir }}/artifacts`. Might be good for debugging and so on

ci_framework/roles/operator_build/vars/main.yml
cjeanner, line 1: if not used, we can remove this file/directory

cjeanner (commented):
Lovely!
Note that, if you switch the `community.general.make` to `ci_make` calls, you'll be able to leverage its `dry_run: true` parameter, which will generate a file with the passed params, and NOT run the make. This means: molecule compatiblity++ :)

ci_framework/roles/operator_build/tasks/build.yml
viroel: sure, I will take a look.

viroel: done

ci_framework/roles/operator_build/README.md
```diff
@@ -0,0 +1,58 @@
...
+        name: operator_build
+      vars:
+        cifmw_operator_build_push_registry: "quay.io"
+        cifmw_operator_build_push_org: "myorg"
+        cifmw_operator_build_meta_build: true
+        cifmw_operator_build_meta_src: "/home/user/openstack-operator"
+        cifmw_operator_build_operators:
+          - name: "mariadb-operator"
+            src: "/home/user/mariadb-operator"
+            pr_owner: "ci-framework/openstack-operator"
```
viroel, line 37: In Zuul, the idea is that we can go through zuul inventory, find all `openstack-k8s-operators` repos, and populate this list of operators. We can get all those info from zuul. (We can get PR number, and with that get PR Owner using GH API)[1]

[1] https://github.com/rdo-infra/review.rdoproject.org-config/blob/142a98841377483222c7e331754f3436cd6ac3b6/playbooks/crc/build_operators.yaml#L8-L34

ci_framework/roles/operator_build/tasks/build.yml
```diff
@@ -0,0 +1,147 @@
...
+    path: /tmp/artifacts
+    state: directory
+
+- name: Call manifests
+  ci_make:
+    chdir: "{{ operator.src }}"
+    output_dir: /tmp/artifacts
+    target: manifests
+  tags:
+    - build
```
viroel, line 78: we may not need build that if we add ci_make dry-run functionality.
we can continue with `push` tag, since we may not need to always push to some registry, so we would skip it.

ci_framework/roles/operator_build/tasks/build.yml
```diff
@@ -0,0 +1,147 @@
...
+    chdir: "{{ cifmw_operator_build_meta_src }}"
+  when:
+    - cifmw_operator_build_meta_build
+    - operator.name != cifmw_operator_build_meta_name
+    - pr_sha is defined
+    - operator.pr_owner is not defined
+
+- name: Get golang container image
+  containers.podman.podman_image:
+    name: docker.io/library/golang:1.19
```
viroel, line 59: Not sure if this is really needed. 
We can't have golang hardcoded here since oko-devs are always updating it.

ci_framework/roles/operator_build/vars/main.yml
viroel, line 1: to be deleted

ci_framework/roles/operator_build/molecule/default/molecule.yml
```diff
@@ -0,0 +1,13 @@
...
+# Mainly used to override the defaults set in .config/molecule/
+# By default, it uses the "config_podman.yml" - in CI, it will use
+# "config_local.yml".
+log: true
+
+provisioner:
+  name: ansible
+  log: true
+  ansible_args:
+    - --skip-tags=build,push
```
viroel, line 11: skipping tags while I don't figure it out how to properly test this.
We can replace with ci_make dry_run too.

ci_framework/roles/operator_build/tasks/build.yml
```diff
@@ -0,0 +1,147 @@
...
+    go mod edit -replace {{ operator_api_path }}=github.com/{{ operator.pr_owner }}/api@{{ operator.pr_sha }}
+    go mod tidy
+  args:
+    chdir: "{{ cifmw_operator_build_meta_src }}"
+  when:
+    - cifmw_operator_build_meta_build
+    - operator.name != cifmw_operator_build_meta_name
+    - operator.pr_owner is defined
+    - operator.pr_sha is defined
+
```
raukadah, line 32: Based on https://github.com/openshift/release/blob/master/ci-operator/step-registry/openstack-k8s-operators/build/openstack-k8s-operators-build-commands.sh
```
go mod edit -replace github.com/${ORG}/${BASE_OP}/api=github.com/${REPO_NAME}/api@${API_SHA}
  go mod tidy
  pushd ./apis/
  go mod edit -replace github.com/${ORG}/${BASE_OP}/api=github.com/${REPO_NAME}/api@${API_SHA}
  go mod tidy
  popd
```
We already have added tasks for {{ operator.name }}/api. We also need add the same task for  {{ operator.name }}/apis/go.mod file.

Some of the operators does not have *apis* directory, we need to make sure first it exists and then apply the above task.

Rest looks good.

ci_framework/roles/operator_build/tasks/build.yml
raukadah: We can parameterize it. In RDO side, if we just specify golang:1.19 in dockerfile, it tries to go over multiple registries and fail. We can specifically pull this image.

raukadah (changes requested):
Thank you @viroel for picking it up. overall looks good, left few comments

cjeanner:
Waiting on @viroel and @raukadah for some review. It's all passing CI, but since this is an unknown area for me... :)

ci_framework/roles/operator_build/tasks/build.yml
cjeanner: ah - I removed the "push" tag - will re-add it, but maybe as a parameter ?

ci_framework/roles/operator_build/tasks/build.yml
cjeanner: Done, it's now a parameter one may override.

ci_framework/roles/operator_build/tasks/build.yml
cjeanner: is that one callable via some `make` target?

ci_framework/roles/operator_build/tasks/build.yml
cjeanner: added a parameter instead, updated molecule.

PR #70, opened 2023-03-22
Removing unnecessary conversion to list

jpodivin:
Glob already returns a list, either empty or with items. There is no need to cast the output to list again.

This PR has:


cjeanner:
/lgtm
(ansible-test will test that patch)

PR #71, opened 2023-03-22
Ensure correct path for action plugins in molecule

cjeanner:
Molecule was acting weirdly because it didn't find its action plugins. This seems to make it defaults to the `ci_make`, although it was NOT in any location molecule should have been able to know about...


cjeanner:
Useless as far as I can say.

PR #72, opened 2023-03-23
Correct a typo in the parameter name

cjeanner:


raukadah:
thank you for fixing it, we also need to update the readme: https://github.com/openstack-k8s-operators/ci-framework/blame/main/ci_framework/roles/rhol_crc/README.md#L15

cjeanner:
@raukadah oh, right, thanks!

PR #76, opened 2023-03-23
Parametrizing verbosity arguments

jpodivin:
Users of the script will now be able to turn the verbosity up or down without having to hack the args.

This PR has:

### DIFF
--- a/scripts/run_molecule
+++ b/scripts/run_molecule
@@ -26,6 +26,7 @@ ROLE_DIR=$1
 USE_VENV=${USE_VENV:-yes}
 MOLECULE_CONFIG=${MOLECULE_CONFIG:-.config/molecule/config_podman.yml}
 TEST_ALL_ROLES=${TEST_ALL_ROLES:-no}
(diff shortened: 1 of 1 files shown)


jpodivin:
For the record I've been thinking about some sort of "smarter" parsing or sanitation. But it doesn't seem warranted in this case. The script would probably just bloat.

scripts/run_molecule
```diff
@@ -26,6 +26,7 @@ ROLE_DIR=$1
 USE_VENV=${USE_VENV:-yes}
 MOLECULE_CONFIG=${MOLECULE_CONFIG:-.config/molecule/config_podman.yml}
 TEST_ALL_ROLES=${TEST_ALL_ROLES:-no}
+TEST_VERBOSITY=${TEST_VERBOSITY:--debug}
```
cjeanner, line 29:
```suggestion
TEST_VERBOSITY=${TEST_VERBOSITY:-'--debug'}
```

jpodivin: double dash, double dash

PR #81, opened 2023-03-27
Better splitting of the mighty edpm-deploy.yml playbook

cjeanner:
Splitting it like that would allow to get a cleaner overview of the file
content, allowing an easier maintenance later.

This patch also introduces some more top-level parameters, allowing to
activate or not RHOL/CRC and Libvirt supports.

### DIFF
--- a/.zuul.d/collect-logs.yml
+++ b/.zuul.d/collect-logs.yml
@@ -15,6 +15,21 @@
             chdir: "{{ ansible_user_dir }}/zuul-output/logs"
             cmd: cp /tmp/report.html .
 
+    - name: Check if we get ci-framework basedir
+      register: cifmw_state
+      ansible.builtin.stat:
+        path: "{{ ansible_user_dir }}/ci-framework"
+
+    - name: Copy ci-framework interesting files
+      when:
+        - cifmw_state.stat.exists | bool
+      ansible.builtin.shell:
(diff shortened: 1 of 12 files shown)


cjeanner:
recheck

Weird.... "merge conflict" while branch is 100% up-to-date compared to main

cjeanner:
/hold

Need to check something first - I think I can make it even better.

.zuul.d/collect-logs.yml
```diff
@@ -15,13 +15,29 @@
...
+      ansible.builtin.stat:
+        path: "{{ ansible_user_dir }}/ci-framework"
+
+    - name: Copy ci-framework interesting files
+      when:
+        - cifmw_state.stat.exists | bool
+      ansible.builtin.shell:
+        chdir: "{{ ansible_user_dir }}/zuul-output/logs/"
+        cmd: |
+          mkdir ci-framework; pushd ci-framework
```
raukadah, line 29: Can we convert directory creation to ansible and then call shell module to copy log files? We can use block for that if we use ansible.

raukadah (approved):
Looks good, few nit picks to address.

.zuul.d/collect-logs.yml
cjeanner: right - #88 will fix this.

PR #82, opened 2023-03-27
New role: operator_deploy

cjeanner:
It will leverage the install_yamls Makefile targets to deploy the wanted
operators only.

It allows to pass down the needed parameters so that it will consume the
wanted image.

This PR has:

### DIFF
--- a/ci_framework/roles/operator_deploy/README.md
+++ b/ci_framework/roles/operator_deploy/README.md
@@ -1,0 +1,29 @@
+## Role: operator_deploy
+Deploy only selected operator(s) on OpenShift
+
+### Privilege escalation
+None
+
+### Parameters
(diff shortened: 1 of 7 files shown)


ci_framework/roles/operator_deploy/vars/main.yml
```diff
@@ -0,0 +1,22 @@
...
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+
+# While options found within the vars/ path can be overridden using extra
+# vars, items within this path are considered part of the role and not
+# intended to be modified.
+
+# All variables within this role should have a prefix of "cifmw_operator_deploy"
```
adrianfusco, line 22: This file is empty. Maybe we can delete.

PR #89, opened 2023-03-30
Run molecule tests and store results for final report

jpodivin:
I don't really like the regex conditional. I'll try to think of something else.

This PR has:

### DIFF
--- a/scripts/run_molecule
+++ b/scripts/run_molecule
@@ -27,6 +27,7 @@ USE_VENV=${USE_VENV:-yes}
 MOLECULE_CONFIG=${MOLECULE_CONFIG:-.config/molecule/config_podman.yml}
 TEST_ALL_ROLES=${TEST_ALL_ROLES:-no}
 TEST_VERBOSITY=${TEST_VERBOSITY:-'--debug'}
+MOLECULE_FAILURES=""
 # Run local test
 (test -d ${ROLE_DIR} || (echo 'No such directory: ${ROLE_DIR}' && exit 1) ) || exit 1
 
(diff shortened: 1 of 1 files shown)


rebtoor (changes requested):
Tested locally and worked fine for me, so I'll request a minor change at the EOF.

scripts/run_molecule
```diff
@@ -45,14 +46,26 @@ for rolepath in ${ROLES_LIST}; do
...
 done
 popd
+
+if [ ${#MOLECULE_FAILURES} -gt 0 ]; then
+    echo "Molecule tests of following roles have failed"
+    echo "==============================="
+    echo -e $MOLECULE_FAILURES
+    echo "==============================="
+    exit 1
+fi
```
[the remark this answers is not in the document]
jpodivin: Done.

scripts/run_molecule
```diff
@@ -45,14 +46,26 @@ for rolepath in ${ROLES_LIST}; do
...
+        case ${USE_VENV} in
+                y|yes|true)
+                    ${HOME}/test-python/bin/molecule -c "${PROJECT_DIR}${MOLECULE_CONFIG}" ${TEST_VERBOSITY} test --all
+                    ;;
+                *)
+                    molecule -c "${PROJECT_DIR}${MOLECULE_CONFIG}" ${TEST_VERBOSITY} test --all
+                    ;;
+        esac ) || true
+        if [[ $MOLECULE_RESULTS =~ failed=[1-9] ]]; then
+            MOLECULE_FAILURES=$MOLECULE_FAILURES"${role}\n"
```
cjeanner, line 59:
```suggestion
            MOLECULE_FAILURES="${MOLECULE_FAILURES}${role}\n"
```
Didn't test the specific `"` location change, but it seems more "bash". At least, the `{}` should be used for sure.

scripts/run_molecule
```diff
@@ -45,14 +46,26 @@ for rolepath in ${ROLES_LIST}; do
...
+    MOLECULE_RESULTS=$(\
+        case ${USE_VENV} in
+                y|yes|true)
+                    ${HOME}/test-python/bin/molecule -c "${PROJECT_DIR}${MOLECULE_CONFIG}" ${TEST_VERBOSITY} test --all
+                    ;;
+                *)
+                    molecule -c "${PROJECT_DIR}${MOLECULE_CONFIG}" ${TEST_VERBOSITY} test --all
+                    ;;
+        esac ) || true
+        if [[ $MOLECULE_RESULTS =~ failed=[1-9] ]]; then
```
cjeanner, line 58: what if we get 10 failures?

cjeanner (changes requested):
sounds OK on the idea, but there are some tiny things.

scripts/run_molecule
jpodivin: I'ts regex. As long as the first number is anything else then zero it will match. 

https://wiki.bash-hackers.org/syntax/ccmd/conditional_expression

cjeanner: ah, right, no full matching. Need more coffee.

PR #93, opened 2023-03-31
Ensure we can re-run the script multiple times

cjeanner:
Until now, we had to trash and recreate the container if we wanted to
iterate on ansible-test runs.

This patch corrects this issue, and we can now run the following without
any issue:
```
podman run --rm -ti \
        --security-opt label=disable \
        -v .:/opt/sources --user root cifmw:latest bash
./scripts/run_ansible_test
./scripts/run_ansible_test
```
It makes the whole process faster and easier for developers.

This PR has:


cjeanner:
Bypassing the CI workflow.

Reason: it seems the whole signature system on centos 9 stream is broken due to the use of sha1 algorithm in gpg related content, while that very same sha1 has been marked as weak:
https://kojihub.stream.centos.org/koji/buildinfo?buildID=31391

PR #96, opened 2023-04-03
Molecule role test failure trigger changed to RC

jpodivin:
Proposed change addresses an issue when failure of the molecule wouldn't be reported if it was not caused by broken test assertion. For example when the tested role contained syntax error.

This PR has:


rebtoor (approved):

jpodivin:
The original patch was reverted.

PR #99, opened 2023-04-05
Workaround gnupg2 issue

cjeanner:
Latest gnupg2 (gnupg2-2.3.3-3) markes SHA-1 as weak, breaking
effectively the whole signature system on CentOS.

This patch ensures we'll be at least able to build the container in CI,
and should be reverted as soon as the issue is corrected on CentOS.

More information:
https://bugzilla.redhat.com/show_bug.cgi?id=2184640


cjeanner:
Forcing merge without waiting on CI.

Reason: the issue this workaround fixes would have prevented prow jobs to work anyway. This workaround should allow to pass other jobs, without actual impact on the code quality.

As stated, it must be reverted once the situation related to the signature mechanism is cleared.

PR #103, opened 2023-04-12
Inject new role path in ansible

cjeanner:
The use_install_yamls role and related modules primary task is to
generate task files embedding the `ci_make` calls matching install_yamls
Makefiles targets.

In order to be able to consume those tasks, we have to ensure Ansible
knows about them. We generate the whole content in a dedicated location
nested in the `cifmw_basedir/artifacts/roles` in order to keep it clean
while allowing to get this generated content in the logs.

We also add some new documentation about the implication of this current
change, especially if the operator wants to output the content somewhere
else in the tree.


raukadah (approved):
/lgtm

PR #115, opened 2023-04-14
Ensure we have ansible log in known location

cjeanner:
This location is then used in order to gather the log file in
cifmw_basedir/logs/ for future reference.


adrianfusco (commented):
/lgtm

PR #116, opened 2023-04-14
Remove "`" from task name

cjeanner:
Apparently the regexp used in ci_make action doesn't strip those chars.
Let's just remove them from the generated tasks, they don't really bring
much value anyway.

From now on, the filename generated will be as simple as:
ci_make_INDEX_run_TARGET.sh

The associated logfile will also match that pattern.


raukadah (approved):
So Earlier it was 
```
ci_make_0_run_`make_edpm_compute`.log
```
and now
```
ci_make_0_run_edpm_compute.log
```
The playbook run looks cleaner now. 
previous
```
install_yamls_makes : Run `make edpm_compute_repos` 
```
and now
```
install_yamls_makes : Run edpm_compute_repos
```

PR #123, opened 2023-04-18
Cleaning up some unused files, and consolidate PATH

cjeanner:
In order to access the CRC provided `oc` and related tools, we have to
ensure ansible tasks know about the specific location.

We now expose a new cacheable fact that may be overridden by the
operator. We'll then be able to set the task or playbook environment
PATH value to that new fact in order to gain access to otherwise unknown
commands.

This patch also takes the opportunity to clean some unused files in the
`operator_build` role, and converts a `shell` call to `command`.

This PR has:


cjeanner:
recheck

viroel:
/lgtm

thanks for that, I was missing it yesterday :)

PR #126, opened 2023-04-18
New role: edpm_prepare

cjeanner:
This role wraps all of the preparation steps/make targets in order to
get the dataplane ready for the actual EDPM deployment.

The goal here is to make a modular role, calling the install_yamls_makes
tasks in a way we may toggle various switches, or even inject operators
early in the game.

This PR has:

### DIFF
--- a/ci_framework/playbooks/06-deploy-edpm.yml
+++ b/ci_framework/playbooks/06-deploy-edpm.yml
@@ -12,6 +12,10 @@
         name: libvirt_manager
         tasks_from: deploy_edpm_compute.yml
 
+    - name: Call predeploy
+      ansible.builtin.include_role:
+        name: edpm_prepare
+
 - name: Run post_deploy hooks
   vars:
     hooks: "{{ post_deploy | default([]) }}"
--- a/ci_framework/roles/edpm_prepare/README.md
+++ b/ci_framework/roles/edpm_prepare/README.md
@@ -1,0 +1,12 @@
+## Role: edpm_prepare
+Prepares the environment to deploy OpenStack control plane and compute nodes.
+
+### Privilege escalation
+This role doesn't need privilege scalation.
+
+### Parameters
+* `cifmw_edpm_prepare_basedir`: (String) Base directory. Defaults to `cifmw_basedir` which defaults to **~/ci-framework**.
+* `cifmw_edpm_prepare_dry_run`: (Boolean) Skips resources installations and waits. Defaults to false.
+* `cifmw_edpm_prepare_manifests_dir`: String) Directory in where install_yamls output manifests will be placed. Defaults to **"{{ cifmw_edpm_prepare_basedir }}/artifacts/manifests"**
+* `cifmw_edpm_skip_openstack_operator:`: (Boolean) Intentionally skips the deployment of the OpenStack metaoperator. Defaults to false.
+* `cifmw_edpm_prepare_wait_subscription_retries`: (Integer) Number of retries, with 5 seconds delays, waiting for the OpenStack subscription to come up. Defaults to 5.
--- a/ci_framework/roles/edpm_prepare/molecule/default/prepare.yml
+++ b/ci_framework/roles/edpm_prepare/molecule/default/prepare.yml
@@ -1,0 +1,30 @@
+---
+# Copyright Red Hat, Inc.
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+
+- name: Prepare
+  hosts: all
+  vars:
+    ansible_user_dir: "{{ lookup('env', 'HOME') }}"
+    cifmw_basedir: "{{ ansible_user_dir }}/ci-framework"
+    cifmw_install_yamls_tasks_out: "{{ ansible_user_dir }}/zuul-jobs/roles/install_yamls_makes/tasks"
+    cifmw_installyamls_repos: "{{ ansible_user_dir }}/src/github.com/openstack-k8s-operators/install_yamls"
+    cifmw_install_yamls_defaults:
+      NAMESPACE: openstack
+  roles:
+    - role: test_deps
+    - role: ci_setup
+    - role: install_yamls
--- a/ci_framework/roles/edpm_prepare/tasks/main.yml
+++ b/ci_framework/roles/edpm_prepare/tasks/main.yml
@@ -1,0 +1,102 @@
+- name: Prepare storage in CRC
+  vars:
+    make_crc_storage_dryrun: "{{ cifmw_edpm_prepare_dry_run }}"
+  when:
+    - cifmw_use_crc is defined
+    - cifmw_use_crc | bool
+  ansible.builtin.include_role:
+    name: 'install_yamls_makes'
+    tasks_from: 'make_crc_storage'
+
+- name: Set install_yamls Makefile environment variables
+  vars:
+    operators_build_output: "{{ (cifmw_operator_build_output | default({'operators':{}})).operators }}"
+  ansible.builtin.set_fact:
+    cifmw_edpm_prepare_common_env:
+      OUT: "{{ cifmw_edpm_prepare_manifests_dir }}"
+    cifmw_edpm_prepare_make_openstack_env: |
+      {% if cifmw_operator_build_meta_name in operators_build_output %}
+      OPENSTACK_IMG: {{ operators_build_output[cifmw_operator_build_meta_name].image }}
+      {% endif %}
+    cifmw_edpm_prepare_make_openstack_deploy_env: |
+      {% if cifmw_operator_build_meta_name in operators_build_output %}
+      OPENSTACK_BRANCH: ""
+      GIT_CLONE_OPTS: "-l"
+      OPENSTACK_REPO: operators_build_output[cifmw_operator_build_meta_name].git_src_dir
+      {% endif %}
+
+- name: Prepare inputs
+  vars:
+    make_input_env: "{{ cifmw_edpm_prepare_common_env }}"
+    make_input_dryrun: "{{ cifmw_edpm_prepare_dry_run }}"
+  ansible.builtin.include_role:
+    name: 'install_yamls_makes'
+    tasks_from: 'make_input'
+
+# TODO: Prepare a proper operator list that dictates what operators to deploy instead of relaying on a simple flag
+- name: OpenStack meta-operator installation
+  when: not cifmw_edpm_skip_openstack_operator
+  block:
+  - name: Install OpenStack operator
+    vars:
+      make_openstack_env: "{{ cifmw_edpm_prepare_common_env |
+        combine(cifmw_edpm_prepare_make_openstack_env | from_yaml)}}"
+      make_openstack_dryrun: "{{ cifmw_edpm_prepare_dry_run }}"
+    ansible.builtin.include_role:
+      name: 'install_yamls_makes'
+      tasks_from: 'make_openstack'
+
+- name: Wait for OpenStack control plane to be installed
+  block:
+    - name: Wait for OpenStack subscription creation
+      ansible.builtin.command:
+        cmd: oc get sub openstack-operator --namespace={{ cifmw_install_yamls_defaults['NAMESPACE'] }} -o=jsonpath='{.status.installplan.name}'
+      until: cifmw_edpm_prepare_wait_installplan_out.rc == 0 and cifmw_edpm_prepare_wait_installplan_out.stdout != ""
+      register: cifmw_edpm_prepare_wait_installplan_out
+      retries: "{{ cifmw_edpm_prepare_wait_subscription_retries }}"
+      delay: 5
+
+    - name: Wait for OpenStack operator to get installed
+      ansible.builtin.command:
+        cmd: >
+          oc wait InstallPlan {{ cifmw_edpm_prepare_wait_installplan_out.stdout }}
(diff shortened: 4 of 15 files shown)


ci_framework/roles/edpm_prepare/tasks/main.yml
```diff
@@ -0,0 +1,100 @@
...
+    tasks_from: 'make_openstack'
+
+- name: Wait for OpenStack subscription creation
+  ansible.builtin.shell: |
+    oc get sub openstack-operator --namespace={{ cifmw_install_yamls_defaults['NAMESPACE'] }} > /dev/null 2>&1 &&
+      oc get sub openstack-operator --namespace={{ cifmw_install_yamls_defaults['NAMESPACE'] }} -o=jsonpath='{.status.installplan.name}'
+  until: cifmw_edpm_prepare_wait_installplan_out.rc == 0 and cifmw_edpm_prepare_wait_installplan_out.stdout != ""
+  register: cifmw_edpm_prepare_wait_installplan_out
+  retries: 30
+  delay: 5
```
raukadah, line 57: or can we move these tasks to a seperate playbook and invoke based on some condition like install_openstack_operator | default ('true') | bool? This way I can define one var in ci and job is done.

raukadah (commented):
Great start, added some comments to handle the need of CI.

ci_framework/roles/edpm_prepare/tasks/main.yml
pablintino: So, you need to choose what operators to deploy? Or is the openstack metaoperator the only one to skip? I can create a variable with a default for that

raukadah: We want to skip the openstack metaoperator installation.

pablintino: I'll add the cifmw_edpm_skip_openstack_operator with a default value of false. Thanks for the point!

pablintino: It can be, but, that's not that easy, as we may need to tune the OpenStackControlPlane CR, is doable and it should be somehting I'd like to see, but that will take some more time. Skipping the metaoperator is an easy task.

ci_framework/roles/edpm_prepare/tasks/main.yml
```diff
@@ -0,0 +1,98 @@
...
+- name: Find the OpenStack CR manifest
+  ansible.builtin.find:
+    paths: "{{ cifmw_edpm_prepare_intall_yamls_output_dir }}/{{ cifmw_install_yamls_defaults['NAMESPACE'] }}/openstack/cr"
+    contains: "kind: OpenStackControlPlane"
+    patterns: "*.yaml"
+  register: cifmw_edpm_prepare_openstack_cr_manifest_paths
+
+- name: Ensure manifest exists
+  ansible.builtin.assert:
+    that: cifmw_edpm_prepare_openstack_cr_manifest_paths.matched == 1
```
cjeanner, line 80: I think we may have to check "> 0" instead here - we may end with custom files in that location, and they may match the same kind. In case we're updating it later... ?

ci_framework/roles/edpm_prepare/tasks/main.yml
```diff
@@ -0,0 +1,98 @@
...
+    quiet: true
+    msg: "Cannot determine OpenStackControlPlane deployment manifest"
+
+- name: Find files out
+  ansible.builtin.set_fact:
+    cifmw_edpm_prepare_ctrlplane_cr_name: |-
+      {{  (
+            lookup('ansible.builtin.template', cifmw_edpm_prepare_openstack_cr_manifest_paths.files[0].path) | from_yaml
```
cjeanner, line 88: warning: lookup is running on the controller (i.e. your laptop). It will therefore fail if we're running against a remote node, such a hypervisor.
You may want to use "ansible.builtin.slurp" instead, then `|b64decode | json_query()` its "content".

ci_framework/playbooks/06-deploy-edpm.yml
```diff
@@ -6,10 +6,15 @@
 - hosts: "{{ cifmw_target_host | default('localhost') }}"
   gather_facts: false
   tasks:
-    - name: Create and provision external computes
```
cjeanner, line 9: Please re-align blocks

ci_framework/roles/edpm_prepare/molecule/default/prepare.yml
cjeanner, line 1: According to molecule, you may be missing this:
msg: '''cifmw_install_yamls_defaults'' is undefined. ''cifmw_install_yamls_defaults'' is undefined'

Please check existing molecule jobs for better comparison.

ci_framework/roles/edpm_prepare/tasks/main.yml
```diff
@@ -0,0 +1,98 @@
...
+  - name: Install OpenStack operator
+    vars:
+      make_openstack_env: "{{ cifmw_edpm_prepare_common_env | from_yaml |
+        combine(cifmw_edpm_prepare_make_openstack_env | from_yaml)}}"
+    ansible.builtin.include_role:
+      name: 'install_yamls_makes'
+      tasks_from: 'make_openstack'
+
+  - name: Wait for OpenStack subscription creation
+    ansible.builtin.shell: |
```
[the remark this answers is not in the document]
pablintino: Well, the namespace it's getting created before, in the make input. Here we are waiting to the OLM sub to be created, that sub will reference an InstallPlan, that when finished, will tell us that the metaoperator is installed and ready. But, I'm thinking that the first command has no sense anymore. I played a lot with this till I get it working and that first command seems like a leftover of my test. I'll try to delete it and keep all in one single command.

ci_framework/roles/edpm_prepare/tasks/main.yml
```diff
@@ -0,0 +1,98 @@
...
+      oc get sub openstack-operator --namespace={{ cifmw_install_yamls_defaults['NAMESPACE'] }} > /dev/null 2>&1 &&
+        oc get sub openstack-operator --namespace={{ cifmw_install_yamls_defaults['NAMESPACE'] }} -o=jsonpath='{.status.installplan.name}'
+    until: cifmw_edpm_prepare_wait_installplan_out.rc == 0 and cifmw_edpm_prepare_wait_installplan_out.stdout != ""
+    register: cifmw_edpm_prepare_wait_installplan_out
+    retries: 30
+    delay: 5
+
+  - name: Wait for OpenStack operator to get installed
```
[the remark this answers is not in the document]
pablintino: Sure, I'll add a varible for this and the final deploy timeout for sure.

ci_framework/roles/edpm_prepare/tasks/main.yml
pablintino: Honestly, I don't like the whole solution, is something that works for a first iteration, but I'll probably create a custom Python module that given a path is able to apply the kustomization and ouput the resulting manifests, with maybe the ability to retrieve only a field given a yaml path. I put 1 as value for the assert cause I expect to have a single control plane CR applied, if there a more than one that's something that I'm not expecting, and in fact, I'll wait only for one control plane.

ci_framework/roles/edpm_prepare/tasks/main.yml
```diff
@@ -0,0 +1,98 @@
...
+      OUT: {{ cifmw_edpm_prepare_intall_yamls_output_dir }}
+    cifmw_edpm_prepare_make_openstack_env: |
+      {% if cifmw_operator_build_meta_name in cifmw_operator_build_output.operators %}
+      OPENSTACK_IMG: {{ cifmw_operator_build_output.operators[cifmw_operator_build_meta_name].image }}
+      {% endif %}
+    cifmw_edpm_prepare_make_openstack_deploy_env: |
+      {% if cifmw_operator_build_meta_name in cifmw_operator_build_output.operators %}
+      OPENSTACK_BRANCH: ""
+      GIT_CLONE_OPTS: "-l"
+      OPENSTACK_REPO: cifmw_operator_build_output.operators[cifmw_operator_build_meta_name].git_src_dir
```
raukadah, line 26: Based on my testing with [EDPM Zuul job pr ](https://github.com/openstack-k8s-operators/ci-framework/pull/154). I am hitting following[ error ](https://logserver.rdoproject.org/54/154/a95182f33941efb8c829a18b600d9fcec54a6b4d/github-check/crc-podified-edpm-deployment/e21972c/job-output.txt)
```
atal: [localhost]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'cifmw_operator_build_output' is undefined. 'cifmw_operator_build_output' is undefined\n\nThe error appears to be in '/home/zuul-worker/src/github.com/openstack-k8s-operators/ci-framework/ci_framework/roles/edpm_prepare/tasks/main.yml': line 14, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Set install_yamls Makefile environment variables\n  ^ here\n"}
```
cifmw_operator_build_output comes from operator_build role. But in the above job, we are not building operators that' why this var is not set during runtime.

raukadah: @pablintino  When we use make_install_yamls role, for each task we have specific vars to manipulate their environment and params for a make target.
For example in case of [make openstack_deploy
](https://logserver.rdoproject.org/58/158/e939b97efe6f02da47c3d1141d537d3e93b9a306/github-check/cifmw-end-to-end/5b0ed1e/ci-framework/artifacts/roles/install_yamls_makes/tasks/make_openstack_deploy.yml)
we have make_openstack_deploy_environment and make_openstack_deploy_params vars to do the same.

If an user defines these vars in their var files then these vars will be ignored if they use the edpm_prepare role.
Can we check if these vars for example make_openstack_deploy_params is defined, skip these kinds of tasks?

Do let me know your thoughts on this.

ci_framework/playbooks/06-deploy-edpm.yml
```diff
@@ -6,10 +6,15 @@
...
-      ansible.builtin.include_role:
-        name: libvirt_manager
-        tasks_from: deploy_edpm_compute.yml
+  - name: Create and provision external computes
+    ansible.builtin.include_role:
+      name: libvirt_manager
+      tasks_from: deploy_edpm_compute.yml
```
raukadah, line 12: Can we call this task after "call predeploy" tasks? It will help the user who just wants to deploy podified control plane

ci_framework/roles/edpm_prepare/tasks/main.yml
```diff
@@ -0,0 +1,98 @@
...
+    name: 'install_yamls_makes'
+    tasks_from: 'make_input'
+
+# TODO: Prepare a proper operator list that dictates what operators to deploy instead of relaying on a simple flag
+- name: OpenStack meta-operator installation
+  when: not cifmw_edpm_skip_openstack_operator
+  block:
+  - name: Install OpenStack operator
+    vars:
+      make_openstack_env: "{{ cifmw_edpm_prepare_common_env | from_yaml |
```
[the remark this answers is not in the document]
pablintino: Good point, done.

ci_framework/roles/edpm_prepare/tasks/main.yml
pablintino: Good catch, done.

ci_framework/playbooks/06-deploy-edpm.yml
cjeanner: not sure - both are OK in yaml, just less readable by humans

ci_framework/playbooks/06-deploy-edpm.yml
```diff
@@ -6,6 +6,10 @@
 - hosts: "{{ cifmw_target_host | default('localhost') }}"
   gather_facts: false
   tasks:
+    - name: Call predeploy
```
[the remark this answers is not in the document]
pablintino: True, I changed this the other way around to acomplish @raukadah suggestions, but, if I end implementing the skip deploying edpm computes in case ```cifmw_libvirt_manager_configuration.vms.compute.amount``` this change has no sense.

ci_framework/roles/edpm_prepare/README.md
```diff
@@ -0,0 +1,11 @@
+## Role: edpm_prepare
+Prepares the environment to deploy OpenStack control plane and compute nodes.
+
+### Privilege escalation
+This role doesn't need privilege scalation.
+
+### Parameters
+* `cifmw_edpm_prepare_basedir`: (String) Base directory. Defaults to `cifmw_basedir` which defaults to **~/ci-framework**.
+* `cifmw_edpm_prepare_intall_yamls_output_dir`: (String) Directory in where install_yamls output manifests will be placed. Defaults to **"{{ cifmw_edpm_prepare_basedir }}/artifacts/kustomize"**
```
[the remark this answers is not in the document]
pablintino: Sure, I alredy knew that https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci_framework/roles/ci_setup/tasks/directories.yml#L9 as that line is the one that needed the https://github.com/openstack-k8s-operators/ci-framework/pull/158 patch. What's your opinion of a global ' ' ' ```cifmw_manifets```? I'd like to remove the 'kustomize' in the name, as there we will put all kind of k8s and kustomize is a tool that is there today but who knows tomorrow, what we can ensure is that in that directory we will put k8s manifests.
The CR folder, the one of, will be something like this:
```
"{{ cifmw_manifets }}/{{ cifmw_install_yamls_defaults['NAMESPACE'] }}/cr"
```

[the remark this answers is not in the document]
pablintino: I've created a new global called ```cifmw_manifests```. It defaults to the value of ```{{ cifmw_basedir}}/manifests```

cjeanner:
recheck

raukadah:
@cjeanner The molecule job parent is base-minimal. The default timeout of [base-minima](https://github.com/rdo-infra/review.rdoproject.org-config/blob/7a22d43c16feed46b094f59fd83d3f31ed76b29c/zuul.d/jobs.yaml#L46)l job is 1800 sec  but the job is taking more than default timeout. We might want to increase the timeout to 3600.

Add timeout: 3600 here https://github.com/openstack-k8s-operators/ci-framework/blob/main/zuul.d/jobs.yaml#LL4C25-L4C25.

It will fix the timeout issue may be.

raukadah (approved):
Great work @pablintino :-)

PR #130, opened 2023-04-19
Use an action plugin instead of a list of tasks

cjeanner:
Using an action plugin should make it a bit faster and easier to call in
the future.

This PR has:

### DIFF
--- a/ci_framework/plugins/action/discover_latest_image.py
+++ b/ci_framework/plugins/action/discover_latest_image.py
@@ -1,0 +1,113 @@
+# Copyright Red Hat, Inc.
+# Apache License Version 2.0 (see LICENSE)
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+DOCUMENTATION = r'''
+---
+action: discover_latest_image
+
+short_description: Discover the latest image available on remote server.
+
+description:
+- Discover latest image matching a specific pattern on remote server.
+
+options: Please refer to ansible.builtin.uri options
+
+'''
+
+EXAMPLES = r'''
+- name: Get latest CentOS 9 Stream image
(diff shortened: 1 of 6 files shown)


raukadah (approved):
/lgtm

ci_framework/plugins/action/discover_latest_image.py
```diff
@@ -0,0 +1,113 @@
...
+        # Prepare matcher
+        matcher = re.compile(rf'{img_prefix}.*\.qcow2$')
+
+        # Start parsing
+        parsed = BeautifulSoup(page['content'], 'html.parser')
+        all_links = parsed.find_all('a')
+        extracted = []
+        for link in all_links:
+            if matcher.match(link.get('href')):
+                extracted.append(link.get('href'))
```
pablintino, line 93: Seems to be a good place to use list comprehension
```python
all_links = [a_elem.get('href') for a_elem in parsed.find_all('a')]
extracted =  [link for link in all_links if matcher.match(link)]
```

ci_framework/plugins/action/discover_latest_image.py
```diff
@@ -0,0 +1,113 @@
...
+
+        if BS4_IMPORT:
+            raise AnsibleError(missing_required_lib('beautifulsoup4'))
+
+        if 'image_prefix' not in module_args:
+            raise AnsibleError('"image_prefix" parameter is mandatory')
+
+        img_prefix = module_args.pop('image_prefix')
+
+        result = {
```
pablintino, line 67: Why not initializing these when used? At a first glance I expected this to be used in the following lines but it's used at near the end, when the whole result can be created in line 105

pablintino (commented):
Only a few style comments

PR #134, opened 2023-04-20
Allow ci_make to work with remote host

cjeanner:
Until now, ci_make was pushing files to the controller instead of remote
host. This prevented operators to run the Framework against a remote
host, since the directory tree (cifmw_basedir) was created remotely, but
the action plugin wanted to write on the controller.

This change now leverage the `ansible.builtin.copy` module instead,
ensuring the content is consistently pushed on the target.

We also take the opportunity to remove the "warning" about potentially
old community.general collection. It was raising false-positive under
some conditions, especially when the `make` failed: in such a case, the
"command" isn't present in the module output, even though the collection
was up to date.

This PR has:


raukadah (approved):
/lgtm

PR #138, opened 2023-04-21
Operator build plugin now outputs a results dictionary

pablintino:
To be able to fecth the operators build in the operators build step the role should pass to the next steps, mainly the deploy one, the images that were build.

As this role can add useful information for further usage like the git hash of each build operator the dictionary can be used for other fields in the future.

This PR has:

### DIFF
--- a/ci_framework/roles/operator_build/tasks/clone.yml
+++ b/ci_framework/roles/operator_build/tasks/clone.yml
@@ -24,3 +24,16 @@
     repo: "https://github.com/{{ cifmw_operator_build_org }}/{{ operator.name }}.git"
     dest: "{{ operator.src }}"
   when: not stat_op_src.stat.exists
+
+- name: "{{ operator.name }} - Fetch operator src commit"
+  ansible.builtin.command:
+    cmd: "git rev-parse HEAD"
+  register: operator_src_hash
+  when: operator.pr_sha is not defined
+
+- name: "{{ operator.name }} - Save operator src hash"
+  ansible.builtin.set_fact:
+    cifmw_operator_build_output: >-
+        {{ cifmw_operator_build_output|combine({'operators': { operator.name: {
+          'commit_hash': operator_src_hash.stdout if operator_src_hash is defined else operator.pr_sha
+        }}}, recursive=True)}}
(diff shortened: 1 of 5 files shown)


ci_framework/roles/operator_build/vars/main.yml
```diff
@@ -0,0 +1,26 @@
...
+
+# While options found within the vars/ path can be overridden using extra
+# vars, items within this path are considered part of the role and not
+# intended to be modified.
+
+# All variables within this role should have a prefix of "cifmw_rhol_crc"
+
+# https://access.redhat.com/documentation/en-us/red_hat_openshift_local/2.14/html/getting_started_guide/installation_gsg#required-software-packages_gsg
+cifmw_operator_build_output:
+  operators: {}
```
cjeanner, line 26: missing newline char at the end of the file - but it must NOT have any newline.. Check with your text editor how to get it (vim works nicely with that btw)

cjeanner (changes requested):

ci_framework/roles/operator_build/tasks/clone.yml
```diff
@@ -24,3 +24,15 @@
     repo: "https://github.com/{{ cifmw_operator_build_org }}/{{ operator.name }}.git"
     dest: "{{ operator.src }}"
   when: not stat_op_src.stat.exists
+
+- name: "{{ operator.name }} - Fetch operator src commit"
```
cjeanner, line 28: IIRC the role allow to pass down a specific commit - we may want to re-use that param if it's set, imho

cjeanner (changes requested):
It would be good to add a molecule test for that new dict, in order to ensure we get the expected content in the end.

cjeanner (changes requested):
in the tasks/main.yml, please add at the end a new artifacts:
```
ansible.builtin.copy:
  dest: "{{ cifmw_operator_build_basedir }}/artifacts/custom-operators.yml"
  content: "{{ cifmw_operator_build_output | to_nice_yaml }}"
  mode: 0644
```
That way, we'll get a trace of the built content that may be re-used for promotion or related tasks - also for debugging.

PR #139, opened 2023-04-22
Allow to pass an alternative image for golang

cjeanner:
With the limitations of docker hub registry, it may be good to provide
an alternative image.

This PR has:


cjeanner:
@raukadah we may want to mirror the golang image on quay.rdoproject.org as well - apparently it's not present ATM. WDYT? @viroel any thoughts?

raukadah (approved):
/lgtm

PR #141, opened 2023-04-24
Introduce Zuul jobs directory structure

raukadah:
Ci-framework is going to be the central repository for holding most of the common job definitions and playbooks.

In order to keep the zuul jobs and playbook cleaner, We need to provide proper directory structure.

Here is the new dir structure:
- ci/playbooks for storing for all CI related playbooks
- zuul.d/jobs.yaml for all the job definitions
- zuul.d/nodeset.yaml for nodeset definitions
- zuul.d/projects.yaml expose the zuul jobs here.

This PR has:


cjeanner:
/lgtm

I'll then do some follow-up work in order to split the molecule jobs and get rid of the (horrible) run_molecule script (or, at least, make it simpler). We'll rely on the zuul detection directly instead of doing some fun things (that doesn't really work).

cjeanner:
/lgtm

with correct nodeset definition it will work :).

arxcruz (approved):

PR #145, opened 2023-04-25
Make generate_make_env module output install_yamls Makefile default values

pablintino:
Since edpm_prepare role needs to be aware of some default values we need to get into Ansible the already parsed Makefile default values.

Overridden values by the 'cifmw_install_yamls_vars' are passed along already overridden.

This PR has:


cjeanner (commented):
It looks nice, but I'm wondering if we should really make the whole override part in the module itself instead of just using `|combine` in ansible.
Keeping the module as simple as possible, with just the "default_values" output, would probably be good enough.

Also.. We may want to just use the ansible_facts (like it was done in previous version), since it would allow to do less things in ansible and get a direct access to the facts from wherever in the run.

ci_framework/roles/install_yamls/tasks/main.yml
```diff
@@ -14,6 +14,12 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+
+- name: "Get environment structure"
+  register: get_makefile_env_output
+  get_makefile_env:
+    makefile_path: "{{ cifmw_install_yamls_repo }}/Makefile"
```
cjeanner, line 21: You'd be missing the devsetup/Makefile content.

ci_framework/plugins/modules/get_makefile_env.py
```diff
@@ -0,0 +1,89 @@
...
+import os
+
+
+# Return install_yamls vars
+def __get_makefile_vars(makefile_path):
+    if not os.path.exists(makefile_path):
+        return None
+
+    with open(makefile_path) as f:
+        lines_split = [line.split('?=') for line in f.readlines() if '?=' in line and not line.startswith('#')]
```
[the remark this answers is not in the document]
pablintino: I don't think so since ?= is a conditional assignment, it assigns it's value only if the value isn't provided from outside i.e a param or an environment variable. := assignments will ignore our given values so I assume that those variables are the ones that are not meant to be overridden.

ci_framework/roles/install_yamls/tasks/main.yml
cjeanner: We'd need to consider the "cost" of calling the module twice. Proposal: what about "find all Makefile in tree" and iterate on them all?
The idea was mostly to avoid calling 2 times the same action, and I think we should stick to it. Plus, if we use a kind of "find", we'd be able to cover future changes to the Makefile locations... ?

pablintino: Sounds good to me. I'll change the behavior

ci_framework/plugins/modules/get_makefiles_env.py
```diff
@@ -0,0 +1,93 @@
...
+author:
+  - Chandan Kumar (@raukadah)
+  - Cedric Jeanneret (@cjeanner)
+  - Pablo Rodriguez (@pablintino)
+'''
+
+EXAMPLES = r'''
+# Recursively get all the variables of the Makefiles under /home/user
+- name: Generate env var fact
+  get_makefiles_env:
```
cjeanner, line 34: I think we should add "register: my_makefile_params" in order to emphasis it won't end in the facts.

ci_framework/roles/install_yamls/tasks/main.yml
```diff
@@ -22,20 +28,34 @@
         path: "{{ cifmw_install_yamls_out_dir }}"
         state: directory
 
-    - name: "Get environment structure"
-      generate_make_env:
-        install_yamls_path: "{{ cifmw_install_yamls_repo }}/Makefile"
-        install_yamls_var: "{{ cifmw_install_yamls_vars }}"
+    - name: Set environment override edpm_env fact
+      ansible.builtin.set_fact:
```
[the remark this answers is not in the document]
cjeanner: You missed re-adding the `cacheable: true`

cjeanner (changes requested):

raukadah (commented):
small nit pick, rest looks good. Thank you Pablo for improving it. :-)

raukadah (commented):
Thank you :-) +2

cjeanner:
Waiting for the e2e job to finish to double-check generated content, but it seems to be OK at first look

PR #147, opened 2023-04-26
Inject kustomize snippet for generated computes

cjeanner:
This new snippet will override the one injected by install_yamls, in
order to cover any amount of compute nodes. install_yamls has a fixed,
hardcoded limit of 2 computes (0, 1) while the Framework has no
limitation.

This patch also takes advantage of the new get_makefiles_env module
output in order to set various content in the kustomize snippet. Since
the kustomize location is also kind of hardcoded in install_yamls, we
therefore need to stick to the directory tree.
Since this directory tree takes the NAMESPACE as a part, we need to get
that data in the ci_setup role - leading to some refactoring of the
01-bootstrap.yml playbook.

Depends-On: https://github.com/openstack-k8s-operators/install_yamls/pull/229


cjeanner:
recheck

cjeanner:
/hold

Generated kustomize snippet:
https://logserver.rdoproject.org/47/147/7012759723f545cde9dedae462bb7951d647e225/github-check/cifmw-end-to-end/1b8aca4/ci-framework/artifacts/openstack/cr/99-cifmw-computes.yaml

Looks OK, but we'll need to test it and, maybe, do some more adjustments and tuning. Note that, in order to test it, we HAVE to get the Depends-On patch merged.

Let's hold this patch until then..

ci_framework/roles/libvirt_manager/templates/kustomize_compute.yml.j2
```diff
@@ -0,0 +1,28 @@
+{%- set compute_range = range(0, compute_config.amount | default(1) | int) | list -%}"
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+namespace: {{ cifmw_install_yamls_defaults['NAMESPACE'] }}
+patches:
+- target:
+    kind: OpenStackControlPlane
```
raukadah, line 8: Kustomize is going to replace these vars using this [cr file](https://github.com/openstack-k8s-operators/install_yamls/blob/master/Makefile#L201)  and the correct[ kind ](https://github.com/openstack-k8s-operators/install_yamls/blob/master/Makefile#L373)is 
```
OpenStackDataPlane
```

ci_framework/roles/libvirt_manager/templates/kustomize_compute.yml.j2
```diff
@@ -0,0 +1,28 @@
...
+  patch: |-
```
raukadah, line 9: I am not sure we need this file. The current install_yamls kustomization works on [this cr file](https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/config/samples/dataplane_v1beta1_openstackdataplane.yaml) . If a particular path does not exist. The kustomization will does not work.

for example:
```
path: /spec/nodes/edpm-compute-2/ansibleHost
```
is not present in the CR file then it will fail.

raukadah (commented):
I am not sure about kustomization template. For updating a cr file, we need to make sure the path exists otherwise kustomize will fail. I have left couple of questions.

ci_framework/roles/libvirt_manager/templates/kustomize_compute.yml.j2
```diff
@@ -0,0 +1,28 @@
+{%- set compute_range = range(0, compute_config.amount | default(1) | int) | list -%}"
```
raukadah, line 1: May be we can check edpm-compute-(compute_config.amout -1) node exists in the cr file and then apply the kustomization. this way we can avoid failing.

PR #149, opened 2023-04-27
Add operator src directory for usage in future deploy stages.

pablintino:
Since the edpm_prepare role needs the actual build src instead of a checkout of a branch (as src can be from a PR and not only the HEAD of a branch) the source path used during operator build should be passed along to make edpm_prepare clone its copy from that path.
Depends-On: https://github.com/openstack-k8s-operators/install_yamls/pull/230


cjeanner:
/assign @viroel 

Seems OK - but would love to get Doug review on that

viroel:
/lgtm

outputs here [1]

[1] https://logserver.rdoproject.org/49/149/65e5fae6b3e31fc45910302a64335069d66ebfd1/github-check/cifmw-end-to-end/80f4229/ci-framework/artifacts/custom-operators.yml

PR #151, opened 2023-04-27
Allow to pass extra variables/parameters to playbook hooks

cjeanner:
Until now, we weren't able to pass any extra parameters to the playbook
we want to run as hooks.

This patch now corrects this limitation, and adds support for "inline
parameters" as well as "environment file". This means that "-e foo=bar"
as well as "-e @my-file.yml" are supported.

In order to leverage this, you'll need to add the following entry in the
hook dict:
```YAML
extra_vars:
  foo: bar
  file: /path/to/my/file
```

It will then be converted to proper parameters for the
`ansible-playbook` command.

This PR has:


raukadah (approved):
/lgtm

PR #152, opened 2023-04-27
Introduce hook callback

cjeanner:
Introduce hook callback

If a hook creates a file such as
`{{ cifmw_basedir }}/artifacts/{{ step }}_{{ hook_name }}.yml`
the run_hook role will load it after the actual run.

`cifmw_basedir`, `hook_name` and `step` parameters are passed down from the
`run_hook` role directly, inheriting them from the main environment.

The generated file will then be loaded using the
ansible.builtin.include_vars, and they will then be available as
top-level parameters inside the main play.

This is especially interesting for hooks generating random content that
need to be feeded back in some other parts of the run. We may also
leverage this in order to pass data between hooks, since we may load the
data from within the hook playbook as well (location is known, nothing
would prevent this).

This PR has:

### DIFF
--- a/ci_framework/playbooks/02-infra.yml
+++ b/ci_framework/playbooks/02-infra.yml
@@ -1,6 +1,7 @@
 - name: Run pre_infra hooks
   vars:
     hooks: "{{ pre_infra | default([]) }}"
+    step: pre_infra
   ansible.builtin.import_playbook: ./hooks.yml
 
 - hosts: "{{ cifmw_target_host | default('localhost') }}"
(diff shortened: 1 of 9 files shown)


cjeanner:
/hold

Need to correct a mistake with file name.

cjeanner:
Victory!
https://logserver.rdoproject.org/52/152/61e30f024f64fe708fa418513d4184c10c8b3f68/github-check/cifmw-molecule-run_hook/b20c63b/ci-framework/artifacts/molecule_with_extra-vars.yml

and the molecule job is successfully testing the availability of ceph_uuid

PR #155, opened 2023-05-02
Skip operator_build role if no operator list is provided

viroel:
This PR has:

### DIFF
--- a/ci_framework/playbooks/05-build-operators.yml
+++ b/ci_framework/playbooks/05-build-operators.yml
@@ -17,6 +17,9 @@
         oc registry login --insecure=true
 
     - name: Build operator and meta-operator
+      when:
+        - cifmw_operator_build_operators is defined
+        - cifmw_operator_build_operators | length > 0
       ansible.builtin.include_role:
         name: operator_build
 


ci_framework/playbooks/05-build-operators.yml
```diff
@@ -17,6 +17,7 @@
         oc registry login --insecure=true
 
     - name: Build operator and meta-operator
+      when: cifmw_operator_build_operators is defined
```
cjeanner, line 20: MAYBE a double-check:
- is defined
- | length > 0
Just to be on the safe side..

pablintino:
Keep in mind that this change will break the edpm_prepare. How are we gonna deal with the defaults of each step? If, for example, the operator_build gets skipped for whatever reason, are we gonna create default values in that step/playbook (here)ci_framework/playbooks/05-build-operators.yml or are we gonna let every user step (the edpm_prepare now) to deal with the defaults? Just asking to know how to approach the fix.

cjeanner:
> Keep in mind that this change will break the edpm_prepare. How are we gonna deal with the defaults of each step? If, for example, the operator_build gets skipped for whatever reason, are we gonna create default values in that step/playbook (here)ci_framework/playbooks/05-build-operators.yml or are we gonna let every user step (the edpm_prepare now) to deal with the defaults? Just asking to know how to approach the fix.

We'd need to ensure we get a sensible default set at some point in the future step supposed to consume the previous data.
For instance, we may do this:
```YAML
- name: Set some needed facts if not set already
  when: cifmw_prepared_fact_from_last_step is not set
  ansible.builtin.set_fact:
    cifmw_prepared_fact_from_last_step: {}
```

Or we can use `| default({})` for instance... @pablintino care to address this?

cjeanner:
> > Keep in mind that this change will break the edpm_prepare. How are we gonna deal with the defaults of each step? If, for example, the operator_build gets skipped for whatever reason, are we gonna create default values in that step/playbook (here)ci_framework/playbooks/05-build-operators.yml or are we gonna let every user step (the edpm_prepare now) to deal with the defaults? Just asking to know how to approach the fix.
> 
> I think that we should consider the scenario where we just want to deploy everything consuming latest containers from default registry.

We may add a second end-to-end job in CI, like `cifmw-e2e-nobuilds` in order to follow the "easy path" without ANY build - no operator, no package, no container.

That may help catching this kind of issues.

pablintino:
@viroel @cjeanner Sure, I think that this is needed, that's why I was asking how are we going to deal with the default scenario. I'll go with the ```| default({}) ``` approach tomorrow when I'm back.

cjeanner:
Just added such a job with #156 - we'll need this current one in before I think, and the rebasing 156 would ensure we're all good.

cjeanner:
LGTM

cjeanner (approved):

PR #158, opened 2023-05-02
Pin ansible-compat in order to work around known issue

cjeanner:
As described in
https://github.com/ansible-community/molecule/issues/3903 the latest
updates in ansible-compat are breaking molecule.

While waiting for the proper fix, let's just pin things up.


raukadah (approved):
/lgtm

PR #159, opened 2023-05-03
Fixing edpm_deploy molecule test default var

pablintino:
Since the addition of openstack-k8s-operators/ci-framework#147 the edpm_deploy molecule test will fail as it will call ci_setup without a default value for the cifmw_install_yamls_defaults variable, needed there to create the CR directory, what is namespace dependant.

This PR has:


pablintino:
Patch applied in https://github.com/openstack-k8s-operators/ci-framework/pull/147 itself

PR #165, opened 2023-05-04
Some quick corrections to zuul and Containerfile.test

cjeanner:
First, we want to enable the cifmw-end-to-end-nobuild job - it was
missing in the patch that created it, so it was never launched until
now.

Second, we want to make the local test container a bit more "like a
system" and allow to leverage sudo from the rootless "prow" user. It may
help debugging some situations.

This PR has:


cjeanner:
/approve

We need that new job in order to ensure a better coverage.

PR #170, opened 2023-05-16
Remove name and override-checkout from required projects

raukadah:
Recently the default branch of all openstack-k8s-operators
    projects is changed from master to main.
    
    We donot need to hardcode the repo name and override-checkout
    in the Zuul job definition. This patch does the same.
    
    It also includes other changes:
    - Adds centos-9-crc-3xl nodeset for edpm jobs and
      also bumps the crc ram to 20 GB.
    - Adds ignore-2.15.txt to fix ansible tests.
    - image_catalog instead of image to reference
      the OPENSTACK_IMG value.
    
This PR has:

### DIFF
--- a/ci_framework/tests/sanity/ignore-2.15.txt
+++ b/ci_framework/tests/sanity/ignore-2.15.txt
@@ -1,0 +1,1 @@
+ignore.txt
--- a/zuul.d/edpm.yaml
+++ b/zuul.d/edpm.yaml
@@ -2,21 +2,19 @@
 # Base job for openstack crc based job
 - job:
     name: cifmw-base-crc-openstack
-    nodeset: centos-9-crc-xxl
+    nodeset: centos-9-crc-3xl
     timeout: 10800
     abstract: true
     parent: base-crc
     required-projects:
-      - github.com/openstack-k8s-operators/install_yamls
+      - openstack-k8s-operators/install_yamls
       - openstack-k8s-operators/openstack-operator
-      - name: github.com/openstack-k8s-operators/ci-framework
-        override-checkout: main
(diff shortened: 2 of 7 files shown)


SeanMooney:
i dont think this shoudl be requried any more.

in general you should avoid using overied_checkout so now that all project have had there default branch renamed form master to main we should probably do the opposite and remove all
override-checkout: main useage

arxcruz (approved):
Although you are also updating the memory flavor, I am not against it.

ci_framework/tests/sanity/ignore-2.15.txt
```diff
@@ -0,0 +1 @@
+ignore.txt
```
pablintino, line 1: I'm gonna add something more flexible for this in a separate PR

pablintino (approved):

zuul.d/edpm.yaml
```diff
@@ -49,4 +47,4 @@
           operators:
               openstack-operator:
                   git_src_dir: /home/zuul-worker/src/github.com/openstack-k8s-operators/openstack-operator
-                  image: quay.io/openstack-k8s-operators/openstack-operator-index:latest
+                  image_catalog: quay.io/openstack-k8s-operators/openstack-operator-index:latest
```
SeanMooney, line 50: +1

SeanMooney (commented):
you have mixed some unrelated changes together but i think over all this looks goo

pablintino:
> you have mixed some unrelated changes together but i think over all this looks good

Can't agree more, we should stop mixing things together, the title of the PR and the commit message is totally irrelevant with this practice

PR #171, opened 2023-05-16
[Content provider]Return cifmw_operator_build_output var to child job

raukadah:
The child job depends on content provider. The content provider
needs to return cifmw_operator_build_output var to the child job
so that child job should consume the proper operator images
to do the deployment.
    
This patch also removes the hardcoded content provider ip
and replace it with var. It also moves the tasks to discover
the content provider ip in pre and pass it as a var to
 the inner ansible call.
    
Note: It also fixes the operator_build role to include
the operator image during make_bundle.
    
This PR has:

### DIFF
--- a/ci/playbooks/content_provider/run.yml
+++ b/ci/playbooks/content_provider/run.yml
@@ -11,13 +11,20 @@
           -e @scenarios/centos-9/base.yml
           -e @scenarios/centos-9/content_provider.yml
           -e @scenarios/centos-9/zuul_inventory.yml
+          -e "cifmw_rp_registry_ip={{ cifmw_rp_registry_ip }}"
 
-    - name: Discover an IPv4 for provider job
+    - name: Include inner ansible vars file
+      ansible.builtin.slurp:
+        src: "{{ cifmw_artifacts_basedir }}/artifacts/ansible-vars.yml"
+      register: inner_ansible
+
+    - name: Get inner ansible vars
       ansible.builtin.set_fact:
-        content_provider_ip: >-
-          {{ hostvars[groups.all[0]].ansible_host if hostvars[groups.all[0]].ansible_host
-          is match("[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+")
-          else hostvars[inventory_hostname]['ansible_default_ipv4']['address'] }}
+        inner_ansible_vars: "{{ inner_ansible.content | b64decode | from_yaml }}"
+
+    - name: Set content provider
+      ansible.builtin.set_fact:
+        content_provider_ip: "{{ cifmw_rp_registry_ip }} "
 
     - name: Return Zuul Data
       ansible.builtin.debug:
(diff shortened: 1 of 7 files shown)


ci/playbooks/content_provider/run.yml
```diff
@@ -31,3 +31,4 @@
           zuul:
             pause: true
           content_provider_registry_ip: "{{ content_provider_ip | default('nowhere') }}"
+          cifmw_operator_build_output: "{{ cifmw_operator_build_output }}"
```
SeanMooney, line 34: this appears to be undefied here

https://review.rdoproject.org/zuul/build/2836b71121c7445493eb4b7495ce95e0/log/job-output.txt#1319-1323

ci_framework/roles/registry_deploy/tasks/main.yml
```diff
@@ -55,5 +55,5 @@
     dest: /etc/containers/registries.conf
     content: |-
       [[registry]]
-      location = "localhost:{{ cifmw_rp_registry_port }}"
+      location = "{{ cifmw_rp_registry_ip }}:{{ cifmw_rp_registry_port }}"
```
SeanMooney, line 58: im not sure if its related to this change but I'm now seign 
Error: trying to reuse blob sha256:e023e0e48e6e29e90e519f4dd356d058ff2bffbd16e28b802f3b8f93aa4ccb17 at destination: pinging container registry localhost:5001: Get "https://localhost:5001/v2/": http: server gave HTTP response to HTTPS client

https://review.rdoproject.org/zuul/build/07bb2baa7f4c4256ae4dfe98bb181fae/log/ci-framework/logs/ci_make_008_nova-operator_-_call_docker-push.log#18

SeanMooney: should this be "http://{{ cifmw_rp_registry_ip }}:{{ cifmw_rp_registry_port }}"

ci/playbooks/content_provider/run.yml
SeanMooney: as noted in slack there is a miss alignment on the key name

https://review.rdoproject.org/zuul/build/ad91003836bc477da6a65fad9886df45/log/zuul-info/inventory.yaml#33-39

the image key for the OpenStack-operator ahs the catalog URL and the base job is now expecting it to be called image_catalog https://review.rdoproject.org/r/c/config/+/48603/1/playbooks/crc/openstack.yaml#37

some eviter we need to revert https://review.rdoproject.org/r/c/config/+/48603/

or fix the cifmw_operator_build_output

SeanMooney (approved):
we have seen this work correctly in both the nova and datapalne operator zuul job patches.

over all this look good to me

cjeanner:
/approve

This will make the whole work far, far better, with a 1:1 match of the "late" content-provided job used in tripleo env. Thank you Chandan!

PR #172, opened 2023-05-16
Add a simple approach to deploy a local test environment based on CS9 VM

pablintino:
This PR has:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,8 @@ CI_CTX_NAME ?= localhost/cifmw:latest
 MOLECULE_CONFIG ?= ${MOLECULE_CONFIG:-.config/molecule/config_podman.yml}
 # Run molecule against all tests
 TEST_ALL_ROLES ?= ${TEST_ALL_ROLES:-no}
+# Ansible options for the local_env_vm target
+LOCAL_ENV_OPTS ?= ""
 
 # target vars for generic operator install info 1: target name , 2: operator name
 define vars
@@ -136,3 +138,12 @@ run_ctx_ansible_test: ci_ctx ## Run molecule check in a container
 enable-git-hooks:
 	git config core.hooksPath "./.githooks"
 	$(warning REMEMBER, YOU MUST HAVE REVIEWED THE CUSTOM HOOKS in .githooks!)
+
+##@ Developper environments
+.PHONY: local_env_create
+local_env_create: ## Create a virtualized lab on your local machine. Nested virt MUST be supported by your system. Use LOCAL_ENV_OPTS param to pass options to ansible-playbook
+	if [ "x${LOCAL_ENV_OPTS}" == "x" ]; then \
+		ansible-playbook -i localhost, -c local dev-local-env.yml; \
+	else \
+		ansible-playbook -i localhost, -c local dev-local-env.yml ${LOCAL_ENV_OPTS}; \
+	fi
--- a/ci_framework/roles/local_env_vm/tasks/main.yml
+++ b/ci_framework/roles/local_env_vm/tasks/main.yml
@@ -1,0 +1,144 @@
+---
+# Copyright Red Hat, Inc.
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+- name: Create image directory
+  ansible.builtin.file:
+    path: "{{ cifmw_local_env_vm_basedir }}/images"
+    state: directory
+
+- name: Debug
+  ansible.builtin.debug:
+    msg: "{{ cifmw_local_env_vm_basedir }}/images/{{ cifmw_discovered_image_name }}"
+
+- name: Manage image
+  block:
+    - name: Check image availability
+      register: base_img
+      ansible.builtin.stat:
+        path: "{{ cifmw_local_env_vm_basedir }}/images/{{ cifmw_discovered_image_name }}"
+
+    - name: Fetch image
+      when:
+        - not base_img.stat.exists
+      register: dload_img
+      ansible.builtin.get_url:
+        url: "{{ cifmw_discovered_image_url }}"
+        dest: "{{ cifmw_local_env_vm_basedir }}/images/{{ cifmw_discovered_image_name }}"
+        checksum: "sha256:{{ cifmw_discovered_sha256 }}"
+      until:
+        - dload_img.status_code is defined
+        - dload_img.status_code == 200
+      retries: 2
+      delay: 5
+
+- name: Set some facts for easier usage
+  ansible.builtin.set_fact:
+    cifmw_backing_img: "{{ cifmw_local_env_vm_basedir }}/images/{{ cifmw_discovered_image_name }}"
+    cifmw_work_img: "{{ cifmw_local_env_vm_basedir }}/images/cifmw-vm.qcow2"
+    cifmw_vm_name: "cifmw-vm"
+
+- name: Generate firstboot script as an artifact
+  ansible.builtin.copy:
+    dest: "{{ cifmw_local_env_vm_basedir }}/artifacts/firstboot.sh"
+    src: "firstboot.sh"
+    mode: 0755
+
+- name: Create image layer
+  register: create_layer
+  ansible.builtin.command:
+    creates: "{{ cifmw_work_img }}"
+    cmd: >-
+      qemu-img create -o backing_file={{ cifmw_backing_img }},backing_fmt=qcow2
+      -f qcow2 {{ cifmw_work_img }} {{ cifmw_local_env_vm_space }}
+
+- name: Prepare the workload image
+  when:
+    - create_layer is changed
+  ansible.builtin.shell:
+    executable: /bin/bash
+    cmd: >-
+      virt-sysprep -a {{ cifmw_work_img }} --network --update --selinux-relabel
+      --ssh-inject "root:file:{{ ansible_user_dir }}/.ssh/authorized_keys"
+      --root-password "password:{{ cifmw_local_env_vm_root_passwd }}"
+      --firstboot "{{ cifmw_local_env_vm_basedir }}/artifacts/firstboot.sh"
+      --copy-in "{{ cifmw_local_env_vm_pullsecret }}:/root/"
+      --install vim,git,tmux,libvirt,libvirt-client,libvirt,libvirt-daemon-kvm,virt-install,qemu-kvm,make,bash-completion,guestfs-tools,libguestfs,python3-pip 2>&1 > {{ cifmw_local_env_vm_basedir }}/logs/virt-syspre.log
+
+- name: Change access rights on layered image
+  become: true
+  ansible.builtin.file:
+    path: "{{ cifmw_work_img }}"
+    mode: 0664
+    group: qemu
+
+- name: Create VM based on the workload image
+  ansible.builtin.command:
+    cmd: >-
+      virt-install --import
+      --connect qemu:///system
+      --name "{{ cifmw_vm_name }}"
+      --memory "{{ cifmw_local_env_vm_memory }}"
+      --vcpus "{{ cifmw_local_env_vm_cpu }}"
+      --cpu host
+      --os-variant=centos-stream9
+      --graphics {{ cifmw_local_env_vm_graphics }}
+      --noautoconsole
+      --network="{{ cifmw_local_env_vm_network }}"
+      --disk "{{ cifmw_work_img }}"
+
+- name: Get local_vm IP
+  register: local_env_vm_ip
+  ansible.builtin.command:
+    cmd: "virsh -c qemu:///system -q domifaddr --source arp --domain {{ cifmw_vm_name }}"
+  until: "'ipv4' in local_env_vm_ip.stdout"
+  retries: 30
+  delay: 5
+
+- name: Set VM ip as a fact for later use
+  ansible.builtin.set_fact:
+    cifmw_local_vm_ip: "{{ local_env_vm_ip.stdout.split()[-1].split('/')[0] }}"
+    cacheable: true
+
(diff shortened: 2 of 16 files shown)


Makefile
```diff
@@ -14,6 +14,10 @@ MOLECULE_CONFIG ?= ${MOLECULE_CONFIG:-.config/molecule/config_podman.yml}
 # Run molecule against all tests
 TEST_ALL_ROLES ?= ${TEST_ALL_ROLES:-no}
 
+CIFMW_LOCAL_ENV_IMAGE_URL ?= https://cloud.centos.org/centos/9-stream/x86_64/images/CentOS-Stream-GenericCloud-9-20230327.0.x86_64.qcow2
+
+CIFMW_LOCAL_ENV_BASE_DIR ?= /opt/ci-framework
```
raukadah, line 19: Since the user is going to run make local_env_create from ci_framework. So can we change like this.
```
CIFMW_LOCAL_ENV_BASE_DIR ?= ${PWD}
```

scripts/local_env_create_vm
```diff
@@ -0,0 +1,118 @@
...
+if [ "$EUID" -ne 0 ]; then
+    echo "Local env needs sudo permissions"
+    exit 1
+fi
+
+SCRIPTPATH="$( cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
+
+CIFMW_LOCAL_ENV_BASE_DIR="${1}"
+CIFMW_LOCAL_ENV_BACKING_IMG_URL="${2}"
+CIFMW_LOCAL_ENV_PULL_SECRET_PATH="${3}"
```
raukadah, line 32: If a user runs like this 
```
CIFMW_LOCAL_ENV_BASE_DIR=${PWD} CIFMW_LOCAL_ENV_PULL_SECRET_PATH=~/pull-secret sudo make local_env_create
```
then it generally hit by following issue.
```
+ CIFMW_LOCAL_ENV_BASE_DIR=/opt/ci-framework
+ CIFMW_LOCAL_ENV_BACKING_IMG_URL=https://cloud.centos.org/centos/9-stream/x86_64/images/CentOS-Stream-GenericCloud-9-20230327.0.x86_64.qcow2
scripts/local_env_create_vm: line 32: 3: unbound variable
```
And if user runs like this
```
sudo CIFMW_LOCAL_ENV_BASE_DIR=${PWD} CIFMW_LOCAL_ENV_PULL_SECRET_PATH=/home/developer/pull-secret make local_env_create
```
It works fine. Can we handle the above issue also?

scripts/local_env_create_vm
```diff
@@ -0,0 +1,118 @@
...
+if [[ ! -f "$CIFMW_LOCAL_ENV_BACKING_IMG" ]]; then
+    mkdir -p "$CIFMW_LOCAL_ENV_IMAGES_DIR"
+    curl -l -o "$CIFMW_LOCAL_ENV_BACKING_IMG" "${CIFMW_LOCAL_ENV_BACKING_IMG_URL}"
+    curl -l -o "${CIFMW_LOCAL_ENV_BACKING_IMG}.SHA256SUM" "${CIFMW_LOCAL_ENV_BACKING_IMG_URL}.SHA256SUM"
+    (cd "$CIFMW_LOCAL_ENV_IMAGES_DIR" ; sha256sum -c "${CIFMW_LOCAL_ENV_BACKING_IMG}.SHA256SUM")
+fi
+
+rm -rf "$CIFMW_LOCAL_ENV_VM_IMG"
+
+qemu-img create -o backing_file="$CIFMW_LOCAL_ENV_BACKING_IMG",backing_fmt=qcow2 -f qcow2 "$CIFMW_LOCAL_ENV_VM_IMG" "120G"
```
raukadah, line 55: On a fresh box, qemu-img, virt-sysprep (coming from guestfs-tools) cli is missing. I need to install these packages separately. 

Can we add the following step at Line 54:
```
sudo dnf -y install qemu-img guestfs-tools
```

scripts/local_env_create_vm
```diff
@@ -0,0 +1,118 @@
...
+  --selinux-relabel \
+  --root-password "password:fooBar" \
+  --firstboot "${SCRIPTPATH}/local_env_vm_first_boot" \
+  --update \
+  --write "/root/pull-secret:$(cat ${CIFMW_LOCAL_ENV_PULL_SECRET_PATH})" \
+  --install vim,git,tmux,libvirt,libvirt-client,libvirt,libvirt-daemon-kvm,virt-install,qemu-kvm,make,bash-completion,guestfs-tools,libguestfs,python3-pip
+
+virt-install --import \
+--name "$CIFMW_LOCAL_ENV_VM_NAME" \
+--memory 22572 \
```
raukadah, line 67: Can we parameterize memory and vcpus and keeping the above values as default. So that user can update it based on their needs?

scripts/local_env_create_vm
```diff
@@ -0,0 +1,118 @@
...
+-i "${CIFMW_LOCAL_ENV_CALLING_USER_HOME}/.ssh/id_rsa" \
+"zuul@$CIFMW_LOCAL_ENV_VM_IP" \
+"$(< ${SCRIPTPATH}/local_env_vm_prepare)"
+
+if [[ $? -ne 0 ]]; then
+    echo "Post installation script failed"
+    exit $?
+fi
+
+echo "Local environment setup success! VM IP: $CIFMW_LOCAL_ENV_VM_IP. Use 'ssh zuul@$CIFMW_LOCAL_ENV_VM_IP' to connect"
```
raukadah, line 118: Can we add one more line to guide the users for further steps.
```
echo " Run deploy-edpm.yml to deploy openstack: cd ~/src/ci-framework; ansible-playbook deploy-edpm.yml -e @scenarios/centos-9/local-env.yml"
```

raukadah (commented):
Awesome work @pablintino . It will improve the whole user experience. I have left few nit picks .

scripts/local_env_create_vm
IgnacioPro: @raukadah  We also now have that documented on the official wiki here:

https://github.com/openstack-k8s-operators/ci-framework/wiki/Quickstart

Makefile
pablintino: Sure? As we are gonna run as root we can end with root file under users $HOME, seems ugly...

scripts/local_env_create_vm
pablintino: Uhhhmm this is not related with the script. With the notation you propose you are passing to make the args as environment variables. As you are putting them before sudo, those values will be exported only in your current shell process and won't be passed to the child, and sudoed, shell unless you add -E to sudo.

scripts/local_env_create_vm
pablintino: Good point, I was expecting to have this at some point out of the first iteration, but is doable. I'll add it.

scripts/local_env_create_vm
pablintino: Good point, I was expecting to have this at some point out of the first iteration, but is doable. I'll add it.

scripts/local_env_create_vm
arxcruz: Agree with Pablo, also, this patch is lacking of a readme, it won't matter to have it in the Makefile if we don't have this information on the readme

arxcruz (changes requested):

scripts/local_env_create_vm
cjeanner, line 1: IMHO that script should leverage the existing roles we have in ci-framework. They will properly handle:
- libvirt dependencies and configuration
- directories
- rights

That may come as a follow-up if we need this asap for users to consume it, but we really, really should keep that in mind.

scripts/local_env_create_vm
```diff
@@ -0,0 +1,137 @@
...
+
+if [ "$EUID" -ne 0 ]; then
+    echo "Local env needs sudo permissions"
+    exit 1
+fi
+
+SCRIPTPATH="$( cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
+
+CIFMW_LOCAL_ENV_BASE_DIR="${1}"
+CIFMW_LOCAL_ENV_BACKING_IMG_URL="${2}"
```
cjeanner, line 50: using the ansible action plugin, we may auto-detect the latest image directly, for instance.

scripts/local_env_create_vm
```diff
@@ -0,0 +1,137 @@
...
+CIFMW_LOCAL_ENV_PULL_SECRET_PATH="${3}"
+CIFMW_LOCAL_ENV_CALLING_USER=$([ -z "$SUDO_USER" ] && echo "root" || echo "$SUDO_USER")
+CIFMW_LOCAL_ENV_CALLING_USER_HOME=$(eval echo "~$CIFMW_LOCAL_ENV_CALLING_USER")
+
+CIFMW_LOCAL_ENV_IMAGES_DIR="${1}/images"
```
cjeanner, line 55: since ${1} is already assigned to a "human readable var name", we should use it. Note that it should be user writable - the default `/opt/ci-framework` isn't writable by unprivileged user.

Makefile
cjeanner: We shouldn't use root nor `sudo` - really. Think "user", and if really, requires the `sudo` only for the needed pieces. This is mostly addressed in the ansible roles actually. One more reason to consume them :)

scenarios/centos-9/local-env.yml
cjeanner, line 1: hmmm I think we shouldn't need to push all of those vars - some are already existing in the other scenario files. Let's consume them, and override/set just the needed bits.

scripts/local_env_create_vm
cjeanner: README and DOC (once we get the proper things set up, apparently using Sphinx and either gh-pages or readthedoc).

scripts/local_env_create_vm
```diff
@@ -0,0 +1,137 @@
...
+done
+
+if [ -z "$CIFMW_LOCAL_ENV_BACKING_IMG" ]; then
+    echo "Unable to find the $CIFMW_LOCAL_ENV_VM_NAME IP"
+    exit 1
+fi
+
+# Copy local ci-framework to target
+rsync -r -l \
+-e "ssh -i ${CIFMW_LOCAL_ENV_CALLING_USER_HOME}/.ssh/id_rsa -o StrictHostKeyChecking=no" \
```
cjeanner, line 114: maybe adding a block in the ~/.ssh/config would make it more convenient for future use?

scripts/local_env_vm_prepare
```diff
@@ -0,0 +1,35 @@
...
+CIFMW_REPO_PATH=/home/zuul/src/ci-framework
+
+if [[ ! -d "$CIFMW_REPO_PATH" ]]; then
+    git clone https://github.com/openstack-k8s-operators/ci-framework "$CIFMW_REPO_PATH"
+fi
+
+git clone https://github.com/openstack-k8s-operators/install_yamls /home/zuul/src/install_yamls
+
+pushd "$CIFMW_REPO_PATH"
+pip install --upgrade pip setuptools bindep
```
cjeanner, line 32: iirc this should be launched with `sudo` in order to expose all the things in the system. "bad", but needed. At least, that's what I did in my demo VM.

cjeanner:
/hold

Working on a addressing my own comments until Tuesday, if I can't make it, let's go with the current state and make some follow-up.

Makefile
cjeanner: new role takes this comment into account.

scripts/local_env_create_vm
cjeanner: The new role/playbook should take care of this. Maybe we'd need to add some more packages in the libvirt_manager role directly.

scripts/local_env_create_vm
cjeanner: new role takes this into account and addresses the comment.

scripts/local_env_create_vm
cjeanner: New role has something like that - and we'll have to update the doc accordingly as well.

dev-local-env.yml
```diff
@@ -0,0 +1,44 @@
...
+    cifmw_discover_latest_image_base_url: "https://cloud.centos.org/centos/9-stream/x86_64/images"
+  tasks:
+    - name: Ensure we have the needed support
+      ansible.builtin.include_role:
+        name: libvirt_manager
+        tasks_from: virtualization_prerequisites
+
+    - name: Get directories and base packages
+      ansible.builtin.include_role:
+        name: ci_setup
```
pablintino, line 18: Seems that with this we are installing "ca-certificates" not needed, and not harmful, but, If we start adding required packages for other envs, we will install more unuseful stuff in developers local envs.

ci_framework/roles/local_env_vm/tasks/main.yml
```diff
@@ -0,0 +1,137 @@
...
+      -f qcow2 {{ cifmw_work_img }} {{ cifmw_local_env_vm_space }}
+
+- name: Prepare the workload image
+  when:
+    - create_layer is changed
+  ansible.builtin.shell:
+    executable: /bin/bash
+    cmd: >-
+      virt-sysprep -a {{ cifmw_work_img }} --network --update --selinux-relabel
+      --ssh-inject "root:file:{{ ansible_user_dir }}/.ssh/authorized_keys"
```
pablintino, line 74: The user may not have an  authorized_keys file. If we assume that we need to document it, and ensure that the quick start guide covers it.

pablintino (commented):
Seems nice to me, only a couple of comments.

pablintino (commented):
Seems nice to me, only a couple of comments.

ci_framework/roles/local_env_vm/tasks/main.yml
IgnacioPro: I think another solution would be to add a pre-run task that checks if the file is present, if not, throw an error prompting the user to create it.

dev-local-env.yml
cjeanner: We may consider filtering packages if we see we're adding more and more. Though, at this point, I really doubt we'll need more "common/base" packages. More over, ca-certificates is mostly present on FC hosts, especially once we get some softwares such as the JRE and other big things (maybe libreoffice itself is depending on it in some way)

ci_framework/roles/local_env_vm/tasks/main.yml
cjeanner: hmm yeah. that would allow to support any ssh format - for instance, I don't have RSA anymore ;). Lemme amend in order to add some pre in the playbook.

cjeanner:
/lgtm
(since Pablo can't do it ;))

PR #173, opened 2023-05-17
Use centos-stream-9 nodeset instead of crc nodeset

raukadah:
CRC nodeset have higher memory and molecule jobs does not
    need higher memory.
    
    centos-stream-9 is a normal nodeset with low memory footprint.
    It will be useful for running molecule job. It will avoid the
    unwanted resources wastage.
    
    operator_deploy and rhol_crc molecule jobs requires crc node
    for testing.
    
This PR has:


raukadah:
recheck

rlandy (approved):
lgtm

PR #178, opened 2023-05-18
Operator status

raukadah:
The current task does not check the status of operators to make
 sure they are successfully deployed. It further breaks the
 openstack services deployment by giving unwanted errors.
    
This patch adds a step to make sure operators are successfully
installed. It will help to fail the job when there is an issue
with the meta operator installation.

Depends-On: #301

This PR has:
- [X] no need for checklist

### DIFF
--- a/ci_framework/roles/edpm_prepare/tasks/main.yml
+++ b/ci_framework/roles/edpm_prepare/tasks/main.yml
@@ -101,6 +101,23 @@
           --namespace={{ cifmw_install_yamls_defaults['OPERATOR_NAMESPACE'] }}
           --for=jsonpath='{.status.phase}'=Complete --timeout=20m
 
+    # TODO(pablintino): rework that part in order to use an Assert
+    # this means mostly: push this to a new task file, and retry to include_tasks
+    # The new task file may use some block/rescue in order to properly display the failure if any
+    - name: Make sure all Openstack operators are deployed
+      environment:
+        KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+        PATH: "{{ cifmw_path }}"
+      ansible.builtin.shell: >-
+        set -o pipefail && oc get csv -l operators.coreos.com/openstack-operator.openstack
+        --namespace={{ cifmw_install_yamls_defaults['OPERATOR_NAMESPACE'] }}
+        --no-headers=true | grep -i "succeeded"
+      register: operator_status
+      until: operator_status.rc == 0
+      changed_when: false
+      retries: "{{ cifmw_edpm_prepare_wait_subscription_retries }}"
+      delay: 30
+
 - name: Install OpenStack service
   vars:
     make_openstack_deploy_env: "{{ cifmw_edpm_prepare_common_env |


SeanMooney (approved):
This has been working well for me in the nova-operator-compute-kit job
so it would be nice to see this merge to enable that to proceed

ci_framework/roles/edpm_prepare/tasks/main.yml
```diff
@@ -77,6 +77,15 @@
         cmd: >
           oc wait InstallPlan {{ cifmw_edpm_prepare_wait_installplan_out.stdout }}
           --namespace={{ cifmw_install_yamls_defaults['NAMESPACE'] }} --for=jsonpath='{.status.phase}'=Complete --timeout=20m
+
+    - name: Make sure all Openstack operators are deployed
+      ansible.builtin.shell: |
+        set -o pipefail && oc get csv -l operators.coreos.com/openstack-operator.openstack --no-headers=true | grep -i "succeeded"
```
pablintino, line 83: Maybe is cleaner by getting the yaml and asserting based on that yaml fields. Specially if we can get from there errors about the actual failure.

cjeanner: :+1:  to @pablintino proposal.

Also, we have to ensure we pass this:
```
environment:
  PATH: "{{ cifmw_path }}"
```
in order to get `oc`.

ci_framework/roles/edpm_prepare/tasks/main.yml
```diff
@@ -77,6 +77,15 @@
...
+      register: operator_status
+      until: operator_status.rc == 0
+      changed_when: false
+      retries: "{{ cifmw_edpm_prepare_wait_subscription_retries }}"
+      delay: 30
   when: not cifmw_edpm_prepare_dry_run
```
cjeanner, line 89: please move it just bellow the `name` key - easier to read :)

cjeanner (changes requested):
PR needs to be rebased against `main`.

I'm more than OK with the idea to validate the deployed service of course, especially since this would fail with a clear error, instead of waiting more time.

Just... it would be good to get it done with actual ansible, so that we can indeed extract more data from the  `oc` command (maybe even dump the content in a file for later debug?)

ci_framework/roles/edpm_prepare/tasks/main.yml
cjeanner: done during the rebase

ci_framework/roles/edpm_prepare/tasks/main.yml
cjeanner: After a quick discussion:
- add a TODO
- merge as-is
- refine in a future iteration.

cjeanner:
recheck

ci_framework/roles/edpm_prepare/tasks/main.yml
```diff
@@ -105,6 +105,23 @@
...
           --for=jsonpath='{.status.phase}'=Complete --timeout=20m
 
+    # TODO(pablintino): rework that part in order to use an Assert
+    # this means mostly: push this to a new task file, and retry to include_tasks
+    # The new task file may use some block/rescue in order to properly display the failure if any
+    - name: Make sure all Openstack operators are deployed
+      environment:
+        KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+        PATH: "{{ cifmw_path }}"
+      ansible.builtin.shell: |
```
cjeanner, line 115:
```suggestion
      ansible.builtin.shell: >-
```
That would avoid the need to use `\`

cjeanner:
(just rebased against main - imho we should be able to merge it if CI is green - WDYT @raukadah ? Still valid?)

raukadah:
thank you everyone for review, I need to close this pr in favor of https://github.com/openstack-k8s-operators/ci-framework/pull/311

PR #183, opened 2023-05-22
Sphinx documentation and version support

dsariel:
1. Sphinx documentation initial commit
2. version display support

Version numbers will be automaticaly updated on the readthedoc page
Subject releases to be tagged as following:
 v1.0
 v2.0
 v2.1

See docs/source/conf.py

This PR has:


docs/source/requirements.rst
```diff
@@ -0,0 +1,10 @@
+requirements
+============
+
+## Minimum requirements
+
+To run the ci-framework locally on your machine you will need:
+
+* 32 GB RAM
```
arxcruz, line 8: 20GB

arxcruz (changes requested):
Where will we publish the html version? Or we won't?

docs/source/requirements.rst
pablintino: If fact, with our current implementation based on a nested VM, I won't recommend anything less that 32GB, as the VM is configured to use 22GB and the direct deployment/non-nested VM is not yet there. Maybe we should rephrase that whole sentence.

arxcruz: Oh, you are right, I was thinking about the VM having 20gb not the whole machine...

dsariel: @pablintino @cjeanner what option do you prefer? github page or readthedoc? 

with readthedoc permission is rewuired to integrate a webhook (https://readthedocs.org/dashboard/ci-framework/integrations/) so maybe github  page for the project is better.

cjeanner: I have not strong preferences. As long as we can automate in a convenient, non-invasive way...

IIRC, ghpage requires pushing to a specific branch - not sure how we may automate this actually.

readthedoc is kind of "the" place to be - many projects are using it in a convenient way. Integrating the webhook isn't a blocker at this stage - especially since it's a well known, heavily used and therefore tested/audited hook.

docs/Makefile
cjeanner, line 1: maybe taking the same kind of structure we have in the main Makefile would be better? Or, maybe, directly integrating a `make doc` in the main Makefile... IMHO that would be better.

docs/source/conf.py
```diff
@@ -0,0 +1,91 @@
...
+
+# If true, `todo` and `todoList` produce output, else they produce nothing.
+todo_include_todos = True
+
+
+# -- Options for HTML output ----------------------------------------------
+
+# The theme to use for HTML and HTML Help pages.  See the documentation for
+# a list of builtin themes.
+html_theme = 'sphinx_rtd_theme'
```
cjeanner, line 80: wondering if there's some "official" theme for Red Hat and/or OpenShift content?

docs/requirements.txt
cjeanner, line 1: maybe nudge it to /doc-requirements.txt instead?

cjeanner (changes requested):

docs/source/requirements.rst
dsariel: great. then we just need to fix access permissions. Who is the guy to ask for the change?

docs/Makefile
dsariel: np. I'll add "make doc" rule to the Makefile

docs/source/conf.py
dsariel: not one I know about. Could be nice to use some thought

They are advsing to use their own scheme to give all docs same look. 
Sphinx_RTD_Theme is a theme specifically designed to mimic the style of the Read the Docs (RTD) documentation platform. It provides a clean and modern look, making your documentation easy to read and navigate. The theme includes features such as responsive design, table of contents sidebar, and customizable color schemes.

docs/requirements.txt
dsariel: done

Makefile
```diff
@@ -147,3 +147,15 @@ local_env_create: ## Create a virtualized lab on your local machine. Nested virt
...
 		ansible-playbook -i localhost, -c local dev-local-env.yml ${LOCAL_ENV_OPTS}; \
 	fi
+
+##@ Generate documentation
+VIRTENV_FOLDER_NAME = virtdocs
+.PHONY: docs
+docs:
+	cd docs && \
+	virtualenv $(VIRTENV_FOLDER_NAME) && \
+	. virtdocs/bin/activate && \
```
cjeanner, line 157: you should use the VIRTENV_FOLDER_NAME instead of hardcoding "virtdocs".

Makefile
```diff
@@ -147,3 +147,15 @@ local_env_create: ## Create a virtualized lab on your local machine. Nested virt
...
+	pip install -r doc-requirements.txt && \
```
cjeanner, line 158: that would be pip install -r ../doc-requirements.txt then, I think?

docs/Makefile
cjeanner, line 1: do we still need that one since the target is in the main Makefile? Maybe RTD requirement?

dsariel:
> Where will we publish the html version? Or we won't?

The initial commit can be seen here: https://ci-framework.readthedocs.io/en/latest/

I'm working on nice charts and they will be submitted soon

docs/source/conf.py
```diff
@@ -0,0 +1,91 @@
...
+# Theme options are theme-specific and customize the look and feel of a theme
+# further.  For a list of options available for each theme, see the
+# documentation.
+#
+# html_theme_options = {}
+
+# -- Options for HTMLHelp output ------------------------------------------
+
+# Output file base name for HTML help builder.
+# htmlhelp_basename = 'doc'
```
cjeanner, line 91: missing newline char - will fail linters

docs/source/quickstart.rst
cjeanner, line 1: you may want to re-import the updated quickstart from the wiki

Makefile
```diff
@@ -147,3 +147,15 @@ local_env_create: ## Create a virtualized lab on your local machine. Nested virt
 	else \
 		ansible-playbook -i localhost, -c local dev-local-env.yml ${LOCAL_ENV_OPTS}; \
 	fi
+
+##@ Generate documentation
+VIRTENV_FOLDER_NAME = virtdocs
+.PHONY: docs
+docs:
+	cd docs && \
+	virtualenv $(VIRTENV_FOLDER_NAME) && \
```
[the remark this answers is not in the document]
dsariel: ack

Makefile
dsariel: doc-requirements.txt is under docs

docs/source/conf.py
dsariel: ack

Makefile
dsariel: guess it is related to the prev push. resolved

dsariel:
@

> Where will we publish the html version? Or we won't?

the version willbe automaticaly taken from git repo tags. So releaser will only need to tag like v1.0 (look at conf.py)
Also I have configured readthedocs to create docs for each release we have a tag for.

docs/source/conf.py
dsariel: done

dsariel:
it works now with .md files: https://ci-framework.readthedocs.io/en/latest/structure.html (just indes.rst must me in rst format)

PR #185, opened 2023-05-23
Pass additional make_edpm_deploy_env var to inner ansible

raukadah:
https://github.com/openstack-k8s-operators/edpm-ansible/pull/161 adds the EDPM based zuul job. We build the ansibleee-runner image in the EDPM zuul job itself and RDO quay registry. This image needs to be passed to the deployment.

In order to do that We need to set DATAPLANE_RUNNER_IMG install_yamls makefile var to manipulate the ansibleee-runner image.

This patch adds a conditional to check for make_edpm_deploy_env var set in pre-playbook and pass it to inner ansible-playbook call.

It will help to test the deployment.
It is being tested here:  https://github.com/openstack-k8s-operators/edpm-ansible/pull/161

This PR has:

### DIFF
--- a/ci/playbooks/edpm/run.yml
+++ b/ci/playbooks/edpm/run.yml
@@ -10,4 +10,7 @@
           -e @scenarios/centos-9/base.yml
           -e @scenarios/centos-9/edpm_ci.yml
           -e @scenarios/centos-9/zuul_inventory.yml
+          {%- if make_edpm_deploy_env is defined %}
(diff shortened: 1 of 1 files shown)


cjeanner (commented):
Sounds ok - just a small "fear" about using that param as-is, I hope there's no kind of override anywhere in ci-framework itself. We'll know more with the on-going test.

arxcruz (approved):

PR #189, opened 2023-05-23
Allow to pass arbitrary extra vars to playbook

cjeanner:
By setting cifmw_extras variable in zuul job definition, we are now
allowed to pass any arbitrary "--extra-vars" option - being a plain
key/value, or pointer to a file.

This PR has:
- [X] Appropriate testing (molecule, python unit tests)
- [X] Appropriate documentation (README in the role, main README is up-to-date)

PR #190, opened 2023-05-24
pass make_edpm_deploy_env as a dict

raukadah:
https://github.com/openstack-k8s-operators/edpm-ansible/pull/161 sets
make_edpm_deploy_env ansible var as dict. This vars needs to be passed
as dict also to inner ansible during edpm deployment otherwise it
will fail the execution complaining dict and str cannot concatenation.
    
This patch fixes the same.

This PR has:

### DIFF
--- a/ci/playbooks/edpm/run.yml
+++ b/ci/playbooks/edpm/run.yml
@@ -11,6 +11,6 @@
           -e @scenarios/centos-9/edpm_ci.yml
           -e @scenarios/centos-9/zuul_inventory.yml
           {%- if make_edpm_deploy_env is defined %}
-          -e "make_edpm_deploy_env={{ make_edpm_deploy_env }}"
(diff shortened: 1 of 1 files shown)


ci/playbooks/edpm/run.yml
```diff
@@ -11,6 +11,6 @@
           -e @scenarios/centos-9/edpm_ci.yml
           -e @scenarios/centos-9/zuul_inventory.yml
           {%- if make_edpm_deploy_env is defined %}
-          -e "make_edpm_deploy_env={{ make_edpm_deploy_env }}"
+          -e "{'make_edpm_deploy_env':{{ make_edpm_deploy_env | from_yaml }}}"
```
cjeanner, line 14: I'm wondering if we shouldn't use a plain file at this point. Passing some yaml/json may lead to really, really unwanted behavior at this point.

What if the "previous step" is generating a plain file, and we allow to inject it on-demand, for instance as proposed in #189 ?

PR #193, opened 2023-05-25
Consolidate and correct README and doc contents

cjeanner:
In order to get a consistent display through the various role README, we
needed to scrub and correct some content.

It's mostly cosmetics, while also remove a deprecated/never-used
parameter.

This PR has:
- [X] Appropriate documentation (README in the role, main README is up-to-date)

PR #195, opened 2023-05-25
Do not kick any molecule nor e2e job on .md editions

cjeanner:
Until now, even if we were only editing .md files (aka doc), we would
end with at least the end-to-end-nobuild job running and, if we were
editing a role README.md, we'd see its molecule job being kicked.

This patch ensures zuul won't be triggered on .md nor on /docs/
modifications, meaning less load and time on zuul infra.

### DIFF
--- a/zuul.d/end-to-end.yaml
+++ b/zuul.d/end-to-end.yaml
@@ -4,8 +4,11 @@
     name: cifmw-end-to-end
     nodeset: centos-9-crc-xxl
     files:
-      - ^ci_framework/roles/.*_build
-      - ^ci_framework/roles/build.*
+      - ^ci_framework/roles/.*_build/(?!meta|README).*
+      - ^ci_framework/roles/build.*/(?!meta|README).*
+    irrelevant-files:
+      - ^.*/*.md
+      - ^ci/templates
     parent: base-simple-crc
     vars:
       crc_parameters: "--memory 20000 --disk-size 120 --cpus 6"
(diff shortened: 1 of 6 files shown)


zuul.d/end-to-end.yaml
```diff
@@ -6,6 +6,10 @@
     files:
       - ^ci_framework/roles/.*_build
       - ^ci_framework/roles/build.*
+    irrelevant-files:
+      - ^.*/*.md
+      - ^zuul.d
```
pablintino, line 11: Is this intentional, the PR is about skipping doc changes, but this wil skip changes in zuul config, what happens if we change a zuul job variable and the job breaks?

cjeanner: I hesitated a bit on that one... Maybe we can keep it, yeah. Probably safer.

cjeanner:
/hold

While testing this patch with a depends-on from #193 molecule still kicks in with README.md modifications... Need to modify a bit things I guess.

cjeanner:
/unhold

actually, it's working as it should - since this patch edits zuul.d/molecule-base.yaml, it kicks all of the

cjeanner:
/approve

It's been actively tested in #193 - and we can already see the benefits in that patch: far, far less jobs are being kicked. We may need to iterate a bit more but, as-is, this patch is helping a lot already.

cjeanner:
recheck

raukadah (approved):
/approve

PR #197, opened 2023-05-25
Refactor documentation and add new content

cjeanner:
This patch changes a bit the doc structure in order to make it more
convenient for both accessing data, and maintaining content.

We also expose an FAQ, allowing people to consume the framework without
being blocked by known problem we can't easily solve.

We also take the opportunity to clean the README as well as USAGE, so
that everything is only in one location.

This patch also exposes the README listing the various ansible plugins
we have, in order to make them easier to understand.


raukadah (approved):
/lgtm

raukadah:
The docs looks nice https://logserver.rdoproject.org/97/197/b649c32e34f3d53bc945b6fd59b9d833983efc8a/github-check/cifmw-doc/0b877b2/doc_build/html/ Thank you @cjeanner and team for making it happen.

PR #198, opened 2023-05-25
Add a skip flag for compute_repos task

viroel:
Some deployments/distros are not suported by repo-setup, so we need a way to skip it.

This PR has:

### DIFF
--- a/ci_framework/roles/libvirt_manager/README.md
+++ b/ci_framework/roles/libvirt_manager/README.md
@@ -23,3 +23,4 @@ Used for checking if:
 * `cifmw_libvirt_manager_installyamls`: (String) install_yamls repository location. Defaults to `cifmw_installyamls_repos` which defaults to `../..`
 * `cifmw_libvirt_manager_dryrun`: (Boolean) Toggle ci_make `dry_run` parameter. Defaults to `false`.
 * `cifmw_libvirt_manager_compute_amount`: (Integer) State the amount of computes you want. Defaults to `1`.
+* `cifmw_libvirt_manager_skip_edpm_compute_repos`: (Boolean) Intentionally skips the configuration of repos on EDPM compute nodes. Defaults to `False`.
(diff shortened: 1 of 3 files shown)


viroel:
recheck

raukadah:
@viroel We support[ rhos-release](https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci_framework/roles/repo_setup/tasks/rhos_release.yml) in repo-setup. If there is any specific usecase missing from repo-setup we can work on that. 

This step is tied with edpm_compute creation. We should not make it optional.

viroel:
> @viroel We support[ rhos-release](https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci_framework/roles/repo_setup/tasks/rhos_release.yml) in repo-setup. If there is any specific usecase missing from repo-setup we can work on that.
> 
> This step is tied with edpm_compute creation. We should not make it optional.

Yes, I know that we have that in the role. But this code doesn't call this code, this code is using install_yamls target, which has limitation on what we can do there. See [1]. There is also the problem that we can't provide any vars to this task at this moment, so there is no way (for now) unless I skip this.

[1] https://github.com/openstack-k8s-operators/install_yamls/blob/main/devsetup/scripts/edpm-compute-repos.sh

raukadah:
> > @viroel We support[ rhos-release](https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci_framework/roles/repo_setup/tasks/rhos_release.yml) in repo-setup. If there is any specific usecase missing from repo-setup we can work on that.
> > This step is tied with edpm_compute creation. We should not make it optional.
> 
> Yes, I know that we have that in the role. But this code doesn't call this code, this code is using install_yamls target, which has limitation on what we can do there. See [1]. There is also the problem that we can't provide any vars to this task at this moment, so there is no way (for now) unless I skip this.
> 
> [1] https://github.com/openstack-k8s-operators/install_yamls/blob/main/devsetup/scripts/edpm-compute-repos.sh

Thank you for clarifying it. Make sense

raukadah (approved):
/approve

PR #200, opened 2023-05-26
Add documentation about molecule tests

cjeanner:
It may happen a molecule job needs CRC - this patch explains how to
consume it.

It also add a doc string in the Makefile in order to expose the
"role_molecule" target whenever we run `make help`.


raukadah (approved):
/lgtm

PR #201, opened 2023-05-26
Use base-simple-crc as a parent for EDPM jobs

raukadah:
This patch drops base-crc zuul job parent where we used to run
`make openstack` in the config repo in favor of base-simple-crc.
    
base-simple-crc job will start the crc node only in pre-run
step. It will execute the complete framework from scratch by
running the whole steps defined in deploy-edpm.yml playbook to
do the EDPM deployment and validate it.
    
This patch drops the vars related to base-crc and uses all the
default defined like network-isolation from the install_yamls.
It also drops cifmw_operator_build_output var which is no longer
needed.
    
It also adds a pre_infra step to install the required tools.
It includes vars related crc to drive the complete workflow.
    

This PR has:
- [ ] ~~Appropriate documentation (README in the role, main README is up-to-date)~~

### DIFF
--- a/ci_framework/roles/edpm_deploy/tasks/main.yml
+++ b/ci_framework/roles/edpm_deploy/tasks/main.yml
@@ -33,31 +33,43 @@
     - not cifmw_edpm_deploy_dryrun | bool
   block:
     - name: Get info about dataplane node
+      environment:
+        PATH: "{{ cifmw_path }}"
       ansible.builtin.shell: |
-        oc get openstackdataplane
-        oc get openstackdataplanerole
-        oc get openstackdataplanenode
-        oc get pods | grep edpm
+        oc get openstackdataplane -n {{ cifmw_install_yamls_defaults['NAMESPACE'] }}
+        oc get openstackdataplanerole -n {{ cifmw_install_yamls_defaults['NAMESPACE'] }}
+        oc get openstackdataplanenode -n {{ cifmw_install_yamls_defaults['NAMESPACE'] }}
+        oc get pods -n {{ cifmw_install_yamls_defaults['NAMESPACE'] }} | grep edpm
       ignore_errors: true
 
     - name: Get nova-edpm-compute-0-deploy-nova pod name
       register: edpm_pod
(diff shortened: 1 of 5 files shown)


raukadah:
recheck

raukadah:
recheck

rlandy:
recheck

raukadah:
recheck

SeanMooney:
feel free to proceed with this change even though its going to break my nova job.
that is not merged currently so i can adapt to this refactor separately.

raukadah:
recheck

Sandeepyadav93:
/lgtm

as per logs:

https://logserver.rdoproject.org/01/201/07344d854fd77d423e7beab87513e3471891bb4a/github-check/ci-framework-crc-podified-edpm-deployment/ca80feb/job-output.txt

ci_framework/roles/edpm_deploy/tasks/main.yml
```diff
@@ -34,30 +34,34 @@
   block:
     - name: Get info about dataplane node
       ansible.builtin.shell: |
```
cjeanner, line 36: I'm surprised that one works as-is - we should add this just to be 100% safe:
```YAML
- name: Get info about dataplane node
  environment:
    PATH: "{{ cifmw_path }}"
  ansible.builtin.shell: | [...]
```

ci_framework/roles/edpm_prepare/tasks/main.yml
```diff
@@ -29,6 +29,15 @@
     name: 'install_yamls_makes'
     tasks_from: 'make_crc_storage'
 
+- name: Attach default network to CRC
+  vars:
+    make_crc_attach_default_interface_env:
+      PATH: "{{ cifmw_path }}"
```
cjeanner, line 35: I think we don't need to pass down the PATH to the `install_yamls_makes` tasks. IIRC @pablintino ensured it's in the `make` environment

cjeanner (approved):
/lgtm

PR #202, opened 2023-05-26
Move data directory to a more understandable name

cjeanner:
This should avoid issues when users clone the repository in the home
directory directly.

This PR has:

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/204

### DIFF
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -1,7 +1,7 @@
 [defaults]
 action_plugins = ./plugins/action:~/plugins/action:/usr/share/ansible/plugins/action
 library = ./plugins/modules:~/plugins/modules:/usr/share/ansible/plugins/modules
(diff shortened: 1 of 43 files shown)


cjeanner:
recheck

afazekas:
Sounds reasonable.
I case we want to have a separate repo for data files, we need some other name.

ci-framework-workdir, or cifmw-workdir might be a good name too.

cjeanner:
> Sounds reasonable. I case we want to have a separate repo for data files, we need some other name.
> 
> ci-framework-workdir, or cifmw-workdir might be a good name too.

Actually... the data should be in there... So I don't really see why we'd need yet another directory outside of that one? The goal is to be as non-invasive as possible.

cjeanner:
recheck

cjeanner:
recheck

cjeanner:
recheck

raukadah (approved):
/lgtm

PR #218, opened 2023-05-30
Correct loop label

cjeanner:
The "item" is now an absolute path, meaning we're seeing something wrong
in the loop, while the code is actually correct.

This PR has:


cjeanner:
recheck

PR #221, opened 2023-05-30
Adding new openshift_login role

pablintino:
This roles enables the framework to perform OpenShift logins in a consistent way by centralizing all login related actions into it. It based on explicit calls to the `oc` cli tool, that manages authentication underneath based on a given set of role variables. Depending on the variables that are passed to the role the `oc login` command arguments are build allowing multiple scenarios, like:
- Login with user, password and API endpoint
- Login with a kubeconfig file only
- Login with a kubeconfig file but overriden user/password Despite the given login variables used, the role sets at the end of it's execution the API URL, user, context, kubeconfig and token that applies to the whole run, taht way, any consumer in the playbook can use those credentials to login.

  - [x] README in the role

### DIFF
--- a/ci_framework/playbooks/05-build-operators.yml
+++ b/ci_framework/playbooks/05-build-operators.yml
@@ -9,20 +9,6 @@
   environment:
     PATH: "{{ cifmw_path }}"
   tasks:
-    - name: Perform cluster login
-      when:
-        - cifmw_use_crc is defined
-        - cifmw_use_crc | bool
-      block:
-        - name: Perform CRC login
-          ansible.builtin.include_role:
-            name: rhol_crc
-            tasks_from: add_crc_creds.yml
-
-        - name: CRC registry login
-          ansible.builtin.command: |
-            oc registry login --insecure=true
-
     - name: Build operator and meta-operator
       when:
         - cifmw_operator_build_operators is defined
--- a/ci_framework/roles/openshift_login/defaults/main.yml
+++ b/ci_framework/roles/openshift_login/defaults/main.yml
@@ -16,6 +16,12 @@
 
 
 # All variables intended for modification should be placed in this file.
-# All variables within this role should have a prefix of "cifmw_openshift_setup"
-cifmw_openshift_setup_dry_run: false
-cifmw_openshift_setup_create_namespaces: []
+# All variables within this role should have a prefix of "cifmw_openshift_login"
+cifmw_openshift_login_basedir: "{{ cifmw_basedir | default(ansible_user_dir ~ '/ci-framework-data') }}"
+cifmw_openshift_login_kubeconfig_default_path: "{{ cifmw_openshift_login_basedir ~ '/.kube/config' }}"
+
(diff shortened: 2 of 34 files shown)


ci_framework/plugins/action/__pycache__/ci_make.cpython-39.pyc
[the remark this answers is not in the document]
pablintino: Done, seems like a legit one for the .gitignore.

ci_framework/roles/openshift_login/defaults/main.yml
```diff
@@ -0,0 +1,27 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+
+# All variables intended for modification should be placed in this file.
+# All variables within this role should have a prefix of "cifmw_openshift_login"
+cifmw_openshift_login_basedir: "{{ cifmw_basedir | default(ansible_user_dir ~ '/ci-framework') }}"
```
raukadah, line 20: replace ci-framework with ci-framework-data as per https://github.com/openstack-k8s-operators/ci-framework/commit/20a684b12ef8ce4494657cbd50abca1021a6d54e

raukadah (commented):
With this role are we thinking about dropping this task file https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci_framework/roles/rhol_crc/tasks/add_crc_creds.yml ?

ci_framework/roles/openshift_login/defaults/main.yml
pablintino: Nice catch, done.

pablintino:
> With this role are we thinking about dropping this task file https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci_framework/roles/rhol_crc/tasks/add_crc_creds.yml ?

That wasn't my intention, as someone for local-env may want to use it. But, now, that it's not used from other roles, since login is performed in a centralized way in the new role and every role gets proper credentials from the facts that this role sets, there is no need to still using that `crc` role task file, that was highly coupled to `crc`.

ci_framework/playbooks/05-build-operators.yml
cjeanner, line 1: hmmm since you're editing this, we should probably get the full "end-to-end" job in order to ensure this part is still working.

Care to amend the zuul.d/end-to-end.yaml file and add something like this to the "files" list:
`- ^ci_framework/playbooks/.*build.*.yml`

That should kick the build job as soon as we edit the build related playbooks.

cjeanner:
/approve

it's looking fine, still some nits and bits, but overall we're close to a good state. Let's mark it as such.

pablintino:
recheck

pablintino:
recheck

raukadah (approved):
/approve

PR #222, opened 2023-05-31
Replace zuul-worker with zuul

raukadah:
Currently on CS9 crc image, the default user is zuul-worker. In dependent job, when data is passed from one job to another. then ansible_user_dir is already returned to parent job's user. It causes issues on dependent jobs.

The infra has fixed the crc node user to zuul. This patch makes sure we replace the zuul-worker to zuul at all places.

As a pull request owner and reviewers, I checked that:
- [X] Appropriate testing is done and actually running
- [ ] ~~Appropriate documentation exists and/or is up-to-date:~~
  - [ ] ~~README in the role~~
  - [ ] ~~Content of the docs/source is reflecting the changes~~

PR #230, opened 2023-05-31
Ensure we expose our custom PATH to command/shell

cjeanner:
If we don't do this, we may end with failures such as "oc: command no
found", leading to a whole crash of the deploy.


cjeanner:
(in case it fails in zuul, it's more than probably due to https://github.com/openstack-k8s-operators/ci-framework/pull/222 - the Depends-On should allow a `recheck` leading to a successful merge, without rebasing)

pablintino:
recheck

PR #238, opened 2023-06-02
add collection of Openstack CR instances.

SeanMooney:
This change extends the artifact collection to
include the output of get and describe on all openstack
CRDs.

This will allow the collection of both the rendered CRs
and the status of them to allow failures to be debugged.

Signed-off-by: Sean Mooney <work@seanmooney.info>

### DIFF
--- a/ci_framework/roles/artifacts/tasks/cluster_info.yml
+++ b/ci_framework/roles/artifacts/tasks/cluster_info.yml
@@ -1,4 +1,8 @@
 ---
+# Note if any of the tasks fail, we will still try to collect the data
+# from the cluster. This is because we want to collect as much data as
+# possible, even if the cluster is not in a good state so we use
+# ignore_errors: true.
 - name: Check for oc command
   ansible.builtin.command: which oc
   environment:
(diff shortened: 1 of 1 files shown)


ci_framework/roles/artifacts/tasks/cluster_info.yml
```diff
@@ -30,13 +30,35 @@
...
+      args:
+        chdir: "{{ cifmw_artifacts_basedir }}/logs/edpm"
+      changed_when: true
+      ignore_errors: true
+
+    - name: Dump all openstack CRs
+      ansible.builtin.shell: |
+        mkdir crs
+        pushd crs
+          all_crds=$(oc get crd | grep openstack | awk '{print $1}' | awk -F '.' '{print $1}')
```
cjeanner, line 55: We can remove the `| grep openstack`:
`oc get crd | awk '/openstack/ { print $1}' | awk -F '.' '{print $1}'`
I'm pretty sure we may even get a one-call awk, but I'm not fluent enough. Reducing the amount of pipe is always a good thing :).

ci_framework/roles/artifacts/tasks/cluster_info.yml
```diff
@@ -30,13 +30,35 @@
...
+          for cr in ${all_crds}; do
+            echo ${crd}
+            oc get -o yaml ${cr} > ${cr}.yaml
+            oc describe ${cr} > ${cr}-describe.txt
+          done
+        popd
       args:
         chdir: "{{ cifmw_artifacts_basedir }}/logs/edpm"
       changed_when: true
```
cjeanner, line 64: why don't you have the `ignore_errors: true` here? Just wanting to ensure we're consistent and doing the right thing :).

ci_framework/roles/artifacts/tasks/cluster_info.yml
SeanMooney: ya i was debating about removing the grep and doing  awk '/openstack/ { print $1}' but i also didn't know how to collapse the two awks so i gave up on both.

ill fix this shortly

ci_framework/roles/artifacts/tasks/cluster_info.yml
SeanMooney: actully looking locally  i do it just did not show the last line in the file by default and you have to click...

have i said that github sucks for code review lately :)

ci_framework/roles/artifacts/tasks/cluster_info.yml
```diff
@@ -30,13 +30,35 @@
...
+          oc get -o yaml ${pod} > ${pod}.yaml
+          oc describe ${pod} > ${pod}-describe.txt
         done
+        # so we rename the directory to "pods" to consistently
+        # use plural for directories.
+        mv pod pods
+      args:
+        chdir: "{{ cifmw_artifacts_basedir }}/logs/edpm"
+      changed_when: true
+      ignore_errors: true
```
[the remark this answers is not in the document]
cjeanner: fair enough

ci_framework/roles/artifacts/tasks/cluster_info.yml
cjeanner: dang....

cjeanner (approved):
/approve

cjeanner:
Improving logging is needed. With Chandan's patch (#232) we should now get a far better overview of the state.

PR #240, opened 2023-06-02
Installing oc  as part of ci_setup

psathyan:
OpenShift client is installed as part of ci_setup role. This is to enable `hive` role to leverage the binary for its task.

  - [x] README in the role

### DIFF
--- a/ci_framework/roles/ci_setup/defaults/main.yml
+++ b/ci_framework/roles/ci_setup/defaults/main.yml
@@ -17,4 +17,6 @@
 
 # All variables intended for modification should be placed in this file.
 # All variables within this role should have a prefix of "cifmw_ci_setup"
+
 cifmw_ci_setup_basedir: "{{ cifmw_basedir | default(ansible_user_dir ~ '/ci-framework-data') }}"
+cifmw_ci_setup_openshift_client_version: "stable"
--- a/ci_framework/roles/ci_setup/molecule/default/converge.yml
+++ b/ci_framework/roles/ci_setup/molecule/default/converge.yml
@@ -14,10 +14,10 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
-
 - name: Converge
   hosts: all
   vars:
+    cifmw_path: "{{ ansible_user_dir }}/.crc/bin:{{ ansible_user_dir }}/.crc/bin/oc:{{ ansible_user_dir }}/bin:{{ ansible_env.PATH }}"
     cifmw_install_yamls_defaults:
       NAMESPACE: openstack
       DATAPLANE_RUNNER_IMG: foo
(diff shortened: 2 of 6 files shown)


ci_framework/roles/ci_setup/tasks/packages.yml
```diff
@@ -8,3 +8,40 @@
...
+      ansible.builtin.command:
+        cmd: "oc version -o yaml"
+
+    - name: Check if version of openshift client is supported
+      vars:
+        oc_version_data: "{{ oc_version.stdout | from_yaml }}"
+        oc_major_version: "{{ oc_version_data.releaseClientVersion | split('.') | first }}"
+      ansible.builtin.assert:
+        that:
+          - oc_major_version | int >= 4
```
cjeanner, line 33: you probably want to consume `cifmw_ci_setup_openshift_minimum_version` parameter here.. ?

ci_framework/roles/ci_setup/tasks/packages.yml
```diff
@@ -8,3 +8,40 @@
...
+  rescue:
+    - name: Ensure openshift client install path is present
```
pablintino, line 35: This task is not needed as it was already created, if needed, [here](https://github.com/openstack-k8s-operators/ci-framework/pull/240/commits/456ecfcf8275ede2f507a82e1ef028c231373802#diff-bca68ddbe5009a351475b8cf4f9aed320d165c7be9403105689b3fad5832e96aR12)

ci_framework/roles/ci_setup/defaults/main.yml
```diff
@@ -17,4 +17,6 @@
 
 # All variables intended for modification should be placed in this file.
 # All variables within this role should have a prefix of "cifmw_ci_setup"
+
 cifmw_ci_setup_basedir: "{{ cifmw_basedir | default(ansible_user_dir ~ '/ci-framework-data') }}"
+cifmw_ci_setup_openshift_client_version: "4.13.0"
```
pablintino, line 22: Are we gonna stick to an specific 4.13.X version? Or maybe is a better idea to stick to the latest, stable, 4.13.Y? Seems like the mirror URL allows downloading the latest stable version for each version. This is not a must, I'm just asking if it's a better, as that will, while guaranteeing compatibility, have all the bugfixes of the cli tool.

ci_framework/roles/ci_setup/tasks/packages.yml
psathyan: That was the purpose of the variable.

Addressed in the next patch set.

ci_framework/roles/ci_setup/defaults/main.yml
psathyan: Good point. Let me switch to using stable instead of hard coding the version. We can switch to a version specific if we encounter issues later.

ci_framework/roles/ci_setup/molecule/default/converge.yml
```diff
@@ -26,32 +26,70 @@
...
+            cmd: "oc version -o yaml"
+
+        - name: Debug - oc version output
+          ansible.builtin.debug:
+            msg: "{{ oc_ver.stdout }}"
+
+        - name: Compare the expected version
+          vars:
+            oc_ver_dict: "{{ oc_ver.stdout | from_yaml }}"
+            oc_min_ver: "{{ oc_ver_dict.releaseClientVersion | split('.') | first }}"
```
cjeanner, line 74: that'd be `oc_maj_ver` I think, with the `|first` ?

cjeanner (commented):
/approve

a potential last nit, but I won't block on that one :)

ci_framework/roles/ci_setup/molecule/default/converge.yml
psathyan: yes... you are correct. Fixing it :)

PR #245, opened 2023-06-05
Remove crc bits from cifmw-base-crc-openstack and rename it

raukadah:
cifmw-base-crc-openstack uses base-simple-crc as a parent
and contains ci-framework and crc related jobs vars.
    
 It does not give a clear seperation between crc and ci-framework.
 This patch breaks the cifmw-base-crc-openstack into two jobs
 1. cifmw-base-crc containing the crc bits
 2. cifmw-base-openstack containing the ci-framework bits with
       pre-run and post-run playbook calls and vars
    
It also keeps the existing cifmw-base-crc-openstack job to avoid
breakage if someone is using it.
    
cifmw-base-openstack base job will provide the complete environment
 to use the ci-framework starting from pre-setup to log collection.
    
The run playbook needs to be called in the main job as it does not
gets overriden when main job is used as a parent.

### DIFF
--- a/zuul.d/edpm.yaml
+++ b/zuul.d/edpm.yaml
@@ -1,11 +1,22 @@
 ---
-# Base job for openstack crc based job
+# Base job for crc based job
 - job:
-    name: cifmw-base-crc-openstack
+    name: cifmw-base-crc
     nodeset: centos-9-crc-3xl
     timeout: 10800
     abstract: true
     parent: base-simple-crc
+    vars:
+      crc_parameters: "--memory 21000 --disk-size 120 --cpus 8"
+      pre_pull_images:
+        - registry.redhat.io/rhosp-rhel9/openstack-rabbitmq:17.0
+
+# Base job for openstack based job containing ci-framework bits
+- job:
+    name: cifmw-base-crc-openstack
+    parent: cifmw-base-crc
+    timeout: 10800
+    abstract: true
     irrelevant-files:
       - .*/*.md
     required-projects:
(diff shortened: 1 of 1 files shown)


zuul.d/edpm.yaml
```diff
@@ -1,11 +1,22 @@
...
     abstract: true
     parent: base-simple-crc
+    vars:
+      crc_parameters: "--memory 21000 --disk-size 120 --cpus 8"
+      pre_pull_images:
+        - registry.redhat.io/rhosp-rhel9/openstack-rabbitmq:17.0
+
+# Base job for openstack based job containing ci-framework bits
+- job:
+    name: cifmw-base-openstack
```
viroel, line 16: I think it's worth keeping the job name as `cifmw-base-crc-openstack` since its parent is setting up a crc environmet. I see that you are keeping the old name for compatibility, but as personal opinion, looks like the old name still represents it better.

viroel (commented):
Thanks for splitting the jobs Chandan.

zuul.d/edpm.yaml
cjeanner: Same thoughts - seems to be corrected, thanks @raukadah !

cjeanner:
/approve

Thanks for splitting!

zuul.d/edpm.yaml
```diff
@@ -17,40 +28,29 @@
...
-      - ci/playbooks/e2e-prepare.yml
-      - ci/playbooks/dump_zuul_vars.yml
-      - ci/playbooks/edpm/run.yml
-    post-run:
-      - ci/playbooks/e2e-collect-logs.yml
-      - ci/playbooks/collect-logs.yml
+    run: ci/playbooks/edpm/run.yml
     vars:
       # This should be set here since is associated with parent job (base-crc) and not the scenario
       cifmw_openshift_kubeconfig: "/home/zuul/.crc/machines/crc/kubeconfig"
```
Sandeepyadav93, line 47: From reading comments, it looks like this was needed for the base-crc parent. Do we still need this?

raukadah: Thank you @Sandeepyadav93 for highlighting. It is no longer required here, I have dropped it.

Sandeepyadav93: thanks for the update.

cjeanner:
/lgtm

no one stepped against it, let's get it in.

PR #246, opened 2023-06-05
Consolidate a bit ansible code

cjeanner:
This patch adds a couple of tags to the code (work isn't finished yet).
It also removes some useless files (mostly `cleanup.yml` that weren't
modified), and corrects some ansible code, mostly missing FQN (i.e.
missing `ansible.builtin` prefix for the modules).

There's also some reordering of parameters.

Basic functions shouldn't be affected by this change - but it's opening
a door for better usage, allowing to run multiple times the same command
and using `--tags` and `--skip-tags` parameters. This should allow a
faster run, as well as a "stage" oriented usage (stage-1: install
packages; stage-2: bootstrap basics; ...).


raukadah (approved):
Thank you @cjeanner , It's a nice addition to the ci-framework. Looks good.

PR #249, opened 2023-06-05
Create placeholder for cookbooks

cjeanner:
Small but efficient receipts are needed and will find place in this
location.

Let's advertise it already, so that users will know we're working on
this kind of content.

### DIFF
--- a/docs/source/Cookbooks/01_comingsoon.md
+++ b/docs/source/Cookbooks/01_comingsoon.md
@@ -1,0 +1,6 @@
+# Coming soon
+
+We're working hard in order to provide small but efficient receipts for a daily
+usage.
+
+Stay tuned!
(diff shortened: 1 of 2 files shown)


docs/source/Cookbooks/01_comingsoon.md
```diff
@@ -0,0 +1,6 @@
+# Coming soon
```
psathyan, line 1: nit: `Cookbooks` we could go with lower cases only for folder names

psathyan (commented):
lgtm

docs/source/Cookbooks/01_comingsoon.md
cjeanner: more as a follow-up, all the directories have a capital letter. Doesn't really change anything, but if we go with all small-caps, we need to update the script creating new roles.

raukadah (approved):
/lgtm

PR #250, opened 2023-06-05
Pass proper namespace while collecting pods info

raukadah:
Currently we do not pass namespace while collecting pods data. It generates empty pod files. Based on controller-manager name in pod, set the proper namespace and then collect the logs.

- [ ] ~~Appropriate documentation exists and/or is up-to-date:~~
  - [ ] ~~README in the role~~
  - [ ] ~~Content of the docs/source is reflecting the changes~~

### DIFF
--- a/ci_framework/roles/artifacts/tasks/cluster_info.yml
+++ b/ci_framework/roles/artifacts/tasks/cluster_info.yml
@@ -24,7 +24,6 @@
       ansible.builtin.shell: |
         oc get {{ item }} > {{ cifmw_artifacts_basedir }}/logs/edpm/{{ item }}.log
       loop:
-        - pods
         - secrets
         - pv
         - events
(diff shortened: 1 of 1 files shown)


ci_framework/roles/artifacts/tasks/cluster_info.yml
```diff
@@ -46,12 +46,18 @@
         # prefixed with "pod/" so that the redirect places the logs
         # in the correct pod directory. That is why we do not cd after
         # creating the directory.
+        search_word="controller-manager"
```
queria, line 49: maybe it would be better to do double loop - as all_pods is generated just above by listing pods in namespaces?

e.g.:
  for ns in openstack-operators openstack; do
     for pod in $(oc get pods ... -n $ns); do
       ....
     done
  done
  # or alternatively 'for ns in $(oc get namespaces -o name)' though it prefixes them with "namespace/" too, and not sure if we would want really all openshift ones recorded here)

psathyan (approved):
lgtm

raukadah:
It is not working as expected.

raukadah:
Fixed here: https://github.com/openstack-k8s-operators/ci-framework/pull/252

PR #253, opened 2023-06-06
Actually enable tags in ansible

cjeanner:
Tags aren't passed down "as-is" when we use `include_role` or
`include_tasks`. This is due to the fact `include_*` dynamically load
the target on runtime.

Changing most of the `include` by `import` helps ensuring we get proper
tags support in the project.
It also should improve the performances, since ansible is able to load
most of its content during the "boot" of the run.

Usually, we should rely on `import` and keep the `include` only for
dynamic loading, such as "load that tasks file in a loop".


raukadah (approved):
Thank for highlighting that import vs include in the commit message. We can document in the README somewhere as best practices somewhere.

arxcruz (approved):

PR #263, opened 2023-06-07
Use cifmw_openshift_setup_registry_tls_enable var to disable tls

raukadah:
openshift-image-registry is an internal registry. We need to use
`tls-verify=false` while logging to internal registry otherwise it
will give
```
pinging container registry default-route-openshift-image-registry.apps-crc.testing:
received unexpected HTTP status: 503 Service Unavailable
```
    
This patch add cifmw_openshift_setup_registry_tls_enable to set the
value of tls-verify during podman login to internal registry.
    
The same is taken from here[1].
    
[1]. https://access.redhat.com/documentation/en-us/red_hat_codeready_containers/1.21/html/getting_started_guide/using-codeready-containers_gsg#accessing-the-internal-openshift-registry_gsg
    

### DIFF
--- a/ci_framework/roles/openshift_setup/README.md
+++ b/ci_framework/roles/openshift_setup/README.md
@@ -8,3 +8,4 @@ No privilege escalation needed.
 ## Parameters
 * `cifmw_openshift_setup_dry_run`: (Boolean) Skips resources creation. Defaults to `false`.
 * `cifmw_openshift_setup_create_namespaces`: (Strings) Namespaces to create beforehand. Defaults to `[]`.
+* `cifmw_openshift_setup_registry_tls_enable`: (Boolean) Enable tls_verify with openshift registy login. Defaults to `false`.
--- a/ci_framework/roles/openshift_setup/tasks/main.yml
+++ b/ci_framework/roles/openshift_setup/tasks/main.yml
@@ -61,4 +61,5 @@
         podman login
         -u {{ cifmw_openshift_user }}
         -p {{ cifmw_openshift_token }}
+        --tls-verify= {{ cifmw_openshift_setup_registry_tls_enable | bool }}
         {{ cifmw_openshift_setup_registry_default_route.resources[0].spec.host }}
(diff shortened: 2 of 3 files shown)


ci_framework/roles/openshift_setup/tasks/main.yml
```diff
@@ -61,4 +61,5 @@
         podman login
         -u {{ cifmw_openshift_user }}
         -p {{ cifmw_openshift_token }}
+        --tls-verify=false
```
cjeanner, line 64: if *really*, please don't hard-code it, but add a new parameter. I'm still convince we shouldn't need it - so adding a parameter allows an easy toggle if it blocks anything in the CI.

raukadah: Done

ci_framework/roles/openshift_setup/README.md
```diff
@@ -8,3 +8,4 @@ No privilege escalation needed.
 ## Parameters
 * `cifmw_openshift_setup_dry_run`: (Boolean) Skips resources creation. Defaults to `false`.
 * `cifmw_openshift_setup_create_namespaces`: (Strings) Namespaces to create beforehand. Defaults to `[]`.
+* `cifmw_openshift_setup_registry_tls_enable`: (Boolean) Enable tls_verify with openshift registy login. Defaults to `false`.
```
cjeanner, line 11: why defaulting to false? we should really, by default, enable security.

Do you happen to have a node on hold with the failure? If it's a crc job, we're probably missing this:
```Bash
oc extract secret/router-ca --keys=tls.crt -n openshift-ingress-operator --confirm
sudo cp -f tls.crt /etc/pki/ca-trust/source/anchors/crc-router-ca.pem
sudo update-ca-trust
```

raukadah: @cjeanner this issue is random, We do not have node hold. I can try to hold the node with above issue.

raukadah:
Based on the discussion, we come up with following solutions:

1. Since [Login into OpenShift internal registry](https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci_framework/roles/openshift_setup/tasks/main.yml#L57) is related to crc internal registry. Internal registry is not used in edpm and baremetal job, we can add a var to skip this task.
2. Due to resources issue, the internal registry might not be up when we run the following task. We can add a wait mechanism to check whether registry is up or not.
3. Or retry the podman login till it successed.

The team is going to propose the same in different prs that's why we are closing this pr.

PR #270, opened 2023-06-07
Receipt: override operator image

cjeanner:

### DIFF
--- a/docs/source/cookbooks/override-operator-images.md
+++ b/docs/source/cookbooks/override-operator-images.md
@@ -1,0 +1,27 @@
+# Override operator images
+
+## Full deploy override
+In order to override images during an actual deploy, you have to:
+
+* clone [openstack-k8s-operators/openstack-operator](https://github.com/openstack-k8s-operators/openstack-operator/)
+* edit [custom-bundle.Dockerfile](https://github.com/openstack-k8s-operators/openstack-operator/blob/main/custom-bundle.Dockerfile) in order to point to the wanted images
+* run `go mod edit` in the [project root](https://github.com/openstack-k8s-operators/openstack-operator/blob/main/go.mod) and update references
+* run `go mod edit` in the [apis directory](https://github.com/openstack-k8s-operators/openstack-operator/blob/main/apis/go.mod#L11) and update references
+
+Then, you'll have to pass down the repository path to the ci-framework, rebuild it, and deploy:
+```YAML
+cifmw_operator_build_meta_src: /path/to/openstack-operator
+cifmw_operator_build_meta_build: true
+cifmw_operator_build_push_registry: "default-route-openshift-image-registry.apps-crc.testing"
+cifmw_operator_build_push_org: "openstack-operators"
+```
+
+### Example
+You want to update the keystone operator, and point to a specific version.
+
+* You edit the [custom-bundle.Dockerfile on line 3](https://github.com/openstack-k8s-operators/openstack-operator/blob/015b6a5293b76cbb2e1e14553168e501477314ef/custom-bundle.Dockerfile#L3)
+* You `go mod edit` [line 16](https://github.com/openstack-k8s-operators/openstack-operator/blob/015b6a5293b76cbb2e1e14553168e501477314ef/go.mod#L16)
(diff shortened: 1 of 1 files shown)


docs/source/cookbooks/override-operator-images.md
```diff
@@ -0,0 +1,25 @@
+# Override operator images
```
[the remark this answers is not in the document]
cjeanner: hmm right, yes. I'll have a look and improve that PR. Lemme pass it as draft for now.

docs/source/cookbooks/override-operator-images.md
```diff
@@ -0,0 +1,27 @@
...
+* clone [openstack-k8s-operators/openstack-operator](https://github.com/openstack-k8s-operators/openstack-operator/)
+* edit [custom-bundle.Dockerfile](https://github.com/openstack-k8s-operators/openstack-operator/blob/main/custom-bundle.Dockerfile) in order to point to the wanted images
+* run `go mod edit` in the [project root](https://github.com/openstack-k8s-operators/openstack-operator/blob/main/go.mod) and update references
+* run `go mod edit` in the [apis directory](https://github.com/openstack-k8s-operators/openstack-operator/blob/main/apis/go.mod#L11) and update references
+
+Then, you'll have to pass down the repository path to the ci-framework, rebuild it, and deploy:
+```YAML
+cifmw_operator_build_meta_src: /path/to/openstack-operator
+cifmw_operator_build_meta_build: true
+cifmw_operator_build_push_registry: quay.io
```
pablintino, line 15: Maybe is worthy to add a note here about the internal OpenShift regsitry, for those that doesn't have an external registry available.

cjeanner: right - updated.

docs/source/cookbooks/override-operator-images.md
```diff
@@ -0,0 +1,27 @@
...
+Then, you'll have to pass down the repository path to the ci-framework, rebuild it, and deploy:
+```YAML
+cifmw_operator_build_meta_src: /path/to/openstack-operator
+cifmw_operator_build_meta_build: true
+cifmw_operator_build_push_registry: "default-route-openshift-image-registry.apps-crc.testing"
+cifmw_operator_build_push_org: "openstack-operators"
+```
+
+### Example
+You want to update the keystone operator, and point to a specific version.
```
afazekas, line 20: At openstack-operators build time currently expected the repo in question is network downloadable.
So local git repos will not be picked up, local only change is not visible.

This method is limited to things which aviailable over network.
 (git can clone from local dir, but if the custom (not meta operator) container dir is not mapped AFAIK)

cjeanner: @viroel care to have a look?

docs/source/cookbooks/override-operator-images.md
```diff
@@ -0,0 +1,27 @@
+# Override operator images
```
Sandeepyadav93, line 1: @cjeanner I haven't tested these steps locally but logically the steps look correct.

I would like to add that we have a few possible scenarios for image override:-

1) The user has already built an OpenStack image that they want to override.

2) The user has already built the individual operator image and only wants to rebuild the OpenStack operator including their individual operator changes/image.

3) When the user has made changes for the individual operator (locally) and wants to rebuild both the individual and meta operators' image with their changes.

We are covering scenario 2) with this PR but we need some information about 1) and 3), which can be added in a follow-up (I don't want to block this PR)

docs/source/cookbooks/override-operator-images.md
viroel: I think we need to clarify here:
 - user must build and push its own operator images to a registry first
 - user must provide a github api reference in go.mod (I am not sure if path to local changes will work here, needs to be tested)

cjeanner:
I'll let people with the right knowledge take over this PR - imho that's the best :). I'm obviously missing some pieces, having either @Sandeepyadav93 or @viroel take over the content would be for the better.

cjeanner:
/hold

so that it won't merge before corrections - and it should allow anyone to unhold it once ready.

cjeanner:
Not relevant anymore

PR #271, opened 2023-06-08
Make sure git package is installed before clonning repo-setup

raukadah:
We need to make sure git package is installed before clonning repo-setup otherwise module will fail with following errors:
```
Failed to find required executable "git" in paths: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```


ci_framework/roles/repo_setup/tasks/install.yml
```diff
@@ -7,6 +7,12 @@
     - tmp
     - artifacts/repositories
 
+- name: Make sure git package is installed
```
cjeanner, line 10: please make sure to add
```YAML
tags:
  - packages
  - bootstrap
```
so that it will be picked up during those runs. One may argue "this should be part of the ci_setup role" - I'm not sure whether any of the other roles are consuming git at some point?

ci_framework/roles/test_deps/tasks/main.yml
```diff
@@ -71,6 +71,7 @@
     cifmw_repo_setup_output: "/etc/yum.repos.d"
     cifmw_repo_setup_branch: "master"
     cifmw_repo_setup_promotion: "current-podified"
+    cifmw_repo_setup_os_release: "centos"
```
cjeanner, line 74: I have the feeling this one is either not part of this PR, or there's some missing details... ?

cjeanner (changes requested):

ci_framework/roles/test_deps/tasks/main.yml
cjeanner: OK - it's in the actual commit message, not in the first comment:
```
Note: Set cifmw_repo_setup_os_release to centos for ubi
```

ci_framework/roles/repo_setup/tasks/install.yml
raukadah: Add tags. For timebeing, I am keeping it here. as we are doing git operation in repo_setup role.

ci_framework/roles/test_deps/tasks/main.yml
raukadah: Yes, Updated in commit message also.

ci_framework/roles/repo_setup/tasks/install.yml
cjeanner: :+1:

raukadah:
recheck

cjeanner (approved):
/approve
/lgtm

PR #272, opened 2023-06-08
Always set ansible_user_dir var when undefined

raukadah:
On using 01-bootstrap.yml playbook seperatly, user does not pass ansible_user_dir. It fails the playbook with undefined var.

ansible_user_dir var is used in almost all the var. We need to make sure we set the same.

  - [ ] README in the role


queria:
It seems weird to set ansible_ prefixed variables, it seems it could be sometimes missing - but in special cases, not with inventory and fact gathering imo.

Especially it seems this might be some quirk in ansible, as after fact_gathering this is indeed known fact - even for localhost!

Without inventory (so just localhost builtin), with play like:
```---
- hosts: localhost
  gather_facts: yes
  tasks:
    #- debug:
    #    var: hostvars['localhost']
    - debug:
        msg: "hello {{ inventory_hostname }} - {{ ansible_user_home | default('__null__') }}"
    - debug:
        msg: "{{ hostvars[inventory_hostname]['ansible_user_dir'] }}"
```
it results in:
```TASK [debug] ******************************************************************************************************************************************************************************************************************************
ok: [localhost] => {
    "msg": "hello localhost - __null__"
}

TASK [debug] ******************************************************************************************************************************************************************************************************************************
ok: [localhost] => {
    "msg": "/var/home/psedlak"
}```

showing that direct variable 'ansible_user_home' is not visible, but it is visible under host specific variables/facts!

So we may need this as workaround i guess?

File: ci_framework/playbooks/01-bootstrap.yml [hunk 1]
```diff
@@ -2,6 +2,15 @@
   hosts: "{{ cifmw_target_host | default('localhost') }}"
   gather_facts: true
   tasks:
+    - name: Set custom ansible_user_dir reusable fact
+      tags:
+        - always
+      when:
+        - ansible_user_dir is not defined
+      ansible.builtin.set_fact:
+        ansible_user_dir: "{{ lookup('env', 'HOME') }}"
```
elfiesmelfie, line 11: lookup('env', 'XXX') looks up things on the deployment host [1].

If you want to get $HOME on the target host, you need to use "{{ ansible_env.HOME }}" 

This is a non issue when using the default cifmw_target_host since it is localhost, but it will become an issue when a different value is set. 

[1] https://docs.ansible.com/ansible/latest/collections/ansible/builtin/env_lookup.html
[2] https://stackoverflow.com/questions/57207045/ansible-lookup-not-working-for-enviroment-variables

elfiesmelfie (changes requested):

ci_framework/playbooks/01-bootstrap.yml
cjeanner: Indeed. It would work only thanks to the "hosts: localhost" usage we currently have. But since we're planing on consuming remote host, we'll hit this kind of issue sooner or later :(. `lookup` is local to the controller only, unfortunately.

Another way would be to ensure we gather minimal facts if this variable isn't set. Probably better imho. WDYT?

cjeanner (changes requested):

ci_framework/playbooks/01-bootstrap.yml
```diff
@@ -2,6 +2,14 @@
   hosts: "{{ cifmw_target_host | default('localhost') }}"
   gather_facts: true
   tasks:
+    - name: Gather minimum facts
```
cjeanner, line 5: adding this would ensure it's running in any conditions:
```YAML
tags:
  - always
```

cjeanner (commented):
tiny "nit", but it should really do the trick.

MAYBE we should consider adding this to all of the playbooks, in order to avoid weird situation?

Still - with the `gather_facts: true` set on #L3 I don't really understand why we don't get that parameter :/.

raukadah:
> tiny "nit", but it should really do the trick.
> 

I hit this issue. When I was trying to call the[ bootstrap playbook](https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci_framework/playbooks/01-bootstrap.yml) directly via zuul. 

> MAYBE we should consider adding this to all of the playbooks, in order to avoid weird situation?
> 
In other playbooks, we are calling multiple plays and hooks. Not sure it is a good idea.

> Still - with the `gather_facts: true` set on #L3 I don't really understand why we don't get that parameter :/.
No idea.

cjeanner:
That patch would be useless actually:
with `gather_facts: true` we indeed get the needed `ansible_user_dir` set. Tested, and it indeed works. It also work if we set it to `false` and run the `ansible.builtin.setup` with `gather_subset: ['!all', 'min']` without any issue.

After some more digging, it seems the 01-bootstrap.yml is called from a playbook, itself called as a plain `ansible-playbook` command:
https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci/playbooks/content_provider/content_provider.yml#L2-L3

The exposed error isn't really precise as to WHERE the parameter is missing:
```
2023-06-12 04:22:25.900687 | controller | ERROR! {{ ansible_user_dir }}/src/github.com/openstack-k8s-operators/ci-framework: 'ansible_user_dir' is undefined. 'ansible_user_dir' is undefined. {{ ansible_user_dir }}/src/github.com/openstack-k8s-operators/ci-framework: 'ansible_user_dir' is undefined. 'ansible_user_dir' is undefined
```

What we can understand here, it's missing from [that call](https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci/playbooks/content_provider/run.yml#L5-L14).

Maybe we should investigate a bit more that specific part. It may also be interesting to check out why [this](https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci/playbooks/e2e-run.yml) is working without issue... While we're consuming the main `deploy-edpm.yml` playbook, we don't set anything weird in there:
```YAML
- name: Bootstrap step
  ansible.builtin.import_playbook: ci_framework/playbooks/01-bootstrap.yml

- name: Import infra entrypoint playbook
  ansible.builtin.import_playbook: ci_framework/playbooks/02-infra.yml
[...]
```

Namely: same code is being consumed. Now, it would be interesting to know a bit more about the specific task that failed. In order to do so, we may want to ensure we're passing the same ansible configuration file consumed by the main run. This can be done like that:
https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci_framework/roles/run_hook/tasks/playbook.yml#L50

That should allow to get the run log in the same ansible.log file, and therefore get all of the needed data.

Care to propose a patch with that, @raukadah, and add the depends-on on the nova job you're trying to get?

raukadah:
It is no longer needed.

PR #273, opened 2023-06-09
Don't build the storage-bundle separately

rabi:
It's not needed after PR[1].

[1]https://github.com/openstack-k8s-operators/openstack-operator/pull/340


Sandeepyadav93:
/lgtm

we need this after: https://github.com/openstack-k8s-operators/openstack-operator/pull/340

Could not find any more reference of storage bundle in my quick grep- looks good to me

slagle (approved):

ci_framework/roles/operator_build/README.md
```diff
@@ -17,7 +17,6 @@ you want to build meta-operator too, so the role can properly replace api refere
 * `cifmw_operator_build_meta_src`: (String) Directory with src code for meta operator. Defaults to `"{{ ansible_user_dir }}/src/github.com/{{ cifmw_operator_build_org }}/{{ cifmw_operator_build_meta_name }}"`
 * `cifmw_operator_build_meta_build`: (Boolean) When set to `true` updates meta-operator's go.mod when build operators and builds meta-operator in the end. Default to `true`.
 * `cifmw_operator_build_meta_image_base`: (String) Name of the service added to be added to meta-operator build. Still limited to a sindle service (and operator). Default to `""`
-* `cifmw_operator_build_extra_bundles`: (List) List of additional bundles to be added in meta operator catalog build. Defaults to `[]`.
```
cjeanner, line 20: Wondering if we shouldn't keep that parameter - @raukadah any thoughts?

raukadah: @cjeanner Thank you for the review, We no longer need this. As extra operator is already added in https://github.com/openstack-k8s-operators/openstack-operator/blob/main/hack/pin-bundle-images.sh#L61

rebtoor (approved):

PR #278, opened 2023-06-10
Use local registry for local bundles in catalog build

rabi:
This would help bundles built with PR from local registry.

### DIFF
--- a/ci_framework/roles/operator_build/tasks/build.yml
+++ b/ci_framework/roles/operator_build/tasks/build.yml
@@ -146,7 +146,7 @@
       IMAGENAMESPACE: "{{ cifmw_operator_build_push_org }}"
       IMAGEREGISTRY: "{{ cifmw_operator_build_push_registry }}"
       IMAGEBASE: "{{ operator.image_base | default('') }}"
-      LOCAL_REGISTRY: "{{ cifmw_operator_build_local_registry | default ('') }}"
+      LOCAL_REGISTRY: "{{ cifmw_operator_build_local_registry | default (0) }}"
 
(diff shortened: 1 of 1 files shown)


rabi:
recheck

rabi:
recheck

raukadah (approved):
/approve
/lgtm

ci_framework/roles/operator_build/tasks/build.yml
```diff
@@ -157,10 +157,6 @@
     params:
       IMG: "{{ operator_img }}"
       BUNDLE_IMG: "{{ operator_img_bundle }}"
-      IMAGENAMESPACE: "{{ cifmw_operator_build_push_org }}"
```
Sandeepyadav93, line 160: we need those image var for bundle and catalog_build both targets only and not bundle_build as openstack-operator bundle does not include other csvs now

ci_framework/roles/operator_build/tasks/build.yml
```diff
@@ -146,7 +146,7 @@
       IMAGENAMESPACE: "{{ cifmw_operator_build_push_org }}"
       IMAGEREGISTRY: "{{ cifmw_operator_build_push_registry }}"
       IMAGEBASE: "{{ operator.image_base | default('') }}"
-      LOCAL_REGISTRY: "{{ cifmw_operator_build_local_registry | default ('') }}"
```
Sandeepyadav93, line 149: looks good

script expect an integer
https://github.com/openstack-k8s-operators/openstack-operator/blob/main/hack/pin-bundle-images.sh#L47

PR #277, opened 2023-06-09
Adds cifmw_extras to kuttl-run playbook

viroel:
This patch adds cifmw_extras to kuttl run playbook to allow passing different scenarios files.
This patch also change the order of extra-vars in ansible-playbook calls. By moving zuul_inventory.yml to the end, it will allow us to overrides vars in zuul job definitions, which can be useful to:
 * avoid too many scenario files when change in the scenario is minimal
 * allow us to testproject a zuul job without changing the original scenario file.

As a pull request owner and reviewers, I checked that:
- [x] Appropriate testing is done and actually running
- [x] Appropriate documentation exists and/or is up-to-date:
  - [x] README in the role
  - [x] Content of the docs/source is reflecting the changes

PR #279, opened 2023-06-12
Run make_crc_attach_default_interface in kuttl tests

raukadah:
In kuttl tests, we call make openstack_deploy_prep which
calls network isolation behind the scenes. If we donot run
make_crc_attach_default_interface in the job, then
nmstate operator installations will fail on crc.
    
In order to fix that, we need to run make_crc_attach_default_interface
always in kuttl job.
    
Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/277

### DIFF
--- a/ci/playbooks/kuttl/e2e-kuttl.yml
+++ b/ci/playbooks/kuttl/e2e-kuttl.yml
@@ -10,6 +10,13 @@
         name: 'install_yamls_makes'
         tasks_from: 'make_download_tools'
 
+    - name: Attach default network to CRC
+      when:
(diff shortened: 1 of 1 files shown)


cescgina:
@raukadah Why is `openstack_deploy_prep` called when running the kuttl tests? As far as I know that target it's not needed for any kuttl test

cescgina:
> @raukadah Why is `openstack_deploy_prep` called when running the kuttl tests? As far as I know that target it's not needed for any kuttl test
/lgtm 
Ah I see, there is a dependence on `openstack_deploy_prep` via `openstack_crds` that's new, so this is needed, sorry for the noise

PR #282, opened 2023-06-12
Module level docs for ci_make

jpodivin:
The ci_make now has appropriate DOCUMENTATION, EXAMPLES and RETURN blocks.

Contents of the RETURN block are derived from documentation of original make and copy modules.

  - [x] README in the role

### DIFF
--- a/ci_framework/plugins/action/ci_make.py
+++ b/ci_framework/plugins/action/ci_make.py
@@ -14,6 +14,125 @@ from ansible.errors import AnsibleActionFail
 from ansible.module_utils import basic
 from ansible.utils.display import Display
 
+
+DOCUMENTATION = r'''
+---
+action: ci_make
+
+short_description:
+
+description:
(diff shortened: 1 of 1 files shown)


ci_framework/plugins/action/ci_make.py
```diff
@@ -16,6 +16,125 @@
...
+
+short_description:
+
+description:
+- This moduel mostly wraps `community.general.make` and `ansible.builtin.copy`.
+- It requires an additional parameter, `output_dir`, in order to output the `make` generated command.
+- It also adds a new optional parameter, `dry_run`,
+- allowing to NOT run `community.general.make` module, but get a file with the passed parameters.
+
+options: Same as `community.general.make` except for output_dir.
```
cjeanner, line 32: `dry_run` is also an addition to the original community.general.make module

cjeanner (approved):

PR #286, opened 2023-06-12
pass ansible_cfg to content provider run playbook

raukadah:

### DIFF
--- a/ci/playbooks/content_provider/run.yml
+++ b/ci/playbooks/content_provider/run.yml
@@ -3,6 +3,8 @@
   gather_facts: true
   tasks:
     - name: Deploy content provider
+      environment:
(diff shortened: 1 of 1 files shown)


ci/playbooks/content_provider/run.yml
```diff
@@ -3,6 +3,8 @@
   gather_facts: true
   tasks:
     - name: Deploy content provider
+      environment:
+        ANSIBLE_CONFIG: "{{ ansible_user_dir }}/src/github.com/openstack-k8s-operators/ci-framework/ansible.cfg"
```
cjeanner, line 7: hmmm - either the one provided by the framework, or the one actually used by the zuul job. IIRC they have a different log file. Might be good to use zuul's instead?

Let's see how it works with that one as a first iteration

PR #287, opened 2023-06-12
Moving documentation build script into separate file

jpodivin:
Moving the script for doc building in docs source directory will allow for cleaner code. The new script also handles one shot virtual environment, cleanup of artifacts from previous build, and errors.

  - [x] README in the role

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,3 @@
-docs/source/_build/
-docs/virtdocs/
+docs/_build/
+docs/_venv/
 **/__pycache__/
--- a/docs/source/build-docs.sh
+++ b/docs/source/build-docs.sh
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
+DOCS_DIR="./docs"
+VENV_DIR=${DOCS_DIR}/_venv
+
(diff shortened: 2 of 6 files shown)


docs/source/build-docs.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
+DOCS_DIR="./docs"
+VENV_DIR="$(mktemp -d)"
```
cjeanner, line 5: frankly... I'd rather be able to get a fixed venv name in order to be able to re-use it. It'd be faster, and would avoid re-builting the venv each time when you're nit-picking on a local change.

jpodivin: I agree in principal. It does take more time, however keeping the venv does have some drawbacks when you are making changes to the build process, as the old venv may interfere with your changes.

That being said it is a personal preference, and if you want a venv with a longer life time I have no problem.

.gitignore
```diff
@@ -1,3 +1,4 @@
-docs/source/_build/
+docs/_build/
+docs/_venv/
 docs/virtdocs/
```
cjeanner, line 3: afaik that was the original venv - we can remove it I guess?

davidsaOpenu (approved):

dsariel (approved):

PR #292, opened 2023-06-13
Fix cifmw_edpm_deploy_baremetal_common_env vars

raukadah:
https://github.com/openstack-k8s-operators/ci-framework/pull/241 drops the legacy global edpm_env var and combines cifmw_install_yamls_environment var with role specific common_env var.

The patch adds the same in baremetal edpm role.


cjeanner (approved):
/approve

We may want to indicate this in the "Contributing" doc. @pablintino care to have a look, since you implemented this. Something like "here's how you can consume parameters in your role/code".

cjeanner:
Forcing merge: we're in a lock situation, with a dependency loop. Let's alleviate it by merging what we can on our side.

PR #293, opened 2023-06-13
Add edpm build image content provider job

bshewale:
  - [x] README in the role


ci_framework/roles/edpm_build_images/molecule/default/prepare.yml
```diff
@@ -0,0 +1,21 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+
+- name: Prepare
+  hosts: all
+  roles:
+    - role: test_deps
```
cjeanner, line 21: if we remove most of the `dry_run` conditions, you may have to push some preparation tasks here - installing dependencies or maybe fetching code, setting some repositories up (might be good to have a look in the other roles, and to the main playbooks, especially the 01-bootstrap.yml for these kind of tasks. usually, adding `ci_setup` role would already help a lot)

cjeanner (changes requested):

PR #295, opened 2023-06-13
Get the pods names with all information

raukadah:
Currently we are listing the pod names in the logs. It does not provide useful information about the status which makes harder for user to debug.

Getting the detailed status about all pods improves the debugging.

### DIFF
--- a/ci_framework/roles/artifacts/tasks/cluster_info.yml
+++ b/ci_framework/roles/artifacts/tasks/cluster_info.yml
@@ -40,8 +40,8 @@
 
     - name: Dump the logs of all pods
       ansible.builtin.shell: |
-        oc get pods -o name -n openstack-operators > operator_pods.txt
-        oc get pods -o name -n openstack > openstack_pods.txt
+        oc get pods -n openstack-operators > operator_pods.txt
+        oc get pods -n openstack > openstack_pods.txt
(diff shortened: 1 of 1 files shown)


pablintino:
What's the purpose of saving the command output is we already have the describe and the yaml of each pod?

raukadah:
> What's the purpose of saving the command output is we already have the describe and the yaml of each pod?

The current change will give the status of all pods for both openstack and operator namespace. From there an user can easily find out specific pod failing with specific status and then user can check for describe and yaml of those pod.

cjeanner (approved):
/lgtm

I get the `oc get pods` acts a bit like `podman ps -a` - showing a nice summary of the container/services. Then, when we see in the "central" log a failed container, we can investigate deeper, using the other generated files.

This may indeed help getting faster on the issues.

jpodivin:
I would suggest we try to refactor these into calls to k8s_info module[0], at least eventually. Inline shell scripts can be difficult to maintain. Although in this case they are, obviously, fairly simple.

[0]https://docs.ansible.com/ansible/latest/collections/kubernetes/core/k8s_info_module.html#ansible-collections-kubernetes-core-k8s-info-module

cjeanner:
/approve

more logs, more details - love it!

PR #296, opened 2023-06-13
Setting defaults for command key of the returned dict

jpodivin:
It appears that the `community.general.make` module may not return the entire expected dictionary in some cases.

This change sets defaults for the value of the 'command' key, in order to prevent errors.

  - [x] README in the role


psathyan (approved):
/lgtm

PR #297, opened 2023-06-13
Negative testing for ci_make

cjeanner:
Under some circustances, community.general.make doesn't return the
"command" key in its dict.
Usually, this happens when the underlying "run_command" is failing,
leading to an early exit of the module, without the generated command.

This patch adds a test in order to ensure such a case is properly caught
in the action plugin, and ensures the generated reproducer script is
correct, i.e. showing the json_dump() of the passed parameters.


jpodivin:
Thanks Cedric. This will make things much firmer going forward.

PR #298, opened 2023-06-13
Remove useless env file

cjeanner:
Especially useless when it's not created due to earlier failure, leading
to a failure to collect logs..


raukadah (approved):
/approve
/lgtm

cjeanner:
recheck

PR #300, opened 2023-06-14
Gather facts in the 99-logs.yml playbook

cjeanner:
Since this specific playbook may be called from another run, we want to
ensure facts are actually gathered.
This should prevent weird cases when the "run" stage in zuul fails to
do anything (or the pre-run is failing), hitting directly the "post-run"
stage.

### DIFF
--- a/ci_framework/playbooks/99-logs.yml
+++ b/ci_framework/playbooks/99-logs.yml
@@ -1,6 +1,6 @@
 - name: Logging playbook
   hosts: "{{ cifmw_target_host | default('localhost') }}"
-  gather_facts: false
+  gather_facts: true
   tasks:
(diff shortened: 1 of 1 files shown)


jpodivin (approved):
Shame we can't just gather a subset. But given the fact that we can't be sure where the necessary variables belong we'll have to get everything.

raukadah (approved):
/approve
/lgtm

cjeanner:
recheck

cjeanner:
Forcing the merge: the error raised in zuul isn't related to the patch, and should actually be fixed by #301. Since we may need the logs in failing jobs, we'd rather ensure we're actually able to get them.

PR #304, opened 2023-06-14
Ensure we're having cifmw_path parameter

cjeanner:
Since this playbook may be running outside of the main deploy-edpm.yml
play, we have to ensure cifmw_path is set.

This is especially true within zuul CI, since this 99-logs.yml is
launched from the post-run stage.

We could consume the generated custom-params.yml file, but since it may
not exist in case of (really) early failure of the "run" stage, we'd
rather statically add the missing parts.

### DIFF
--- a/ci_framework/playbooks/99-logs.yml
+++ b/ci_framework/playbooks/99-logs.yml
@@ -2,6 +2,15 @@
   hosts: "{{ cifmw_target_host | default('localhost') }}"
   gather_facts: false
   tasks:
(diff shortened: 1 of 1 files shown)


pablintino:
OK from my side. I was thinking that this is gonna suffer from the same KUBECONFIG issues that we had inside the main playbook. Honestly, I always thought that executing this separately was a bad idea, a quick fix, for sure, but a bad idea. We will need to apply the same corrections we are gonna apply there.

Apart from that, for me this fix is OK.

/approve

PR #305, opened 2023-06-14
Protect 99-logs and subsequent artifacts role

cjeanner:
It may happen some of the needed pieces are missing from the environment
parameters - especially in zuul jobs, since the 99-logs.yml playbook is
running standalone.

This patch should ensure we won't break the log gathering, especially
when it comes to `oc` calls - where we may need a crafted PATH among
things.


raukadah (approved):
/approve
/lgtm

PR #308, opened 2023-06-15
Grab CRC VM logs from its disk image

cjeanner:
As commented in the files directly, this must be the last action done
against CRC: after those tasks are done, the VM is stopped.


pablintino:
Force merged as this PR is needed to debug other authentication errors and the `cifmw-end-to-end-nobuild-tagged` is broken beacause of another issue.

PR #309, opened 2023-06-15
Support post-config tweaks for CRC

bogdando:
Provide a customizable list of the post-config CRC commands. By default, it replicates the list of tweaks applied in OCP Prow CI: https://github.com/openshift/release/pull/34536

  - [x] README in the role

### DIFF
--- a/ci_framework/roles/rhol_crc/tasks/main.yml
+++ b/ci_framework/roles/rhol_crc/tasks/main.yml
@@ -90,6 +90,9 @@
             - name: Start RHOL/CRC
               ansible.builtin.command: "{{ cifmw_rhol_crc_binary }} start"
               register: cifmw_rhol_crc_cmd_start
+            
+            - name: Apply CRC post-configuration tweaks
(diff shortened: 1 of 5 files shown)


pablintino:
Seems like something that can be done with a hook?

ci_framework/roles/rhol_crc/tasks/main.yml
cjeanner, line 1: there's apparently a trailing whitespace at the end of one of the new lines you added.

cjeanner (changes requested):
Couple of notes.

I'm OK with those configuration being part of the role. Hook wouldn't make sense here imho, even the `post_infra` one. Adding a new parameter is OK - passing down some defaults via a "static" var (compared to the "default") is also OK.

Reading the passed default config, I'm wondering if they wouldn't help alleviating the resource shortage we're hitting every now and then... ?

pablintino:
Closed as replaced by https://github.com/openstack-k8s-operators/ci-framework/pull/327

PR #311, opened 2023-06-15
[WIP] consume wait params from install_yamls

raukadah:
Depends-On: https://github.com/openstack-k8s-operators/install_yamls/pull/326

### DIFF
--- a/ci_framework/roles/edpm_prepare/tasks/main.yml
+++ b/ci_framework/roles/edpm_prepare/tasks/main.yml
@@ -101,6 +101,13 @@
           --namespace={{ cifmw_install_yamls_defaults['OPERATOR_NAMESPACE'] }}
           --for=jsonpath='{.status.phase}'=Complete --timeout=20m
 
+    - name: Wait for openstack operator to be deployed
+      vars:
+        make_openstack_wait_env: "{{ cifmw_install_yamls_environment | combine({'PATH': cifmw_path}) }}"
+      ansible.builtin.include_role:
+        name: 'install_yamls_makes'
+        tasks_from: 'make_openstack_wait'
+
 - name: Install OpenStack service
   vars:
     make_openstack_deploy_env: "{{ cifmw_edpm_prepare_common_env |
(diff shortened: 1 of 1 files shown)


ci_framework/roles/edpm_prepare/tasks/main.yml
```diff
@@ -101,6 +101,11 @@
           --namespace={{ cifmw_install_yamls_defaults['OPERATOR_NAMESPACE'] }}
           --for=jsonpath='{.status.phase}'=Complete --timeout=20m
 
+    - name: Wait for openstack operator to be deployed
```
pablintino, line 104: Can I ask why we do this if already wait on line 94?

ci_framework/roles/edpm_prepare/tasks/main.yml
```diff
@@ -142,3 +147,8 @@
           --namespace={{ cifmw_install_yamls_defaults['NAMESPACE'] }}
           --for=condition=ready
           --timeout=30m
+
+    - name: Wait for openstack services to be up and running
```
pablintino, line 151: Can I ask why we do this if already wait on line 109?

ci_framework/roles/edpm_prepare/tasks/main.yml
raukadah: I check with Fabricio, `make wait` is added for operator specific installation. But these waits`make_openstack_wai` are added for deployment.

raukadah: I check with Fabricio, `make wait` is added for operator specific installation. But these waits`make_openstack_wai` are added for deployment.

raukadah: @pablintino Thank you for reviewing it, the goal is of this utility is to replace this https://github.com/openstack-k8s-operators/ci-framework/pull/178 pr with above new target.

It will help to find out the issues after install plan and to know whether operator installation is successful.

pablintino:
This one can now be closed, the approach of how we handle the deploy of the OpenStackControlplane resource and how we wait has changed a lot. Mainly:
- We are now not using the `openstack_deploy` target and we use the `_prep` one. We need to kustomize the CR before deploying, so we use the base target instead to generate the CR and the install_yamls kustomization. After that we apply our own kustomization (if available) and finally apply the CR.
- The wait condition is now identical, so no difference now.  https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci_framework/roles/edpm_prepare/tasks/main.yml#L165

PR #315, opened 2023-06-16
Temporary disable the "admin_setup" step

cjeanner:


afazekas:
Show must go on

PR #318, opened 2023-06-16
Ensure libguestfs is installed when gathering CRC VM logs

pablintino:


raukadah (approved):
/approve

PR #310, opened 2023-06-15
Retry getting k8s resource

afazekas:
Looks like some cases the requested resource
is not provided at this point of time.

`better say sorry than ask for permission` retry
strategy put here until better solution
comes.

As a pull request owner and reviewers, I checked that:
- [x] Appropriate testing is done and actually running
- [x] Appropriate documentation exists and/or is up-to-date:
  - [x] README in the role
  - [x] Content of the docs/source is reflecting the changes

PR #320, opened 2023-06-19
[PoC - do not merge] os-net-setup: enable and add retries + pre-debug

queria:
Re-enable os-net-setup role,
while adding retry-looping for problematic tasks
as also early debug step to get more insight of what might be missing at this stage of setup.

  - [ ] README in the role

### DIFF
--- a/ci_framework/roles/os_net_setup/tasks/main.yml
+++ b/ci_framework/roles/os_net_setup/tasks/main.yml
@@ -1,3 +1,31 @@
+- name: pre-debug oc get clouds_yaml or keystone commands
+  failed_when: false
+  register: debug_oc_get
+  shell:
+    cmd: |
+      set -x
+
+      eval $(${HOME}/ci-framework-data/bin/crc oc-env)
+      export KUBECONFIG="${HOME}/.crc/machines/crc/kubeconfig"
+
+      ons_clouds_yaml=$(oc get configmap -n openstack openstack-config -o 'jsonpath={.data.clouds\.yaml}')
+      ons_secret=$(oc get secret -n openstack openstack-config-secret -o 'jsonpath={.data.secure\.yaml}')
+
+      KEYSCT=$(oc get keystoneapi keystone -o 'jsonpath={ .spec.secret }')
+      KEYSEL=$(oc get keystoneapi keystone -o 'jsonpath={ .spec.passwordSelectors.admin }')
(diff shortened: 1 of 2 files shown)


ci_framework/roles/os_net_setup/tasks/main.yml
```diff
@@ -1,3 +1,31 @@
...
+    cmd: |
+      set -x
+
+      eval $(${HOME}/ci-framework-data/bin/crc oc-env)
+      export KUBECONFIG="${HOME}/.crc/machines/crc/kubeconfig"
+
+      ons_clouds_yaml=$(oc get configmap -n openstack openstack-config -o 'jsonpath={.data.clouds\.yaml}')
+      ons_secret=$(oc get secret -n openstack openstack-config-secret -o 'jsonpath={.data.secure\.yaml}')
+
+      KEYSCT=$(oc get keystoneapi keystone -o json | jq -r .spec.secret)
```
psathyan, line 14: In the gate, I noticed `jq` was unavailable. You may want to think about  using it. This could help
```
oc get keystoneapi keystone -o jsonpath='{ .spec.secret}'
```

psathyan: Likewise in the below lines... we could opt to use `-o jsonpath` instead. This would avoid the dependency on `jq`.

ci_framework/roles/os_net_setup/tasks/main.yml
```diff
@@ -1,3 +1,31 @@
+- name: pre-debug oc get clouds_yaml or keystone commands
+  failed_when: false
+  register: debug_oc_get
+  shell:
+    cmd: |
+      set -x
+
+      eval $(${HOME}/ci-framework-data/bin/crc oc-env)
+      export KUBECONFIG="${HOME}/.crc/machines/crc/kubeconfig"
```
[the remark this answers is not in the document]
queria: thanks, though atm i do not plan to have this part of code merged, it is meant just to reveal/indicate if there is issue with the actual oc get ... command output or with some other part (ansible variable reference or such)

ci_framework/roles/os_net_setup/tasks/main.yml
queria: thanks good point, yes using jsonpath seems way better, no need for extra command/processing afterwards if we can query the exact detail directly

queria:
abandoning in favor of #310

PR #322, opened 2023-06-20
Add ci_script action plugin

pablintino:

### DIFF
--- a/ci_framework/plugins/action/ci_script.py
+++ b/ci_framework/plugins/action/ci_script.py
@@ -1,0 +1,177 @@
+#!/usr/bin/python
+
+# Copyright Red Hat, Inc.
+# Apache License Version 2.0 (see LICENSE
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+
+DOCUMENTATION = r'''
+---
+action: ci_script
+
+short_description: Runs a given script in string format remotely while providing logging
+
+description:
+    - Runs a string given script remotly while providing logging.
+    - Logs are written to the `output_dir`/../logs folder.
+
+options:
+    output_dir:
+        description: The path in where the script will be copied
+        required: true
+        type: str
+    script:
+        description: The shell script content to be run
+        required: true
+        type: str
+    creates:
+        description: A filename on the remote node, when it already exists, this step will not be run.
+        type: str
+    decrypt:
+        description: This option controls the autodecryption of source files using vault.
+        type: bool
+        default: true
+    executable:
+        description: Name or path of a executable to invoke the script with.
+        type: str
+    removes:
+        description: A filename on the remote node, when it does not exist, this step will not be run.
+        type: str
+    debug:
+        description: If true the script will, in addition, use bash tracing.
+        type: bool
+        default: false
+'''
+
+EXAMPLES = r'''
+- name: Run custom script
+  register: script_output
+  ci_script:
+    output_dir: "/home/zuul/ci-framework-data/artifacts"
+    cmd: |
+      mkdir /home/zuul/test-dir
+      cd /home/zuul/test-dir
+      git clone https://github.com/openstack-k8s-operators/ci-framework.git
+'''
+
+RETURN = r'''
+changed:
+   description: Always true.
+   returned: always
+   type: bool
+failed:
+   description: True if the execution failed.
+   returned: always
+   type: bool
+rc:
+   description: Script return code.
+   returned: always
+   type: int
+stderr:
+    description: stderr output as string.
+    returned: always
+    type: str
+stderr_line:
+    description: stderr output as lines.
+    returned: always
+    type: list[str]
+stdout:
+    description: stdout output as string.
+    returned: always
+    type: str
+stdout_line:
+    description: stdout output as lines.
+    returned: always
+    type: list[str]
+'''
+
+import glob
+import pathlib
+import re
+import uuid
+
+from ansible.plugins.action import ActionBase
+from ansible.errors import AnsibleActionFail
+
+
+TMPL_SCRIPT = '''#!/bin/bash
+set -euo pipefail
+%(opts)s
+exec > >(tee -i %(logpath)s) 2>&1
+%(content)s
+'''
+
+
+class ActionModule(ActionBase):
+
+    def __init__(self, **kwargs):
+        super(ActionModule, self).__init__(**kwargs)
+        self.__script_file_path = pathlib.Path().home().joinpath('ansible', 'tmp').joinpath(uuid.uuid4().hex).absolute()
+        self.__script_file_path.parent.mkdir(parents=True, exist_ok=True)
+
+    @staticmethod
+    def __build_options(task_vars):
+        if task_vars.get('cifmw_debug', False) or task_vars.get('cifmw_ci_script_debug', False):
+            return 'set -x'
+        return ''
+
+    def run(self, tmp=None, task_vars=None):
+
+        super(ActionModule, self).run(tmp, task_vars)
+
+        if 'output_dir' not in self._task.args:
+            raise AnsibleActionFail('output_dir parameter is missing')
+
+        if 'script' not in self._task.args:
+            raise AnsibleActionFail('script parameter is missing')
+
+        # Ensure chdir is not used
+        if 'chdir' in self._task.args:
+            raise AnsibleActionFail('chdir parameter is not supported')
+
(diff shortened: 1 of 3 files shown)


ci_framework/plugins/action/ci_script.py
```diff
@@ -0,0 +1,150 @@
...
+DOCUMENTATION = r'''
+---
+action: ci_script
+
+short_description: Runs a string given script remotly while providing logging
+
+description:
+- Discover latest image matching a specific pattern on remote server.
+
+options:
```
cjeanner, line 19: Since we're consuming `ansible.builtin.script` module under the hood, I'd rather support the module parameter as listed here:
https://docs.ansible.com/ansible/latest/collections/ansible/builtin/script_module.html

Some of them are really nice, such as the `creates` and `removes`.
Maybe the `chdir` is better "masked", since it should be overridden within this action plugin anyway.

ci_framework/plugins/action/ci_script.py
```diff
@@ -0,0 +1,150 @@
...
+  ci_script:
+    output_dir: "/home/zuul/ci-framework-data/artifacts"
+    cmd: |
+      mkdir /home/zuul/test-dir
+      cd /home/zuul/test-dir
+      git clone https://github.com/openstack-k8s-operators/ci-framework.git
+'''
+
+RETURN = r'''
+changed:
```
cjeanner, line 46: if the `creates` and `removes` are supported, this may also be "false".

ci_framework/plugins/action/ci_script.py
```diff
@@ -0,0 +1,150 @@
...
+import re
+import uuid
+
+from ansible.plugins.action import ActionBase
+from ansible.errors import AnsibleActionFail
+
+
+TMPL_SCRIPT = '''#!/bin/bash
+%(opts)s
+exec > >(tee -ia %(logpath)s)
```
cjeanner, line 87: I'd avoid using `-a` option for `tee` - unless you're expecting multiple runs to happen, meaning appending things to the log file?

ci_framework/plugins/action/ci_script.py
```diff
@@ -0,0 +1,150 @@
...
+class ActionModule(ActionBase):
+
+    def __init__(self, **kwargs):
+        super(ActionModule, self).__init__(**kwargs)
+        self.__script_file_path = pathlib.Path('~/.ansible/tmp').joinpath(uuid.uuid4().hex).absolute()
+        self.__script_file_path.parent.mkdir(parents=True, exist_ok=True)
+
+    @staticmethod
+    def __build_options(task_args):
+        opts = 'set -euo pipefail'
```
cjeanner, line 100: I'd hard-code that one in the template directly. Would avoid the `\n` as shown on L102

ci_framework/plugins/action/ci_script.py
```diff
@@ -0,0 +1,150 @@
...
+        description: The path in where the script will be copied
+        required: true
+        type: str
+    cmd:
+        description: The shell script content to be run
+        required: true
+        type: str
+    debug:
+        description: Enables bash tracing in the script
+        required: true
```
cjeanner, line 30: You apparently set a default to True later - though imho we shouldn't default to True, nor require that parameter, unless the CI Framework adds a new `ROLENAME_debug` parameter, with a global `cifmw_debug` and a default value to `False`.
Reason to default to `False` : in case of a command aiming to extract data, the `-x` will break the output, and create a really messy situation.
For such cases, we MAY want to be smarter with the stdout and, probably, split a bit things (i.e. using PS4 or something like that...)

ci_framework/plugins/action/ci_script.py
```diff
@@ -0,0 +1,150 @@
...
+
+    def __init__(self, **kwargs):
+        super(ActionModule, self).__init__(**kwargs)
+        self.__script_file_path = pathlib.Path('~/.ansible/tmp').joinpath(uuid.uuid4().hex).absolute()
+        self.__script_file_path.parent.mkdir(parents=True, exist_ok=True)
+
+    @staticmethod
+    def __build_options(task_args):
+        opts = 'set -euo pipefail'
+        if task_args.get('debug', True):
```
cjeanner, line 101: here - defaulting to True. We may want to avoid that default.

ci_framework/plugins/action/ci_script.py
```diff
@@ -0,0 +1,150 @@
...
+           'logpath': log_path.as_posix(),
+           'opts': self.__build_options(self._task.args)
+        }
+
+        script_path_str = self.__script_file_path.as_posix()
+        with open(self.__script_file_path, 'w') as tmp_script_file:
+            script_content = TMPL_SCRIPT % script_template_data
+            tmp_script_file.write(script_content)
+
+        file_task = self._task.copy()
```
cjeanner, line 131: sooo here, you're basically passing the same parameters to the ansible.builtin.script module - and since you didn't remove the `output_dir` and `script` parameters, it will fail because of unknown parameters. Check what was done in the `ci_make` action plugin.

cjeanner (commented):
In order to test this action plugin, you may leverage ansible-test. Check how it was done for `ci_make` here:
https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci_framework/tests/integration/targets/make/tasks/main.yml

basically, you'll have to create a tests/integration/targets/script/tasks/main.yml and inject your tests in it.
Be careful, it's running on Prow, meaning read-only unless you're in /tmp, and no privileges at all.

ci_framework/plugins/action/ci_script.py
pablintino: Well, maybe this is part of Ansible not being consistent, but this works, there is no error regarding unknown parameters, seems that every action/module is designed differently. I can change it to be consistent anyway.

ci_framework/plugins/action/ci_script.py
```diff
@@ -0,0 +1,169 @@
...
+DOCUMENTATION = r'''
+---
+action: ci_script
+
+short_description: Runs a string given script remotly while providing logging
+
+description:
+    - Runs a string given script remotly while providing logging.
+    - Logs are written to the `output_dir`/../logs folder.
+    - If `cifmw_ci_script_debug` or `cifmw_debug` variables are set to true output and will be traced.
```
cjeanner, line 19: IMHO this should be an actual parameter, such as `debug`, of the action plugin. Doing so, we'd allow to be a bit more precise as to what task we want to debug.

queria, line 19: could this line be rephrased a bit please, maybe it's just me, but i'm not sure i understand if it means:
- without any of _debug=true there will be no logs
- or if with _debug=true set there will be extra trace info in the logs

ci_framework/plugins/action/ci_script.py
```diff
@@ -0,0 +1,169 @@
...
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+
+DOCUMENTATION = r'''
+---
+action: ci_script
+
+short_description: Runs a string given script remotly while providing logging
```
queria, line 14: remotely

(and maybe 'runs a script given in form of a string' or such?)

ci_framework/plugins/action/ci_script.py
```diff
@@ -0,0 +1,169 @@
...
+
+        # Ensure chdir is not used
+        if 'chdir' in self._task.args:
+            raise AnsibleActionFail('chdir parameter is not supported')
+
+        # Remove cmd if not passed, we are going to use _raw_params to pass the cmd we create here
+        if 'cmd' in self._task.args:
+            self._task.args.pop('cmd')
+
+        output_dir = pathlib.Path(self._task.args.pop('output_dir'))
```
cjeanner, line 135: At this point, I'd ensure output_dir and log_dir exist. And fail if they don't

ci_framework/plugins/action/ci_script.py
```diff
@@ -0,0 +1,169 @@
...
+
+        fnum = len(glob.glob(f'{output_dir}/ci_script_*'))
+        t_name = re.sub(r'([^\x00-\x7F]|\s)+', '_', self._task._name).lower()
+        log_path = output_dir.parent.joinpath('logs', f'ci_script_{fnum:03}_{t_name}.log')
```
cjeanner, line 139: so building log_path out of `os.path.join(log_dir, f'ci_script_...')`

raukadah (approved):
Great addition to the toolchest
/approve

PR #328, opened 2023-06-23
Dynamically retrieve base module name

Sandeepyadav93:
With this commit, we update `operator_build` role to dyynamically retrieve base module name, mod can be either /api or /apis. For ex. infra-operator have apis.

### DIFF
--- a/ci_framework/roles/operator_build/tasks/build.yml
+++ b/ci_framework/roles/operator_build/tasks/build.yml
@@ -14,17 +14,31 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
-- name: "{{ operator.name }} - Set default api path"
-  ansible.builtin.set_fact:
-    operator_api_path: "github.com/{{ cifmw_operator_build_org }}/{{ operator.name }}/api"
+- name: "{{ operator.name }} - Get base module name and set the api path"
+  when:
+    - cifmw_operator_build_meta_build
+    - operator.name != cifmw_operator_build_meta_name
+  block:
+    - name: "{{ operator.name }} - Read go.mod file contents of Openstack Operator"
+      ansible.builtin.slurp:
+        src: "{{ cifmw_operator_build_meta_src }}/go.mod"
+      register: go_mod_out
+
+    - name: "{{ operator.name }} - Get base module name from go.mod"
+      ansible.builtin.set_fact:
+        operator_base_module_name: "{{ go_mod_out['content'] | b64decode | regex_search(cifmw_operator_build_org + '/' + operator.name + '/(\\w*)\\s', '\\1') | first }}"
+
+    - name: "{{ operator.name }} - Set default api path"
+      ansible.builtin.set_fact:
+        operator_api_path: "github.com/{{ cifmw_operator_build_org }}/{{ operator.name }}/{{ operator_base_module_name }}"
 
 - name: "{{ operator.name }} - Update the go.mod file in meta operator for provided PR_SHA"
   ansible.builtin.shell: |
-    go mod edit -replace {{ operator_api_path }}=github.com/{{ operator.pr_owner }}/api@{{ operator.pr_sha }}
+    go mod edit -replace {{ operator_api_path }}=github.com/{{ operator.pr_owner }}/{{ operator_base_module_name }}@{{ operator.pr_sha }}
     go mod tidy
     if [ -d ./apis ]; then
       pushd ./apis/
-      go mod edit -replace {{ operator_api_path }}=github.com/{{ operator.pr_owner }}/api@{{ operator.pr_sha }}
+      go mod edit -replace {{ operator_api_path }}=github.com/{{ operator.pr_owner }}/{{ operator_base_module_name }}@{{ operator.pr_sha }}
       go mod tidy
       popd
     fi


ci_framework/roles/operator_build/tasks/build.yml
```diff
@@ -14,9 +14,16 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+- name: "{{ operator.name }} - Get base module name from go.mod"
+  ansible.builtin.shell: |
+    grep {{ cifmw_operator_build_org }}/{{ operator.name }}/api go.mod | xargs basename
```
rabi, line 19: `| xargs basename` is not needed as you'e again appending it in L26(?)

rabi: Ok you have to split the version then.

rabi (commented):
/lgtm

Sandeepyadav93:
/hold

need further changes

Sandeepyadav93:
/unhold

Content provider which was earlier failing on PR[1] passed with depends-on this parch.

[1] https://github.com/openstack-k8s-operators/infra-operator/pull/66

Sandeepyadav93:
recheck

failed on tempest

https://logserver.rdoproject.org/28/328/7b6d4a6afec56f6399a0a8312b6435fd92a91eb7/github-check/cifmw-end-to-end/2539e88/job-output.txt

rabi (commented):
/lgtm

Sandeepyadav93:
~~~~

]$ git clone https://github.com/openstack-k8s-operators/openstack-operator.git

$ cd openstack-operator/

$ grep github.com/openstack-k8s-operators/infra-operator/api go.mod
	github.com/openstack-k8s-operators/infra-operator/apis v0.0.0-20230615131203-9a25d21f088c


$ basename github.com/openstack-k8s-operators/infra-operator/apis v0.0.0-20230615131203-9a25d21f088c
apis

~~~

ci_framework/roles/operator_build/tasks/build.yml
```diff
@@ -14,17 +14,29 @@
...
+        grep {{ cifmw_operator_build_org }}/{{ operator.name }}/api go.mod | head -n 1 | xargs basename
+      args:
+        chdir: "{{ cifmw_operator_build_meta_src }}"
+      register: operator_base_module_name
+
+    - name: "{{ operator.name }} - Set default api path"
+      ansible.builtin.set_fact:
+        operator_api_path: "github.com/{{ cifmw_operator_build_org }}/{{ operator.name }}/{{ operator_base_module_name.stdout }}"
+  when:
+    - cifmw_operator_build_meta_build
```
pablintino, line 30: Dumb style thing, we are trying to put when clauses on top.

Sandeepyadav93: thanks, done

ci_framework/roles/operator_build/tasks/build.yml
```diff
@@ -14,17 +14,29 @@
...
-  ansible.builtin.set_fact:
-    operator_api_path: "github.com/{{ cifmw_operator_build_org }}/{{ operator.name }}/api"
+- name: "{{ operator.name }} - Get base module name and set the api path"
+  when:
+    - cifmw_operator_build_meta_build
+    - operator.name != cifmw_operator_build_meta_name
+  block:
+    - name: "{{ operator.name }} - Get base module name from go.mod"
+      ansible.builtin.shell: |
+        grep {{ cifmw_operator_build_org }}/{{ operator.name }}/api go.mod | head -n 1 | xargs basename
```
pablintino, line 24: I agree with @marios comment on the review meeting, why not use Ansible? This can be perfectly done with lineinfile module and a regex such as this one (example, should be parameterized obviously) `github\.com\/openstack-k8s-operators\/infra-operator\/(\w*)\s`

rlandy: ack - but can be done in a follow up if this is blocking

viroel: I don't have a preference here, I think that the shell was a quick solution to have a fix landing asap.
If it is something that must merge asap, we can think on a follow up?

pablintino: Sure, please, create a subtask here https://issues.redhat.com/browse/OSP-25624

Sandeepyadav93: I will fix it here instead of adding another card to the backlog. 

Just for my understanding, @marios @pablintino could you please elaborate your reasoning why do you think ansible way is better? So that next time I will prefer one over the other.

pablintino: Is just best practices, here you have and intro: https://www.ansible.com/blog/ansible-best-practices-essentials (point four). There they talk about error management, but we can talk about idempotence too as most modules properly compute the `changed` property, about consistent run in multiple environments, here the use case is very basic, but think about commands that behaves differently between versions, the exact same role will behave differently, but, with native modules, as all is done in the ansible implementation of the module and versions are fixed, the results of the module run will be equal. We have tried the most to use native modules because of those reasons, so, if that reasons are not enough, we can talk about consistency too, as we are trying to use them whenever possible, why not try to apply that rule everywhere when possible?

Sandeepyadav93: I discussed this with Pablo, I read more about lineinfile but all the example I found on internet are either adding/deletion/replacing things and not just search which I wanted, looks like the `lineinfile` module is primarily used for modifying lines in a file rather than searching.

https://www.middlewareinventory.com/blog/ansible-lineinfile-examples/

we have created https://issues.redhat.com/browse/OSP-26014 to improve this in a followup. lets merge this as it is and I will improve this in a followup.

Sandeepyadav93:
Rebasing the patch to fix tempest issue on end-to-end job

Sandeepyadav93:
recheck

ci_framework/roles/operator_build/tasks/build.yml
Sandeepyadav93: marking as resolved as we have changes the code

ci_framework/roles/operator_build/tasks/build.yml
Sandeepyadav93: Thanks Pablo for your earlier explanation and link why using ansible module is better and for suggesting the alternative module.

I have updated the code.

ci_framework/roles/operator_build/tasks/build.yml
```diff
@@ -14,17 +14,31 @@
...
 
-- name: "{{ operator.name }} - Set default api path"
-  ansible.builtin.set_fact:
-    operator_api_path: "github.com/{{ cifmw_operator_build_org }}/{{ operator.name }}/api"
+- name: "{{ operator.name }} - Get base module name and set the api path"
+  when:
+    - cifmw_operator_build_meta_build
+    - operator.name != cifmw_operator_build_meta_name
+  block:
+    - name: {{ operator.name }} - Read go.mod file contents of Openstack Operator
```
marios, line 22: does this task name need the quotes because of {{ 

I have had this fail in the past ... but perhaps it is dependent on ansible version or something (i.e. have you run this like that and it is OK?)

marios (commented):
lgtm but one inline doubt about ""

ci_framework/roles/operator_build/tasks/build.yml
Sandeepyadav93: thanks, marios, fixed it.

I couldn't test earlier as the image build had issues due to missing nova hash(this issue got fixed today)

cjeanner (approved):

PR #329, opened 2023-06-23
OSP-25630 Generate artifact "operators"

rachael-george:
Generate artifacts containing images used in the current deployment

  - [x] README in the role


cjeanner:
ah. great... so "update branch" from github UI is actually creating a merge......... how NOT convenient.

Sorry @rachael-george - I'll clean my mess and review.

ci_framework/roles/artifacts/tasks/cluster_info.yml
```diff
@@ -89,3 +89,52 @@
         chdir: "{{ cifmw_artifacts_basedir }}/logs/edpm"
       changed_when: true
       ignore_errors: true
+
+    - name: Create a file to store the operator images
+      changed_when: true
+      ignore_errors: true
+      ansible.builtin.shell: |
```
cjeanner, line 96: we can use ansible.builtin.copy, if really:
```YAML
ansible.builtin.copy:
  dest: "{{ cifmw_artifacts_basedir }}/artifacts/operator_images.yaml"
  content: "cifmw_install_yamls_vars:"
```

ci_framework/roles/artifacts/tasks/cluster_info.yml
```diff
@@ -89,3 +89,52 @@
...
+        echo "cifmw_install_yamls_vars:" > {{ cifmw_artifacts_basedir }}/artifacts/operator_images.yaml
+
+    - name: Get operator images from "{{ cifmw_install_yamls_defaults['OPERATOR_NAMESPACE'] }}" namespace
+      changed_when: true
+      ignore_errors: true
+      environment:
+        KUBECONFIG: "{{ cifmw_install_yamls_environment['KUBECONFIG'] }}"
+        PATH: "{{ cifmw_path }}"
+      ansible.builtin.shell: |
```
cjeanner, line 105: using the new `ci_script` plugin may be good for that - we'd get the generated script and its logs, like we get for `ci_make`.
You can see examples of `ci_script` calls here:
https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci_framework/tests/integration/targets/script/tasks/main.yml

ci_framework/roles/artifacts/tasks/cluster_info.yml
```diff
@@ -89,3 +89,52 @@
...
+            echo "  ${operator_name}_OP_IMG: ${image_name}" >> {{ cifmw_artifacts_basedir }}/artifacts/operator_images.yaml;
+         done
+
+    - name: Get openstack service images from "{{ cifmw_install_yamls_defaults['NAMESPACE'] }}" namespace
+      changed_when: true
+      ignore_errors: true
+      environment:
+        KUBECONFIG: "{{ cifmw_install_yamls_environment['KUBECONFIG'] }}"
+        PATH: "{{ cifmw_path }}"
+      ansible.builtin.shell: |
```
cjeanner, line 132: ditto - using `ci_script` would give better logging and reproducer capabilities.

rachael-george: Updated

ci_framework/roles/artifacts/tasks/cluster_info.yml
rachael-george: updated

ci_framework/roles/artifacts/tasks/cluster_info.yml
rachael-george: updated

ci_framework/roles/env_op_images/README.md
```diff
@@ -0,0 +1,14 @@
+# env_op_images
+A role to gather the container images used in the openstack deployment with specific tags. The file created as part of this role can be used with override_op_image role during deployment to recreate the openstack environment
+
+## Parameters
+* `cifmw_env_op_images_dir`: (String) Directory where the operator_images.yaml will be stored. Defaults to `~/ci-framework-data/artifacts`
+
```
cjeanner, line 6: blank line - will break list during the doc generation

ci_framework/roles/env_op_images/tasks/main.yml
```diff
@@ -0,0 +1,110 @@
...
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+- name: Ensure directory is present
+  ansible.builtin.file:
+    path: "{{ cifmw_env_op_images_dir }}"
+    state: directory
+
+- name: Check for oc command
```
cjeanner, line 22: What's the use? I don't think you check the oc_installed parameter ?

ci_framework/roles/env_op_images/tasks/main.yml
```diff
@@ -0,0 +1,110 @@
...
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  ansible.builtin.shell: |
+    oc get OpenStackControlPlane -n {{ cifmw_install_yamls_defaults['NAMESPACE'] }} --no-headers=true | awk '{print $2}'
+  register: openstack_installed
+
+- name: Get openstack operator images
+  when:
+    - oc_installed is defined
+    - oc_installed.rc == 0
```
cjeanner, line 40: OK, here you use it, but.. it would already fail with the 2 previous tasks :). I don't think it makes sense to do those tests.

ci_framework/roles/override_op_image/tasks/main.yml
```diff
@@ -0,0 +1,45 @@
...
+- name: Update CSVs with custom images
+  when:
+  - cifmw_openstack_svc_image is defined
+  - cifmw_openshift_kubeconfig is defined
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  block:
+    - name: Dump OpenStackControlPlane CR into a file
+      ansible.builtin.shell: |
```
cjeanner, line 31: As a general consideration, please use the `ci_script` instead of shell/command (I saw some other ansible.builtin.command calls in other task files).

rachael-george: Updated here. In the other task, it was used to run a  single command, should I change that to use `ci_script` as well?

ci_framework/roles/env_op_images/README.md
rachael-george: Updated

ci_framework/roles/env_op_images/tasks/main.yml
```diff
@@ -0,0 +1,101 @@
...
+- name: Ensure directory is present
+  ansible.builtin.file:
+    path: "{{ cifmw_env_op_images_dir }}"
+    state: directory
+
+- name: Check if OpenStackControlPlane is setup
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  ansible.builtin.shell: |
```
cjeanner, line 26: OK, we can keep the "shell" here, unless there are valuable outputs we may want to log in a dedicated file. IMHO, it's fine as-is.
@pablintino any thoughts?

ci_framework/roles/env_op_images/tasks/main.yml
```diff
@@ -0,0 +1,101 @@
...
+  block:
+    - name: Create a file to store the images
+      ansible.builtin.copy:
+        dest: "{{ cifmw_env_op_images_dir }}/{{ cifmw_env_op_images_file  }}"
+        content: "cifmw_install_yamls_vars:"
+
+    - name: Get openstack operator-index image
+      environment:
+        KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+        PATH: "{{ cifmw_path }}"
```
cjeanner, line 44: you can move that "environment" definition to the block itself. It would avoid repetition.

ci_framework/roles/override_op_image/molecule/default/prepare.yml
```diff
@@ -0,0 +1,28 @@
...
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+
+- name: Prepare
+  hosts: all
+  vars:
+    ansible_user_dir: "{{ lookup('env', 'HOME') }}"
+    cifmw_basedir: "{{ ansible_user_dir }}/ci-framework-data"
```
cjeanner, line 22: once you correct the "defaults/main.yml" of that role in order to ensure some default basedir is set, you should be all good.

ci_framework/roles/override_op_image/defaults/main.yml
```diff
@@ -0,0 +1,21 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+
+# All variables intended for modification should be placed in this file.
+# All variables within this role should have a prefix of "cifmw_override_op_image"
+
+cifmw_override_op_image_dir:  "{{ cifmw_basedir }}/artifacts/manifests/openstack"
```
cjeanner, line 21: same as the other role - cifmw_basedir may not exist, and you want to ensure artifacts *and* logs tree exist.

ci_framework/roles/override_op_image/tasks/main.yml
```diff
@@ -0,0 +1,49 @@
...
+    - name: Replace openstack service images in the yaml file
+      with_dict: "{{cifmw_openstack_svc_image}}"
+      ansible.builtin.replace:
+        path: "{{cifmw_override_op_image_dir}}/osp_cp_cr.yaml"
+        regexp: "{{ item.key }}\\[[^\\]]+\\]"
+        replace: "{{ item.key}}{{item.value}}"
+
+    - name: Apply the changes to OpenStackControlPlane CR
+      ci_script:
+        output_dir:
```
cjeanner, line 46: missing value for output_dir :)

ci_framework/roles/override_op_image/tasks/main.yml
cjeanner: ci_script allows to get proper logging in a dedicate file as well as a reproducer script, for debugging purpose or just a reference. Using it over `command` or `shell` is really a matter of "would a log or reproducer provide valuable information?" - I'd usually push for `ci_script`, but there are cases where, indeed, it doesn't really make sense (commented a `shell` call just before actually :) ).

ci_framework/roles/env_op_images/tasks/main.yml
rachael-george: Done

ci_framework/roles/env_op_images/tasks/main.yml
```diff
@@ -0,0 +1,92 @@
...
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  ansible.builtin.shell: |
+    oc get OpenStackControlPlane -n {{ cifmw_install_yamls_defaults['NAMESPACE'] }} --no-headers=true | awk '{print $2}'
+  register: openstack_installed
+
+- name: Get openstack operator images
+  when:
+    - cifmw_openshift_kubeconfig is defined
+    - openstack_installed.stdout | bool
```
marios, line 36: just wondering how that is meant to work...  does the output of this command ever give us a boolean value?

e.g in an environment with no controlplane:

```
$ oc get OpenStackControlPlane -n openstack --no-headers=true | awk '{print $2}'
No resources found in openstack namespace.
```

so that one would evaluate to false with |bool
when would it be true though? i am missing what is given from the awk '{print $2}' probably?

marios (commented):
stared for a bit, minor question/comment inline
its a lot to review so I'll have another look on my next reviews time if this is not merged then

pablintino:
Can you please squash all the changes before merging?

ci_framework/roles/env_op_images/tasks/main.yml
rachael-george: This is the output when  OpenStackControlPlane CR is present in the cluster and is successfully deployed.
```
$ oc get OpenStackControlPlane -n openstack --no-headers=true | awk '{print $2}'
True
```
So, it will collect the container images only when the cluster is deployed

ci_framework/roles/override_op_image/README.md
```diff
@@ -0,0 +1,14 @@
...
+A role used to override the operator images used in the deployment. A `yaml` file is passed to the ansible playbook (usually the file generated by `env_op_images` role from a previous deployment) containing the images with tags to be used for the current deployment
+
+## Parameters
+* `cifmw_override_op_image_dir`: (String) The directory where the updated `OpenstackControlPlane` CR yaml used for the deployment is stored. Defaults to `{{ cifmw_basedir }}/artifacts/manifests/openstack`
+
+## Examples
+```YAML
+- name: Override openstack service images
+  when:
+  - cifmw_openstack_svc_image is defined
```
cjeanner, line 11: apparently pySpelling is crying due to that "svc" mention. Care to add it to the dictionary, as described in the doc here: https://ci-framework.readthedocs.io/en/latest/internals/02_testing.html#spellchecking

ci_framework/roles/env_op_images/tasks/main.yml
```diff
@@ -0,0 +1,92 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+- name: Ensure directory is present
+  ansible.builtin.file:
+    path: "{{ cifmw_env_op_images_dir }}"
```
cjeanner, line 19: hmm, you don't mention the `{{ item }}` so you'll create twice the exact same directory.
You want
```suggestion
    path: "{{ cifmw_env_op_images_dir }}/{{ item }}"
```

ci_framework/roles/env_op_images/tasks/main.yml
cjeanner: iirc, if it's not an empty string, it will be considered as true.. Maybe we can get a stronger check on that command output?

ci_framework/roles/install_yamls/defaults/main.yml
```diff
@@ -35,3 +35,22 @@ cifmw_install_yamls_whitelisted_vars:
   - OUTPUT_BASEDIR
   - OUTPUT_DIR
   - SSH_KEY
+  - OPENSTACK_OP_IMG
```
cjeanner, line 38: I don't really get why those parameters have to be listed here. AFAIK they aren't taken by install_yamls, are they? I'd expect the list to be provided by some ansible env file that we load and consume?

ci_framework/roles/override_op_image/tasks/main.yml
```diff
@@ -0,0 +1,52 @@
...
+      ansible.builtin.replace:
+        path: "{{cifmw_override_op_image_dir}}/osp_cp_cr.yaml"
+        regexp: "{{ item.key }}\\[[^\\]]+\\]"
+        replace: "{{ item.key}}{{item.value}}"
+
+    - name: Apply the changes to OpenStackControlPlane CR
+      ci_script:
+        output_dir: "{{cifmw_override_op_image_dir}}"
+        script: |
+          sed '/resourceVersion/d' {{cifmw_override_op_image_dir}}/osp_cp_cr.yaml > {{cifmw_override_op_image_dir}}/osp_cp_cr.yaml
```
cjeanner, line 51: you can use `sed -i` in order to write the file in-place

ci_framework/roles/override_op_image/tasks/main.yml
```diff
@@ -0,0 +1,52 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+- name: Ensure directory exists
+  ansible.builtin.file:
+    path: "{{ cifmw_basedir | default(ansible_user_dir ~ '/ci-framework-data') }}/{{item}}"
```
cjeanner, line 19:
```suggestion
    path: "{{ cifmw_override_op_image_dir }}/{{ item }}"
```

ci_framework/roles/install_yamls/defaults/main.yml
rachael-george: Yes, they are not used by install_yamls. From what I understand, the only image used for deployment is `OPENSTACK_IMG`.  I collected all the other images, just to be able to cross check the values in the new deployment. I can move them under a different env variable. Does that sound okay?

cjeanner: I don't think using environment variables is needed/correct actually. Since the comparison is done in ansible, using actual ansible parameter should be sufficient ? You can create a new parameter as a dict and populate it, or even write a file in the artifacts directory (from the env_op_images for instance), and load it later (from the override_op_image). This would allow people to inject their own overrides, in addition.

What do you think?

ci_framework/roles/env_op_images/tasks/main.yml
```diff
@@ -0,0 +1,92 @@
...
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  ansible.builtin.shell: |
+    oc get OpenStackControlPlane -n {{ cifmw_install_yamls_defaults['NAMESPACE'] }} --no-headers=true | awk '{print $2}'
+  register: openstack_installed
+
+- name: Get openstack operator images
+  when:
+    - cifmw_openshift_kubeconfig is defined
```
cjeanner, line 35: that condition is useless - you already consume that variable in the previous one actually.

ci_framework/roles/env_op_images/tasks/main.yml
```diff
@@ -0,0 +1,92 @@
...
+             image_name=$(oc get pod $pod -n {{ cifmw_install_yamls_defaults['OPERATOR_NAMESPACE'] }} -o jsonpath='{.status.containerStatuses[1].image}' );
+             echo "  ${operator_name}_OP_IMG: ${image_name}" >> "{{ cifmw_env_op_images_dir }}/artifacts/{{ cifmw_env_op_images_file  }}"
+          done
+
+    - name: Get openstack service images
+      ci_script:
+        output_dir: "{{ cifmw_env_op_images_dir }}/artifacts"
+        script: |
+          echo -e "\ncifmw_openstack_svc_image:"  >> "{{ cifmw_env_op_images_dir }}/artifacts/{{ cifmw_env_op_images_file  }}"
+          for pod in $(oc get pod  --no-headers=true  -n {{ cifmw_install_yamls_defaults['NAMESPACE'] }} | grep -v Completed | awk '{print $1}');  do
```
cjeanner, line 86: @pablintino guess there's no cleaner way to get the data without nested loop? Care to have a look, you know more than I do :)

cjeanner:
Please squash the commits into one only. And I would like to get @pablintino view on that one - it looks correct, but he knows more than me :)

cjeanner:
/lgtm

I'll let @pablintino have a look and approve it if all good.

ci_framework/roles/env_op_images/tasks/main.yml
```diff
@@ -0,0 +1,91 @@
...
+          image=$(oc get pod ${pod_name} -n {{ cifmw_install_yamls_defaults['OPERATOR_NAMESPACE'] }} -o jsonpath='{.status.containerStatuses[0].imageID}');
+          echo "  RABBITMQ_OP_IMG: ${image}" >> "{{ cifmw_env_op_images_dir }}/artifacts/{{ cifmw_env_op_images_file  }}"
+
+    - name: Get rest of the operator images
+      ci_script:
+        output_dir: "{{ cifmw_env_op_images_dir }}/artifacts"
+        script: |
+          for pod in $(oc get pod -l openstack.org/operator-name --no-headers=true -n {{ cifmw_install_yamls_defaults['OPERATOR_NAMESPACE'] }} | grep -v 'openstack-operator-controller-manager' | awk '{print $1}');  do
+             operator_name=$(oc get pod $pod -o jsonpath='{.metadata.labels.openstack\.org/operator-name}' -n {{ cifmw_install_yamls_defaults['OPERATOR_NAMESPACE'] }} --no-headers=true | tr '[:lower:]' '[:upper:]');
+             image_name=$(oc get pod $pod -n {{ cifmw_install_yamls_defaults['OPERATOR_NAMESPACE'] }} -o jsonpath='{.status.containerStatuses[1].image}' );
```
pablintino, line 76: Why we fetch here the full ID (sha265 one) and in the rest of the examples before we fetch the plain name and tag?

ci_framework/roles/env_op_images/tasks/main.yml
```diff
@@ -0,0 +1,91 @@
...
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  ignore_errors: true
+  block:
+    - name: Create a file to store the images
+      ansible.builtin.copy:
+        dest: "{{ cifmw_env_op_images_dir }}/artifacts/{{ cifmw_env_op_images_file  }}"
+        content: "cifmw_install_yamls_vars:"
+
+    - name: Get openstack operator-index image
```
pablintino, line 46: Why not making a single call instead of 4x2? One single call to fetch those pods as yaml will be enough, with the retrieved data we can post process it here using plain Ansible. That way we can reduce the number of calls to kubeapi and make the whole snippet more Ansible friendly instead of having a tone of bash shells. If you want to fetch the data in a single call you can try the k8s module, that fetches the content directly as an Ansible variable, or use `oc` and parse with a `from_yaml`/`from_json` afterwards.

Thank way you will end up with a proper formated file with all the needed info in a single api call. With a format like this one:
```yaml
apiVersion: v1
items:
- apiVersion: v1
  kind: Pod
  metadata:
  ... pod meta
  spec:
   ... pod spec
  status:
    conditions:
     ... pod conditions
    containerStatuses:
    - containerID: cri-o://83e932c9a1672cd14c4f05792ffb4d1c96f264d4b0bb55121f601cc01a78cde7
      image: quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:7370b4b4acca491815ac6e617fa04aff824c62ad1e94e19d55505ba510539798
      imageID: quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:7370b4b4acca491815ac6e617fa04aff824c62ad1e94e19d55505ba510539798
      lastState: {}
      name: extract
      ready: false
      restartCount: 0
      started: false
      state:
        terminated:
          containerID: cri-o://83e932c9a1672cd14c4f05792ffb4d1c96f264d4b0bb55121f601cc01a78cde7
          exitCode: 0
          finishedAt: "2023-08-24T09:02:29Z"
          reason: Completed
          startedAt: "2023-08-24T09:02:27Z"
    hostIP: 192.168.126.11
    initContainerStatuses:
    ... init container status
    phase: Succeeded
    podIP: 10.217.0.80
    podIPs:
    - ip: 10.217.0.80
    qosClass: Burstable
    startTime: "2023-08-24T09:01:53Z"
- apiVersion: v1
  kind: Pod
  metadata:
    annotations:`
... other pod data
```
Using plain Ansible filters like select/selectattr with equalto and match will be enough to correctly point to the needed field for every Pod.

ci_framework/roles/env_op_images/tasks/main.yml
```diff
@@ -0,0 +1,91 @@
...
+      ci_script:
+        output_dir: "{{ cifmw_env_op_images_dir }}/artifacts"
+        script: |
+          for pod in $(oc get pod -l openstack.org/operator-name --no-headers=true -n {{ cifmw_install_yamls_defaults['OPERATOR_NAMESPACE'] }} | grep -v 'openstack-operator-controller-manager' | awk '{print $1}');  do
+             operator_name=$(oc get pod $pod -o jsonpath='{.metadata.labels.openstack\.org/operator-name}' -n {{ cifmw_install_yamls_defaults['OPERATOR_NAMESPACE'] }} --no-headers=true | tr '[:lower:]' '[:upper:]');
+             image_name=$(oc get pod $pod -n {{ cifmw_install_yamls_defaults['OPERATOR_NAMESPACE'] }} -o jsonpath='{.status.containerStatuses[1].image}' );
+             echo "  ${operator_name}_OP_IMG: ${image_name}" >> "{{ cifmw_env_op_images_dir }}/artifacts/{{ cifmw_env_op_images_file  }}"
+          done
+
+    - name: Get openstack service images
```
pablintino, line 80: My previous comment applies even better for this use case, when we make a tone of calls, some shell `cut` and even `jq`. TBO, can be perfectly done in a single call fetching all pods of the `openstack` namespace, filtering in ansible to not pick the ones from jobs (I assume that's why the `Completed` grep is there) and from there, plain Ansible filtering and mapping.

ci_framework/roles/edpm_prepare/tasks/main.yml
```diff
@@ -154,6 +154,12 @@
         output_dir: "{{ cifmw_edpm_prepare_basedir }}/artifacts"
         script: "oc apply -f {{ cifmw_edpm_prepare_openstack_cr_manifest_paths.files[0].path }}"
 
+    - name: Override openstack service images
+      when:
+      - cifmw_openstack_svc_image is defined
```
pablintino, line 159: Where does this variable come from? It's not part of a role and it's not documented as a global one.

ci_framework/roles/override_op_image/tasks/main.yml
pablintino, line 1: Uhmmm I don't see this approach. This role is called only from edpm_prepare, so I'll assume for the rest of the comment that it has been created only for that purpose.
Main main concern, and the reason I don't see this approach as a valid one, why are we making edpm_prepare deploy a thing that we already know that is not what we won't? Will be far far better to make use a kustomization like the one done here: https://github.com/openstack-k8s-operators/ci-framework/pull/329/files#diff-1360e3a211e5e823d01cff09e8d8a233caaf61abfbc56bc319d1f783af994394R141-L146 (now that call only accepts a single kustomization, but I'm about to add support for multiple ones) and deploy, in a single shot the desired versions.

ci_framework/roles/override_op_image/molecule/default/converge.yml
pablintino, line 1: Not really testing that the desired image are the ones deployed.

pablintino (changes requested):

ci_framework/roles/edpm_prepare/tasks/main.yml
rachael-george: When the images are collected from an existing cluster, it is stored in a yaml file under this key.

pablintino:
What's the status of this PR? Any news?

ci_framework/roles/override_op_image/tasks/main.yml
rachael-george: The approach that I currently have is that once the openstackcontrolplane  CR is applied to the cluster, I search for all the images in the CR which has a corresponding key (image) in the dictionary with custom images (specified under `cifmw_openstack_svc_image`) , replace the entire image value (image+tag) in the CR and apply the changes.
The issue I am facing with using `edpm_kustomize`  to achieve the same. The image fields are not available in the CR yaml files before deployment  to be able to easily perform a replace option.  Also, the keys used for images in the CR itself are not uniform across operators/components and needs to be dealt with individually. 

Any suggestions on how to resolve this?

pablintino: OK I get your point, so the difficult point here is that the container image fields are not standard and are not present in the default CR? Isn't that field always in `.spec[serviceName].template.containerImage`?
This thing is hard to follow to be honest, I mean, I get your point, but I'm not even able to understand, how this works https://github.com/openstack-k8s-operators/ci-framework/pull/329/files#diff-562090fa2191e194b77085536f96ce0559f752ade496d5df5625e92b8e2a4b20R40-R45? What's the appearance of one of the generated regexes there? Just to be sure if what you are trying to do has any caveats.
So, for me OK to patch the existing one if it's difficult to create those fields on the fly in the default CR, but I'd prefer to use a proper tool (Ansible, jinja, kustomize, those kind of tools) to properly manipulate a manifests instead of relaying on manipulating it as a string with regexes, but that's other discussion.

ci_framework/roles/override_op_image/tasks/main.yml
```diff
@@ -0,0 +1,52 @@
...
+
+- name: Update CSVs with custom images
+  when:
+  - cifmw_openstack_svc_image is defined
+  - cifmw_openshift_kubeconfig is defined
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  block:
+    - name: Dump OpenStackControlPlane CR into a file
```
pablintino, line 33: At this point, if this role is always called from edpm_prepare, you already have the controlplane CR loaded in a fact, as a result of the kustomization, and, at this point, is already deployed, why making to calls to fetch something that's already there?

ci_framework/roles/override_op_image/tasks/main.yml
```diff
@@ -0,0 +1,52 @@
...
+        output_dir: "{{ cifmw_override_op_image_dir }}/artifacts/manifests/openstack"
+        script: |
+          cp_name=$(oc get OpenStackControlPlane -n {{ cifmw_install_yamls_defaults['NAMESPACE'] }} --no-headers=true | awk '{print $1}')
+          oc get OpenStackControlPlane { cp_name } -n {{ cifmw_install_yamls_defaults['NAMESPACE'] }} -o yaml > {{ cifmw_override_op_image_dir }}/artifacts/manifests/openstack/osp_cp_cr.yaml
+
+    - name: Replace openstack service images in the yaml file
+      with_dict: "{{cifmw_openstack_svc_image}}"
+      ansible.builtin.replace:
+        path: "{{ cifmw_override_op_image_dir }}/artifacts/manifests/openstack/osp_cp_cr.yaml"
+        regexp: "{{ item.key }}\\[[^\\]]+\\]"
```
pablintino, line 44: This is the regex I said I don't really understand. Anyway, can't this be done in Ansible instead of manipulating the manifest based on regexes?

ci_framework/roles/env_op_images/tasks/main.yml
```diff
@@ -0,0 +1,111 @@
...
+        field_selectors:
+          - status.phase=Running
+      register: pod_list
+
+    - name: Retreive openstack-operator-index pod
+      set_fact:
+        selected_pod: "{{ pod_list.resources| selectattr('metadata.generateName', 'equalto', 'openstack-operator-index-') | list|first}}"
+
+    - name: Write openstack-operator-index image to file
+      ansible.builtin.lineinfile:
```
pablintino, line 59: Is this generating a yaml file? Why not using proper Ansible? This is hacky, I mean, there are enough builtin tools to do this. The thing that I expected here is a fact, created with set_fact and where you can add, by using the combine filter, every key-value with each image, and, at the very end, dump that fact to a file.

ci_framework/roles/override_op_image/tasks/main.yml
rachael-george: The file containing the images, looks something like this:
```
cifmw_openstack_svc_image:
  quay.io/podified-antelope-centos9/openstack-ceilometer-central: sha256:5904247ce6322e2fb56830572e868e6d2937490ab7167865be5fb01d3f1f780d
  quay.io/podified-antelope-centos9/openstack-ceilometer-notification: sha256:8e3b56862a5f047b799edaa30a8e0b342b402f5181d397120327d3d9c4272564
  quay.io/infrawatch/sg-core: sha256:4befc0268975239bf67f895d9a5b9304e4733182e61ece527aba53e55582883d
  quay.io/podified-antelope-centos9/openstack-cinder-api: sha256:c8cfce053db833ba4ff141598e281213d001f14b6f23340caa0c140e7f58db3c
```
The regex searches for any string in the CR that matches the key (which also happens to be the image name) and replaces that with image name (key) and image_tag (value)  from the file.

ci_framework/roles/override_op_image/tasks/main.yml
rachael-george: For most of the services it is defined `.spec[serviceName].template.containerImage`. But there are a few service that are a bit different, eg: 
```  
ovn:
    enabled: true
    template:
      ovnController:
        debug:
          service: false
        external-ids:
          enable-chassis-as-gateway: true
          ovn-bridge: br-int
          ovn-encap-type: geneve
          system-id: random
        networkAttachment: tenant
        ovnContainerImage: quay.io/podified-antelope-centos9/openstack-ovn-controller@sha256:cb0466d67590977d17c5ec7094fe0b6726b39990b42fb6381bac283b7a9bcb9c
        ovsContainerImage: quay.io/podified-antelope-centos9/openstack-ovn-base@sha256:b950e89935362fc4b488189a6963630ca557c71c90f8b0cfc27d3a257f7b917f
```
I have added an explanation for the regex below

ci_framework/roles/env_op_images/tasks/main.yml
rachael-george: Since openstack-operator-index and rabbitmq images require different image extraction, I kept the code as is. Changed it for other operators

ci_framework/roles/env_op_images/tasks/main.yml
```diff
@@ -0,0 +1,111 @@
...
+    - name: Create a file to store the images
+      ansible.builtin.copy:
+        dest: "{{ cifmw_env_op_images_dir }}/artifacts/{{ cifmw_env_op_images_file  }}"
+        content: |
+          cifmw_install_yamls_vars:
+          cifmw_openstack_operator_images:
+          cifmw_edpm_prepare_extra_vars:
+
+    - name: Get images from the csv
+      ansible.builtin.shell: |
```
pablintino, line 50: Why using the shell module instead of command?

ci_framework/roles/env_op_images/tasks/main.yml
```diff
@@ -0,0 +1,111 @@
...
+        oc get csv -n {{ cifmw_install_yamls_defaults['OPERATOR_NAMESPACE'] }} openstack-operator.v0.0.1 -o=jsonpath='{.spec.install.spec.deployments[0].spec.template.spec.containers[1].env[*]}'
```
pablintino, line 51: Hardcoded version will be an issue soon. Why not omit the CSV name and fetch the images using `.items[*].spec`?

ci_framework/roles/env_op_images/tasks/main.yml
```diff
@@ -0,0 +1,111 @@
...
+      register: csv_images
+
+    - name: Format the output
```
pablintino, line 54: This task and the next can be merged into one single task

ci_framework/roles/env_op_images/tasks/main.yml
```diff
@@ -0,0 +1,111 @@
...
+        csv_image_list: "{{ csv_image_list | default([]) + [item | from_json] }}"
+
+    - name: Write images to file
+      loop: "{{ csv_image_list }}"
+      ansible.builtin.lineinfile:
+        path: "{{ cifmw_env_op_images_dir }}/artifacts/{{ cifmw_env_op_images_file  }}"
+        line: "  {{item.name}}: {{item.value}}"
+        insertafter: "cifmw_edpm_prepare_extra_vars:"
+
+    - name: Get all the pods in openstack-operator namespace
```
pablintino, line 70: I see that you trust that the block KUBECONFIG provides auth info, that can do the job, but  I'd prefer to be consistent all across the framework. Can you please check how we use this module?

ci_framework/roles/env_op_images/tasks/main.yml
```diff
@@ -0,0 +1,111 @@
...
+      kubernetes.core.k8s_info:
+        kind: Pod
+        namespace: "{{ cifmw_install_yamls_defaults['OPERATOR_NAMESPACE'] }}"
+        field_selectors:
+          - status.phase=Running
+      register: pod_list
+
+    - name: Retreive openstack-operator-index pod
```
pablintino, line 78: Retrieve*

ci_framework/roles/env_op_images/tasks/main.yml
```diff
@@ -0,0 +1,111 @@
...
+        selected_pods: "{{ pod_list.resources| rejectattr('metadata.generateName', 'contains','openstack-operator-index-') | rejectattr('metadata.generateName', 'contains','rabbitmq-cluster-operator-') }}"
+
+    - name: Add operator images to the dictionary
+      loop: "{{ selected_pods }}"
+      ansible.builtin.set_fact:
+        operator_images_dict: "{{operator_images_dict | combine({ item.metadata.labels['openstack.org/operator-name'] | upper ~ '_OP_IMG': item.status.containerStatuses[1].imageID}) }}"
+
+    - name: Write openstack operator images to file
+      loop: "{{ operator_images_dict | dict2items }}"
+      ansible.builtin.lineinfile:
```
pablintino, line 108: Please, use `ansible.builtin.copy` and the `to_nice_yaml` filter to properly create a YAML file within Ansible.

ci_framework/roles/env_op_images/tasks/main.yml
```diff
@@ -0,0 +1,111 @@
...
+      ansible.builtin.set_fact:
+        selected_pod: "{{ pod_list.resources| selectattr('metadata.generateName', 'equalto', 'openstack-operator-index-') | list|first}}"
+
+    - name: Write openstack-operator-index image to file
+      ansible.builtin.lineinfile:
+        path: "{{ cifmw_env_op_images_dir }}/artifacts/{{ cifmw_env_op_images_file  }}"
+        line: "  OPENSTACK_IMG: {{selected_pod.status.containerStatuses[0].imageID}}"
+        insertafter: "cifmw_install_yamls_vars:"
+
+    - name: Retrieve rabbitmq-operator pod
```
pablintino, line 88: The next three tasks can be perfectly merged into a single one.

ci_framework/roles/env_op_images/tasks/main.yml
pablintino: For the `lineinfile` module usage, please, use `ansible.builtin.copy` and the `to_nice_yaml` filter to properly create a YAML file within Ansible.

ci_framework/roles/env_op_images/tasks/main.yml
```diff
@@ -0,0 +1,111 @@
...
+
+- name: Ensure directory is present
+  ansible.builtin.file:
+    path: "{{ cifmw_env_op_images_dir }}/{{ item }}"
+    state: directory
+  loop:
+    - artifacts
+    - logs
+
+- name: Check if OpenStackControlPlane is setup
```
pablintino, line 25: You are already using the `k8s_info` module, why not using it to check if there is a OpenStackControlPlane CR in k8s instead of using oc and bash?

ci_framework/roles/env_op_images/tasks/main.yml
```diff
@@ -0,0 +1,111 @@
...
+  when:
+    - openstack_installed.stdout == "True"
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  block:
+    - name: Create a file to store the images
+      ansible.builtin.copy:
+        dest: "{{ cifmw_env_op_images_dir }}/artifacts/{{ cifmw_env_op_images_file  }}"
+        content: |
```
pablintino, line 44: No need to treat yaml content as a plain string, you can declare the yaml in a variable or a fact and magically format it and passing it to the copy module by: `content: "{{ _my_dict | to_nice_yaml }}"`

ci_framework/roles/env_op_images/tasks/main.yml
```diff
@@ -0,0 +1,111 @@
...
+- name: Get openstack operator images
+  when:
+    - openstack_installed.stdout == "True"
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  block:
+    - name: Create a file to store the images
+      ansible.builtin.copy:
+        dest: "{{ cifmw_env_op_images_dir }}/artifacts/{{ cifmw_env_op_images_file  }}"
```
pablintino, line 43: Double space at the end.

pablintino (changes requested):

ci_framework/roles/env_op_images/tasks/main.yml
rachael-george: Since there are two commands now, sticking to the shell module

ci_framework/roles/env_op_images/tasks/main.yml
rachael-george: Added the auth parameters

rachael-george:
Created a new PR https://github.com/openstack-k8s-operators/ci-framework/pull/772 instead of this. 
Hence, closing the PR

PR #216, opened 2023-05-30
Create tempest role

arxcruz:
This is a *very* simple role to run tempest.
At the moment it does have the following:
* tempest_list_allowed to list tests to be executed
* tempest_list_skipped to list tests to be excluded
* Configure the include and exclude list of tests
* Run tempest via podman command line In the future, the podman run will be replace with a pod.

This PR has:
- [x] Appropriate testing (molecule, python unit tests)
- [x] Appropriate documentation (README in the role, main README is up-to-date)

PR #330, opened 2023-06-23
Set cifmw_run_tests to false by default and run tempest on e2e jobs

raukadah:
https://github.com/openstack-k8s-operators/ci-framework/pull/216
enable tempest on end to end and edpm job.

Currently edpm and baremetal jobs run on multiple projects. Since it
is added recently. it is failing on EDPM job. It will block the prs.

This change changes the default value for cifmw_run_tests to false
and enable it in end to end jobs only.

### DIFF
--- a/deploy-edpm.yml
+++ b/deploy-edpm.yml
@@ -30,7 +30,7 @@
 
 - name: Import run test playbook
   ansible.builtin.import_playbook: ci_framework/playbooks/08-run-tests.yml
-  when: cifmw_run_tests | default('true') | bool
+  when: cifmw_run_tests | default('false') | bool
 
 - name: Run log related tasks
   ansible.builtin.import_playbook: ci_framework/playbooks/99-logs.yml
--- a/zuul.d/edpm.yaml
+++ b/zuul.d/edpm.yaml
@@ -48,8 +48,6 @@
     name: cifmw-crc-podified-edpm-baremetal
     parent: cifmw-base-crc-openstack
     run: ci/playbooks/edpm_baremetal_deployment/run.yml
-    vars:
-      cifmw_run_tests: false
 
 # Install Yamls specific job
 - job:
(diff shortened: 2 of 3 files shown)


deploy-edpm.yml
```diff
@@ -30,7 +30,7 @@
 
 - name: Import run test playbook
   ansible.builtin.import_playbook: ci_framework/playbooks/08-run-tests.yml
-  when: cifmw_run_tests | default('true') | bool
+  when: cifmw_run_tests | default('false') | bool
```
viroel, line 33: Now you need to enable it on edpm job, right? You are removing from both edpm and baremetal job.
Is edpm job passing with tempest?

viroel (changes requested):

zuul.d/edpm.yaml
```diff
@@ -48,8 +48,6 @@
     name: cifmw-crc-podified-edpm-baremetal
     parent: cifmw-base-crc-openstack
     run: ci/playbooks/edpm_baremetal_deployment/run.yml
-    vars:
-      cifmw_run_tests: false
 
 # Install Yamls specific job
 - job:
```
arxcruz, line 53: Need to add here cifmw_run_tests: true

deploy-edpm.yml
raukadah: @viroel @pablintino thank you for the reviews. The job was running on edpm and end to end job and skipped on bm job. Since edpm job is running on another projects also. tempest fails there it will block the other project's pr. In order to avoid we are making this change. Once it is stabilized. We will enable it everywhere.

zuul.d/edpm.yaml
raukadah: Added it to the scenario files

deploy-edpm.yml
viroel: ok, I didn't get that when reading PR message at the first time.

PR #333, opened 2023-06-26
edpm_prepare: custom timeout for ControlPlane

queria:
Enable setting custom timeout for ControPlane deployment getting Ready. In some cases (esp nested virtualization) it seems it might need extra time.

  - [ ] README in the role


queria:
This seems as not needed - the actual time used in case of success run was about 1.5min, so the timeout failure which can happen in that step is completely different issue.

PR #339, opened 2023-06-28
Patch CSV for ansible runner image

rabi:

### DIFF
--- a/ci_framework/roles/edpm_deploy_baremetal/tasks/main.yml
+++ b/ci_framework/roles/edpm_deploy_baremetal/tasks/main.yml
@@ -28,14 +28,24 @@
       {% endif %}
     cifmw_edpm_deploy_baremetal_operators_build_output: "{{ operators_build_output }}"
 
-- name: Install OpenStack operator
+- name: Install Dataplane with baremetal
   vars:
-    make_edpm_compute_baremetal_env: "{{ cifmw_edpm_deploy_baremetal_common_env |
+    make_edpm_baremetal_env: "{{ cifmw_edpm_deploy_baremetal_common_env |
       combine(cifmw_edpm_deploy_baremetal_make_openstack_env | from_yaml)}}"
-    make_edpm_compute_baremetal_dryrun: "{{ cifmw_edpm_deploy_baremetal_dry_run }}"
+    make_edpm_baremetal_dryrun: "{{ cifmw_edpm_deploy_baremetal_dry_run }}"
   ansible.builtin.include_role:
     name: 'install_yamls_makes'
-    tasks_from: 'make_edpm_compute_baremetal'
+    tasks_from: 'make_edpm_baremetal'
+
+- name: Wait for Ironic to be ready
+  when: not cifmw_edpm_deploy_baremetal_dry_run
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  ansible.builtin.command:
+    cmd: >-
+      oc wait pod -l name=ironic -n baremetal-operator-system --for=condition=Ready
+      --timeout={{ cifmw_edpm_deploy_baremetal_wait_ironic_timeout_mins }}m
 
 - name: Wait for OpenStack Provision Server pod to be created
   when: not cifmw_edpm_deploy_baremetal_dry_run
@@ -63,18 +73,6 @@
       --for condition=Available
       --timeout={{ cifmw_edpm_deploy_baremetal_wait_provisionserver_timeout_mins }}m
 
-- name: Get list of baremetal nodes
-  when: not cifmw_edpm_deploy_baremetal_dry_run
-  environment:
-    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
-    PATH: "{{ cifmw_path }}"
-  ansible.builtin.command:
-    cmd: >-
-      oc get baremetalhost
-      -n {{ cifmw_install_yamls_defaults['NAMESPACE'] }}
-      -o=jsonpath='{.items[*].metadata.name}'
-  register: cifmw_edpm_deploy_baremetal_bmh_output
-
 - name: Wait for baremetal nodes to reach 'provisioned' state
   when: not cifmw_edpm_deploy_baremetal_dry_run
   environment:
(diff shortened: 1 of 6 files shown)


ci_framework/roles/edpm_prepare/tasks/main.yml
```diff
@@ -101,6 +101,19 @@
...
+    - not cifmw_edpm_prepare_dry_run
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  ansible.builtin.command:
+    cmd: >-
+      oc patch csv -n openstack-operators openstack-ansibleee-operator.v0.0.1
+      --type='json' -p='[{"op":"replace",
+      "path":"/spec/install/spec/deployments/0/spec/template/spec/containers/1/env/0",
+      "value": {"name": "ANSIBLEEE_IMAGE_URL_DEFAULT", "value": ${ANSIBLEEE_IMAGE_URL_DEFAULT}}}]'
```
Sandeepyadav93, line 115: @rabi I don't think I can update your PR as it was sent from your local fork. I was thinking of updating with something like below using ansible role var here:

~~~
-      "value": {"name": "ANSIBLEEE_IMAGE_URL_DEFAULT", "value": ${ANSIBLEEE_IMAGE_URL_DEFAULT}}}]'
+      "value": {"name": "ANSIBLEEE_IMAGE_URL_DEFAULT", "value": {{cifmw_edpm_prepare_ansible_runner_image}}}}]'
~~~


and specify in defaults/main.yaml

~~~
+cifmw_edpm_prepare_ansible_runner_image: "http://quay.io/openstack-k8s-operators/openstack-ansibleee-runner:latest"
~~~

and maybe we can only patch based on condition
~~~
when: cifmw_edpm_prepare_ansible_runner_image != "http://quay.io/openstack-k8s-operators/openstack-ansibleee-runner:latest"
~~~

ci_framework/roles/edpm_deploy_baremetal/tasks/main.yml
```diff
@@ -28,14 +28,14 @@
       {% endif %}
     cifmw_edpm_deploy_baremetal_operators_build_output: "{{ operators_build_output }}"
 
-- name: Install OpenStack operator
+- name: Install Dataplane with baremetal
   vars:
     make_edpm_compute_baremetal_env: "{{ cifmw_edpm_deploy_baremetal_common_env |
```
raukadah, line 33: Since we are changing the target name here. We also need to change the var names associated with `make_edpm_baremetal` target.
Here is the generated [taskfile](https://logserver.rdoproject.org/39/339/1c9ae0dbd553a046a537672445a9725bff4c90a5/github-check/ci-framework-crc-podified-edpm-baremetal/3385591/ci-framework-data/artifacts/roles/install_yamls_makes/tasks/make_edpm_baremetal.yml) for `make_edpm_baremetal`

We need to change make_edpm_compute_baremetal_env to make_edpm_baremetal_env and same for dry run var.

rabi: ack, changed.. But seems to have  not created any issue without those vars but failed attaching the interface as we did run the attach_default_interface make target in edpm_prepare. 

https://logserver.rdoproject.org/39/339/a173897b0f87282b7c5ddcd6c5bbffbd8abc961b/github-check/ci-framework-crc-podified-edpm-baremetal/5af6a1a/ci-framework-data/logs/ci_make_004_run_edpm_baremetal.log

raukadah: Yes, `make crc_attach_default_interface` is enabled by default in edpm-prepare. Thank you for disabling it.

raukadah: New vars are applied here defined under[ BM scenario file]( https://github.com/openstack-k8s-operators/ci-framework/blob/main/scenarios/centos-9/edpm_baremetal_deployment_ci.yml#L4) : https://logserver.rdoproject.org/39/339/b5bdd79e0a50731af3b14a4af5b4358bbac96ef2/github-check/ci-framework-crc-podified-edpm-baremetal/fb1c415/ci-framework-data/logs/ci_make_003_run_edpm_baremetal.log

raukadah:
recheck due to "error: You must be logged in to the server (Unauthorized)"

Sandeepyadav93:
recheck

rabi:
recheck

ci_framework/roles/edpm_prepare/defaults/main.yml
```diff
@@ -23,3 +23,5 @@ cifmw_edpm_prepare_skip_openstack_operator: false
 cifmw_edpm_prepare_wait_subscription_retries: 30
 cifmw_edpm_prepare_dry_run: false
 cifmw_edpm_prepare_skip_crc_storage_creation: false
+cifmw_edpm_prepare_default_ansible_runner_image: "http://quay.io/openstack-k8s-operators/openstack-ansibleee-runner:latest"
```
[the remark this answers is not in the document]
rabi: We would remove DATAPLANE_RUNNER_IMG from install yamls, the default is hardcoded everywhere https://github.com/openstack-k8s-operators/openstack-ansibleee-operator/blob/main/api/v1alpha1/openstack_ansibleee_types.go#L31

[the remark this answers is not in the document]
rabi: Also DATAPLANE_RUNNER_IMG can be changed, we would need to patch the csv in every case where it does not match with the operator default[1] 

[1] https://github.com/openstack-k8s-operators/openstack-ansibleee-operator/blob/main/config/default/manager_default_images.yaml#L15

rabi:
recheck

rabi:
recheck

rabi:
recheck

rabi:
recheck

kajinamit:
/lgtm

We'd want to use CR input at some point. However the current design does not allow easy kusomize, so I'm ok with this.

rabi:
> /lgtm
> 
> We'd want to use CR input at some point. However the current design does not allow easy kusomize, so I'm ok with this.

You mean patch all service CRs with the dynamically generated image. I think the ENV vars are there for this and patching is the CSV is the legit approach[1].

[1] https://github.com/openstack-k8s-operators/docs/blob/main/webhooks.md#changing-webhook-defaults-in-olm-context

PR #340, opened 2023-06-28
Porting create_role_molecule script to Python

adrianfusco:
  - [x] README in the role


Makefile
```diff
@@ -41,7 +41,7 @@ create_new_role:
 
 .PHONY: role_molecule
 role_molecule: ## Regenerate the molecule jobs configuration
-	bash scripts/create_role_molecule
+	python3 scripts/create_role_molecule.py
```
pablintino, line 44: I'd prefer to skip this by making the script executable, just add the proper shebang `#!/usr/bin/env python3
` and add exec permissions to the file. With that you can avoid changing this line.

requirements.txt
```diff
@@ -1,2 +1,3 @@
 pbr>=1.6
 requests
+jinja2
```
pablintino, line 3: This requirements file is used for all the ci-framework, what means that jinja2 (that is already in use, but embedded in ansible) will be added there too, that's why I preferred to try to not to use external libs. Anyway, we can get read of this by launching the script in a properly configure container using podman, like other scripts in the framework.

scripts/create_role_molecule.py
```diff
@@ -0,0 +1,124 @@
...
+    molecule_jobs_content = [
+        '# DO NOT EDIT - generated using make role_molecule'
+    ]
+
+    projects_yaml_template_path = TEMPLATES_DIR / 'projects.yaml'
+    try:
+        with open(projects_yaml_template_path, 'r') as projects_template_file:
+            projects_yaml_content = projects_template_file.read().splitlines()
+    except (FileNotFoundError, PermissionError, OSError) as e:
+        print(f"Error reading projects.yaml template file: {e}")
```
pablintino, line 83: I'd like to see the return code of the script being set to non-zero in case of failure.

scripts/create_role_molecule.py
```diff
@@ -0,0 +1,124 @@
...
+
+    molecule_jobs_content = [
+        '# DO NOT EDIT - generated using make role_molecule'
+    ]
+
+    projects_yaml_template_path = TEMPLATES_DIR / 'projects.yaml'
+    try:
+        with open(projects_yaml_template_path, 'r') as projects_template_file:
+            projects_yaml_content = projects_template_file.read().splitlines()
+    except (FileNotFoundError, PermissionError, OSError) as e:
```
pablintino, line 82: Redundant, as FileNotFoundError and PermissionError are both subclasses of OSError.

scripts/create_role_molecule.py
```diff
@@ -0,0 +1,124 @@
...
+
+        molecule_jobs_content.append(iteration_result.strip())
+
+        readme_path = role_directory / 'README.md'
+        link_path = SOURCE_ROLES_DOCS_DIR / f"{role_name}.md"
+
+        if not create_symbolic_link(readme_path, link_path, role_name):
+            print(f"Failed to create symbolic link ({link_path}) for the role {role_name}.")
+
+        projects_yaml_content.append(f'        - cifmw-molecule-{role_name}')
```
pablintino, line 109: Seems like a dirty way of appending content... I wouldn't expect to take indentation into account here.

scripts/create_role_molecule.py
```diff
@@ -0,0 +1,124 @@
...
+        return
+
+    molecule_jobs_content = [
+        '# DO NOT EDIT - generated using make role_molecule'
+    ]
+
+    projects_yaml_template_path = TEMPLATES_DIR / 'projects.yaml'
+    try:
+        with open(projects_yaml_template_path, 'r') as projects_template_file:
+            projects_yaml_content = projects_template_file.read().splitlines()
```
pablintino, line 81: Why not loading and treating it like yaml? That way you can avoid tricky things like https://github.com/openstack-k8s-operators/ci-framework/pull/340/commits/abcde36d1c1c355561290db37fbe44045aa8779d#diff-5d3878a84b579ecca7e889822946d5e58c6df4c4ef136f7646baea74e85d8ec7R109

scripts/create_role_molecule.py
```diff
@@ -0,0 +1,124 @@
...
+PROJECT_DIR = Path(__file__).resolve().parent.parent
+SCRIPT_DIR = PROJECT_DIR / 'scripts'
+TEMPLATES_DIR = PROJECT_DIR / 'ci' / 'templates'
+ROLES_DIR = PROJECT_DIR / 'ci_framework' / 'roles'
+ZUUL_JOB_DIR = PROJECT_DIR / 'zuul.d'
+SOURCE_ROLES_DOCS_DIR = PROJECT_DIR / 'docs' / 'source' / 'roles'
+
+# These roles will use centos-9-crc-xl nodeset and it will be used to replace the content 
+# of the Jinja template.
+ROLES_NEED_CRC_XML = [
```
pablintino, line 16: How are gonna provide custom properties like timeouts?

scripts/create_role_molecule.py
```diff
@@ -0,0 +1,124 @@
...
+    for role_directory in ROLES_DIR.iterdir():
+        if not role_directory.is_dir():
+            print("Skipping. Not a role directory")
+            continue
+
+        role_name = role_directory.name
+        iteration_result = template.render(ROLE=role_name)
+
+        if role_name in ROLES_NEED_CRC_XML:
+            iteration_result = re.sub(
```
pablintino, line 95: I wouldn't expect to use regexes, and in fact, I doubt if jinja gives here any value. The same thing can be done easier by just loading the template as a yaml, and replacing there, in the resulting dict, each needed key and value.

requirements.txt
cjeanner: +1 - I'd rather use `f'foo {blah}'` or any core templating feature.

scripts/create_role_molecule.py
cjeanner: maybe getting a YAML file as a source of customization is a thing?

i.e:
```YAML
# custom-molecule.yml
operator_deploy:
  nodeset: crc_xxl
  timeout: 4800
rhol_crc:
  nodeset: crc_xxl
....
```

nested in the `ci` directory, and used in that script. And if a job (or specific parameters) aren't defined there, we'll fallback on the default things.

scripts/create_role_molecule.py
cjeanner: we can either edit the YAML structure, or just override the existing file without any thoughts. I'd go for the latter, easier to maintain in the future.

scripts/create_role_molecule.py
cjeanner: indeed. Proper templating, please? :)

scripts/create_role_molecule.py
cjeanner: right. treating the data as yaml and making use of `yaml.dump()` should be preferred. thus `yaml.loads` is a key element.

scripts/create_role_molecule.py
pablintino: Exactly, that was the idea we discussed, seems like easy to do and maintain

ci/config/molecule.yaml
```diff
@@ -0,0 +1,14 @@
...
+    nodeset: centos-9-crc-xl
+- job:
+    name: cifmw-molecule-openshift_provisioner_node
+    nodeset: centos-9-crc-xl
+- job:
+    name: cifmw-molecule-openshift_setup
+    nodeset: centos-9-crc-xl
+- job:
+    name: cifmw-molecule-rhol_crc
+    nodeset: centos-9-crc-xlss
```
pablintino, line 12: Typo

scripts/create_role_molecule.py
```diff
@@ -0,0 +1,126 @@
...
+    """
+    Generate YAML structure for a Zuul job based on given role.
+
+    Args:
+        role_name (str): Name of the role.
+
+    Returns:
+        dict: YAML structure for the role.
+    """
+    zuul_job_base_yaml = {
```
pablintino, line 25: Why not putting this on a file? It will be easier to maintain. Maybe this can be a jinja template... That way will be, get the jinja2 template, from that create the base yaml, override the values from the yaml with your current yaml merge logic and append it to the resulting molecule.yml

adrianfusco:
@pablintino @cjeanner I've modified the script.

- I have a function `generate_zuul_job_role_yaml` with the basic zuul job structure, all the roles will be created by default with this information.
- I've created a `ci/config/molecule.yaml` file with the extra information of each job. If we need to add extra directives to a zuul jobs it could be done here, e.g. if we need to add a nodeset and timeout to a specific job we will have in this YAML the following structure:

```
- job:
    name: cifmw-molecule-openshift_login
    nodeset: centos-9-crc-xl
    timeout: 3600
 ```
 And the information of this file will be merged with the jobs that are being generated also in the script.
 I think we can avoid using templates and just use `yaml` (as we talked @pablintino ).
 
 Let me know what do you think about this approach. Thank you.

File: scripts/create_role_molecule.py [hunk 11]
```diff
@@ -0,0 +1,126 @@
...
+
+    for role_directory in ROLES_DIR.iterdir():
+        if not role_directory.is_dir():
+            print("Skipping. Not a role directory")
+            continue
+
+        try:
+            role_yaml = generate_zuul_job_role_yaml(role_directory.name)
+            molecule_yaml_zuul_jobs.append(role_yaml)
+        except Exception as e:
```
pablintino, line 86: Why is this needed? Re raising it here means nothing...

ci/config/molecule.yaml
adrianfusco: Removed. Thx.

scripts/create_role_molecule.py
adrianfusco: True. Removed. Thx.

scripts/create_role_molecule.py
adrianfusco: Changed into a template.

scripts/create_role_molecule.py
```diff
@@ -0,0 +1,109 @@
...
+
+    role_directories = [entry.name for entry in ROLES_DIR.iterdir() if entry.is_dir()]
+    molecule_yaml_zuul_jobs = generate_zuul_job_role_yaml(role_directories)
+
+    molecule_ci_config_yaml = CI_CONFIG_DIR / 'molecule.yaml'
+    try:
+        with open(molecule_ci_config_yaml, 'r') as file:
+            molecule_ci_jobs_info = yaml.safe_load(file)
+    except FileNotFoundError as e:
+        raise e
```
pablintino, line 77: What's the point of this?

scripts/create_role_molecule.py
```diff
@@ -0,0 +1,109 @@
...
+    try:
+        with open(project_yaml_path, 'r') as file:
+            projects_jobs_info = yaml.safe_load(file)
+    except FileNotFoundError as e:
+        raise e
+
+    # Add each role name after the generated content
+    for role_directory in ROLES_DIR.iterdir():
+        if not role_directory.is_dir():
+            print("Skipping. Not a role directory")
```
pablintino, line 50: Maybe add some context, like the role_directory to the log. PS: I'm (and most python devs too) against using plain print, maybe this can be added as a warning using the builtin logger

scripts/create_role_molecule.py
```diff
@@ -0,0 +1,109 @@
...
+    Args:
+        roles_name (str): Name of the role.
+
+    Returns:
+        dict: YAML structure for the role.
+    """
+    template_loader = FileSystemLoader(searchpath=CI_TEMPLATES_DIR)
+    template_env = Environment(loader=template_loader)
+    template = template_env.get_template('molecule.yaml.j2')
+    zuul_rendered_template_jobs = yaml.safe_load(template.render(role_names=roles_name))
```
pablintino, line 30: Style: Why not merging assignation and return clauses? Just a question, no need to change if you don't want to.

scripts/create_role_molecule.py
```diff
@@ -0,0 +1,109 @@
...
+    return zuul_rendered_template_jobs
+
+
+def regenerate_projects_zuul_jobs_yaml():
+    """
+    Regenerate zuul.d/projects.yaml file with the name of each role
+    under ci-framework/roles
+    """
+    project_yaml_path = CI_TEMPLATES_DIR / 'projects.yaml'
```
pablintino, line 39: If it's constant can be perfectly moved to the top of the file with the other "constants" https://github.com/openstack-k8s-operators/ci-framework/pull/340/files#diff-5d3878a84b579ecca7e889822946d5e58c6df4c4ef136f7646baea74e85d8ec7R10-R14

ci/config/molecule.yaml
```diff
@@ -0,0 +1,12 @@
+- job:
```
cjeanner, line 1: I'd rather use the role name instead of the job name. It would avoid some typos imho.

scripts/create_role_molecule.py
cjeanner: in addition... I'd use `os.path.join()`

ci/config/molecule.yaml
adrianfusco: Sorry, I don't get this. This YAML will be used for being merged to what the j2 generates based on the name of the job.

```
{% for role_name in role_names %}
- job:
    name: cifmw-molecule-{{ role_name }}
    parent: cifmw-molecule-base
    vars:
      TEST_RUN: {{ role_name }}
    files:
      - ^ansible-requirements.txt
      - ^molecule-requirements.txt
      - ^ci_framework/roles/{{ role_name }}/(?!meta|README).*
      - ^ci/playbooks/molecule.*
{% endfor %}
```

scripts/create_role_molecule.py
```diff
@@ -0,0 +1,109 @@
...
+from jinja2 import Environment, FileSystemLoader
+
+
+PROJECT_DIR = Path(__file__).resolve().parent.parent
+SCRIPT_DIR = PROJECT_DIR / 'scripts'
+CI_TEMPLATES_DIR = PROJECT_DIR / 'ci' / 'templates'
+CI_CONFIG_DIR = PROJECT_DIR / 'ci' / 'config'
+ROLES_DIR = PROJECT_DIR / 'ci_framework' / 'roles'
+ZUUL_JOB_DIR = PROJECT_DIR / 'zuul.d'
+SOURCE_ROLES_DOCS_DIR = PROJECT_DIR / 'docs' / 'source' / 'roles'
```
cjeanner, line 14: I'd use standard `os.path.join()`

scripts/create_role_molecule.py
```diff
@@ -0,0 +1,109 @@
...
+
+    Args:
+        roles_name (str): Name of the role.
+
+    Returns:
+        dict: YAML structure for the role.
+    """
+    template_loader = FileSystemLoader(searchpath=CI_TEMPLATES_DIR)
+    template_env = Environment(loader=template_loader)
+    template = template_env.get_template('molecule.yaml.j2')
```
cjeanner, line 29: wondering why we'd need to rely on an external template and use jinja2, when we may be able to get it in plain "inline" python ?

scripts/create_role_molecule.py
adrianfusco: I think is okay since we are not doing low level path manipulation and this lib is included from Python 3.4

scripts/create_role_molecule.py
```diff
@@ -0,0 +1,109 @@
...
+            yaml.dump(projects_jobs_info, projects_file)
+    except Exception as e:
+        raise e
+
+
+def regenerate_molecule_zuul_jobs_yaml():
+    """
+    Regenerate zuul.d/molecule.yaml file with role-specific YAML structures.
+    """
+    molecule_yaml_zuul_jobs = []
```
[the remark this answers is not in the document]
adrianfusco: Deleted. Thx.

scripts/create_role_molecule.py
```diff
@@ -0,0 +1,109 @@
...
+    Regenerate zuul.d/projects.yaml file with the name of each role
+    under ci-framework/roles
+    """
+    project_yaml_path = CI_TEMPLATES_DIR / 'projects.yaml'
+
+    try:
+        with open(project_yaml_path, 'r') as file:
+            projects_jobs_info = yaml.safe_load(file)
+    except FileNotFoundError as e:
+        raise e
```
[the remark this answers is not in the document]
adrianfusco: I've removed the `raises`s. Thx.

scripts/create_role_molecule.py
cjeanner: fair enough.

cjeanner:
Since the code is a bit more complicated than before, it may be useful to get a unit test in order to ensure we're generating the content properly. Care to work on that and integrated it in the CI directly? That would also ensure future changes won't break existing content.

pablintino:
> Since the code is a bit more complicated than before, it may be useful to get a unit test in order to ensure we're generating the content properly. Care to work on that and integrated it in the CI directly? That would also ensure future changes won't break existing content.

+1

ci_framework/tests/unit/test_create_role_molecule.py
cjeanner, line 1: Beware - this tree here isn't for "standard python tests" - it's for ansible only. Meaning it's running via ansible-test; and ansible-test is running in a "sealed" environment, relying only on core python libraries.

In order to unit-test the script itself, we'd need to either create `/tests`, or maybe `/scripts/tests` (better imho) and bootstrap some venv and, probably, call out `tox` or related.

ci/config/molecule.yaml
cjeanner: oh, like that. OK, makes sense then.

docs/source/development/02_molecule.md
```diff
@@ -6,15 +6,29 @@ structure, including the molecule part.
 
 [More information about molecule](https://molecule.readthedocs.io/)
 
+# Testing
```
cjeanner, line 9: this should probably be an h2, not h1 (so `## Testing`)

scripts/setup_env
```diff
@@ -80,7 +80,7 @@ if [ ${USE_VENV} == 'yes' ]; then
 fi
 
 # Run bindep
-"${PIP}" install pip setuptools bindep --upgrade
+"${PIP}" install pip setuptools bindep tox --upgrade
```
cjeanner, line 83: `tox` should be in the test-requirements.txt file instead.

Makefile
cjeanner, line 1: Apparently you're missing the "tox" target in there? Care to ensure you're rebased against `main` (avoid the "merge", because... well, we don't want to get "merge" in a PR).

You'll also need to update the `make tox` target to actually consume the tests.

adrianfusco: In this case the tests have been consumed. In fact yesterday I had an error in the tests from this same PR: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openstack-k8s-operators_ci-framework/340/pull-ci-openstack-k8s-operators-ci-framework-main-tox/1678427964779270144

I fixed and now it passes but the job in Prow doesn't show detailed information when it passes.

scripts/create_role_molecule.py
```diff
@@ -0,0 +1,137 @@
...
+    """
+    Regenerate zuul.d/projects.yaml file with the name of each role
+    under ci-framework/roles
+    """
+    with open(generated_paths['ci_templates_dir'] / 'projects.yaml', 'r') as file:
+        projects_jobs_info = yaml.safe_load(file)
+
+    # Add each role name after the generated content
+    for role_directory in generated_paths['roles_dir'].iterdir():
+        print(role_directory)
```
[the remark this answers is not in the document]
adrianfusco: Deleted print that was for debugging. Thx.

scripts/create_role_molecule.py
```diff
@@ -0,0 +1,137 @@
...
+# License for the specific language governing permissions and limitations
+# under the License.
+
+from pathlib import Path
+import yaml
+import logging
+from jinja2 import Environment, FileSystemLoader
+
+
+def get_project_paths(project_dir=Path(__file__).resolve().parent.parent):
```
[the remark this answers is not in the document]
adrianfusco: You are right. I'll keep in mind. Thx.

cjeanner:
recheck

PR #341, opened 2023-06-28
Collect output of cluster-info command

raukadah:
It will be useful to collect more logs about
the cluster.

### DIFF
--- a/ci_framework/roles/artifacts/tasks/cluster_info.yml
+++ b/ci_framework/roles/artifacts/tasks/cluster_info.yml
@@ -78,3 +78,13 @@
         chdir: "{{ cifmw_artifacts_basedir }}/logs/edpm"
       changed_when: true
       ignore_errors: true
+
+    - name: Dump cluster info for debugging
+      ansible.builtin.shell: |
+        mkdir cluster_info
+        oc cluster-info dump --all-namespaces -o yaml --output-directory={{ cifmw_artifacts_basedir }}/logs/cluster_info
+        oc cluster-info -o yaml > {{ cifmw_artifacts_basedir }}/logs/cluster_info.yaml
+      args:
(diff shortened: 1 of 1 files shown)


ci_framework/roles/artifacts/tasks/cluster_info.yml
```diff
@@ -78,3 +78,13 @@
         chdir: "{{ cifmw_artifacts_basedir }}/logs/edpm"
       changed_when: true
       ignore_errors: true
+
+    - name: Dump cluster info for debugging
+      ansible.builtin.shell: |
+        mkdir cluster_info
+        oc cluster-info dump --all-namespaces -o yaml --output-directory={{ cifmw_artifacts_basedir }}/logs/cluster_info
+        oc cluster-info -o yaml > {{ cifmw_artifacts_basedir }}/logs/cluster_info.yaml
```
cjeanner, line 86: since you use the `chdir` parameter, there's no need to put the full path for the output-direcory, nor the shell redirect... ?

cjeanner (commented):
It may be good to ensure we're not having duplicated logs - I see we're gathering a lot of things from crc/openshift in various ways.

Maybe better to collaborate with @Sandeepyadav93 and the "must-gather" thingy so that we consolidate all of the openshift logs/data/info in one place?

raukadah:
> It may be good to ensure we're not having duplicated logs - I see we're gathering a lot of things from crc/openshift in various ways.
> 
> Maybe better to collaborate with @Sandeepyadav93 and the "must-gather" thingy so that we consolidate all of the openshift logs/data/info in one place?

Thanks you @cjeanner, closing this pr in favor of https://github.com/openstack-k8s-operators/ci-framework/pull/359

PR #343, opened 2023-06-30
Collect logs for BMO and ironic

rabi:
We need to collect the logs for baremetal-operator and ironic used for baremetal provisioning.


cjeanner (commented):
/approve

Thanks @rabi for that new addition!

PR #348, opened 2023-07-03
Use bootstrap playbook before building edpm images

raukadah:
In cifmw-base-edpm-build-images job, we run repo-setup before running edpm_build_images role.

In downstream, we run periodic version of build images using cifmw-base-edpm-build-images job.

we need to install and configure ca_cert. install_ca role already takes care of that. repo_setup and install_ca roles are called as a part of 01-bootstrap.yml playbook using `packages` tag.

This patch calls 01-bootstrap.yml playbook using `packages` tag and drop repo_setup role call.


ci/playbooks/edpm_build_images/run.yml
```diff
@@ -16,3 +16,4 @@
           {%-   endfor %}
           {%- endif %}
           -e @scenarios/centos-9/zuul_inventory.yml
+          --tags packages
```
Sandeepyadav93, line 19: We don't have this tag here everywhere in edpm_build_images role, ex not on these plays: https://github.com/openstack-k8s-operators/ci-framework/blob/d002a528481727d5d673d4053bed943819082f3a/ci_framework/roles/edpm_build_images/tasks/run.yml

raukadah:
recheck

cjeanner:
recheck

cjeanner:
recheck

PR #349, opened 2023-07-03
[DNM] repo_setup env variable

raukadah:
  - [ ] README in the role


cjeanner:
Shouldn't we consume the install_ca role instead, since it would ensure the internal CA(s) are properly installed and, then, consumed? Or is this, once again, linked to python-request not consuming CA from the system?

raukadah:
> Shouldn't we consume the install_ca role instead, since it would ensure the internal CA(s) are properly installed and, then, consumed? Or is this, once again, linked to python-request not consuming CA from the system?

I tried to use install_ca role but it was not working then I tried to pass the env variable at import_task level where [repo_setup](https://github.com/openstack-k8s-operators/ci-framework/commit/9703e051d23c3176bda291a02d1c18585aa806a2#diff-2e7ef878da552db89bb522eb96d7187e9aa1be6d6f7478831478a29c885c1701R39) is called. In both cases it was not working. that's why i went via this route. It might be linked with python-request.

cjeanner:
/lgtm

python-requests is a pain since they switched to certifi library X(.

cjeanner:
/lgtm

even with doc - that's perfect :)

ci_framework/roles/repo_setup/tasks/artifacts.yml
```diff
@@ -1,15 +1,17 @@
...
+      {{ cifmw_repo_setup_basedir }}/venv/bin/repo-setup-get-hash
+      --dlrn-url {{ cifmw_repo_setup_dlrn_uri[:-1] }}
+      --os-version {{ cifmw_repo_setup_os_release }}{{ cifmw_repo_setup_dist_major_version }}
+      --tag {{cifmw_repo_setup_promotion }}
+      --release {{ cifmw_repo_setup_branch }}
+      {% if cifmw_repo_setup_component is defined %}
+      --component {{ cifmw_repo_setup_component }}
+      {% endif %}
       --json
+  environment: "{{ cifmw_repo_setup_env | default ({}) }}"
```
Sandeepyadav93, line 14: We already have defaults {} in defaults/main.yaml so | default ({}) is redundant so okay to keep it.

Sandeepyadav93 (commented):
/lgtm

PR #350, opened 2023-07-03
Set proper image element files based on distro

raukadah:
Image element files in edpm-image-builder repo are ending with 9-stream.yaml for centos and -9.yaml for Red Hat. We need to add conditional in order to make sure proper image elements file are passed to diskimage builder otherwise edpm image building will fail.


cjeanner:
/lgtm

makes sense.

bshewale:
looks good to me :)

Thanks @raukadah

PR #353, opened 2023-07-03
Avoid "always" tag on install_yamls content

cjeanner:
install_yamls role should be launched only in two cases:
- we add --tags boostrap
- we don't add any tag

Apparently, with the "always" being there, the tasks were launched even
though we wanted only the "packages" tag.


cjeanner:
recheck

cjeanner:
@raukadah if you use that patch for #348, you should get out of your issues. Lemme know!

cjeanner:
OK, so #348 is now green with this last iteration. IMHO we can go with it.
The thing I understand:
`import_role` is "flattened" during catalogue generation, and the `tags` associated to that task are added to the nested `tasks`. Since we had `always` tag in the install_yamls role, the `bootstrap` tag was then *added* to them, meaning all tasks in install_yamls had `bootstrap,always` as tags. Meaning, basically, "run always unless we --skip-tags bootstrap" or the like.

Not sure if this would be different with the `include_role` though, since it dynamically include at runtime, but I'd rather stick to the current `import_role` for the sake of clarity and, to some extend, speed.

psathyan (approved):

PR #358, opened 2023-07-04
Add a new tag we can then filter on

cjeanner:
Since adding --tags packages prevent running the second playbook tasks,         
let's take advantage of the behavior of import_role with tags: it will          
basically apply the passed tags down to the nested content, allowing            
then to filter on a second, dedicated tag without the need to actually          
edit the role itself.


raukadah (approved):
/approve

Sandeepyadav93:
/lgtm

thank you for the fix!

PR #359, opened 2023-07-05
Add os_must_gather role

raukadah:
os_must_gather role uses openstack-must-gather project and
container image to gather logs from openshift cluster related
to openstack.
    
This role drops artifacts/tasks/cluster_info.yml task file
in favor of above role.


raukadah:
recheck

ci_framework/roles/os_must_gather/tasks/main.yml
```diff
@@ -0,0 +1,40 @@
...
+    cmd: >-
+      oc adm must-gather --image {{ cifmw_os_must_gather_image }}
+      {% if cifmw_os_must_gather_timeout is defined %}
+      --timeout {{ cifmw_os_must_gather_timeout }}
+      {% endif %}
+      {% if cifmw_os_must_gather_host_network | bool %}
+      --host-network=true
+      {% endif %}
+      --dest-dir {{ cifmw_os_must_gather_output_dir }}/logs > os_must_gather.log
+  args:
```
cjeanner, line 38: if you use the "cmd" parameter, you don't need to use the "args" level - just append the chdir and related at the same level.

ci_framework/playbooks/99-logs.yml
```diff
@@ -31,6 +31,12 @@
     - name: Generate artifacts
       ansible.builtin.import_role:
         name: artifacts
+      when: not cifmw_os_must_gather|bool | default('false')
```
cjeanner, line 34: IMHO one may not exclude the other - let's have a discussion with the must_gather folks and see where we do have overlap, where we don't, and how to properly converge.

cjeanner:
/hold

We're about to get a discussion with the must_gather team, let's check how we may converge, overlaps and all.

ci_framework/roles/os_must_gather/tasks/main.yml
```diff
@@ -0,0 +1,72 @@
...
+    - cifmw_openshift_kubeconfig is defined
+  block:
+    - name: Run openstack-must-gather command
+      environment:
+        KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+        PATH: "{{ cifmw_path }}"
+      ansible.builtin.command:
+        cmd: >-
+          oc adm must-gather --image {{ cifmw_os_must_gather_image }}
+          {% if cifmw_os_must_gather_timeout is defined %}
```
cjeanner, line 44: it's always defined, since it's in the defaults/main.yml

ci_framework/roles/os_must_gather/tasks/main.yml
```diff
@@ -0,0 +1,72 @@
...
+          oc delete pod -l app=kube-controller-manager -n openshift-kube-controller-manager
+
+    - name: Re-run openstack-must-gather command
+      environment:
+        KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+        PATH: "{{ cifmw_path }}"
+      ansible.builtin.command:
+        cmd: >-
+          oc adm must-gather --image {{ cifmw_os_must_gather_image }}
+          {% if cifmw_os_must_gather_timeout is defined %}
```
cjeanner, line 66: ditto - always defined

cjeanner (commented):
/unhold

must_gather will be used to collect openshift data from the API, while the "artifacts" will collect logs from the disk/system directly, so that we cover potential dead API.

raukadah:
@cjeanner Below are the logs from two scenario:
1. When must-gather task passes: https://logserver.rdoproject.org/59/359/04971bedfca1edfbcd63fcb74449c12b52becc52/github-check/ci-framework-crc-podified-edpm-baremetal/594dd84/ci-framework-data/logs/ 
2. When must-gather tasks fails due to any issue on cluster, then rescue tasks run: https://logserver.rdoproject.org/59/359/04971bedfca1edfbcd63fcb74449c12b52becc52/github-check/ci-framework-crc-podified-galera-deployment/6039c5b/ci-framework-data/logs/

Please have a look, Thank you :-)

ci_framework/roles/os_must_gather/tasks/main.yml
```diff
@@ -0,0 +1,57 @@
...
+- name: Check for oc command
+  ansible.builtin.command: command -v oc
+  environment:
+    PATH: "{{ cifmw_path | default(ansible_env.PATH) }}"
+  register: oc_installed
+  ignore_errors: true
+
+- name: Running openstack-must-gather tool
+  when:
+    - oc_installed is defined
```
pablintino, line 33: If it's registered it will be defined. I don't see how that fact cannot be defined.

ci_framework/roles/os_must_gather/tasks/main.yml
```diff
@@ -0,0 +1,57 @@
...
+          --host-network={{ cifmw_os_must_gather_host_network }}
+          --dest-dir {{ cifmw_os_must_gather_output_dir }}/logs > {{ cifmw_os_must_gather_output_dir }}/logs/os_must_gather.log
+  rescue:
+    - name: Inspect the cluster after must-gather failure
+      ignore_errors: true
+      environment:
+        KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+        PATH: "{{ cifmw_path }}"
+      ansible.builtin.shell: |
+        mkdir -p {{ cifmw_os_must_gather_output_dir }}/logs/oc_inspect
```
pablintino, line 54: Why not use ansible for this loop? The list of NSs to lookup can even be a variable with a default value.

ci_framework/roles/os_must_gather/tasks/main.yml
cjeanner: old habits - if you run ansible in check mode, commands/shells aren't executed, leading to undefined resources.

ci_framework/roles/os_must_gather/tasks/main.yml
cjeanner: each ansible task is time consuming - we may reconsider this, but I'm fine with this kind of loop in a shell.

pablintino: /lgtm

PR #360, opened 2023-07-06
Remove openstack_ci_bootstrap as it will not used now.

pablintino:
The role was added as part of a PoC to remove nested CRC but it wasn't used. Since the role is not yet documented and properly tested we remove it till proper doc, testing and use case arrives.

- [x] Appropriate testing is done and actually running.
- [x] Appropriate documentation exists and/or is up-to-date.


cjeanner:
/approve

Please re-generate the molecule job list, else it will still point to the non-existing role and crash.

cjeanner:
Oh and:
` docs/source/roles/openstack_ci_bootstrap.md: Broken symlink`

pablintino:
> /approve
> 
> Please re-generate the molecule job list, else it will still point to the non-existing role and crash.

Seems that the original PR missed to add that change, is not there.

PR #365, opened 2023-07-10
Create EDPM vm when deploy_edpm is set to true

raukadah:
deploy-edpm.yml is used in edpm, baremetal and podified deployment job. In podified deployment job, we do not do the EDPM deployment but vms were getting created.

It eats the space and lets the unwanted failure.

This patch disables the EDPM vm creation task and create it only when required.


cjeanner:
/approve

sounds legit - thanks for that one!

PR #367, opened 2023-07-10
Reload generated environment

cjeanner:
Doing so, we should ensure we're able to authenticate and call `oc`
command


raukadah (approved):
/approve

PR #361, opened 2023-07-06
Introduce new CI layout for end-to-end jobs

pablintino:
In order to improve stability while also allowing more flexibility with
"what we want to test", this patch introduces a new multinode layout for
CRC based end-to-end jobs.

cifmw-podified-edpm-base-multinode-crc provides the basis for the
3-nodes layout, providing one "ansible controller", a CRC instance and
one Compute node.
At the end of that base job, everything is ready to start running the CI
Framework - dependencies are installed, and you should get all of the
needed repositories, also for proper Depends-On support.
    
cifmw-podified-edpm-deployment-multinode-crc is an actual end-to-end
job, running the standard EDPM deployment with some tests.

- [X] new layout is in use and tested

PR #369, opened 2023-07-10
Ensure we're running the hook against localhost only

cjeanner:
the involved playbook runs against localhost. Though, in this case, it
may not properly be passed down - so let's just pass a static string as
inventory.


cjeanner:
With the new ability to set the connection type, we can properly ensure where the hook is running

cjeanner:
/approve

It now properly works and we shouldn't see any issue with the #361 change. We may depends-on in order to double-check that.

https://review.rdoproject.org/zuul/build/ef73c8f154994ac89a671f1bbfa1c713/log/ci-framework-data/logs/ansible.log-2023-07-22-06:51#805

raukadah (approved):
/approve

PR #370, opened 2023-07-11
Gather logs from standalone CRC over network

cjeanner:
Not using guestfish should even make it faster/better in the "nested
virt" case.


cjeanner:
recheck

PR #378, opened 2023-07-13
Some cleaning in ansible

cjeanner:
While trying to find why we get a weird "artifacts/artifacts" empty
directory, I found some other inconsistent content, like a double "/"
and a parameter that doesn't seem to be used anywhere.

  - [X] README in the role


cjeanner:
recheck

raukadah (approved):

PR #386, opened 2023-07-20
Fixed the syntax error

bshewale:
As in the "ci-script" plugins [1] we don't have 'chdir' option and that is causing syntax error in the periodic jobs [2].

[1]: https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci_framework/plugins/action/ci_script.py#L20
[2]: https://logserver.rdoproject.org/65/49365/1/check/periodic-edpm-container-image-quay-push-centos-9-master/cd9d1d4/job-output.txt

  - [x] README in the role

### DIFF
--- a/ci_framework/roles/copy_container/tasks/main.yml
+++ b/ci_framework/roles/copy_container/tasks/main.yml
@@ -38,8 +38,10 @@
   register: go_build
   ci_script:
     output_dir: "{{ cifmw_artifacts_basedir }}/artifacts"
-    cmd: go build
-    chdir: "{{ temporary_copy_container_dir.path }}"
+    cmd: |
+      pushd "{{ temporary_copy_container_dir.path }}"
+      go build
+      popd
   changed_when: false
 
 - name: Copy binary to /usr/local/bin (CentOS 8)


ci_framework/roles/copy_container/tasks/main.yml
```diff
@@ -38,8 +38,9 @@
   register: go_build
   ci_script:
     output_dir: "{{ cifmw_artifacts_basedir }}/artifacts"
-    cmd: go build
-    chdir: "{{ temporary_copy_container_dir.path }}"
```
cjeanner, line 42: ci_script should support the same parameters as the original ansible.builtin.script:
https://docs.ansible.com/ansible/latest/collections/ansible/builtin/script_module.html#parameter-chdir

there's something else then... Do you have some link?

cjeanner:
/hold

there's a need for investigation here - chdir SHOULD be supported.

ci_framework/roles/copy_container/tasks/main.yml
cjeanner: dang... ok, I'm wrong:
https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci_framework/plugins/action/ci_script.py#L131

ci_framework/roles/copy_container/tasks/main.yml
```diff
@@ -38,8 +38,9 @@
...
+    cmd: |
+      go build
```
cjeanner, line 42: you may want to
```
pushd {{ temporary_copy_container_dir.path }}"
go build
popd
```

Else, it would run the `go build` in the wrong place.

cjeanner:
/hold

Let's see if #388 solves the issue.

cjeanner:
We could get rid of the issues related to the `ci_script` and even properly use it in the periodic jobs. Closing this PR since it's not needed anymore.

PR #390, opened 2023-07-21
Store edpm-ansible scenario file in artifacts directory.

raukadah:
It will be useful for debugging.
It also fixes the case for zuul_change_list is undefined when e2e-prepare playbook is not used in the ci job.


raukadah:
recheck

cjeanner:
recheck

cjeanner (approved):
/approve

PR #391, opened 2023-07-21
Added missing namespaces in os_must_gather role

raukadah:
https://github.com/openstack-k8s-operators/openstack-must-gather/pull/3 adds cert-manager and openshift-nmstate namespace in openstack-must-gather.

The same needs to be updated to collect logs from these namespace if must-gather command fails.

It also includes metallb-system namespace in the same.


kajinamit:
recheck

PR #392, opened 2023-07-21
Use ci_script instead of shell in edpm_deploy role

raukadah:
It will help to log the output in a file, useful for debugging later.


cjeanner:
/approve

Consuming ci_script for that case makes sense. Note that we'll need to double-check the reproducer script actually exposes the environment variables.

raukadah:
recheck

cjeanner (approved):

PR #397, opened 2023-07-24
Remove dependency to community.libvirt from artifacts role

cjeanner:
Since we rely on ssh to gather logs from CRC directly, we don't really
care about the fact it's a VM or not. Its XML domain definition has no
real value either, so we can really just drop it.


raukadah (approved):

afazekas:
In case we want to collect libvirt info , likely we should do it for all vm  and network, where crc is not more equals then the others.

PR #399, opened 2023-07-24
Ensure copy_container role follows cifmw conventions

cjeanner:

Testproject: https://review.rdoproject.org/r/c/testproject/+/49365

### DIFF
--- a/ci_framework/roles/copy_container/defaults/main.yml
+++ b/ci_framework/roles/copy_container/defaults/main.yml
@@ -1,3 +1,9 @@
 ---
+# LEGAGY names
 enable_cron: true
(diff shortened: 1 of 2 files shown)


pablintino:
/lgtm

The question now is, how is using the legacy names? Just to track and delete them, and come back to delete those when finished

cjeanner:
> /lgtm
> 
> The question now is, how is using the legacy names? Just to track and delete them, and come back to delete those when finished

sooo - for now we have to make sure role is converged. Next step may be to add a "warning" and, eventually, a "fail" in case old names are used. We have to fix jobs, not break them further, at least as a first step :D

PR #400, opened 2023-07-24
Correct ci_script call

cjeanner:
while replacing "shell" and "command" by "ci_script", we must ensure we
also correct the actual parameters.

Test-project: https://review.rdoproject.org/r/c/testproject/+/49365


cjeanner:
The role isn't used/tested by the zuul job, and we have to unblock CI and promotion, failing due to the issues corrected in this PR.

PR #408, opened 2023-07-25
Get spellcheck workflow

cjeanner:


cjeanner:
@Sandeepyadav93 @lewisdenny any thoughts on that PR? We shouldn't get THAT many doc update, meaning new content SHOULD be small enough to not be a burden to spell-check... ?

.github/workflows/pyspell.yml
```diff
@@ -0,0 +1,23 @@
+# This is workflow for spell checking using PySpelling lib (https://pypi.org/project/pyspelling/)
+name: Spellcheck
+# Controls when the action will run.
+on:
```
Sandeepyadav93, line 4: @cjeanner I wonder if we want to optimize this job to only run when we change README.md using the path filter.

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore

but this is a light job we can run on all changes too.

cjeanner: it's now filtered on only .md files. Thanks for the headsup!

lewisdenny:
> @Sandeepyadav93 @lewisdenny any thoughts on that PR? We shouldn't get THAT many doc update, meaning new content SHOULD be small enough to not be a burden to spell-check... ?

I think a very small burden is worth professional looking docs. Now the typos are fixed, going forward it will be no issue :)

cjeanner:
recheck

Sandeepyadav93:
/lgtm

thanks cedric

cjeanner:
recheck

raukadah (approved):
/approve

PR #409, opened 2023-07-26
Set edpm-ansible.yml var file if file exists

raukadah:
Currently existance of edpm-ansible.yml var file depends on zuul_change_list var. This var is set in different playbook but it does not passed in another playbook i.e. `run.yml`.

It leads to failure of testing edpm-ansible changes.

By making sure edpm-ansible.yml var file exists, it fixes the issue.

### DIFF
--- a/ci/playbooks/edpm/run.yml
+++ b/ci/playbooks/edpm/run.yml
@@ -2,6 +2,11 @@
 - hosts: all
   gather_facts: true
   tasks:
+    - name: Check for edpm-ansible.yml file
+      ansible.builtin.stat:
+        path: "{{ ansible_user_dir }}/ci-framework-data/artifacts/edpm-ansible.yml"
+      register: edpm_file
+
     - name: Run Podified EDPM deployment
       ansible.builtin.command:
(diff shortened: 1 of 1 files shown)


afazekas:
I wonder is there a way to solve this without increasing the number of tasks to call ;-)

/lgtm

rabi (commented):
/lgtm

ci/playbooks/edpm/run.yml
```diff
@@ -2,14 +2,19 @@
 - hosts: all
   gather_facts: true
   tasks:
```
rabi, line 4: I missed this earlier . It should also be changed in https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci/playbooks/edpm_baremetal_deployment/run.yml as the baremetal jobs still don't use built images.

[1] https://logserver.rdoproject.org/43/243/80432abaa4078afd51ab64d80ec2a32f97fbee9a/github-check/edpm-ansible-crc-podified-edpm-baremetal/3c3771e/ci-framework-data/logs/crc/pods/openstack-operators_openstack-ansibleee-operator-controller-manager-647c77cb746tldg_30f7ca6f-621d-4287-a365-eb7039b1df42/manager/6.log

PR #412, opened 2023-07-26
Add some irrelevant files

kajinamit:
... to avoid running jobs unnecessarily.

  - [x] README in the role

### DIFF
--- a/zuul.d/edpm.yaml
+++ b/zuul.d/edpm.yaml
@@ -19,6 +19,14 @@
     abstract: true
     irrelevant-files:
       - .*/*.md
+      - ^LICENSE$
+      - ^OWNERS$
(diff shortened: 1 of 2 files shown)


zuul.d/edpm.yaml
```diff
@@ -19,6 +19,12 @@
     abstract: true
     irrelevant-files:
```
cjeanner, line 20: wondering if we should get "files" and run only if ci_framework/* is changed? Seems a bit more suitable.. WDYT?

kajinamit: This job template is used to run edpm job in idividual repo, and we have to run it in any logic change.

raukadah:
recheck

raukadah (approved):
/approve

PR #414, opened 2023-07-27
Ansible action plugin for oc commands

adrianfusco:
Small Ansible action plugin that loads kubeconfig, search `oc` binary and execute `oc` commands.

- [ ] Appropriate documentation exists and/or is up-to-date

### DIFF
--- a/ci_framework/plugins/action/oc.py
+++ b/ci_framework/plugins/action/oc.py
@@ -4,8 +4,6 @@
 import os
 from pathlib import Path
 from ansible.errors import AnsibleError
-from ansible.module_utils.basic import get_bin_path
-from ansible.module_utils.six import ensure_text
 from ansible.plugins.action import ActionBase
 
 DOCUMENTATION = r'''
@@ -15,13 +13,13 @@ action: oc
 short_description: Execute OpensShift Client commands - oc -.
 
 description:
-- This module searches the oc binary, load the kubeconfig and execute the oc provided commands.
+- This module searches the oc binary in, load the kubeconfig and execute the oc provided commands.
 '''
 
 EXAMPLES = r'''
 - name: Display the current authenticated user
-    oc_cmd:
-    command: whoami -c
+    oc:
+    cmd: whoami -c
 - name: Display the current authenticated user and get cluster information
       oc:
         cmd: |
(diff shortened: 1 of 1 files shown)


ci_framework/plugins/action/oc.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def run(self, tmp=None, task_vars=None):
+        super().run(tmp, task_vars)
+        module_args = self._task.args.copy()
+
+        if 'cmd' not in module_args or not module_args['cmd']:
+            raise AnsibleError('"cmd" parameter is mandatory')
+
+        # Get remote home, local or remote depending of the hosts
+        user_home = Path(task_vars['ansible_user_dir']).resolve()
```
pablintino, line 55: Keep in mind that this will fail if gather_facts is false, so I'll prefer this line to be evaluated only if really needed, in case that KUBECONFIG is not set as an environment var or cifmw_kubeconfig is not present.

ci_framework/plugins/action/oc.py
```diff
@@ -0,0 +1,113 @@
...
+
+        # Get cluster auth config path or take by default crc one
+        kubeconfig_path = task_vars.get(
+            'cifmw_kubeconfig', False) or user_home / ".crc/machines/crc/kubeconfig"
```
pablintino, line 59: Use None, not false.

ci_framework/plugins/action/oc.py
```diff
@@ -0,0 +1,113 @@
...
+        os.environ["KUBECONFIG"] = str(kubeconfig_path)
```
pablintino, line 60: This may work for localhost, but not for remotes. Please check `_compute_environment_string` and `_execute_module` https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/action/__init__.py#L1000

ci_framework/plugins/action/oc.py
```diff
@@ -0,0 +1,113 @@
...
+
+        # Additional default paths
+        additional_exec_paths = [
+            str(user_home / '.crc/bin'),
+            str(user_home / '.crc/bin/oc'),
+        ]
+
+        # We can provide additional PATHs
+        cifmw_oc_path = task_vars.get('cifmw_oc_path', False)
```
pablintino, line 69: Why not defaulting to None?

ci_framework/plugins/action/oc.py
adrianfusco: I changed the gathering to `explicit` and it still works. Any idea?

PR #418, opened 2023-07-28
Fixed the spellcheck for newly added roles

raukadah:
It fixes the spellcheck for ceph related roles and add their tech workds to docs/dictionary/en-custom.txt.


viroel:
/lgtm

Spellcheck is passing, thanks!

PR #419, opened 2023-07-31
Add support of `--build-args` agrs in edpm_build_images role

bshewale:
This PR adds support of --build-args while
Packaging edpm and ipa image inside container image.

  - [x] README in the role

### DIFF
--- a/ci_framework/roles/edpm_build_images/README.md
+++ b/ci_framework/roles/edpm_build_images/README.md
@@ -29,6 +29,7 @@ None
 * `cifmw_edpm_build_images_push_registry_namespace`: (String) Namespace on registry where we want to push container images. Default: `podified-main-centos9`.
 * `cifmw_edpm_build_images_cert_path`: (String) Cert path. Default: `/etc/pki/ca-trust/source/anchors/rh.crt`
 * `cifmw_edpm_build_images_cert_install`: (Boolean) Whether to install cert in the image. Default: false
+* `cifmw_edpm_build_images_base_image`: (String) Base image to package the edpm and ipa qcow2 images into the container images for rhel distro.
 
 ### Example
 ```YAML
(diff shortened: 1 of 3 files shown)


ci_framework/roles/edpm_build_images/README.md
```diff
@@ -29,6 +29,7 @@ None
 * `cifmw_edpm_build_images_push_registry_namespace`: (String) Namespace on registry where we want to push container images. Default: `podified-main-centos9`.
 * `cifmw_edpm_build_images_cert_path`: (String) Cert path. Default: `/etc/pki/ca-trust/source/anchors/rh.crt`
 * `cifmw_edpm_build_images_cert_install`: (Boolean) Whether to install cert in the image. Default: false
+* `cifmw_edpm_build_images_base_image`: (String) Base image to package the edpm and ipa qcow2 images into the container images. Default: quay.io/centos/centos:stream9-minimal
```
pablintino, line 32: That default value is something that needs maintenance. If someone changes this https://github.com/openstack-k8s-operators/edpm-image-builder/blob/main/Containerfile.image#L1 we need to change the value here. Are we even sure that we are gonna realize that someone changed the value there?
Maybe is better to say, hey, if you provide this value we will override the default one of the image.

cjeanner:
recheck

ci_framework/roles/edpm_build_images/tasks/package.yml
```diff
@@ -10,6 +10,9 @@
     cmd: >-
       buildah bud -f ./Containerfile.image
       -t edpm-hardened-uefi:{{ cifmw_edpm_build_images_tag }}
+      {% if ansible_distribution | lower == 'rhel' %}
```
rlandy, line 13: why would we not allow overriding for all cases - not just rhel.  Sometimes we have to use a diff centos version - so can't we just default {{ cifmw_edpm_build_images_base_image }} and always keep this clause?

raukadah (approved):
/approve

PR #430, opened 2023-08-08
Allow to reproduce a CI run locally using libvirt

cjeanner:
With the new job layout involving 3 nodes (CRC, Compute and an Ansible
controller), we have to provide people a way to reproduce such a job.

This patch extends the libvirt_manager role, as well as create a new
kind of "scenario" for reproducers. The intend there is to get multiple
"layouts" in order to mimic CI runs on a local environment.

  - [X] README in the role

### DIFF
--- a/ci_framework/roles/libvirt_manager/templates/domain.xml.j2
+++ b/ci_framework/roles/libvirt_manager/templates/domain.xml.j2
@@ -1,0 +1,106 @@
+<domain type='kvm'>
+  <name>cifmw-{{ vm_type }}-{{ vm_id }}</name>
+  <memory unit='GB'>{{ vm_data.value.memory | default(2) }}</memory>
+  <vcpu placement='static'>{{ vm_data.value.cpu | default(2) }}</vcpu>
+  <os>
+    <type arch='x86_64' machine='q35'>hvm</type>
+    <boot dev='hd'/>
+    <bootmenu enable='no'/>
+  </os>
+  <cpu mode='host-passthrough' check='none'/>
+  <clock offset='utc'>
+    <timer name='rtc' tickpolicy='catchup'/>
+    <timer name='pit' tickpolicy='delay'/>
+    <timer name='hpet' present='no'/>
+  </clock>
+  <features>
+    <acpi/>
+    <apic/>
+  </features>
+  <devices>
+    <emulator>/usr/libexec/qemu-kvm</emulator>
+    <disk type='file' device='disk'>
+      <driver name='qemu' type='qcow2'/>
+      <source file='{{ cifmw_reproducer_basedir }}/workload/{{ vm_type }}-{{ vm_id }}.qcow2'/>
+      <target dev='sda' bus='sata'/>
+    </disk>
+    {% for network in vm_data.value.nets %}
+    <interface type='network'>
+      <source network='cifmw-{{ network }}'/>
+      <mac address='{{ '24:42:%02i'|format(vm_id|int) |random_mac }}'/>
+      <model type="virtio"/>
+    </interface>
+    {% endfor %}
+    <controller type='usb' index='0' model='ich9-ehci1'>
+      <alias name='usb'/>
+      <address type='pci' domain='0x0000' bus='0x00' slot='0x06' function='0x7'/>
+    </controller>
+    <controller type='usb' index='0' model='ich9-uhci1'>
+      <alias name='usb'/>
+      <master startport='0'/>
+      <address type='pci' domain='0x0000' bus='0x00' slot='0x06' function='0x0' multifunction='on'/>
+    </controller>
+    <controller type='usb' index='0' model='ich9-uhci2'>
+      <alias name='usb'/>
+      <master startport='2'/>
+      <address type='pci' domain='0x0000' bus='0x00' slot='0x06' function='0x1'/>
+    </controller>
+    <controller type='usb' index='0' model='ich9-uhci3'>
+      <alias name='usb'/>
+      <master startport='4'/>
+      <address type='pci' domain='0x0000' bus='0x00' slot='0x06' function='0x2'/>
+    </controller>
+    <controller type='pci' index='0' model='pcie-root'>
+      <alias name='pci.0'/>
+    </controller>
+    <controller type='virtio-serial' index='0'>
+      <alias name='virtio-serial0'/>
+      <address type='pci'/>
+    </controller>
+    <serial type='pty'>
+      <source path='/dev/pts/1'/>
+      <log file="/var/log/libvirt/qemu/{{ vm_type }}-{{ vm_id }}-serial.log" append="off"/>
+      <target type='isa-serial' port='0'>
+        <model name='isa-serial'/>
+      </target>
+      <alias name='serial0'/>
+    </serial>
+    <console type='pty' tty='/dev/pts/1'>
+      <log file="/var/log/libvirt/qemu/{{ vm_type }}-{{ vm_id }}-console.log" append="off"/>
+      <source path='/dev/pts/1'/>
+      <target type='serial' port='0'/>
+      <alias name='serial0'/>
+    </console>
+    <input type='tablet' bus='usb'>
+      <alias name='input0'/>
+      <address type='usb' bus='0' port='1'/>
+    </input>
+    <input type='mouse' bus='ps2'>
+      <alias name='input1'/>
+    </input>
+    <input type='keyboard' bus='ps2'>
+      <alias name='input2'/>
+    </input>
+    <video>
+      <model type='virtio' vram='16384' heads='1' primary='yes'/>
+      <alias name='video0'/>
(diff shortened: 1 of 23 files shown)


ci_framework/roles/libvirt_manager/templates/domain.xml.j2
```diff
@@ -0,0 +1,106 @@
+<domain type='kvm'>
+  <name>cifmw-{{ vm_type }}-{{ vm_id }}</name>
+  <memory unit='GB'>{{ vm_data.value.memory | default(2) }}</memory>
+  <vcpu placement='static'>{{ vm_data.value.cpu | default(2) }}</vcpu>
+  <os>
+    <type arch='x86_64' machine='pc-q35-rhel9.2.0'>hvm</type>
```
afazekas, line 6: IMHO.

Try not specify exact machine, q35 is the generic machine type.

Reasonable to expect the default works, the exact might have issues on the long run.

ci_framework/roles/libvirt_manager/templates/domain.xml.j2
```diff
@@ -0,0 +1,106 @@
...
+    {% for network in vm_data.value.nets %}
+    <interface type='network'>
+      <source network='{{ network }}'/>
+      <mac address='{{ '24:42:%02i'|format(vm_id|int) |random_mac }}'/>
+      <model type="virtio"/>
+    </interface>
+    {% endfor %}
+    <controller type='usb' index='0' model='ich9-ehci1'>
+      <alias name='usb'/>
+      <address type='pci' domain='0x0000' bus='0x00' slot='0x06' function='0x7'/>
```
afazekas, line 36: Exact addresses might not be needed, but it is ok.

ci_framework/roles/libvirt_manager/templates/domain.xml.j2
cjeanner: indeed - updated.

ci_framework/roles/libvirt_manager/templates/domain.xml.j2
cjeanner: it's an old template I "migrate" from one project to another :]. The less I edit it, the better, to be fair. "it works" is the expected state, and since it works... ;)

docs/source/cookbooks/reproducer.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Inventory
+This feature can be launched against your own desktop or laptop, but also
+against a remote hypervisor (preferred due to resources). You can therefore
+create your own inventory as follow:
+```YAML
+all:
+  hosts:
+    hypervisor:
+     ansible_user: virtuser
```
afazekas, line 92: zuul ?

cjeanner: "whatever" - that's on the hypervisor. I don't have "zuul" on the hypervisor, for instance. That's an example.

scenarios/reproducers/3-nodes.yml
```diff
@@ -0,0 +1,66 @@
...
+          </dhcp>
+        </ip>
+      </network>
+    crc: |-
+      <network>
+        <name>crc</name>
+        <forward mode='nat'/>
+        <bridge name='crc' stp='on' delay='0'/>
+        <mac address='52:54:00:fd:be:d0'/>
+        <ip family='ipv4' address='192.168.122.1' prefix='24'>
```
afazekas, line 64: This is the default ip range for the default libvirt network.
Should we use different range (192.168.130.1) for crc or use the default network ?

ci_framework/roles/libvirt_manager/templates/domain.xml.j2
```diff
@@ -0,0 +1,106 @@
...
+    <emulator>/usr/libexec/qemu-kvm</emulator>
+    <disk type='file' device='disk'>
+      <driver name='qemu' type='qcow2'/>
+      <source file='{{ cifmw_reproducer_basedir }}/workload/{{ vm_type }}-{{ vm_id }}.qcow2'/>
+      <target dev='sda' bus='sata'/>
+    </disk>
+    {% for network in vm_data.value.nets %}
+    <interface type='network'>
+      <source network='{{ network }}'/>
+      <mac address='{{ '24:42:%02i'|format(vm_id|int) |random_mac }}'/>
```
afazekas, line 30: The offset in the nets might be one octet in the mac.

scenarios/reproducers/3-nodes.yml
```diff
@@ -0,0 +1,66 @@
...
+  networks:
+    default: |-
+      <network>
+        <name>default</name>
+        <forward mode='nat'/>
+        <bridge name='virbr0' stp='on' delay='0'/>
+        <mac address='52:54:00:6a:f2:dc'/>
+        <ip family='ipv4' address='192.168.100.1' prefix='24'>
+          <dhcp>
+            <range start='192.168.100.10' end='192.168.100.100'/>
```
afazekas, line 54: it is possible to make reservation for each machine for each mac.

afazekas: "Grab IPs for nodes type"  is not needed if you know the address in advanced.

scenarios/reproducers/3-nodes.yml
```diff
@@ -0,0 +1,66 @@
+---
+cifmw_install_yamls_repo: /home/cjeanner/work/github/openstack-k8s-operators/install_yamls
```
afazekas, line 2: I actually do not mind having a specific dir in the $HOME as documented recommendation.
Probably more specific like 'cifmw' instead of work .
Might be put here by var.

scenarios/reproducers/3-nodes.yml
cjeanner: so I've seen crc using .122 but also .130 range - it seems to depend on the version of crc. Thing is, in install_yamls, many assumptions are done about the used range, and many things are hardcoded to .122 range (routes, for instance).

ci_framework/roles/libvirt_manager/templates/domain.xml.j2
cjeanner: in that case we have some more randomness, not a big deal. We ensure we also have enough difference using the vm_id/index, there are less chances we'll hit any collision.
In the past, in tripleo-lab, I *had* collisions before using the VM index as a part of the MAC.

scenarios/reproducers/3-nodes.yml
afazekas: maybe just commenting out and stating it an example is good.

scenarios/reproducers/3-nodes.yml
cjeanner: maybe. But we'd still need to wait for ssh to come up. I'd rather avoid going with static IPs in there, to be fair.

scenarios/reproducers/3-nodes.yml
afazekas: I create a crc image by just setup and stop.

It leads to I have :
cat /etc/hosts |grep crc
192.168.130.11 api.crc.testing canary-openshift-ingress-canary.apps-crc.testing console-openshift-console.apps-crc.testing default-route-openshift-image-registry.apps-crc.testing downloads-openshift-console.apps-crc.testing oauth-openshift.apps-crc.testing


Which step supposed to make either the existing record working,
or updating it to a new one ?

scenarios/reproducers/3-nodes.yml
cjeanner: ah, right, that's a missed thing with my own env, I forgot to update it before pushing.

docs/source/cookbooks/reproducer.md
afazekas: ansible_host: 127.0.0.1
          
As an example ?
Ir keeping there ansible_host:  <You (remote) machines address>

cjeanner: (will update with some more explicit placeholder)

scenarios/reproducers/3-nodes.yml
cjeanner: the record on the hypervisor won't be used actually. So I don't think we should temper with it.

afazekas:
/lgtm

We will improve it later.

cjeanner:
e2e job failed because CRC failed to start - but it's not related to this change.

I won't obsess on rechecks.

afazekas:
I wonder can you swap the "default" and the "crc"  ip ranges than the network order.
in-order to leave the "default" as-is .

afazekas:
Looks like If I use the "default" network as this change state
for example the https://ci-framework.readthedocs.io/en/latest/quickstart/04_non-virt.html  way stops working AS-IS.

I  restore the original default net.

marios (approved):
/lgtm

discussed again in reviews call just now

we need to merge and iterate

PR #432, opened 2023-08-09
Disable tempest in edpm check job and enable it in periodic

raukadah:
Currently edpm check job is not stable. We are moving the tempest tests to periodic line till crc standalone work finishes.


cjeanner:
/approve
/lgtm

and force-merge since it will unblock devs.

PR #435, opened 2023-08-10
Ensure we keep correct order

cjeanner:
Sorting project name should ensure we don't get any random output,
allowing a smoother diff whenever we regenerate the project list.


raukadah (commented):
/approve

PR #439, opened 2023-08-11
Make edpm_prepare and deploy accept kustomizations

pablintino:

### DIFF
--- a/ci_framework/roles/edpm_deploy/tasks/main.yml
+++ b/ci_framework/roles/edpm_deploy/tasks/main.yml
@@ -18,16 +18,63 @@
   tags:
     - always
   ansible.builtin.set_fact:
-    cifmw_edpm_deploy_env: "{{ cifmw_install_yamls_environment | combine({'PATH': cifmw_path }) }}"
+    cifmw_edpm_deploy_env: >-
+      {{
+        cifmw_install_yamls_environment |
+        combine({'PATH': cifmw_path}) |
+        combine(cifmw_edpm_deploy_extra_vars | default({}))
+      }}
     cacheable: true
 
-- name: Deploy EDPM
+- name: Save hostvars
+  ansible.builtin.copy:
+    dest: "{{ ansible_user_dir }}/ci-framework-data/artifacts/hostvars-dump.yml"
+    content: "{{ hostvars | to_nice_yaml }}"
+
+- name: Prepare OpenStack Dataplane CR
   vars:
-    make_edpm_deploy_env: "{{ cifmw_edpm_deploy_env }}"
-    make_edpm_deploy_dryrun: "{{ cifmw_edpm_deploy_dryrun | bool }}"
+    make_edpm_deploy_prep_env: "{{ cifmw_edpm_deploy_env }}"
+    make_edpm_deploy_prep_dryrun: "{{ cifmw_edpm_deploy_dryrun | bool }}"
   ansible.builtin.include_role:
     name: 'install_yamls_makes'
-    tasks_from: 'make_edpm_deploy'
+    tasks_from: 'make_edpm_deploy_prep'
+
+- name: Kustomize and deploy OpenStackDataPlane
+  when:
+    - not cifmw_edpm_deploy_dryrun | bool
+  environment:
+    PATH: "{{ cifmw_path }}"
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+  block:
+    - name: Find the OpenStack Dataplane CR manifest
+      ansible.builtin.find:
+        paths: "{{ cifmw_edpm_deploy_manifests_dir }}/{{ cifmw_install_yamls_defaults['NAMESPACE'] }}/dataplane/cr"
+        contains: "kind: OpenStackDataPlane"
+        excludes: "kind: Kustomization"
+        patterns: "*.yaml"
+      register: cifmw_edpm_deploy_cr_manifest_paths
+
+    - name: Ensure manifest exists
+      ansible.builtin.assert:
+        that: cifmw_edpm_deploy_cr_manifest_paths.matched == 1
+        quiet: true
+        msg: "Cannot determine OpenStackDataPlane deployment manifest"
+
+    - name: Perform kustomizations to the OpenStackDataPlane CR
+      vars:
+        cifmw_edpm_kustomize_cr_path: "{{ cifmw_edpm_deploy_cr_manifest_paths.files[0].path }}"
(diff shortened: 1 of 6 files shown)


ci_framework/roles/edpm_deploy/tasks/main.yml
```diff
@@ -18,24 +18,64 @@
...
+      ansible.builtin.assert:
+        that: cifmw_edpm_deploy_cr_manifest_paths.matched == 1
+        quiet: true
+        msg: "Cannot determine OpenStackDataPlane deployment manifest"
+
+    - name: Perform kustomizations to the OpenStackDataPlane CR
+      vars:
+        cifmw_edpm_kustomize_cr_path: "{{ cifmw_edpm_deploy_cr_manifest_paths.files[0].path }}"
+        cifmw_edpm_kustomize_content: "{{ cifmw_edpm_deploy_dataplane_cr_kustomization | default('{}') }}"
+      ansible.builtin.include_role:
```
viroel, line 67: can we skip when cifmw_edpm_deploy_dataplane_cr_kustomization is empty or not defined?

ci_framework/roles/edpm_prepare/tasks/main.yml
```diff
@@ -122,12 +138,28 @@
         quiet: true
         msg: "Cannot determine OpenStackControlPlane deployment manifest"
 
-    - name: Find files out
+    - name: Perform kustomizations to the OpenStackControlPlane CR
+      vars:
+        cifmw_edpm_kustomize_cr_path: "{{ cifmw_edpm_prepare_openstack_cr_manifest_paths.files[0].path }}"
+        cifmw_edpm_kustomize_content: "{{ cifmw_edpm_prepare_openstack_cr_kustomization | default('{}') }}"
+      ansible.builtin.include_role:
+        name: edpm_kustomize
```
viroel, line 146: same question as in edpm deploy

ci_framework/roles/edpm_deploy/tasks/main.yml
pablintino: That's the current behavior,  the https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci_framework/roles/edpm_kustomize/tasks/main.yml#L49 is only executed if a kustomization is given, but, since we use now the `openstack_deploy_prep` target instead of the plain `openstack_deploy`  we need to apply the install_yamls generated kustomization ourselfs as done in https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci_framework/roles/edpm_kustomize/tasks/main.yml#L30-L46.

ci_framework/roles/edpm_deploy/README.md
```diff
@@ -14,7 +14,7 @@ None
 * `cifmw_edpm_deploy_retries`: (Integer) Number of retries for edpm deploy status. Defaults to `100`.
 * `cifmw_edpm_deploy_run_validation`: (Boolean) Run validation on EDPM node. Defaults to `False`.
 * `cifmw_edpm_deploy_dryrun`: (Boolean) Do a dry run on make edpm_deploy command. Defaults to `False`.
-* `cifmw_edpm_deploy_nova_pod_name`: (String) Name of the nova deployment pod started by an EDPM deploy. Defaults to `nova-edpm-compute-0-deploy-nova`.
```
cjeanner, line 17: Sanity check: this was recently added. Are you sure you want to remove that? If so, shouldn't it be part of its own PR?

ci_framework/roles/edpm_deploy/tasks/main.yml
cjeanner: I think @viroel was more thinking about a `when: ` condition in this task/block, so that we avoid the role inclusion all-together.

ci_framework/roles/edpm_deploy/tasks/main.yml
```diff
@@ -18,55 +18,74 @@
...
       ci_script:
         output_dir: "{{ cifmw_edpm_deploy_basedir }}/artifacts"
-        script: |-
-          oc get openstackdataplane -n {{ cifmw_install_yamls_defaults['NAMESPACE'] }}
-          oc get openstackdataplanerole -n {{ cifmw_install_yamls_defaults['NAMESPACE'] }}
-          oc get openstackdataplanenode -n {{ cifmw_install_yamls_defaults['NAMESPACE'] }}
-          oc get pods -n {{ cifmw_install_yamls_defaults['NAMESPACE'] }} | grep edpm
-      ignore_errors: true
-
-    - name: Get EDPM nova deployment pod name
```
cjeanner, line 49: linked to the removal of the `cifmw_edpm_deploy_nova_pod_name` - might be best in a standalone PR. Not sure why we need to remove this in order to enable kustomization?

ci_framework/roles/edpm_deploy/tasks/main.yml
```diff
@@ -18,55 +18,74 @@
...
-          -n {{ cifmw_install_yamls_defaults['NAMESPACE'] }} |
-          grep -m1  {{ cifmw_edpm_deploy_nova_pod_name }}
-      until: edpm_pod.rc == 0
-      retries: "{{ cifmw_edpm_deploy_retries }}"
-      delay: 50
+        script: "oc apply -f {{ cifmw_edpm_deploy_cr_manifest_paths.files[0].path }}"
 
-    - name: Wait for EDPM nova deployment pod to finish
-      register: deploy_status
+    - name: Wait for OpenStackDataplane to be deployed
```
cjeanner, line 79: guess that's the replacement of the "wait for nova pod".

pablintino: Exactly, instead of waiting for an specific pod the idea now is to wait on the higher level resource to avoid situations like the one happening now, that is that the first deploy pod may fail and we don't realize that the next attempt succeed.

ci_framework/roles/edpm_deploy/tasks/main.yml
pablintino: Maybe, the thing is that we need to call if even if `cifmw_edpm_deploy_dataplane_cr_kustomization` is empty, as we need to apply the install_yamls kustomization, so is not possible to create and easy when condition.

ci_framework/roles/edpm_deploy/README.md
pablintino: I saw it, and yes, I'm sure. Ok, I'll open a separate PR for that part.

ci_framework/roles/edpm_deploy/tasks/main.yml
cjeanner: ah, ok. Right, we need to run the role anyway. Guess we're good then.

cjeanner (commented):
/approve

PR #440, opened 2023-08-11
[dlrn_report] escape the spaces in shell task

raukadah:
Due to use of `|` in shell tasks, each line is treated as a single command, leading to failure of the task. Let' use `>-` to avoid the same.


rlandy (commented):
/lgtm

viroel:
overriding rdo jobs since "cifmw-molecule-dlrn_report SUCCESS in 5m 32s" and this role is not being used in any other job

PR #441, opened 2023-08-14
Consider `None` value of hash variables in conditional check

raukadah:
In dlrn_report role, repo_setup provides dlrn _hash related vars. Each of these vars either return a hash value or None string.

In order to make sure, dlrn hash vars are not empty, we also need for `None` string case in conditional check otherwise the dlrn_report task will fail.


cjeanner:
recheck

PR #442, opened 2023-08-14
Wait OpenStackDataplane resource instead of the deploy pod.

pablintino:
Till now we were waiting for the deployment of EDPM to finish by waiting for the deployment pod, that may fail and can be relaunched by the deploy job. If the pod fails to deploy we weren't aware that maybe the next pod succeed and we marked the entire run as failed.

### DIFF
--- a/ci_framework/roles/edpm_deploy/tasks/main.yml
+++ b/ci_framework/roles/edpm_deploy/tasks/main.yml
@@ -68,68 +68,31 @@
         name: edpm_kustomize
 
     - name: Apply the OpenStackDataPlane CR
-      environment:
-        KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
-        PATH: "{{ cifmw_path }}"
       when: not cifmw_edpm_deploy_dryrun
       ci_script:
         output_dir: "{{ cifmw_edpm_deploy_basedir }}/artifacts"
         script: "oc apply -f {{ cifmw_edpm_deploy_cr_manifest_paths.files[0].path }}"
 
-- name: Wait for EDPM compute deployment to finish
-  when:
-    - not cifmw_edpm_deploy_dryrun | bool
-  environment:
-    PATH: "{{ cifmw_path }}"
-    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
-  block:
-    - name: Get info about dataplane node
-      ci_script:
-        output_dir: "{{ cifmw_edpm_deploy_basedir }}/artifacts"
-        script: |-
-          oc get openstackdataplane -n {{ cifmw_install_yamls_defaults['NAMESPACE'] }}
-          oc get openstackdataplanerole -n {{ cifmw_install_yamls_defaults['NAMESPACE'] }}
-          oc get openstackdataplanenode -n {{ cifmw_install_yamls_defaults['NAMESPACE'] }}
-          oc get pods -n {{ cifmw_install_yamls_defaults['NAMESPACE'] }} | grep edpm
-      ignore_errors: true
-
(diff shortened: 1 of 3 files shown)


ci_framework/roles/edpm_deploy/tasks/main.yml
```diff
@@ -68,52 +68,18 @@
         name: edpm_kustomize
 
     - name: Apply the OpenStackDataPlane CR
-      environment:
-        KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
-        PATH: "{{ cifmw_path }}"
```
cjeanner, line 73: ah right, those are set in the upper level block. Good cleaning!

ci_framework/roles/edpm_deploy/tasks/main.yml
```diff
@@ -68,52 +68,18 @@
...
-          {{ edpm_pod.stdout |trim }}
-      until: deploy_status.rc == 0
-      retries: "{{ cifmw_edpm_deploy_retries }}"
-      delay: 20
+          oc wait OpenStackDataplane {{ cifmw_edpm_kustomize_last_cr_content.metadata.name }}
+          --namespace={{ cifmw_install_yamls_defaults['NAMESPACE'] }}
+          --for=condition=ready
+          --timeout={{ cifmw_edpm_deploy_timeout }}m
 
 - name: Get the logs of EDPM Deploy
```
cjeanner, line 84: while at it, can't we push that under the same upper level block exposing the `environment` ?

afazekas:
Last week I seen ntp related failures in edpm ansible playbooks, I switched to the internal NTP.
The current wait still hidden the issue, unless it happened 4+ times , otherwise it was a time increase.
It left some pods in Error state.

I wonder when thing otherwise stabilize,
should we check for errored pods and consider  the job failed if any exists even tough  tests were passing ?

Having Error -ed things is bad user facing thing.

pablintino:
> 

Why should CI consider a deployment failed if the product itself was able to recover and correctly deploy EDPM?

jpodivin (approved):

afazekas:
Operators find confusing even when the service logs contains ERROR level logs,
there were approach to fail a job when any service logs anything as ERROR.
(Logs typically even did  had message for the error was solved even on Debug level)

Also this kind of ERROR /retry things are time consuming stuff. if we see
one we might either silence it or improve it.

When anybody comes to fixing/debug something first looking for "Error" ,
they may waste a lot of time on a not an error.

pablintino:
> Operators find confusing even when the service logs contains ERROR level logs, there were approach to fail a job when any service logs anything as ERROR. (Logs typically even did had message for the error was solved even on Debug level)
> 
> Also this kind of ERROR /retry things are time consuming stuff. if we see one we might either silence it or improve it.
> 
> When anybody comes to fixing/debug something first looking for "Error" , they may waste a lot of time on a not an error.

But, shouldn't that change go into the operator condition compute itself instead of workaround it here? We saw deployments that took 2 iterations to succeed, and that's something that it's legit, why dropping that internal rety support? If there is an error condition that makes the deployment unusable, maybe is the operator itself, that is the one with enough info about the jobs it manages, the one that needs to flag it as non-ready instead of us, that are simple users.

pablintino:
recheck

afazekas:
The new wait does not cantinas internal name grepping wait, IMHO it sounds nicer, not just a "workaround".

Those "Error"  things definitely in an indicator for a potential of area of improvement.
 - either ci infra should be more stable
 - product should find a way for not having errors in a good env

pablintino:
recheck

PR #443, opened 2023-08-14
Fix the spacing issue in condition check

raukadah:
Usage of `-` in conditional check is cancating all the dlrn report arguments leading to random failures. It also fixes the dlrn hash vars having "" returned by repo-setup role.
```
"cifmw_repo_setup_commit_hash": "",
        "cifmw_repo_setup_distro_hash": "",
        "cifmw_repo_setup_dlrn_api_url": "https://trunk.rdoproject.org/api-centos9-antelope",
        "cifmw_repo_setup_extended_hash": "",
        "cifmw_repo_setup_full_hash": "43bc77e1054a2e185169a239919a5207",
        "cifmw_repo_setup_release": "antelope"
```
Now it uses length to check whether an dlrn hash var is empty or not.

### DIFF
--- a/ci_framework/roles/dlrn_report/tasks/dlrn_report_results.yml
+++ b/ci_framework/roles/dlrn_report/tasks/dlrn_report_results.yml
@@ -9,34 +9,34 @@
     name: repo_setup
 
 - name: Report results to dlrn for the tested hash
-  ansible.builtin.shell:
+  ansible.builtin.command:
     cmd: >-
-      {% if zuul_success is defined and zuul_success |bool %}
+      {% if zuul_success is defined and zuul_success |bool -%}
       echo "REPORTING SUCCESS TO DLRN API";
-      {% else %}
+      {% else -%}
       echo "REPORTING FAILURE TO DLRN API";
-      {% endif %}
+      {% endif -%}
       export SSL_CA_BUNDLE="/etc/pki/ca-trust/extracted/openssl/ca-bundle.trust.crt";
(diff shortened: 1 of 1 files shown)


ci_framework/roles/dlrn_report/tasks/dlrn_report_results.yml
cjeanner, line 1: maybe adding the `-` at the end of the `if` `else` `endif` lines 14, 16 and 18 may be good, for the sake of consistency: it would ensure no empty block of lines persist at the beginning.

Also, I'm wondering if we shouldn't generate the script using a proper template, and run it with a simple `ansible.builtin.command` or equivalent. That way, we would get the script written in the artifacts (easier to check), and we may even use same more fancy logging capabilities. WDYT?

rlandy (approved):

ci_framework/roles/dlrn_report/tasks/dlrn_report_results.yml
cjeanner: hm, in that case, replacing the `shell` by `command` will fail, because there are actually 3 commands in there (the `echo` followed by the `export` and finally the `dlrnapi` call).

So that won't work as-is. I meant "generate a file via a template and call that file with `command`"... Or call ci_script?

raukadah: We cannot use ci_script there. Let me switch back to shell . It will work.

PR #445, opened 2023-08-14
Switch back to shell module instead of command

raukadah:


cjeanner:
/lgtm
/approve

correcting a tiny misunderstanding - can be nudged without further work.

PR #448, opened 2023-08-15
Allow to chose target for CI playbooks

cjeanner:
With the new multinode layout, "all" target will fail since zuul would
try to run things on crc and compute (and any other nodes).

This patch introduces a new `cifmw_zuul_target_host` parameter in those
CI specific playbooks, in order to avoid any potential conflict with the
existing `cifmw_host_target` top level parameter.

Please note this patch takes care of the "all" targets only. Some plays
were already using "controller", and some other were using
cifmw_host_target (with a default to localhost).
In order to not break things for now, we discard those and will address
them in a follow-up, if needed.


raukadah (approved):
/approve

PR #454, opened 2023-08-16
Ensure CI playbooks have a name

cjeanner:
Using the play path as a name allows an easy understanding in case of
issues, since we can easily find what play failed.

### DIFF
--- a/ci/playbooks/build_runner_image.yml
+++ b/ci/playbooks/build_runner_image.yml
@@ -1,5 +1,6 @@
 ---
-- hosts: "{{ cifmw_zuul_target_host | default('all') }}"
+- name: "Run ci/playbooks/build_runner_image.yml"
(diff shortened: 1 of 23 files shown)


cjeanner:
recheck

raukadah:
recheck

raukadah (commented):
/approve

raukadah:
recheck

cjeanner:
Both failures are unrelated to the current patch... Not sure re-running endlessly will improve the results :/.

Still, the kuttl one is failing on keystone on a steady trend, maybe there's something in there, but unrelated to this patch...

Getting those playbook names will improve readability and understanding of the logs, providing a good overview of what play is failing for instance, making the whole debugging thing easier.

Shall we discard those 2 failures and force this one in the tree?

PR #456, opened 2023-08-16
Filter hosts in CI playbook

cjeanner:
We cannot use a parameter as `hosts` value in zuul playbooks. The
parameters as passed via the "vars" and related job parameters are
injected in the inventory, and not exposed during the ansible-playbook
startup, but only once it's loaded and ready.

At that stage, the playbook already defaulted to the `all` value,
preventing clean and nice filtering.

We'd need to pass the parameter as actual --extra-vars, but this isn't
possible in zuul.

We therefore have to leverage the `end_host` meta to skip the host if we
set the cifmw_zuul_target_host parameter.

### DIFF
--- a/ci/playbooks/build_runner_image.yml
+++ b/ci/playbooks/build_runner_image.yml
@@ -1,6 +1,13 @@
 ---
 - hosts: "{{ cifmw_zuul_target_host | default('all') }}"
   tasks:
+    - name: Filter out host if needed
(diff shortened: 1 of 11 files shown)


cjeanner:
Note: I've kept the parameter in case people want to be able to run the playbook locally (I doubt it, but heh...). It doesn't change anything.

PR #457, opened 2023-08-16
Run os_must_gather only in case of failure

dasm:
Run os_must_gather only in case of a failure which happened during "run" step in ZUUL.
By doing so, we're reducing post-run job by about 450-600s, also preserving disk space.

  - [x] README in the role

### DIFF
--- a/ci_framework/playbooks/99-logs.yml
+++ b/ci_framework/playbooks/99-logs.yml
@@ -32,10 +32,6 @@
       ansible.builtin.import_role:
         name: artifacts
 
-    - name: Run os_must_gather
-      ansible.builtin.import_role:
-        name: os_must_gather
-
     - name: Ensure ansible.log is at the expected location
       register: ansible_log_state
       ansible.builtin.stat:
--- a/ci_framework/roles/artifacts/tasks/main.yml
+++ b/ci_framework/roles/artifacts/tasks/main.yml
@@ -14,6 +14,11 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+- name: Verify if "success" flag exists after successful tests execution
+  register: cifmw_success_flag
+  ansible.builtin.stat:
+    path: "{{ ansible_user_dir }}/cifmw-success"
+
 - name: Ensure we have at least minimal ansible facts
   tags:
     - always
@@ -44,8 +49,15 @@
     - always
   ansible.builtin.import_tasks: cleanup.yml
 
-- name: Gather CRC logs
-  ansible.builtin.import_tasks: crc.yml
-
-- name: Gather EDPM logs
-  ansible.builtin.import_tasks: edpm.yml
+- name: Gather logs when forced or tests failed
+  when: cifmw_artifacts_gather_logs | bool or not cifmw_success_flag.stat.exists
+  block:
+    - name: Run os_must_gather
+      ansible.builtin.import_role:
+        name: os_must_gather
+  rescue:
+    - name: Get CRC logs if os_must_gather failed
+      ansible.builtin.import_tasks: crc.yml
+  always:
+    - name: Get EDPM logs
+      ansible.builtin.import_tasks: edpm.yml
--- a/deploy-edpm.yml
+++ b/deploy-edpm.yml
@@ -6,6 +6,14 @@
 # openstack-k8s-operators/install_yaml repository in order to leverage its
 # own methods.
 
+- name: Remove status flag
+  hosts: "{{ cifmw_target_host | default('localhost') }}"
+  tasks:
+    - name: Delete success flag if exists
+      ansible.builtin.file:
+        path: "{{ ansible_user_dir }}/cifmw-success"
+        state: absent
+
 - name: Bootstrap step
   ansible.builtin.import_playbook: ci_framework/playbooks/01-bootstrap.yml
 
@@ -30,6 +38,14 @@
 - name: Import run test playbook
   ansible.builtin.import_playbook: ci_framework/playbooks/08-run-tests.yml
   when: cifmw_run_tests | default('false') | bool
+
+- name: Inject status flag
+  hosts: "{{ cifmw_target_host | default('localhost') }}"
+  tasks:
+    - name: Inject success flag
+      ansible.builtin.file:
+        path: "{{ ansible_user_dir }}/cifmw-success"
+        state: touch
 
 - name: Run log related tasks
   ansible.builtin.import_playbook: ci_framework/playbooks/99-logs.yml
--- a/ci_framework/roles/artifacts/README.md
+++ b/ci_framework/roles/artifacts/README.md
@@ -10,6 +10,7 @@ None - writes happen only in the user home.
 * `cifmw_artifacts_crc_host`: (String) Hostname of the CRC instance. Defaults to `api.crc.testing`.
 * `cifmw_artifacts_crc_user`: (String) Username to connect to the CRC instance. Defaults to `core`.
 * `cifmw_artifacts_crc_sshkey`: (String) Path to the private SSH key to connect to CRC. Defaults to `~/.crc/machines/crc/id_ecdsa`.
+* `cifmw_artifacts_gather_logs`: (Boolean) Force gathering of logs, even in the case of successful test run. Defaults to `false`
 
 ## Examples
 Usually we'll import the role as-is at the very start of the playbook, and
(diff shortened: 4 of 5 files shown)


ci/playbooks/edpm/run.yml
```diff
@@ -30,3 +30,8 @@
           {%-   endfor %}
           {%- endif %}
           -e @scenarios/centos-9/zuul_inventory.yml
+
```
cjeanner, line 33: I know I pointed this file for this task, but after some more reflection, we should be a bit more generic.

The "flag" addition should happen elsewhere. I see two possibilities there:
either we add an actual playbook block in here:
https://github.com/openstack-k8s-operators/ci-framework/blob/main/deploy-edpm.yml
for instance starting on Line 33, something like:
```YAML
- name: Inject status flag
  hosts: {{ cifmw_target_host | default('localhost') }}"
  tasks:
    - name: Inject success flag
      ansible.builtin.file:
        path: "{{ ansible_user_dir }}/cifmw-sucess
```

or maybe add the task in one of the sub-playbook - but not the "test" one since it may not be included...

ci/playbooks/edpm/run.yml
```diff
@@ -30,3 +30,8 @@
...
+    - name: Create success file
+      ansible.builtin.file:
+        path: "{{ ansible_user_dir }}/success"
```
cjeanner, line 36: We should probably chose a better name, such as `cifmw-success` or the like.

Note that, maybe, we want to ensure it's absent before deploying. Just to be on the safe side, especially for people running it on their own env.

ci_framework/playbooks/99-logs.yml
```diff
@@ -28,13 +28,19 @@
...
+        path: "{{ ansible_user_dir }}/success"
+
     - name: Generate artifacts
       ansible.builtin.import_role:
         name: artifacts
 
     - name: Run os_must_gather
       ansible.builtin.import_role:
         name: os_must_gather
+      when: success_file.stat.exists is not defined
```
cjeanner, line 43: that would be:
```YAML
when:
  - not success_file.stat.exists
```
It will always be defined, as a boolean.

pablintino:
Just asking, there is no zuul fact set that tells us the result of the job?

cjeanner:
> Just asking, there is no zuul fact set that tells us the result of the job?

I'd rather avoid that, too zuul centric IMHO.

ci_framework/playbooks/99-logs.yml
```diff
@@ -32,9 +32,15 @@
...
-    - name: Run os_must_gather
+    - name: Check if "success" flag exists
+      register: cifmw_success_flag
+      ansible.builtin.stat:
+        path: "{{ ansible_user_dir }}/cifmw-success"
+
+    - name: Run os_must_gather only when there is no cifmw-success flag or if it's forced
       ansible.builtin.import_role:
         name: os_must_gather
+      when: (cifmw_os_must_gather_run | default('false') | bool) or (not cifmw_success_flag.stat.exists)
```
cjeanner, line 43: that's a good first approach. Just a tiny note in the doc about that new `cifmw_os_must_gather_run` global parameter and we should be OK.
Though.. moving that call (the whole import_role) in the artifacts role may help with:
- cifmw_artifacts_must_gather_run as an "artifacts parameter"
- running [artifacts/crc.yml](https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci_framework/roles/artifacts/tasks/crc.yml) only if must_gather fails
- run that same artifacts/crc.yml only in case of failure as well

So checking for the flag would be in the artifacts role directly, for instance here:
https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci_framework/roles/artifacts/tasks/main.yml#L46

with the import of os_must_gather role right after (with the conditional) and, why not, in a block like this:
```YAML
name: Run os_must_gather if needed, rescue with scp
block:
  - name: Run os_must_gather
     when: cifmw_artifacts_run_must_gather | bool or not cifmw_success_flag.stat.exists
     ansible.builtin.import_role:
      name: os_must_gather
rescue:
  - name: Gather CRC logs from disk
     ansible.builtin.import_tasks: crc.yml
```

What do you think?

deploy-edpm.yml
```diff
@@ -6,6 +6,14 @@
 # openstack-k8s-operators/install_yaml repository in order to leverage its
 # own methods.
 
+- name: Remove status flag
```
dasm, line 9: @cjeanner I'm not sure if that's the right place for removing and injecting the flag. Is there any better (more generic) location?

ci_framework/roles/artifacts/tasks/main.yml
```diff
@@ -44,8 +49,15 @@
...
+- name: Gather logs when forced or tests failed
+  when: cifmw_artifacts_gather_logs | default('false') | bool or not cifmw_success_flag.stat.exists
+  block:
+    - name: Run os_must_gather
+      ansible.builtin.import_role:
+        name: os_must_gather
+  rescue:
+    - name: Get CRC logs if os_must_gather failed
+      ansible.builtin.import_tasks: crc.yml
+  always:
```
dasm, line 61: IIUC @cjeanner you want to skip *any* logs collection on successful run.
I created a block around `os_must_gather` and `crc.yml` with `edpm.yml` . I hope it's going to allow us to spare a disk space and time execution as well.

ci_framework/playbooks/99-logs.yml
```diff
@@ -32,10 +32,6 @@
       ansible.builtin.import_role:
         name: artifacts
 
-    - name: Run os_must_gather
```
dasm, line 35: This block has been moved to `artifacts/tasks/main.yml` as part of log collection.

ci_framework/roles/artifacts/tasks/main.yml
```diff
@@ -14,6 +14,11 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+- name: Verify if "success" flag exists after successful tests execution
```
dasm, line 17: This seems to be dirty way of passing status. However I don't know about any better way of passing statuses between `pre`, `run` and `post` steps.

raukadah:
recheck

dasm:
@cjeanner can you give me anouther round of review?

ci_framework/roles/artifacts/tasks/main.yml
cjeanner: that also work outside of zuul itself, so it's probably not too wrong in the end :)

ci_framework/roles/artifacts/tasks/main.yml
```diff
@@ -44,8 +49,15 @@
     - always
   ansible.builtin.import_tasks: cleanup.yml
 
-- name: Gather CRC logs
-  ansible.builtin.import_tasks: crc.yml
-
-- name: Gather EDPM logs
-  ansible.builtin.import_tasks: edpm.yml
+- name: Gather logs when forced or tests failed
+  when: cifmw_artifacts_gather_logs | default('false') | bool or not cifmw_success_flag.stat.exists
```
cjeanner, line 53: please add this one to the defaults/main.yml in the artifacts role. Doing so, we ensure people not reading the README but checking the defaults/main.yml know about that parameter, and you can also remove the `| default('false')` part.

ci_framework/roles/artifacts/tasks/main.yml
cjeanner: we probably don't need the `always` block for the edpm.yml inclusion.

But that's more cosmetics at this point. Should do the trick.

deploy-edpm.yml
cjeanner: Sounds like the right place. Another way would be with a hook, but it would require people to know about it and inject it. Here, it's written in plain in the main entrypoint playbook, so it should be all fine.

We can't really use the sub-playbooks (well, we can clean in 01-bootstrap), since some aren't always included, such as the 08-run-tests.yml.. So I don't have better alternative.

cjeanner (commented):
/approve

ci_framework/roles/artifacts/tasks/main.yml
dasm: Done

ci_framework/roles/artifacts/tasks/main.yml
cjeanner: hmm yep, by adding the same `when` condition :).

dasm:
recheck

PR #459, opened 2023-08-17
Exporting cifmw_openshift_login_skip_tls_verify as global

pablintino:
  - [x] README in the role


cjeanner:
/lgtm

while it's not using the usual pattern we have with top-level parameters and default, it will work as is.

cjeanner:
/approve
/lgtm

better that way :).

cjeanner:
tempest issue, unrelated to the change. We need it for the new layout, in order to get an easier way to disable TLS check (until we inject CRC CA in the systems, at least).

PR #460, opened 2023-08-17
Ensure we don't fail on stopped VM

cjeanner:
It may happen the VM are already stopped (for instance after a
hypervisor reboot). This patch ensures we don't fail the whole cleanup
in that case.

### DIFF
--- a/ci_framework/roles/libvirt_manager/tasks/clean_layout.yml
+++ b/ci_framework/roles/libvirt_manager/tasks/clean_layout.yml
@@ -14,11 +14,15 @@
     var: cleanup_vms
 
 - name: Destroy machine
+  register: vm_destroy
   community.libvirt.virt:
     command: destroy
     name: "{{ item }}"
     uri: "qemu:///system"
   loop: "{{ cleanup_vms }}"
+  failed_when:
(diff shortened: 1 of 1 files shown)


ci_framework/roles/libvirt_manager/tasks/clean_layout.yml
```diff
@@ -14,11 +14,15 @@
...
 - name: Destroy machine
+  register: vm_destroy
   community.libvirt.virt:
     command: destroy
     name: "{{ item }}"
     uri: "qemu:///system"
   loop: "{{ cleanup_vms }}"
+  failed_when:
+    - vm_destroy.rc is defined
+    - vm_destroy.rc > 1
```
dasm, line 25: For the record, `vm_destroy.rc > 1` because when we're trying to destroy stopped VM, it errors with `rc=1`
```
[dasm@sys ~]$ sudo virsh list --all
 Id   Name             State
---------------------------------
 -    test             shut off

[dasm@sys ~]$ sudo virsh destroy test
error: Failed to destroy domain 'test'
error: Requested operation is not valid: domain is not running

[dasm@sys ~]$ echo $?
1
```

PR #462, opened 2023-08-21
Set proper container registry url and hash for edpm_deploy

raukadah:
edpm_deploy lacks the support for setting proper registry url and container hash.

In downstream and RDO periodic job, we use different container registries. In order to test proper content, we need to set the proper registries otherwise we will see weired issues.

### DIFF
--- a/ci_framework/roles/edpm_deploy/tasks/main.yml
+++ b/ci_framework/roles/edpm_deploy/tasks/main.yml
@@ -22,6 +22,8 @@
       {{
         cifmw_install_yamls_environment |
         combine({'PATH': cifmw_path}) |
+        combine({'DATAPLANE_REGISTRY_URL': cifmw_edpm_deploy_registry_url }) |
+        combine({'DATAPLANE_CONTAINER_TAG': cifmw_repo_setup_full_hash | default(cifmw_install_yamls_defaults['DATAPLANE_CONTAINER_TAG']) }) |
         combine(cifmw_edpm_deploy_extra_vars | default({}))
       }}
     cacheable: true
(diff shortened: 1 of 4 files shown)


raukadah:
recheck

raukadah:
recheck

raukadah:
recheck

ci_framework/roles/edpm_deploy/tasks/main.yml
```diff
@@ -14,6 +14,14 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+- name: Set EDPM container registry related vars
+  tags:
+    - always
+  ansible.builtin.set_fact:
```
pablintino, line 20: If these var is used only in a single place why creating a fact for a single use? Can be perfectly placed in https://github.com/openstack-k8s-operators/ci-framework/blob/7c493a6d767b51339bc59fc2168b40b0f64e7ee3/ci_framework/roles/edpm_deploy/tasks/main.yml#L25-L36

ci_framework/roles/edpm_deploy/tasks/main.yml
```diff
@@ -14,6 +14,14 @@
...
+    cifmw_edpm_deploy_registry:
+      DATAPLANE_REGISTRY_URL: "{{ cifmw_edpm_deploy_registry_url }}"
```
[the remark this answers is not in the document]
raukadah: Good point, Done

ci_framework/roles/edpm_deploy/tasks/main.yml
raukadah: Done

raukadah:
recheck

PR #478, opened 2023-08-28
Change base container image ubi9->centos

steveb:
A FIPS enabled EDPM node is currently failing because libvirt doesn't start[1], logging this message:

  error : virRandomBytes:113 : internal error: failed to generate byte
          stream: An error has been detected in the library and cannot continue
          operations.

In any FIPS enabled environment this issue can be replicated in a container:

  # podman run --rm quay.io/podified-antelope-centos9/openstack-nova-libvirt:current-podified gnutls-cli-debug
  Error in GnuTLS initialization: Error while performing self checks.
  global state initialization error

The root cause of this is that the centos gnutls package installed in the container doesn't work with a ubi9 base image when FIPS is enforced. The centos gnutls works with a centos base image, and the ubi gnutls package works with a ubi9 base image[2].

This change switches the base image to centos:stream9 specifically to fix this FIPS issue, but there may be other problems with mixing packages across distros.

[1] http://pastebin.test.redhat.com/1108025
[2] http://pastebin.test.redhat.com/1108270

  - [x] README in the role

### DIFF
--- a/ci_framework/roles/build_containers/defaults/main.yml
+++ b/ci_framework/roles/build_containers/defaults/main.yml
@@ -43,5 +43,5 @@ cifmw_build_containers_volume_mounts:
 # cifmw_build_containers_build_timeout
 cifmw_build_containers_repo_dir: "{{ cifmw_build_containers_basedir }}/artifacts/repositories"
 cifmw_build_containers_image_tag: current-podified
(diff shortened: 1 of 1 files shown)


raukadah:
recheck

raukadah (approved):
/approve

raukadah:
recheck

PR #484, opened 2023-08-30
Allow to override repo-setup in downstream

Sandeepyadav93:
With this PR, we do the following:-

* Create the repo-setup-downstream OpenStackDataPlaneService (which is
  created in downstream pre-playbook)

* Determine if repo-setup service is present in OpenStackDataPlane
or not. We need this check because install_yamls is tagged and we 
don't have repo-setup service in OpenStackDataPlane in v0.1.0 tag

* Add/Replace repo-setup-downstream with repo-setup service.

* We will do a cleanup in future once we drop dev-preview#1 job to 
  add repo-setup and will only keep replacing logic here.

  - [x] README in the role

### DIFF
--- a/ci_framework/roles/edpm_deploy_baremetal/tasks/main.yml
+++ b/ci_framework/roles/edpm_deploy_baremetal/tasks/main.yml
@@ -115,6 +115,63 @@
   ansible.builtin.debug:
     var: compute_nodes_output.stdout_lines
 
+- name: Patch OpenStackDataPlane to add repo-setup-downstream service
+  when:
+    - cifmw_edpm_deploy_baremetal_repo_setup_override
+    - not cifmw_edpm_deploy_baremetal_dry_run
+  block:
+    # This file will be created in downstream job's pre-playbook
+    - name: Create repo-setup-downstream OpenStackDataPlaneService
+      environment:
+        KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+        PATH: "{{ cifmw_path }}"
+      ansible.builtin.command:
+        cmd: >-
+          oc apply -n {{ cifmw_install_yamls_defaults['NAMESPACE'] }}
+          -f "{{ cifmw_installyamls_repos }}/devsetup/edpm/services/dataplane_v1beta1_openstackdataplaneservice_reposetup_downstream.yaml"
+
+    # We can drop this step once we drop dev-preview#1 jobs in downstream
+    # This is added because install_yamls is tagged and we don't
+    # have repo-setup service in OpenStackDataPlane in v0.1.0 tag
+    - name: Get list of services defined under OpenStackDataPlane resource
+      environment:
+        KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+        PATH: "{{ cifmw_path }}"
+      ansible.builtin.command:
+        cmd: >-
+          oc get openstackdataplane openstack-edpm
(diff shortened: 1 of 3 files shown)


zuul.d/projects.yaml
```diff
@@ -1,55 +1,4 @@
...
-      - cifmw-molecule-polarion
-      - cifmw-molecule-registry_deploy
-      - cifmw-molecule-repo_setup
-      - cifmw-molecule-reproducer
-      - cifmw-molecule-rhol_crc
-      - cifmw-molecule-run_hook
-      - cifmw-molecule-set_openstack_containers
-      - cifmw-molecule-tempest
-      - cifmw-molecule-test_deps
-    name: openstack-k8s-operators/ci-framework
```
danpawlik, line 55: this should be not removed, or?

danpawlik, line 55: this should not be removed, or?

ci_framework/roles/edpm_deploy_baremetal/tasks/main.yml
```diff
@@ -115,6 +115,28 @@
   ansible.builtin.debug:
     var: compute_nodes_output.stdout_lines
 
+- name: Create repo-setup-downrstream OpenStackDataPlaneService
```
rlandy, line 118: typo: repo-setup-downstream

Sandeepyadav93:
recheck

cjeanner:
/lgtm

Why hold it, @Sandeepyadav93 ?

Sandeepyadav93:
@cjeanner it's ready to go, I missed removing the hold flag earlier.

ci_framework/roles/edpm_deploy_baremetal/tasks/main.yml
```diff
@@ -115,6 +115,63 @@
   ansible.builtin.debug:
     var: compute_nodes_output.stdout_lines
 
+- name: Patch OpenStackDataPlane to add repo-setup-downstream service
```
pablintino, line 118: Just a couple of suggestions:

1. If we can add the service before applying the CR will be nice, not so sure that for a this service it will make any real difference, but, just an idea.
2. We should create a task for tracking the dev-preview things, so when it's possible come bakc here and  remove it. Here you have an example for a similar situation. https://issues.redhat.com/browse/OSPCIX-58

Sandeepyadav93: @pablintino 

Thanks for the suggestion.

1>> If we can add the service before applying the CR will be nice, not so sure that for a this service it will make any real difference, but, just an idea.

Based on my understanding, since the deployStrategy is set to false by default in the openstackdataplane CR, when we creating the service should not make any difference as long as it's done before we change the deployStrategy to false, which we do on line 175 in this file. The dev-preview job is green with this change, so I believe we are in good shape.

2>> We should create a task for tracking the dev-preview things, so when it's possible come bakc here and remove it. Here you have an example for a similar situation. https://issues.redhat.com/browse/OSPCIX-58

While there are inline code comments (regarding the to-do of removing these extra logins from dev-preview#1), I agree with your suggestion that creating a Jira card would be beneficial for tracking. I've gone ahead and created a card (OSP-28082) under our dev-preview#2 story to eliminate the dev-preview#1 workaround once we remove that line.

pablintino: Thanks for considering the suggestions, I'll approve it

PR #486, opened 2023-08-30
Ony report DLRN results when var is true

rlandy:
cifmw_dlrn_report_result should be set to true
otherwise skip the whole role.

  - [x] README in the role

### DIFF
--- a/ci_framework/roles/dlrn_report/tasks/main.yml
+++ b/ci_framework/roles/dlrn_report/tasks/main.yml
@@ -14,8 +14,11 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
-- name: Install dlrnapi-client
-  ansible.builtin.import_tasks: install.yml
+- name: Only report DLRN results when var is set
+  when: cifmw_dlrn_report_result | bool
+  block:
+    - name: Install dlrnapi-client
+      ansible.builtin.import_tasks: install.yml
 
-- name: Report DLRN results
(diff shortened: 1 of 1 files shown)


ci_framework/roles/dlrn_report/tasks/main.yml
```diff
@@ -14,8 +14,11 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
-- name: Install dlrnapi-client
-  ansible.builtin.import_tasks: install.yml
+- name: Only report DLRN results when var is set
+  when: cifmw_dlrn_report_result | bool
```
dasm, line 18: FTR, this is a default value: https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci_framework/roles/dlrn_report/README.md?plain=1#L19

https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci_framework/roles/dlrn_report/defaults/main.yml#L24

pablintino:
recheck

dasm (commented):
/lgtm

ci_framework/roles/dlrn_report/tasks/main.yml
pablintino: Agree, should be nice to have it written there

rlandy: That's ok - the base jobs set to false

PR #488, opened 2023-08-31
Collect pod status and their crs file

raukadah:
In order to debug and check resources in successful logs, We need pod status and their cr files.

### DIFF
--- a/ci_framework/roles/artifacts/tasks/crc.yml
+++ b/ci_framework/roles/artifacts/tasks/crc.yml
@@ -40,3 +40,25 @@
           scp -v -r -i {{ cifmw_artifacts_crc_sshkey }}
           root@{{ cifmw_artifacts_crc_host }}:/ostree/deploy/rhcos/var/log/pods
           {{ cifmw_artifacts_basedir }}/logs/crc/
+
+    - name: Collect pods status and cr files
+      ignore_errors: true
+      environment:
+        KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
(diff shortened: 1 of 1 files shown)


ci_framework/roles/artifacts/tasks/crc.yml
```diff
@@ -40,3 +40,26 @@
           scp -v -r -i {{ cifmw_artifacts_crc_sshkey }}
           root@{{ cifmw_artifacts_crc_host }}:/ostree/deploy/rhcos/var/log/pods
           {{ cifmw_artifacts_basedir }}/logs/crc/
+
+    - name: Collect pods status and cr files
```
cjeanner, line 44: the whole file is included if there's a failure or if we instruct the run to gather logs even in case of success:
https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci_framework/roles/artifacts/tasks/main.yml#L67

You probably want to either push that task into the main.yml file, or rework a bit the conditionals, like moving it in this file here, with a block around the whole file but your new task, for instance.

raukadah: @cjeanner When a job successed, we collect crc logs. If a job fails, we run os_must_gather. The above cr log collection will run when the job successed. In case of failure, it will be skipped as crc task files is skipped.
os_must_gather already captures the cr infos.  I hope we should not move it to main.yml.

cjeanner: not sure we should collect crc logs in case of success. anyway....

PR #489, opened 2023-08-31
Add Dev-preview workaround to finish EDPM deployment

raukadah:
It is a know issue with the dev-preview nova-operator that 
it runs the libvirt and nova-compute deployment twice but 
reports ready status after the first run. So with this sleep 
the job waits for the deployment to really stabilize before 
start running the tempest tests.
    
Adding 5 sleep before running tests as a workaround fixes
the issue.

### DIFF
--- a/deploy-edpm.yml
+++ b/deploy-edpm.yml
@@ -35,6 +35,17 @@
 - name: Import admin setup related playbook
   ansible.builtin.import_playbook: ci_framework/playbooks/07-admin-setup.yml
 
+  # Note(Chandan): It is a know issue with the dev-preview nova-operator that
+  # it runs the libvirt and nova-compute deployment twice but reports ready
+  # status after the first run. So with this sleep the job waits for the
+  # deployment to really stabilize before start running the tempest tests.
+- name: Sleep for 5 mins to finish th EDPM deployment
+  hosts: "{{ cifmw_target_host | default('localhost') }}"
+  tasks:
+    - name: Add a sleep for 5 mins
+      ansible.builtin.pause:
+        minutes: 5
+
 - name: Import run test playbook
   ansible.builtin.import_playbook: ci_framework/playbooks/08-run-tests.yml
   when: cifmw_run_tests | default('false') | bool


deploy-edpm.yml
```diff
@@ -35,6 +35,16 @@
 - name: Import admin setup related playbook
   ansible.builtin.import_playbook: ci_framework/playbooks/07-admin-setup.yml
 
+  # Note(Chandan): Add 5 mins sleep to finish the EDPM deployment
+  # before running tempest to avoid no-valid host.
+  # It is a known issue with dev-preview dataplane operator.
```
gibizer, line 40: I would say:
... It is a know issue with the dev-preview nova-operator that it runs the libvirt and nova-compute deployment twice but reports ready status after the first run. So with this sleep the job waits for the deployment to really stabilize before start running the tempest tests.

deploy-edpm.yml
```diff
@@ -35,6 +35,17 @@
...
 
+  # Note(Chandan): It is a know issue with the dev-preview nova-operator that
+  # it runs the libvirt and nova-compute deployment twice but reports ready 
+  # status after the first run. So with this sleep the job waits for the 
+  # deployment to really stabilize before start running the tempest tests.
+- name: Sleep for 5 mins to finish th EDPM deployment
+  hosts: "{{ cifmw_target_host | default('localhost') }}"
+  tasks:
+    - name: Add a sleep for 5 mins
+      ansible.builtin.pause:
```
cjeanner, line 46: no other way than a hard-coded pause?

raukadah: Nope, This will be probably fixed in the main branch. Once Dev preview is released. We can revert this patch.

pablintino: And there is no other way? There is no log trace that points that is ready? No service exposed? Nothing?

rlandy: If there is nothing to poll for, I'd approve this.

rlandy:
recheck

rlandy:
/lgtm
agreed to keep this until nova libvirt rework is done

deploy-edpm.yml
gibizer: if you want to complicate things then you can try to watch for running deployment jobs for 5 minutes after the deployment was ready. But as it also has the 5 minutes magic number I don't think it worth the complexity.

raukadah:
recheck

pablintino:
recheck

raukadah:
recheck

rlandy:
recheck

rlandy:
recheck

viroel:
recheck

PR #490, opened 2023-08-31
Remove galera job

cjeanner:
This job isn't really useful - it's better to add the parameter into the
edpm-deployment job instead.

We also add a small doc about nodeset override in the Zuul job
definition, since it may not be that obvious.


zuul.d/edpm_multinode.yaml
```diff
@@ -123,49 +123,6 @@
...
-      nodes:
-        - name: controller
-          label: cloud-centos-9-stream-tripleo-vexxhost
-        - name: crc
-          label: coreos-crc-extracted-xxl
     vars:
-      cifmw_deploy_edpm: false
-      podified_validation: true
-      cifmw_run_tests: false
       make_openstack_deploy_params:
```
pablintino, line 127: No one is using the openstack_deploy target, so this parameter (important, parameter, not env variable) is not gonna be used. Anyway, I don't like to put here explicit role variables from install_yamls, as we are making the jobs "aware" of the internal calls to install_yamls that we are using, that can change at any time without notice and should't brake. For this purpose we should use `cifmw_install_yamls_vars`.

cjeanner: updated - thanks. I'm lost with all of those parameters :]

cjeanner:
recheck

raukadah (approved):
/approve

zuul.d/edpm_multinode.yaml
```diff
@@ -112,6 +112,8 @@
       cifmw_extras:
         - '@scenarios/centos-9/ci.yml'
         - '@scenarios/centos-9/multinode-ci.yml'
+      cifmw_install_yamls_vars:
+        GALERA_REPLICAS: 3
```
pablintino, line 116: This will override the setting in ci.yml:
````
cifmw_install_yamls_vars:
  BMO_SETUP: false
```

raukadah:
I have added hold to confirm with @viroel whether this job is used in dev preview?

zuul.d/edpm_multinode.yaml
cjeanner: ... we probably want to introduce the hash policy thingy allowing ansible to merge instead of replacing....

cjeanner:
@raukadah multinode isn't used downstream

zuul.d/edpm_multinode.yaml
raukadah: scenarios/centos-9/ci.yml is used in old e2e job only and new crc edpm job. Can we move `BMO_SETUP: false` and `GALERA_REPLICAS: 3` under cifmw_install_yamls_vars var in ci.yml? this way we donot need to define in the job definition.

PR #491, opened 2023-08-31
Add adoption base job to use extracted crc

cescgina:
Add a base job to use the extracted crc layout in adoption CI. The
running job will still reside in rdo-jobs, and will probably port it in
the future, but this change should make it easier to consume the new
layout.


cescgina:
In order to to reuse the `06-deploy-edpm` playbook in adoption to deploy the control plane considering the multinode layout, this PR adds a variable to skip a few tasks from the `fetch_compute_facts` hook

marios (commented):
seems reasonable

PR #496, opened 2023-09-01
Remove hosts file entries for CRC

pablintino:
Since we are now using crc's DNS there is no need to manually add CRC hostnames into the controller hosts file.


cjeanner:
/lgtm

we'll need to do some more cleaning - maybe better to go step by step rather than one big thing ;)

raukadah:
recheck

PR #504, opened 2023-09-01
Continue ci_script implementation replacing shell and command built-i…

adrianfusco:
Continue ci_script implementation replacing shell and command built-in modules.
Role: install_ca

  - [X] README in the role

### DIFF
--- a/ci_framework/roles/install_ca/defaults/main.yml
+++ b/ci_framework/roles/install_ca/defaults/main.yml
@@ -20,3 +20,4 @@
 cifmw_install_ca_trust_dir: '/etc/pki/ca-trust/source/anchors/'
 cifmw_install_ca_bundle_inline: ''
 cifmw_install_ca_bundle_src: ''
+cifmw_install_ca_output_dir: "{{ cifmw_basedir | default(ansible_user_dir ~ '/ci-framework-data') }}"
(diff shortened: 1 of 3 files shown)


ci_framework/roles/run_hook/tasks/playbook.yml
```diff
@@ -73,7 +74,7 @@
           -e "hook_name={{ hook_name }}"
           -e "playbook_dir={{ playbook_path | dirname }}"
           {{ playbook_path }}
-        create: "{{ hook.creates | default(omit) }}"
```
adrianfusco, line 76: @cjeanner @pablintino the build is giving error because is trying to execute `noop.yml.sh` (because is transfering `hooks/playbooks/noop.yml`) since I changed the `create` parameter that was wrong. 

Is this parameter necessary? If not I can delete it. Thx.

cjeanner: `creates` parameter allows to ensure we don't run the same hook twice by allowing to check for the existence of some location. We therefore should keep it, in order to ensure idempotent runs (or, at least, try to)

cjeanner:
hmm, maybe run_hook is that kind of role we shouldn't move the the ci_script: the run location and related thing IS an issue imho, and we may face more issues than really needed.
The hook log, since pure ansible, is in the main ansible.log file already, allowing a "clean" debugging already. IMHO, we shouldn't change that one.

adrianfusco:
It doesn't provide relevant information that can be logged.

PR #508, opened 2023-09-04
Continue ci_script implementation

adrianfusco:
Continue ci_script implementation replacing shell and command built-in modules
Roles: **openshift_setup**, **openshift_login**
Some command builti-in hasn't been ported since they don't give any important or relevant info that can be useful in logs.

  - [X] README in the role

### DIFF
--- a/ci_framework/roles/openshift_login/tasks/try_login.yml
+++ b/ci_framework/roles/openshift_login/tasks/try_login.yml
@@ -34,8 +34,9 @@
   environment:
     PATH: "{{ cifmw_path }}"
     KUBECONFIG: "{{ cifmw_openshift_login_kubeconfig }}"
-  ansible.builtin.command:
-    cmd: >-
+  ci_script:
+    output_dir: "{{ cifmw_openshift_login_basedir }}/artifacts"
+    script: >-
       oc login
       {%- if cifmw_openshift_login_provided_token is not defined %}
       {%- if cifmw_openshift_login_user is defined %}
--- a/ci_framework/roles/openshift_login/tasks/main.yml
+++ b/ci_framework/roles/openshift_login/tasks/main.yml
@@ -14,6 +14,13 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+- name: Ensure output directory exists
+  tags:
+    - always
+  ansible.builtin.file:
+    path: "{{ cifmw_openshift_login_basedir }}/artifacts"
+    state: directory
+
 - name: OpenShift login
   ansible.builtin.include_tasks:
     file: "login.yml"
--- a/ci_framework/roles/openshift_setup/defaults/main.yml
+++ b/ci_framework/roles/openshift_setup/defaults/main.yml
@@ -17,6 +17,7 @@
 
 # All variables intended for modification should be placed in this file.
 # All variables within this role should have a prefix of "cifmw_openshift_setup"
+cifmw_openshift_setup_basedir: "{{ cifmw_basedir | default(ansible_user_dir ~ '/ci-framework-data') }}"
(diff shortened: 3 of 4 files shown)


ci_framework/roles/openshift_login/tasks/try_login.yml
```diff
@@ -34,8 +34,9 @@
   environment:
     PATH: "{{ cifmw_path }}"
     KUBECONFIG: "{{ cifmw_openshift_login_kubeconfig }}"
-  ansible.builtin.command:
-    cmd: >-
+  ci_script:
+    output_dir: "{{ cifmw_openshift_login_basedir }}/artifacts"
```
cjeanner, line 38: is that location actually created in the role?

adrianfusco: Is it necessary? Because the `ci_script` module fails if this folder doesn't exist, isn't? Here is the artifact of this task: https://logserver.rdoproject.org/08/508/4ccac5a1eefc11ae08101c7a9054dd69f9f5da3f/github-check/cifmw-molecule-openshift_login/b236ea0/ci-framework-data/artifacts/ (Maybe I'm wrong.)

lewisdenny: Hi @adrianfusco 

I think Cedric means like [1] so while the ci_script module will fail if the directory doesn't exist it would be best to ensure it does exist. 

[1] https://github.com/lewisdenny/ci-framework/blob/9c8cdc4609a8252bd42983d440c2ea23e4554505/ci_framework/roles/libvirt_manager/tasks/main.yml#L17-L18

adrianfusco: Hi @lewisdenny ack. I'll update the PR

lewisdenny: Thanks for addressing this, lgtm 👍🏻

lewisdenny (approved):

lewisdenny:
recheck

ci_framework/roles/openshift_login/tasks/try_login.yml
dasm, line 38: Hey @adrianfusco I wouldn't be surprised if you answered this question somewhere already.
Why are you replacing `ansible.builtin.command` with `ci_script` plugin? IIUC we're not gathering any logs here, hence there is no immediate need to change that.

Additionally, this piece of code uses `environment`. I checked the code [1] but I don't see passing of environment variables there. @pablintino can `ci_script` be used with `environment` variables?

[1]: https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci_framework/plugins/action/ci_script.py

lewisdenny: Hi @dasm, great thought. I quickly tested this and can confirm it's fine.

Environment sets the shell environment vars so the ci_script module gets the support for free.

```
- name: test ci_script
  hosts: localhost
  tasks:
    - name: test ci_script
      environment:
        TESTVAR: "hello"
      ci_script:
        output_dir: "/Users/ldenny/Code/ci/ci-framework/plugins/action/test"
        script: >-
          echo $TESTVAR ;
          env | grep TESTVAR
```

```
changed: [localhost] => {
    "changed": true,
    "rc": 0,
    "stderr": "",
    "stderr_lines": [],
    "stdout": "hello\nTESTVAR=hello\n",
    "stdout_lines": [
        "hello",
        "TESTVAR=hello"
    ]
}
```

dasm: Neat, thanks!

PR #511, opened 2023-09-04
Add ci_kustomize action plugin

pablintino:
This action plugin is a direct replacement of the edpm_kustomize role. It takes a path to the file you want to kustomize and applies a set of kustomizations given though it's kustomizations parameter or through file ones placed on the same directory where the target manifest is placed.

  - [x] README in the role

### DIFF
--- a/ci_framework/plugins/action/ci_kustomize.py
+++ b/ci_framework/plugins/action/ci_kustomize.py
@@ -1,0 +1,762 @@
+#!/usr/bin/env python3
+
+# Copyright Red Hat, Inc.
+# Apache License Version 2.0 (see LICENSE)
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+
+DOCUMENTATION = r"""
+---
+action: ci_kustomize
+
+short_description: Applies a set of k8s Kustomizations to a set of manifests
+
+description:
+- Allows applying a set of Kustomizations to a given set of manifests using
+  the kustomize or oc tools.
+- kustomize or oc should be discoverable by in the PATH.
+- This modules takes a set of manifest files pointed by I(target_path) and
+  applies to them the set of kustomizations, if available.
+- The kustomization result is always saved in a single file and passed to
+  the caller in the O(result) field if success.
+- Kustomizations can be passed by file and/or by I(kustomizations), they get
+  applied one by one.
+- Filesystem Kustomizations are searched by default in the I(target_path), but
+  more search paths can be added by passing I(kustomizations_paths).
+- Filesystem Kustomizations are applied by strictly alphabetical order.
+- Kustomizations passed by I(kustomizations) are applied by apparition order.
+- Kustomizations passed by I(kustomizations) are applied after the ones from
+  filesystem by default, if the contrary is not told so by
+  I(kustomization_files_goes_first).
+- I(kustomizations) accepts a wide range of input types. More precisly: plain
+  string, list of strings, dict with a single kustomization or list of dicts.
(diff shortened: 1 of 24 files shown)


ci_framework/plugins/action/ci_kustomize.py
```diff
@@ -0,0 +1,755 @@
...
+---
+action: ci_kustomize
+
+short_description: Applies a set of k8s Kustomizations to a set of manifests
+
+description:
+- Allows applying a set of Kustomizations to a given set of manifests using
+  the kustomize or oc tools.
+- kustomize or oc should be discoverable by in the PATH.
+- This modules takes a set of manifest files pointed by I(target_path) and
```
marios, line 21: this "I(" must be part of module documentation generation? I see it in a few places here

ci_framework/plugins/action/ci_kustomize.py
```diff
@@ -0,0 +1,755 @@
...
+  applies to them the set of kustomizations, if available.
+- The kustomization result is always saved in a single file and passed to
+  the caller in the O(result) field if success.
```
marios, line 24: but is that one typo? O(

marios (commented):
first pass and didn't spot anything major... but the module is pretty big so I need to revisit and stare again if it is still around ;)

ci_framework/plugins/action/ci_kustomize.py
pablintino: And this one references and `output` field

ci_framework/plugins/action/ci_kustomize.py
pablintino: `I`is supposed to reference a input arg, that's what docs say, but I haven't checked the generated doc.

cjeanner: (not sure we generate any doc out of those kind of content, to be fair)

.pre-commit-config.yaml
```diff
@@ -53,6 +53,7 @@ repos:
     rev: 23.9.1
     hooks:
       - id: black
+        entry: black --line-length 79 --preview
```
cjeanner, line 56: I think you should use this:
https://pre-commit.com/#pre-commit-configyaml---hooks

ci_framework/plugins/action/ci_kustomize.py
marios: thanks

ci_framework/plugins/action/ci_kustomize.py
marios: thank you

ci_framework/plugins/action/ci_kustomize.py
```diff
@@ -0,0 +1,731 @@
...
+      - /home/user/source/prod-kustomizations
+"""
+
+RETURN = r"""
+count:
+    description: Total number of Kustomizations applied
+    returned: success
+    type: int
+    sample: 10
+kustomizations_paths: typing.List[str]
```
raukadah, line 148: I think it should be `kustomizations_paths:` without typing.List[str]

raukadah (commented):
Thank you @pablintino for working on this. Overall it looks good from first glance. I will go over the code one more time.

ci_framework/plugins/action/ci_kustomize.py
pablintino: Good catch!

pablintino:
recheck

marios (approved):
/lgtm

PR #513, opened 2023-09-05
Skip EDPM jobs on content provider irrelevant_files

raukadah:
We are seeing Zuul job freeze issue against different operator changes due to difference in irrelevant_files on content provider and EDPM job.

Keeping the same irrelevant_files list fixes the following issue.
```
Unable to freeze job graph: Job podified-multinode-edpm-deployment-crc
depends on openstack-k8s-operators-content-provider which was not run.
```

### DIFF
--- a/zuul.d/edpm_multinode.yaml
+++ b/zuul.d/edpm_multinode.yaml
@@ -19,10 +19,25 @@
           label: cloud-centos-9-stream-tripleo-vexxhost
         - name: crc
           label: coreos-crc-extracted-xxl
+    # Note: Always sync irrelevant-files from content provider base job
     irrelevant-files:
       - .*/*.md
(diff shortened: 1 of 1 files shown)


zuul.d/edpm_multinode.yaml
```diff
@@ -19,10 +19,26 @@
...
     irrelevant-files:
       - .*/*.md
-      - .*/*.svg
-      - docs/source/*
+      - ^.github/.*$
+      - ^LICENSE$
+      - ^OWNERS$
+      - ^OWNERS_ALIASES$
+      - ^PROJECT$
+      - ^README.md$
```
cjeanner, line 30: that one may be covered by `.*/*.md` ? or we may correct that one on line 24?

PR #517, opened 2023-09-06
Add support to set concurrency value in tempest

arxcruz:
At this moment, tempest is using all cpus available, and in some test cases this leads to timeout in some resources. Adding the hability to set the proper concurrency will fix this problems.

  - [x] README in the role


raukadah (approved):
/approve

PR #523, opened 2023-09-07
Correct syntax for var creation

cjeanner:


cjeanner:
We may have to force-merge since it's part of the co-dependent PRs related to CRD management... Zuul will probably be red on the deploy jobs.

cjeanner:
Not needed anymore, another fix was in allowing to make the thing pass.

PR #524, opened 2023-09-07
Speed up build_containers molecule job

raukadah:
It will build only keystone containers to validate the build_containers role.

### DIFF
--- a/ci_framework/roles/build_containers/molecule/default/converge.yml
+++ b/ci_framework/roles/build_containers/molecule/default/converge.yml
@@ -19,5 +19,6 @@
   hosts: all
   vars:
     cifmw_build_containers_cleanup: true
+    cifmw_build_containers_config_file: "{{ ansible_user_dir }}/src/github.com/openstack-k8s-operators/ci-framework/roles/build_containers/files/containers.yaml"
   roles:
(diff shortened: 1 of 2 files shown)


arxcruz (approved):
/approve

ci_framework/roles/build_containers/molecule/default/containers.yaml
cjeanner, line 1: care to move that file under a subdirectory named "files"?

ci_framework/roles/build_containers/molecule/default/converge.yml
```diff
@@ -19,5 +19,6 @@
   hosts: all
   vars:
     cifmw_build_containers_cleanup: true
+    cifmw_build_containers_config_file: "{{ ansible_user_dir }}/src/github.com/openstack-k8s-operators/ci-framework/roles/build_containers/molecule/default/containers.yaml"
```
cjeanner, line 22: Once the file is moved, you can use `files/containers.yaml` instead for the path.

cjeanner: (that file has to be seen by TCIB, can't use relative path - sorry, my bad)

PR #525, opened 2023-09-07
Fix the way we call ini lookup on the `initial.ceph` config file

fmount:
With the old syntax we experienced some failures that seem to be fixed by this way of doing the lookup.

- [x] README in the role

### DIFF
--- a/ci_framework/playbooks/ceph.yml
+++ b/ci_framework/playbooks/ceph.yml
@@ -163,7 +163,7 @@
     # public network always exist because is provided by the ceph_spec role
     - name: Get Storage network range
       ansible.builtin.set_fact:
-        cifmw_cephadm_rgw_network: "{{ lookup('ansible.builtin.ini', 'public_network', section='global',  file=cifmw_cephadm_bootstrap_conf) }}"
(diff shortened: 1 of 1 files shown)


fultonj (approved):
That funky lookup plugin.

Thanks Francesco

katarimanojk (commented):
LGTM, Thanks for fixing this.

katarimanojk (approved):

fultonj:
This playbook isn't yet CI'd so I'm not sure we need to block merging this fix if it's limited only to playbooks.ceph.yml

I see this recent push added the following which isn't related though AFAICT

https://github.com/openstack-k8s-operators/ci-framework/compare/762425e6e12304dc3ba137f75c59572f58d564f1..60e744a8ada033a4c26efc6a233ffbacfedb383a

fmount:
> This playbook isn't yet CI'd so I'm not sure we need to block merging this fix if it's limited only to playbooks.ceph.yml

I agree, even though we have a green CI this part is not tested because we have no jobs deploying Ceph atm.
Let's recheck and see if we get a successful execution, otherwise we'll merge anyway.

fmount:
recheck

pablintino:
recheck

fultonj (approved):
Tested. It fixes the bug for me.

PR #526, opened 2023-09-07
FQCN fix for calls to ansible.utils.ipaddr in ceph playbook

fultonj:
Update ceph playbook to use ansible.utils.ipaddr in place of just ipaddr.

  - [ ] README in the role

### DIFF
--- a/.spellcheck.yml
+++ b/.spellcheck.yml
@@ -9,7 +9,7 @@ matrix:
       - ./**/*.md
     pipelines:
       - pyspelling.filters.markdown:
-        dictionary:
-          wordlists:
-            - docs/dictionary/en-custom.txt
-        output: docs/_build/en-custom.dic
+    dictionary:
+      wordlists:
+        - docs/dictionary/en-custom.txt
+      output: docs/_build/en-custom.dic


fultonj:
This PR should just focus on using the FQCN fix for calls to ansible.utils.ipaddr so I'm taking it out of draft. It's based on a larger linter change so it might take time to merge.

While looking into this fix we discovered another issue but it's not related. I'll document what I found on that other below if you want details.

DETAILS:

I was able to reproduce the following which @fmount hit.
```
"Unrecognized type <<class 'ansible.vars.hostvars.HostVars'>> for ipaddr filter <value>"
```
Provided I have:

  https://github.com/openstack-k8s-operators/ci-framework/pull/525

The unrecognized type error looks very similar to:

 https://github.com/ansible-collections/ansible.utils/issues/162

The fix for the above is:

  https://github.com/martin-schwenke/ansible.utils/commit/c91a3de55fa8104069fff7126c23c4d5ca82fe71

I can see the same fix in 2.9

  https://github.com/ansible-collections/ansible.utils/blob/2.9.0/plugins/filter/ipaddr.py#L254-L272

My RHEL8.8 system does not have the fix however.

```
[fultonj@hamfast lib]$ sudo find . -name ipaddr.py
./python3.6/site-packages/ansible/plugins/filter/ipaddr.py
[fultonj@hamfast lib]$ 

[fultonj@hamfast lib]$ fgrep 'isinstance' /usr/lib/python3.6/site-packages/ansible/plugins/filter/ipaddr.py
    elif isinstance(value, (list, tuple, types.GeneratorType)):
    if not isinstance(amount, int):
        if isinstance(value, (list, tuple, types.GeneratorType)):
[fultonj@hamfast lib]$
```
So I think that other issue is fixed by ansible itself in a newer version and I hit it because I don't have that version. The reproduction is also intermittent for me.

Regardless, it's not related to using `ansible.utils.ipaddr` vs `ipaddr` so I'm taking this out of draft. If the unrecognized type error comes back and we need to workaround it for an older version of ansible, then it can be solved in a different PR.

pablintino:
@fultonj Hi! the base commits that this PR uses are now in main, you can just remove them

fultonj:
Obsoleted by https://github.com/openstack-k8s-operators/ci-framework/pull/546

PR #528, opened 2023-09-07
Switch ssh key to our own custom one, create cloud-admin

cjeanner:
It seems that "cloud-admin" is the user to use in a really close
future[1],so let's ensure it's present on the compute. Also, it
seems "root" is being used for now, and it's probably good to use
our own key, since we ensure it's present in all the needed users.

[1] https://github.com/openstack-k8s-operators/dataplane-operator/pull/398/files

Depends-On: https://github.com/openstack-k8s-operators/install_yamls/pull/523

### DIFF
--- a/ci_framework/hooks/playbooks/fetch_compute_facts.yml
+++ b/ci_framework/hooks/playbooks/fetch_compute_facts.yml
@@ -161,7 +161,7 @@
                                 {{ '{{' }} lookup("vars", networks_lower[network] ~ "_ip") {{ '}}' }}/{{ '{{' }} lookup('vars', networks_lower[network] ~ '_cidr') {{ '}}' }}
                             routes: {{ '{{' }} lookup('vars', networks_lower[network] ~ '_host_routes') }}
                         {{ '{%' }} endfor {{ '%}' }}
-                        {% endraw -%}
+                        {% endraw %}
 
                   - op: replace
                     path: /spec/nodeTemplate/ansible/ansibleUser
--- a/ci_framework/hooks/playbooks/fetch_compute_facts.yml
+++ b/ci_framework/hooks/playbooks/fetch_compute_facts.yml
@@ -88,46 +88,46 @@
               namespace: {{ cifmw_install_yamls_defaults.NAMESPACE }}
               patches:
               - target:
-                  kind: OpenStackDataPlane
+                  kind: OpenStackDataPlaneNodeSet
                 patch: |-
                   - op: remove
-                    path: /spec/roles/edpm-compute/nodeTemplate/ansibleVars/edpm_network_config_template
+                    path: /spec/nodeTemplate/ansible/ansibleVars/edpm_network_config_template
 
                   - op: replace
-                    path: /spec/roles/edpm-compute/nodeTemplate/ansibleVars/neutron_public_interface_name
+                    path: /spec/nodeTemplate/ansible/ansibleVars/neutron_public_interface_name
                     value: "{{ crc_ci_bootstrap_networks_out.compute.default.iface | default('') }}"
 
                   - op: replace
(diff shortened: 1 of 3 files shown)


raukadah (approved):

dasm (approved):
/lgtm

fao89 (approved):

cjeanner:
there are some other issues to correct, with the "kind" changing - it wasn't updated in all the needed places :(

ci_framework/hooks/playbooks/fetch_compute_facts.yml
```diff
@@ -193,7 +193,7 @@
             namespace: {{ cifmw_install_yamls_defaults.NAMESPACE }}
             patches:
             - target:
-                kind: OpenStackControlPlane
+                kind: OpenStackDataPlaneNode
```
fao89, line 196:
```suggestion
                kind: OpenStackDataPlaneNodeSet
```

slagle:
recheck

slagle:
https://github.com/openstack-k8s-operators/install_yamls/pull/523 might fix the error with the filename ending up in /etc/os-net-config/config.yaml. But what I still don't get is why the kustomization in ci_framework/hooks/playbooks/fetch_compute_facts.yml doesn't overwrite that as it's using op: replace. So, I'm still not sure the kustomizations are being applied in the right order by the edpm_kustomize role.

I added a depends-on for  https://github.com/openstack-k8s-operators/install_yamls/pull/523 here, and rechecked.

ci_framework/hooks/playbooks/fetch_compute_facts.yml
```diff
@@ -161,14 +161,14 @@
                                 {{ '{{' }} lookup("vars", networks_lower[network] ~ "_ip") {{ '}}' }}/{{ '{{' }} lookup('vars', networks_lower[network] ~ '_cidr') {{ '}}' }}
                             routes: {{ '{{' }} lookup('vars', networks_lower[network] ~ '_host_routes') }}
                         {{ '{%' }} endfor {{ '%}' }}
-                        {% endraw %}
+                        {% endraw -%}
```
slagle, line 164: This change is currently what is causing the issue. By chomping all the whitespace from here to the end of the line, it eats the new line. So, this "op" is munged with the next one and the spacing is not right. oc kustomize just silently ignores this "op". You can see how the ending of it looks different from the others in:

https://logserver.rdoproject.org/98/398/c228d4f63078573a53f2a4ddc84a8c4441ec7b12/github-check/podified-multinode-edpm-deployment-crc/08d07b0/controller/ci-framework-data/artifacts/manifests/openstack/dataplane/cr/kustomization.yaml

Notice only one \n and then a bunch of whitespace after this op, when all the others look different. When I fixed this locally and ran oc kustomize, I then got networkConfig added.

slagle: I pushed a new commit to hopefully address the whitespace issue. If it doesn't fix it, I suggest double checking that newtorkConfig.template got applied to the manifest being created for openstackdataplanenodeset.

cjeanner: erk, that was tricky. Thanks @slagle ! I'll merge as-is since CI is all happy again :).

cjeanner:
recheck

cjeanner:
OK - we can merge this patch. We apparently having once again an interlocked dependency. This is a proof of work I'm happy with:
https://review.rdoproject.org/zuul/buildset/4779fce14ccc4800902d4b9239f3c2e1

Merging on our side, we'll need to get the following in as well to resolve our current issues:
- https://github.com/openstack-k8s-operators/dataplane-operator/pull/398
- https://github.com/openstack-k8s-operators/install_yamls/pull/523

PR #529, opened 2023-09-07
Update remaining reference to OpenStackDataPlane

slagle:
Signed-off-by: James Slagle <jslagle@redhat.com>

  - [x] README in the role


fao89 (approved):

cjeanner:
Thanks you @slagle - I've imported that on in #528 in a distinct patch, so we should be OK. Both are needed anyway to unlock the situation.
Closing this one, since it should be in the mentioned PR.

PR #530, opened 2023-09-07
Implement sosreport role for log collection

dasm:
Replace `ci_script` action with a "sos" log collector. By using industry-provided standard, we ensure that collected logs adhere to standards.

  - [x] README in the role


dasm:
It is not creating `sos.conf` file correctly. Needs further investigation.

raukadah:
@dasm sosreport provides an option `--config-file` to `Specify alternate configuration file.`  We can use this option and generate sos.conf during roles execution?

Issue #531, opened 2023-09-07
ci_script broke ceph playbook


fultonj:
After [1] merged I've been unable to use the ceph playbook in a standalone context as documented in the `cifmw_cephadm`
role readme [2]. Instead I get:
```
  Permission denied: '/root/ci-framework-data/artifacts'
```
I can't even override it easily by passing something like:
```
  -e cifmw_cephadm_basedir=/tmp
```
I tried adding a pre_task to ensure `cifmw_cephadm_basedir` (and `cifmw_cephadm_basedir/artifacts`) exists but this lead my down a rabbit hole of having to create other directories for it until it mysteriously failed.
```
TASK [cifmw_cephadm : Run cephadm bootstrap output_dir={{ cifmw_cephadm_basedir }}/artifacts, script={{ cifmw_cephadm_bin }} \
{% if not cifmw_cephadm_default_container %}--image {{ cifmw_cephadm_container_ns + '/' + cifmw_cephadm_container_image + ':' + cifmw_cephadm_container_tag|string }} \{% endif %}
bootstrap \
--skip-firewalld \
--skip-prepare-host \
--ssh-private-key /home/{{ cifmw_cephadm_ssh_user }}/.ssh/id_rsa \
--ssh-public-key /home/{{ cifmw_cephadm_ssh_user }}/.ssh/id_rsa.pub \
--ssh-user {{ cifmw_cephadm_ssh_user }} \
--allow-fqdn-hostname \
--output-keyring {{ cifmw_cephadm_admin_keyring }} \
--output-config {{ cifmw_cephadm_conf }} \
--fsid {{ cifmw_cephadm_fsid }} \
{% if cifmw_cephadm_registry_url|length > 0 %}--registry-url {{ cifmw_cephadm_registry_url }} \{% endif %}
{% if cifmw_cephadm_registry_username|length > 0 %}--registry-username {{ cifmw_cephadm_registry_username }} \{% endif %}
{% if cifmw_cephadm_registry_password|length > 0 %}--registry-password {{ cifmw_cephadm_registry_password }} \{% endif %}
{% if cifmw_cephadm_spec_on_bootstrap %}--apply-spec {{ cifmw_cephadm_spec }} \{% endif %}
{% if cifmw_cephadm_assimilate_conf_stat.stat.exists %}--config {{ cifmw_cephadm_assimilate_conf }} \{% endif %}
[...]

PR #532, opened 2023-09-07
Revert "Continue ci_script implementation replacing shell and command built-in modules"

fultonj:
Revert ceph playbook move to ci_script.

This reverts commit 0df6fa4cef78a8147405e40d3ebaa9672040846b.

Please see https://github.com/openstack-k8s-operators/ci-framework/issues/531

- [x] README in the role

### DIFF
--- a/ci_framework/roles/cifmw_cephadm/defaults/main.yml
+++ b/ci_framework/roles/cifmw_cephadm/defaults/main.yml
@@ -64,7 +64,6 @@ cifmw_cephadm_single_host_defaults: false
 cifmw_cephadm_extra_args: ""
 cifmw_cephadm_pacific_filter: "16.*"
 cifmw_cephadm_dashboard_enabled: false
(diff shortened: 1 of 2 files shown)


fmount:
I see, I think it worked for me because I was using a standalone node.

fmount (approved):

fmount:
recheck

cjeanner:
currently CI is still broken, need to double-check why a job failed on #528 - we need it with the new CRD things.

fmount:
> currently CI is still broken, need to double-check why a job failed on #528 - we need it with the new CRD things.

technically the ceph playbook is still not executed in CI: not saying we should bypass the CI job, but except molecule we don't have a Ceph deployment atm.

cjeanner:
/approve

Thank you @fultonj for the detailed explanations [here](https://github.com/openstack-k8s-operators/ci-framework/issues/531#issuecomment-1710786382) - I was wondering how it ended using the "root" user.

PR #537, opened 2023-09-11
Revert "Add Dev-preview workaround to finish EDPM deployment"

raukadah:


gibizer:
If this PR is targeting the current main branch of nova- and dataplane-operator then I agree that the extra wait is not needed any more as that was needed only for the dev-preview branch version

rlandy:
We are officially done with dev-preview 1 so ack we can remove it if the wait is not required in master versions

rlandy:
recheck

rabi (approved):
/lgtm

PR #539, opened 2023-09-12
Add flake8 linting and black formatting to CI

lewisdenny:
Currently we have flake8 configured as pre-push hook but not configured in
tox which gets ran in Prow. [1]

This change adds flake8 linting to the tox config, updates the tox Makefile
target to run the flake8 job and fixes errors reported by flake8.

This change also enables the pep8 `max-line-length` check and bumps it
to 88 as recommended by black[2]

This change also enables the black formatter in CI and pre-commit.

[1] https://github.com/openshift/release/blob/f0d768950a3510ca1a9ee7d79b476bca1c9d8e8e/ci-operator/config/openstack-k8s-operators/ci-framework/openstack-k8s-operators-ci-framework-main.yaml#L38-L41
[2] https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length


pablintino:
Many thanks for submitting this, I already mentioned something related not so far ago. I'd like to introduce python style guides in all of our python code as well as pylint, but I had a different, maybe complementary idea.
I'm one those developers who usually considers manual formatting code a waste of time, that's why I don't care about style guides on my own and I usually pick an autoformater, usually the most used and widely available for the language I'm working on, but, unfortunatly, for python is a mess.
I'd like to see this in, at the same time we provide an auto format tool, like autopep8 or black (I usually pick that one, it's defaults are sane) with a rule set that allows the code formatted by that tool to pass the set of rules we enable in flake8. That way no one is gonna loose time fighting broken flak8 rules.

lewisdenny:
> Many thanks for submitting this, I already mentioned something related not so far ago. I'd like to introduce python style guides in all of our python code as well as pylint, but I had a different, maybe complementary idea. I'm one those developers who usually considers manual formatting code a waste of time, that's why I don't care about style guides on my own and I usually pick an autoformater, usually the most used and widely available for the language I'm working on, but, unfortunatly, for python is a mess. I'd like to see this in, at the same time we provide an auto format tool, like autopep8 or black (I usually pick that one, it's defaults are sane) with a rule set that allows the code formatted by that tool to pass the set of rules we enable in flake8. That way no one is gonna loose time fighting broken flak8 rules.

Hi @pablintino, I completely agree and was planning to purpose black linting in an immediate follow up PR to reduce the review complexity. 

Please let me know if that suits otherwise I'm happy to add it to this PR as well.

raukadah:
@lewisdenny @pablintino Instead of adding multiple linters for Python code, can we switch to [ruff](https://beta.ruff.rs/docs/) ? It provides replacements for multiple linters. https://github.com/astral-sh/ruff-pre-commit provides the pre-commit hook also,

pablintino:
> @lewisdenny @pablintino Instead of adding multiple linters for Python code, can we switch to [ruff](https://beta.ruff.rs/docs/) ? It provides replacements for multiple linters. https://github.com/astral-sh/ruff-pre-commit provides the pre-commit hook also,

Uhmm the idea was to use only flake8 I think. I've never seen that one, but seems popular and may be a good replacement for flake, feel free to test if it works better for us.
@lewisdenny is to add the linter properly I think and my proposal is just, OK, ensure the linter runs, but I would like to have a formater that automatically formats my code into something that is valid for our linter.

pablintino:
@lewisdenny I've discussed this with the team and we all agree that ruff sounds nice and promising but we feel that flake is for now enough and it's well known, so, let's proceed with this PR. Next step will be adding black.

lewisdenny:
@raukadah Thanks for the suggestion, I will certainly be using Ruff for any of my personal projects, it looks really sweet. 

@pablintino Thank you for following this up with the team, I decided to go with flake because it's well used within the OpenStack projects already.

I'll add black support to this PR ASAP, cheers!

pablintino:
recheck

PR #542, opened 2023-09-12
Add more project template and improve it

raukadah:
This patch renames the existing project-template podified-multinode-edpm-pipeline to podified-multinode-edpm-baremetal-pipeline (It will be used for run both edpm and baremetal jobs). It updates the baremetal job name to use cifmw-crc-podified-edpm-baremetal.

It also adds a seperate template podified-multinode-edpm-pipeline to run only EDPM jobs.

Both will be useful where both jobs or only EDPM jobs are runned.

Note: It also updates the missing irrelevant-files list from dataplane-operator and ansibleee-operator.

### DIFF
--- a/zuul.d/content_provider.yaml
+++ b/zuul.d/content_provider.yaml
@@ -21,6 +21,18 @@
       - contribute/.*
       - tests
       - roles/.*/molecule/.*
+      # ci-framework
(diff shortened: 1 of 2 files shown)


pablintino:
recheck

PR #543, opened 2023-09-12
Generate gating.repo to install built packages

raukadah:
build_openstack_packages is building the rpm package and creating the repo metadata in a directory. In order to install the package from metadata, we need to add .repo file so that dnf install the package.

This patch adds the gating.repo to install the built packages.

### DIFF
--- a/ci_framework/roles/build_openstack_packages/tasks/create_repo.yml
+++ b/ci_framework/roles/build_openstack_packages/tasks/create_repo.yml
@@ -30,8 +30,19 @@
 
 - name: Run createrepo on generated rpms
   ansible.builtin.command:
-    cmd: createrepo gating_repo
-    chdir: '{{ cifmw_bop_gating_repo | ansible.builtin.dirname }}'
+    cmd: createrepo .
+    chdir: '{{ cifmw_bop_gating_repo }}'
+
+- name: Add gating.repo file to install the required built packages
+  ansible.builtin.copy:
+    content: |
+      [gating-repo]
+      # All gating repo contents must be present in distro.repos.d directory
+      baseurl=file:///etc/distro.repos.d
+      enabled=1
+      gpgcheck=0
+      priority=1
+    dest: "{{ cifmw_bop_gating_repo }}/gating.repo"
 
 - name: Compress the repo
   ansible.builtin.command:  # noqa: command-instead-of-module


ci_framework/roles/build_openstack_packages/tasks/create_repo.yml
```diff
@@ -33,6 +33,17 @@
     cmd: createrepo gating_repo
```
viroel, line 33: I was looking at those lines, doesn't look like we do in quickstack.
if rpms are copied to `{{ cifmw_bop_gating_repo }}`, the createrepo cmd should be:
`createrepo {{ cifmw_bop_gating_repo }}` or `createrepo .` since you already do a `chdir` here.

Did I get it wrong?

ci_framework/roles/build_openstack_packages/tasks/create_repo.yml
```diff
@@ -33,6 +33,17 @@
...
     chdir: '{{ cifmw_bop_gating_repo | ansible.builtin.dirname }}'
 
+- name: Add gating.repo file to install the required built packages
+  ansible.builtin.copy:
+    content: |
+      [gating-repo]
+      # All gating repo contents must be present in distro.repos.d directory
+      baseurl=file:///etc/distro.repos.d
```
viroel, line 41: ok, so you are going to extract the repo always in `/etc/distro.repos.d` ?
maybe `/etc/distro.repos.d/gating_repo` ?

viroel (commented):
Hi @raukadah - few questions inline.

ci_framework/roles/build_openstack_packages/tasks/create_repo.yml
raukadah: In quickstart we run `createrepo` here https://opendev.org/openstack/tripleo-quickstart-extras/src/branch/master/roles/build-test-packages/tasks/main.yml#L296 and we generate the repo files here: https://opendev.org/openstack/tripleo-quickstart-extras/src/branch/master/roles/install-built-repo/templates/install-built-repo.sh.j2#L8

ci_framework/roles/build_openstack_packages/tasks/create_repo.yml
raukadah: Currently we are copying the gating repo in [ci framework repos ](https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci/playbooks/tcib/tcib.yml#L30) directory and in tcib the same repo gets mounted to  `/etc/distro.repos.d` that's why I have done like this.

ci_framework/roles/build_openstack_packages/tasks/create_repo.yml
viroel: Yes, and as you can see, rpms are copied to gating_repo dir[1]. That is the point that I was trying to bring here,

[1] https://opendev.org/openstack/tripleo-quickstart-extras/src/commit/34cb52dfd23cc3314cd121aef8e8c56a99b314bb/roles/build-test-packages/tasks/main.yml#L292

ci_framework/roles/build_openstack_packages/tasks/create_repo.yml
viroel: ok, thanks

ci_framework/roles/build_openstack_packages/tasks/create_repo.yml
viroel: ok, just a heads-up if something is not working as expected, we can revisit these lines. But they are actually merged here, so nothing to do in this PR.

viroel: cmd `createrepo .` should resolve that for us I think. Thanks

PR #544, opened 2023-09-12
Pass extra install_yamls variables to the set_openstack_containers role

pablintino:
Since the introduction of the multinode layout the defaults set by install_yamls refering to networking are no longer valid. In the edpm_prepare role we are since then passing custom variables for the MTU and the interface name using my NMstate, but till this PR set_openstack_containers is not passing them.


pablintino:
recheck

raukadah (commented):
/approve

pablintino:
recheck

PR #547, opened 2023-09-12
Rely on zuul var to set cifmw_operator_build_operators var

raukadah:
Currently `Set var for cifmw_operator_build_operators var` is getting skipped as project var is not defined. project var is a key of zuul var.

Zuul inventory provides these zuul vars. So the correct var is zuul.project. otherwise the task will be skipped.

This patch fixes the same.

Note: It also set image_base for both operators ending with 2 and 3 words,


viroel:
/approve

testing here: https://github.com/openstack-k8s-operators/dataplane-operator/pull/403

PR #548, opened 2023-09-12
Pin molecule to version 5

pablintino:
Child molecule jobs of the ci-framework are failing because molecule version has jumped to 6 but they are not prepare to use that version. For that reason, just pin the version for now.


raukadah (commented):
/approve

PR #549, opened 2023-09-12
Fix Ceph Spec play to correctly map hostname to storage IP

fultonj:
This patch updates the "Build a dict mapping hostname to its IP which is in storage network range" task to correctly retrieve the storage IP from the hostvars.

Fixes bug introduced by 2de93b92fc26f844b75681ddc7ec71f8279fb786

In breaking commit, `hostvars[item][all_addresses]` was changed to `hostvars['ansible_hostname']` which does not contain the same value. Runs of the play after then failed with "'bool' object is not iterable".

Also, add `when` to prevent an undefined key in the dictionary.

  - [x] README in the role

### DIFF
--- a/ci_framework/playbooks/ceph.yml
+++ b/ci_framework/playbooks/ceph.yml
@@ -98,9 +98,18 @@
   pre_tasks:
     - name: Build a dict mapping hostname to its IP which is in storage network range
       ansible.builtin.set_fact:
-        host_to_ip: >-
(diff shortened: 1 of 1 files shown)


pablintino:
recheck

raukadah (approved):
/approve

fmount:
Late here but +2, thanks John!

PR #561, opened 2023-09-15
Add management networks for Ceph SSH and OSD replication

fultonj:
Introduce ssh_network_range (default 192.168.122.0/24) to ceph.yaml playbook for building the Ceph spec. The Ceph orchestrator will use this network to SSH into hosts as the ceph-admin user. This network should be used in the Ceph spec for adding hosts.

Introduce storage_mgmt_network_range to ceph.yaml playbook. If storage_network_range and storage_mgmt_network_range are undefined, then their respective cifmw_ceph_spec_public_network and cifmw_ceph_spec_private_network, in the cifmw_ceph_spec role, will be set to the ssh_network_range for the public network and omitted for the private network respectively in the initial ceph.conf. This ensures that a public network can always be looked up by other plays from the initial ceph.conf while keeping the ceph playbook working without network isolation by default. However, values for these networks can be easily uncommented in the ceph.yaml playbook (or passed as overrides) for when network isolation is used. Commented out values for these networks are based on the project defaults [1].

Molecule tests have been added to cifmw_ceph_spec role to ensure this desired behavior is enforced.

[1] https://github.com/openstack-k8s-operators/docs/blob/main/networking.md

  - [x] README in the role

### DIFF
--- a/ci_framework/playbooks/ceph.yml
+++ b/ci_framework/playbooks/ceph.yml
@@ -88,23 +88,24 @@
   roles:
     - role: cifmw_block_device
 
-- name: Build Ceph spec from gathered IP addresses of EDPM inventory group
+- name: Build Ceph spec and conf from gathered IPs of EDPM inventory group
   tags: spec
   hosts: localhost
   gather_facts: true
   vars:
+    ssh_network_range: 192.168.122.0/24
     # storage_network_range: 172.18.0.0/24
-    storage_network_range: 192.168.122.0/24
+    # storage_mgmt_network_range: 172.20.0.0/24
     all_addresses: ansible_all_ipv4_addresses # change if you need IPv6
   pre_tasks:
-    - name: Build a dict mapping hostname to its IP which is in storage network range
+    - name: Build a dict mapping hostname to its IP which is in management network range
       ansible.builtin.set_fact:
         host_to_ip:
           "{{ host_to_ip | default({}) |
             combine(
               {
                 hostvars[item]['ansible_hostname'] :
-                hostvars[item][all_addresses] | ansible.utils.ipaddr(storage_network_range) | first
+                hostvars[item][all_addresses] | ansible.utils.ipaddr(ssh_network_range) | first
               }
             )
            }}"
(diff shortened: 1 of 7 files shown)


fultonj:
[podified-multinode-edpm-e2e-nobuild-tagged-crc](https://review.rdoproject.org/zuul/build/2cb6650127e74916870eeb3f98254304) failed but it does not use this code.

fultonj:
recheck

ci_framework/playbooks/ceph.yml
```diff
@@ -88,23 +88,24 @@
...
-    - name: Build a dict mapping hostname to its IP which is in storage network range
+    - name: Build a dict mapping hostname to its IP which is in management network range
       ansible.builtin.set_fact:
         host_to_ip:
           "{{ host_to_ip | default({}) |
             combine(
               {
                 hostvars[item]['ansible_hostname'] :
-                hostvars[item][all_addresses] | ansible.utils.ipaddr(storage_network_range) | first
+                hostvars[item][all_addresses] | ansible.utils.ipaddr(ssh_network_range) | first
```
fmount, line 108: +1 on this: by having ssh_network_range we can make sure it matches with the ip address range used to enroll hosts within the Ceph cluster.

ci_framework/roles/cifmw_ceph_spec/templates/initial_ceph.conf.j2
```diff
@@ -1,5 +1,10 @@
 [global]
 osd pool default size = 1
+{% if cifmw_ceph_spec_public_network %}
```
fmount, line 3: +1, this matches what I had in mind, we can enroll nodes by using the `ssh_network` and both `public` and `cluster` network are passed through the `initial_ceph.conf` assimilated at bootstrap time.

fmount (approved):
+2,  no concerns on my side as it makes sense using the ssh_network to enroll nodes and passing both `public` and `cluster` network via `initial_ceph`.

fmount (approved):
Thanks John, /lgtm

katarimanojk:
Thanks John for the PR.

PR #571, opened 2023-09-19
Allow to gather passed hooks in the custom params

cjeanner:
This would make the whole reproducer easier: we'll only need to inject
the custom-params.yml file in the environment instead of having to
gather the scenario contents.

### DIFF
--- a/ci_framework/playbooks/01-bootstrap.yml
+++ b/ci_framework/playbooks/01-bootstrap.yml
@@ -19,7 +19,7 @@
           {{
             hostvars[inventory_hostname] |
             dict2items |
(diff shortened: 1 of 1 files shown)


cjeanner:
Patch is working on its own:
https://logserver.rdoproject.org/71/571/ab28e269c5502980fef3fd02993c14b966520c0b/github-check/podified-multinode-edpm-deployment-crc/f6c6a61/controller/ci-framework-data/artifacts/parameters/custom-params.yml

we can see the hooks being integrated now.

cjeanner:
recheck

PR #582, opened 2023-09-22
Automatically skip bootstrapped network cleanup if autohold

pablintino:
Allows automatically skipping the cleanup of the bootstrapped network if there is an autohold for the current job.


raukadah:
recheck

raukadah (approved):
will be useful to avoid post cleanup step while holding the node. Already tested in another patch.

PR #584, opened 2023-09-22
Bump flavor for edpm baremetal job

Sandeepyadav93:
centos-9-stream-crc-4xlarge have 16vcpu_48GB.

### DIFF
--- a/zuul.d/edpm.yaml
+++ b/zuul.d/edpm.yaml
@@ -11,8 +11,11 @@
 # Virtual Baremetal job with CRC and single compute node.
 - job:
     name: cifmw-crc-podified-edpm-baremetal
+    nodeset: centos-9-crc-4xlarge
     parent: cifmw-base-crc-openstack
     run: ci/playbooks/edpm_baremetal_deployment/run.yml
+    vars:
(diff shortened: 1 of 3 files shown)


zuul.d/edpm.yaml
```diff
@@ -11,6 +11,7 @@
 # Virtual Baremetal job with CRC and single compute node.
 - job:
     name: cifmw-crc-podified-edpm-baremetal
+    nodeset: centos-9-stream-crc-4xlarge
```
cjeanner, line 14: what about downstream?

Sandeepyadav93: Sry, didn't get what about downstream?

You mean need to create same nodeset in downstream? In downstream job is stable with smaller flavor.

cjeanner: ~> DS has an override already, so it won't crash with this change

rabi:
recheck

Sandeepyadav93:
/hold

need some changes for crc spec

rlandy:
recheck

rlandy:
recheck

PR #585, opened 2023-09-22
Fix the value for cifmw_dlrn_promote_hash_in_promote_target

raukadah:
Since we use slurp so there is no stdout key in the above var. It skips the resulting tasks where it is used as a conditional.

This patch fixes the same for the conditionals.

### DIFF
--- a/ci_framework/roles/dlrn_promote/tasks/main.yml
+++ b/ci_framework/roles/dlrn_promote/tasks/main.yml
@@ -28,8 +28,7 @@
 - name: Run DLRN promote hash
   when:
     - cifmw_dlrn_promote_hash_in_promote_target is defined
-    - "'stdout' in cifmw_dlrn_promote_hash_in_promote_target"
-    - cifmw_dlrn_promote_hash_in_promote_target.stdout == '[]'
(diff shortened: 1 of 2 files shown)


ci_framework/roles/dlrn_promote/tasks/main.yml
```diff
@@ -28,8 +28,7 @@
 - name: Run DLRN promote hash
   when:
     - cifmw_dlrn_promote_hash_in_promote_target is defined
-    - "'stdout' in cifmw_dlrn_promote_hash_in_promote_target"
-    - cifmw_dlrn_promote_hash_in_promote_target.stdout == '[]'
+    - cifmw_dlrn_promote_hash_in_promote_target |length == 0 or cifmw_dlrn_promote_hash_in_promote_target == '[]'
```
pablintino, line 31: What's the point of the `or` condition if you already compared the length?
PS: the second one will be always false as `cifmw_dlrn_promote_hash_in_promote_target` is a list, not a string like the right side of the operand (but maybe it even fails cause you are comparing different types).

raukadah: Good point, let me remove list string comparison.

PR #586, opened 2023-09-22
New role: ci_network

cjeanner:
This role will apply network configuration on the hosts.

While it has a default network layout, it is expected to
get data from a previous step, such as the Zuul job bootstrap,
in a specific format matching the default layout.

In the case of the job reproducer, the data will be generated by mixing
data from the actual CI job, and the ones matching the reproducer
infra.

  - [X] README in the role

### DIFF
--- a/ci_framework/roles/ci_network/defaults/main.yml
+++ b/ci_framework/roles/ci_network/defaults/main.yml
@@ -1,0 +1,106 @@
+---
+# Copyright Red Hat, Inc.
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+
+cifmw_network_pre_cleanup: true
+cifmw_network_nm_config: []
+cifmw_network_nm_config_file: "/etc/NetworkManager/NetworkManager.conf"
+cifmw_network_generated_layout: "/etc/ci/env/network-layout.yml"
+# This parameter is provided as an exemple of the expected content.
+# Usually it will be generated in Zuul job, and the reproducer role
+# will generate it accordingly and inject it in the environment.
+cifmw_network_layout:
+  controller:
+    default:
+      iface: "eth1"
+      mac: omit
+      mtu: 1500
+      connection: "ci-private-network"
+      ip4: "192.168.122.10/24"
+      gw4: "192.168.122.1"
+      dns: omit
+  crc:
+    default:
+      iface: "enp2s0"
+      mac: omit
+      mtu: 1500
+      connection: "ci-private-network"
+      ip4: "192.168.122.11/24"
+      gw4: "192.168.122.1"
+      dns: omit
+    vlan19:
+      iface: "eth1.vlan20"
+      vlan: "19"
+      parent_iface: "eth1"
+      connection: "tenant"
+      mac: omit
+      mtu: 1500
+      ip4: "172.19.0.11/24"
+      dns: omit
+    vlan20:
+      iface: "eth1.vlan20"
+      vlan: "20"
+      parent_iface: "eth1"
+      connection: "internal-api"
+      mac: omit
+      mtu: 1500
+      ip4: "172.20.0.11/24"
+      dns: omit
+    vlan21:
+      iface: "eth1.vlan21"
+      vlan: "21"
+      parent_iface: "eth1"
+      connection: "storage"
+      mac: omit
+      mtu: 1500
+      ip4: "172.21.0.11/24"
+      dns: omit
+  compute-0:
+    default:
+      iface: "eth1"
+      mac: omit
+      mtu: 1500
+      connection: "ci-private-network"
+      ip4: "192.168.122.100/24"
+      gw4: "192.168.122.1"
+      dns: omit
+    vlan19:
+      iface: "eth1.vlan20"
+      vlan: "19"
+      parent_iface: "eth1"
+      connection: "tenant"
+      mac: omit
+      mtu: 1500
+      ip4: "172.19.0.100/24"
+      dns: omit
+    vlan20:
+      iface: "eth1.vlan20"
+      vlan: "20"
+      parent_iface: "eth1"
+      connection: "internal-api"
+      mac: omit
+      mtu: 1500
+      ip4: "172.20.0.100/24"
+      dns: omit
+    vlan21:
+      iface: "eth1.vlan21"
+      vlan: "21"
+      parent_iface: "eth1"
+      connection: "storage"
+      mac: omit
+      mtu: 1500
+      ip4: "172.21.0.100/24"
+      dns: omit
--- a/ci_framework/roles/ci_network/molecule/default/converge.yml
+++ b/ci_framework/roles/ci_network/molecule/default/converge.yml
@@ -1,0 +1,51 @@
+---
+# Copyright Red Hat, Inc.
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+
+- name: Converge
+  hosts: all
+  roles:
+    - role: "ci_network"
+  tasks:
+    - name: Get nmcli connections
+      register: nmcli_existing_connections
+      ansible.builtin.command:
+        cmd: nmcli -t -f NAME,DEVICE,ACTIVE con show
+
+    - name: Debug existing connections
+      ansible.builtin.debug:
+        var: nmcli_existing_connections.stdout_lines
+
+    - name: Ensure we have the expected vlan20
+      ansible.builtin.assert:
+        that:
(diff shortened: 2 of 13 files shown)


ci_framework/roles/ci_network/tasks/apply.yml
```diff
@@ -0,0 +1,24 @@
+---
+- name: "Apply network {{ network.key }}"
+  become: true
+  when:
+    - network.value.config_nm | default(true) | bool
+  community.general.nmcli:
```
pablintino, line 6: Uhmmm, I'd say that `autoconnect` is maybe needed, not mandatory, but a nice to have.

ci_framework/roles/ci_network/tasks/apply.yml
```diff
@@ -0,0 +1,24 @@
...
+    # Needed for some specific conditions
+    gw4: "{{ network.value.gw4 | default(omit) }}"
+    # Needed only for IPv6 support
+    ip6: "{{ network.value.ip6 | default(omit) }}"
+    # Needed only for IPv6 support
+    gw6: "{{ network.value.gw6 | default(omit) }}"
+    # Needed for VLAN support
+    parent_iface: "{{ network.value.parent_iface | default(omit)}}"
+    # Needed for some specific infra
+    mtu: "{{ network.value.mtu | default(1500) }}"
```
pablintino, line 22: Are we assuming that the MTU is not always passed? I coded the current openstack logic to pass the MTU always, I don't expect others to not continuing doing so.

ci_framework/roles/ci_network/tasks/nm-cleanup.yml
```diff
@@ -0,0 +1,30 @@
...
+  ansible.builtin.command:
+    cmd: ip -j ro ls
+
+- name: Get nm existing ethernet connections
+  register: nm_ethernet
+  ansible.builtin.shell:  # noqa risky-shell-pipe
+    cmd: nmcli -t -f TYPE,DEVICE,NAME con | grep ethernet
+
+- name: Prepare network data for proper consumption
+  ansible.builtin.set_fact:
```
pablintino, line 13: This task can be removed by adding the declared facts as variables of this one: https://github.com/openstack-k8s-operators/ci-framework/pull/586/files#diff-7bd038433c1242ed1f5b21ab2af6c230f41e76370a5ccde53ac3ef7435e9522cR23

ci_framework/roles/ci_network/tasks/main.yml
```diff
@@ -0,0 +1,68 @@
...
+  ansible.builtin.stat:
+    path: "{{ cifmw_network_generated_layout }}"
+
+- name: Load generated network if exists
+  when:
+    - env_file.stat.exists
+  ansible.builtin.include_vars:
+    file: "{{  cifmw_network_generated_layout }}"
+
+- name: Chose appropriate data source
```
pablintino, line 28: Not really needed is this facts moves to https://github.com/openstack-k8s-operators/ci-framework/pull/586/files#diff-0f14100a05f379e84695ebb0a68b8a4b4b16bc145b0127b0450206dc9d941fbcR63 as a variables or maybe goes directory into the loop condition.

ci_framework/roles/ci_network/tasks/apply.yml
cjeanner: might be OK. Trying to find a reason not to have it, I don't see any, since we're able to NOT configure a connection with nmcli, like the current way.

ci_framework/roles/ci_network/tasks/apply.yml
cjeanner: imho we shouldn't enforce MTU argument. Any reason to enforce it?

ci_framework/roles/ci_network/tasks/main.yml
cjeanner: right. I used it in the early iteration for debugging

ci_framework/roles/ci_network/tasks/apply.yml
pablintino: maybe in that role is not need, but with the current implementation is always filled with the real MTU from the link

cjeanner: We may reconsider if needed, but keeping it optional seems legit for now

ci_framework/roles/ci_network/tasks/apply.yml
```diff
@@ -0,0 +1,43 @@
...
+- name: "Apply network {{ network.key }}"
+  become: true
+  when:
+    - network.value.config_nm | default(true) | bool
+  community.general.nmcli:
+    autoconnect: true
+    state: "{{ network.value.state | default('present') }}"
+    conn_name: "{{ network.value.connection | default(network.key) }}"
+    ifname: "{{ network.value.iface }}"
+    type: "{{ network.key is match('^vlan.*') | ternary('vlan', 'ethernet') }}"
```
pablintino, line 27: Why should VLAN networks start with the vlan prefix? Sounds like a limitation. I think that a better criteria is to just check if `network.value.vlan` is defined, and if it is, assume it's a vlan interface.

ci_framework/roles/ci_network/tasks/apply.yml
```diff
@@ -0,0 +1,43 @@
...
+
+- name: "Apply network {{ network.key }}"
+  become: true
+  when:
+    - network.value.config_nm | default(true) | bool
+  community.general.nmcli:
+    autoconnect: true
+    state: "{{ network.value.state | default('present') }}"
+    conn_name: "{{ network.value.connection | default(network.key) }}"
+    ifname: "{{ network.value.iface }}"
```
pablintino, line 26: Not needed for VLANs, not sure how this module behaves if passed when type is vlan.

ci_framework/roles/ci_network/molecule/default/converge.yml
```diff
@@ -0,0 +1,35 @@
...
+        cmd: nmcli -t -f NAME,DEVICE con show
+
+    - name: Debug existing connections
+      ansible.builtin.debug:
+        var: nmcli_existing_connections.stdout_lines
+
+    - name: Ensure we have the expected vlan20
+      ansible.builtin.assert:
+        that:
+          - "'molecule-test:eth0.20' in nmcli_existing_connections.stdout_lines"
```
pablintino, line 35: Should we ensure is UP? Should we check if the VLAN id is the expected one and that the IP too?

ci_framework/roles/ci_network/tasks/main.yml
```diff
@@ -0,0 +1,70 @@
...
+
+- name: Load generated network if exists
+  when:
+    - env_file.stat.exists
+  ansible.builtin.include_vars:
+    file: "{{  cifmw_network_generated_layout }}"
+
+- name: Configure NetworkManager service
+  vars:
+    _net_layout: >-
```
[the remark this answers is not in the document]
cjeanner: ah, right... I was a bit too enthusiastic when adding the missing var to some of the tasks :]

ci_framework/roles/ci_network/tasks/apply.yml
cjeanner: hmm apparently it's working fine, since molecule is green

ci_framework/roles/ci_network/tasks/nm-cleanup.yml
```diff
@@ -0,0 +1,30 @@
+---
+- name: Fetch routes as a json
+  register: ip_routes
+  ansible.builtin.command:
+    cmd: ip -j ro ls
+
+- name: Get nm existing ethernet connections
+  register: nm_ethernet
+  ansible.builtin.shell:  # noqa risky-shell-pipe
```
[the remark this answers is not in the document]
cjeanner: :+1:

ci_framework/roles/ci_network/tasks/apply.yml
cjeanner:
```
The ifname argument is mandatory for all connection types except bond, team, bridge, vlan and vpn.

This parameter defaults to conn_name when left unset for all connection types except vpn that removes it.
```
Guess we can therefore safely keep it as-is.

ci_framework/roles/ci_network/tasks/apply.yml
cjeanner: let's assume ppl know what they do and let's check if "vlan" is set.

ci_framework/roles/ci_network/molecule/default/converge.yml
cjeanner: added a second assertion block based on the interface config output from nmcli (as a second call)

cjeanner:
recheck

cjeanner:
recheck

ci_framework/roles/ci_network/defaults/main.yml
```diff
@@ -0,0 +1,106 @@
...
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+
+cifmw_network_pre_cleanup: true
+cifmw_network_nm_config: []
+cifmw_network_nm_config_file: "/etc/NetworkManager/NetworkManager.conf"
+cifmw_network_generated_layout: "/etc/ci/env/network-layout.yml"
+# This parameter is provided as an exemple of the expected content.
```
lewisdenny, line 22: exemple -> example

ci_framework/roles/ci_network/tasks/main.yml
```diff
@@ -0,0 +1,65 @@
...
+  register: env_file
+  ansible.builtin.stat:
+    path: "{{ cifmw_network_generated_layout }}"
+
+- name: Load generated network if exists
+  when:
+    - env_file.stat.exists
+  ansible.builtin.include_vars:
+    file: "{{  cifmw_network_generated_layout }}"
+
```
lewisdenny, line 27: One suggestion I have is to split out the valitation task from apply.yml into a new task file called something like validate.yml and call it after `Load generated network if exists`

lewisdenny (commented):
Hi Cedric, I left some comments. 

Thank you, this looks great. Would be nice to take this logic out of the config repo.

rlandy:
recheck

ci_framework/roles/ci_network/defaults/main.yml
cjeanner: I'll make a follow-up, CI is at last green.

ci_framework/roles/ci_network/tasks/main.yml
cjeanner: Maybe something to consider - as a follow-up though.

PR #590, opened 2023-09-25
Fix CoreDNS loading 127.0.0.1 as nameserver

pablintino:

### DIFF
--- a/ci/playbooks/multinode-customizations.yml
+++ b/ci/playbooks/multinode-customizations.yml
@@ -3,8 +3,21 @@
   name: "Tweak CRC node"
   gather_facts: false
   tasks:
+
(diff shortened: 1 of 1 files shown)


ci_framework/hooks/playbooks/fetch_compute_facts.yml
```diff
@@ -30,6 +30,8 @@
   connection: local
   tasks:
     - name: Load parameters
+      delegate_to: crc
```
cjeanner, line 33: sooo we'd then take only those from the CRC host? Expecting CRC to get all the data for all the existing nodes in the deploy?

cjeanner (commented):
/approve

PR #591, opened 2023-09-25
Correct parameter value

cjeanner:
This is the "quick-fix", the "right" way would be to modify the hook in
order to `include_vars` on the generated environment file, allowing to
access all of the data such as the path, namespace and so on.


arxcruz (approved):

raukadah (commented):
/approve

PR #593, opened 2023-09-25
Fix RGW default value

fmount:
When Ceph is deployed using the HCI model in the EDPM nodes, we need to point by default to the keystone-internal service instead of the public Route.
This value can be overriden in all the other use cases and a doc patch aligned with this change has already been proposed [1].

[1] https://github.com/openstack-k8s-operators/docs/pull/63

- [x] README in the role


fultonj (approved):
Yes, we need this change. Thank you Francesco.

lewisdenny (approved):

PR #597, opened 2023-09-26
Add test dir under irrelevant_files list

raukadah:
test dir in operator codebase contains tests. We donot want to run content provider on it. Adding it under irrelevant_files skips the content provider job.

### DIFF
--- a/zuul.d/base.yaml
+++ b/zuul.d/base.yaml
@@ -17,14 +17,12 @@
       - ^PROJECT$
       - ^README.md$
       - ^renovate.json$
-      - ^tests/.*$
       - ^kuttl-test.yaml$
       - molecule/.*
       - molecule-requirements.txt
       - .github/workflows
       - docs/.*
       - contribute/.*
-      - tests
       - roles/.*/molecule/.*
       # ci-framework
       - .readthedocs.yaml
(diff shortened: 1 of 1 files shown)


zuul.d/base.yaml
```diff
@@ -35,6 +35,7 @@
       - .gitignore
       - .golangci.yaml
       - .pre-commit-config.yaml
+      - test
```
cjeanner, line 38: either `tests?/` as a proper regexp, or at least group "tests" and "test"

Note that we have `- tests` on Line 27, but also `^tests/.*$` on line 20 - sounds like duplications.

cjeanner (commented):
/approve

thanks :)

zuul.d/base.yaml
```diff
@@ -17,14 +17,13 @@
...
       - ^renovate.json$
-      - ^tests/.*$
       - ^kuttl-test.yaml$
       - molecule/.*
       - molecule-requirements.txt
       - .github/workflows
       - docs/.*
       - contribute/.*
-      - tests
+      - tests?
```
gibizer, line 26: We run kuttl test in zuul, and the kuttl test cases are under test(s)/kuttl in most cases so we should not ignore that directory.

raukadah: Done

PR #610, opened 2023-09-28
Fixing get_makefiles_env when special chars are in place

pablintino:
Some values may contain special chars like % that may affect the underneath printf command. By replacing printf with echo that is way more simple and doesn't format the input, something that is not needed here as it was only used to add line breaks, special chars should pass correctly without causing issues.

### DIFF
--- a/ci_framework/plugins/modules/get_makefiles_env.py
+++ b/ci_framework/plugins/modules/get_makefiles_env.py
@@ -55,11 +55,13 @@ import subprocess
 import tempfile
 
 
+__FILES_ENCODING = "utf-8"
+__JOIN_CHARS = "|#_:_#|"
 __TMP_TARGET_NAME = "cifmw_dump_vars"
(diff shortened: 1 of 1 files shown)


cjeanner:
/approve

Makes sense. Also good to enable the `check` in the subprocess, in order to raise errors on due time. It should help debugging in the future.

rabi:
This does not run the baremetal job, so should be tested with a depends-on patch in dataplane-operator or other repos.

raukadah (commented):
+2 looks good.

cjeanner:
/test pre-commit
Doesn't seem to be caused by the PR

cjeanner:
/lgtm

Let's wait for the last pre-commit one, and manually merge. Zuul issue is unrelated.

cjeanner:
Locally ran the pre-commit job, it passed. Prow seems to have an issue. Merging.

PR #611, opened 2023-09-28
Ensure we fetch generated parameters

cjeanner:
Until now, we were missing an important piece of the CI job bootstrap,
generated by the bootstrap playbooks. Thost information describe the
network layout and are later consumed by a hook in the case of multinode
+ compute jobs.

Ensuring we get a hold on those data for later reference is important,
since it may help debugging networking related issues.


cjeanner:
recheck

cjeanner:
recheck

PR #613, opened 2023-09-28
Make sure parameters directory exists on install_yamls role

raukadah:
When we call install_yamls role seperately, It fails with parameters directory does not exist. This pr makes sure the directory exists before copying files in parameters directory.

### DIFF
--- a/ci_framework/roles/install_yamls/tasks/main.yml
+++ b/ci_framework/roles/install_yamls/tasks/main.yml
@@ -23,6 +23,7 @@
   loop:
     - "{{ cifmw_install_yamls_out_dir }}"
     - "{{ cifmw_install_yamls_tasks_out }}"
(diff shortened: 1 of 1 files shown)


ci_framework/roles/install_yamls/tasks/main.yml
```diff
@@ -119,6 +119,13 @@
   ansible.builtin.debug:
     var: cifmw_generate_makes
 
+- name: Ensure parameters directory exists
+  tags:
+    - bootstrap
+  ansible.builtin.file:
```
cjeanner, line 125: care to move that at the top, in the same task creating the rest of the tree on L17?

cjeanner (changes requested):

cjeanner (approved):

cjeanner:
recheck

PR #615, opened 2023-09-28
Remove the 3-replicas Galera configuration

cjeanner:
There's no sense to deploy a 3-replicas galera on a single CRC node.
This is more a layout for the Validated Architecture.


cjeanner:
recheck

viroel:
recheck

PR #624, opened 2023-10-03
Enable new parameter support for compute hook

cjeanner:
Since we intend to support IPv6 sooner or later, it's better to be
consistent with the parameter names. Since the ci_network role
introduces the "ip6" and "ip4" as well as "dns6" and "dns4", let's make
the hook aware of them for the reproducer case.

This will also make the switch to ci_network role easier in the near
future.

Regarding the "!crc,!crc": those exclusions ensure we don't try to
deploy repositories on the CRC. Both names are mandatory:
- "crc" is the name in the CI
- "crcs" is a group name, present in the reproducer.

### DIFF
--- a/ci_framework/hooks/playbooks/fetch_compute_facts.yml
+++ b/ci_framework/hooks/playbooks/fetch_compute_facts.yml
@@ -13,7 +13,7 @@
         label: "{{ item }}"
 
 - name: Push repositories hook
-  hosts: all
+  hosts: all,!crcs,!crc
   gather_facts: true
   tasks:
     - name: Copy repositories from controller to compute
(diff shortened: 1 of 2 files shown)


ci_framework/hooks/playbooks/fetch_compute_facts.yml
```diff
@@ -13,7 +13,7 @@
         label: "{{ item }}"
 
 - name: Push repositories hook
-  hosts: all
+  hosts: all,!crcs
```
raukadah, line 16: I checked the zuul console logs, it is working. I think the excluded host would be `!crc` not `!crcs`. I might be wrong here. Does hostname with plural will work?

cjeanner: it will work with the reproducer - there I have a plural for the auto-generated groups. We may add "!crc" actually. Lemme amend it. It will generate a warning, but won't crash.

cjeanner: added "!crc" to match the CI, and updated commit message to explain that little hack :)

PR #574, opened 2023-09-19
Enable CI layout reproduction

cjeanner:
The new "ci_data.yml" tasks file in the "reproducer" role allows to
fetch data from an existing run output. It will fetch the needed bits,
and get everything ready for the actual deploy, by loading all of the
generated pieces.

It will detect how many compute were engaged in the run, and reflect it
in the generated layout.

The new "ci_job.yml" tasks file will ensure repositories are present,
especially ci-framework and install_yamls, at the expected revision.

It will then generate a playbook, more or less in the same way zuul does
on its own, aggregating the CI playbooks passed via the "run" job
parameter.

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/624
Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/625

As a pull request owner and reviewers, I checked that:
- [x] Appropriate testing is done and actually running
- [X] Appropriate documentation exists and/or is up-to-date:
  - [X] README in the role
  - [X] Content of the docs/source is reflecting the changes

PR #625, opened 2023-10-03
Modify libvirt_manager role for proper reproducer capabilities

cjeanner:
This patch extend the libvirt_manager role in order to create the full
reproducer job layout.
In such a case, there are some extended needs compared to the "deploy a
layout" easy use-case: we need to inject some keys, create network
layout files/configuration and so on.

This patch, in conjuction with some others coming later, will allow
users to reproduce a 1:1 CI job, with the layout and networking.

  - [X] README in the role

### DIFF
--- a/ci_framework/roles/libvirt_manager/tasks/deploy_layout.yml
+++ b/ci_framework/roles/libvirt_manager/tasks/deploy_layout.yml
@@ -1,8 +1,17 @@
 ---
 - name: Create needed worload directory
   ansible.builtin.file:
-    path: "{{ cifmw_reproducer_basedir }}/workload"
+    path: "{{ cifmw_reproducer_basedir }}/{{ item }}"
     state: directory
+  loop:
+    - workload
+    - reproducer-inventory
+    - reproducer-network-env
+
+- name: Chose right parameter for layout definition
+  ansible.builtin.set_fact:
+    cacheable: true
+    _layout: "{{ cifmw_libvirt_manager_configuration_gen | default(cifmw_libvirt_manager_configuration) }}"
 
 - name: Ensure networks are defined
   community.libvirt.virt_net:
@@ -10,7 +19,7 @@
     name: "cifmw-{{ item.key }}"
     xml: "{{ item.value | replace(item.key, 'cifmw-' ~ item.key) }}"
     uri: "qemu:///system"
-  loop: "{{ cifmw_libvirt_manager_configuration.networks | dict2items }}"
+  loop: "{{ _layout.networks | dict2items }}"
   loop_control:
     label: "{{ item.key }}"
 
@@ -19,7 +28,7 @@
     command: create
     name: "cifmw-{{ item.key }}"
     uri: "qemu:///system"
-  loop: "{{ cifmw_libvirt_manager_configuration.networks | dict2items }}"
+  loop: "{{ _layout.networks | dict2items }}"
   loop_control:
     label: "{{ item.key }}"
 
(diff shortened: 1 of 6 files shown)


ci_framework/roles/libvirt_manager/tasks/deploy_layout.yml
```diff
@@ -56,23 +65,51 @@
...
+
+- name: Slurp private key for later use
+  register: priv_ssh_key
+  ansible.builtin.slurp:
+    path: .ssh/cifmw_reproducer_key
+
+- name: Create fact holding network data for VMs
+  ansible.builtin.set_fact:
+    cacheable: true
+    cifmw_reproducer_network_data: {}
```
pablintino, line 91: Where is this one used? In another PR?

ci_framework/roles/libvirt_manager/tasks/deploy_layout.yml
```diff
@@ -56,23 +65,51 @@
       --ssh-inject {{ _admin_user }}:file:{{ ansible_user_dir }}/.ssh/authorized_keys
       | tee -a {{ item.value.image_local_dir }}/{{ item.value.disk_file_name }}.log
     creates: "{{ item.value.image_local_dir }}/{{ item.value.disk_file_name }}.log"
-  loop: "{{ cifmw_libvirt_manager_configuration.vms | dict2items }}"
+  loop: "{{ _layout.vms | dict2items }}"
   loop_control:
     label: "{{ item.key }}"
 
+- name: Create temporary ssh keypair
```
pablintino, line 72: Should we explicetly use `ansible_user_dir` as the base path of the SSH key?

ci_framework/roles/libvirt_manager/templates/inventory.yml.j2
```diff
@@ -0,0 +1,10 @@
+{{ vm_type }}s:
+  hosts:
+{% for host in hosts %}
+    {{ vm_type }}-{{ host.item }}:
+      ansible_host: {{ (host.stdout.split())[3] | replace('/24','') }}
+      ansible_user: {{ admin_user }}
+{% if admin_user == 'core' %}
+      ansible_ssh_private_key_file: ~/.ssh/crc_key
```
raukadah, line 8: There is no mention of `crc_key` in the ci-framework. is this something coming in the new pr?

ci_framework/roles/libvirt_manager/tasks/deploy_layout.yml
cjeanner: in the reproducer, yeah. It's kind of inter-dependent

ci_framework/roles/libvirt_manager/tasks/deploy_layout.yml
cjeanner: I had so many issues with the `ansible_user_dir` I abandoned when it came to that part. With the delegation and all, it was "slightly" messy in the end, so I ended up with relative path to the execution location, which happens to be in the user dir on the target host. Bad, good, no idea, it works... :]

ci_framework/roles/libvirt_manager/templates/inventory.yml.j2
cjeanner: that inventory thing is then consumed in the reproducer role - and the key is also generated/managed in that other role

cjeanner:
@pablintino @raukadah so, things are really, really inter-connected with the reproducer (see #574). It's hard to make that whole thing completely independent - but I didn't want to push everything in one single PR...

raukadah:
> @pablintino @raukadah so, things are really, really inter-connected with the reproducer (see #574). It's hard to make that whole thing completely independent - but I didn't want to push everything in one single PR...

Let's get this in!

PR #628, opened 2023-10-03
[dlrn_promote] Add missing brackets to fix command failure

raukadah:


dasm (approved):
/lgtm
/approve

dasm:
Due to `environment` use, we cannot change that to `ansible.builtin.command`. Added `noqa`

PR #630, opened 2023-10-04
Ensure we can get Tempest image before running tests

cjeanner:
This new task should ensure we're able to fetch the tempest image, since
it may happen image registry has some hiccups.


cjeanner:
/retest

issue doesn't seem related to the patch.

cjeanner:
recheck

raukadah (approved):
/approve

PR #633, opened 2023-10-04
Add notice that framework can't be ran as root user.

lewisdenny:
This is due to [1] and fails in a rather silent way.

[1] https://github.com/openstack-k8s-operators/install_yamls/blob/main/devsetup/scripts/crc-setup.sh#L5

  - [X] README in the role


cjeanner (commented):
/approve

PR #634, opened 2023-10-04
Allow to not manage firewall

cjeanner:
In the reproducer, we don't have any nftables deployed. We therefore
don't need to allow anything.
More over, since the firewall isn't configured at all, the rule
insertion is leading to an error, because the INPUT chain doesn't exist.

  - [X] README in the role


lewisdenny (approved):

raukadah (approved):
/approve

PR #637, opened 2023-10-05
Introduce new tag in libvirt_manager

cjeanner:
This tag is mostly used in the reproducer case, when we want to iterate
on a deploy without having to wait for the libvirt steps to pass again
(when it's already deployed). It's allowing a faster run.

As a pull request owner and reviewers, I checked that:
- [X] Appropriate testing is done and actually running

PR #638, opened 2023-10-05
Reproducer: add content-provider support

cjeanner:
This patch introduce content-provider job-like support.

The goal here is to enable job reproduction in cases such as the
podified-multinode-edpm-deployment-crc.
This specific job requires a content-provider job in order to build and
provide the operator/container images.

With this patch, such a need is detected by the role, and a step to get
the content-provider running on the ansible controller node is then
launched.

This patch also corrects some wrong assumption with the way zuul
clones/manages the dependency repositories, and is closer to its
behavior than before.

Finaly, we also introduce a new tag, "boostrap_layout" in order to
re-run only specific steps.

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/637

- [X] Proper documentation update is in place

### DIFF
--- a/ci_framework/roles/reproducer/templates/play.yml.j2
+++ b/ci_framework/roles/reproducer/templates/play.yml.j2
@@ -1,4 +1,83 @@
 ---
+{% if use_content_provider | default(false) | bool -%}
+- name: Configure CRC for insecure registry
+  hosts: crcs
+  vars:
+    content_provider_registry_ip: {{ cifmw_reproducer_ctl_ip4 }}
+  tasks:
+    - name: Set insecure registry on crc node
+      ansible.builtin.import_tasks: ci/playbooks/tasks/set_crc_insecure_registry.yml
+
+- name: "Content Provider"
+  hosts: localhost
+  gather_facts: true
+  vars:
+    cifmw_rp_registry_ip: {{ cifmw_reproducer_ctl_ip4 }}
+    cifmw_rp_registry_firewall: false
+    job_id: "{{ job_id }}"
+  tasks:
+{% raw %}
+    - name: Load env variables
+      ansible.builtin.include_vars:
+        file: "{{ item }}"
+      loop:
+        - "{{ ansible_user_dir }}/{{ job_id }}-params/zuul-params.yml"
+        - "./scenarios/centos-9/base.yml"
+        - "./scenarios/centos-9/content_provider.yml"
+        - "./scenarios/centos-9/zuul_inventory.yml"
+
+    - name: Install necessary tools
+      ansible.builtin.import_role:
+        name: install_yamls_makes
+        tasks_from: make_download_tools
+
+    - name: Deploy registry
+      ansible.builtin.import_role:
+        name: registry_deploy
+
+    # Imported from the ci/playbooks/content_provider/content_provider.yml play
+    # Ensure we get consistent name for the operator, linked to
+    # https://github.com/openstack-k8s-operators/ci-framework/pull/621
+    - name: Set var for cifmw_operator_build_operators var
+      when:
+        - zuul is defined
+        - "'project' in zuul"
+        - "'short_name' in zuul.project"
+      ansible.builtin.set_fact:
+        cifmw_operator_build_operators:
+          - name: "openstack-operator"
+            src: "~/src/github.com/{{ cifmw_operator_build_org }}/openstack-operator"
+            image_base: >-
+              {{ zuul.project.short_name | split('-') | reject('search','operator') | join('-') }}
+
+    - name: Build Operators
+      ansible.builtin.include_role:
+        name: operator_build
+
+    - name: Get the containers list from container registry
+      ansible.builtin.uri:
+        url: "http://{{ cifmw_rp_registry_ip }}:5001/v2/_catalog"
+        return_content: true
+      register: cp_imgs
+
+    - name: Add the container list to file
+      ansible.builtin.copy:
+        content: "{{ cp_imgs.content }}"
+        dest: "{{ ansible_user_dir }}/local_registry.log"
+        mode: "0644"
+
+    - name: Output needed content into consumable environment
+      ansible.builtin.copy:
+        dest: "{{ ansible_user_dir }}/{{ job_id }}-params/content-provider.yml"
+        content: |-
+          {{
+            {'cifmw_operator_build_output': cifmw_operator_build_output,
+             'content_provider_registry_ip': cifmw_rp_registry_ip
+            } | to_nice_yaml
+          }}
+{% endraw %}
+{% endif %}
+
 {% for play in zuul_plays %}
 - name: "Reproducer for {{ job_id }}: {{ play }}"
   vars:
@@ -6,6 +85,9 @@
       - '@~/{{ job_id }}-params/custom-params.yml'
       - '@~/{{ job_id }}-params/install-yamls-params.yml'
       - '@~/{{ job_id }}-params/reproducer_params.yml'
+{% if use_content_provider | bool %}
+      - '@~/{{ job_id }}-params/content-provider.yml'
+{% endif %}
       - cifmw_zuul_target_host=controller-0
       - cifmw_openshift_user=kubeadmin
       - cifmw_openshift_kubeconfig=/home/zuul/.crc/machines/crc/kubeconfig
--- a/docs/source/reproducers/03-zuul.md
+++ b/docs/source/reproducers/03-zuul.md
@@ -6,21 +6,29 @@ a job that doesn't rely on the content-provider job.
 It also has to run against the latest PR content output, since you won't
 have access to the actual commit hash on github.
 
-For now, it has been tested only against
-*podified-multinode-edpm-e2e-nobuild-tagged-crc* job.
+For now, it has been tested only the following jobs:
+* *podified-multinode-edpm-e2e-nobuild-tagged-crc*
+* *podified-multinode-edpm-deployment-crc* (which includes content-provider)
 
 ## Words of advice
 It is strongly advised to run this reproducer against a dedicated hypervisor
 with enough resources. The current configuration will require close to 40G
 of RAM and at least 24 CPUs. You can of course edit the specs,
 but keep in mind *CRC* has its own needs.
- [Please refer to its minimal spec](https://crc.dev/crc/getting_started/getting_started/installing/#_for_openshift_container_platform).
+[Please refer to its minimal spec](https://crc.dev/crc/getting_started/getting_started/installing/#_for_openshift_container_platform).
 
 ## How does it work (high level explanation)
 It will fetch a few files from the Zuul job output, read and combine them
 in order to update the layout to a matching size (especially, setting the
 right amount of computes), create the virtual machines, configure them,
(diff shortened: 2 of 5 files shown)


ci_framework/roles/reproducer/templates/play.yml.j2
```diff
@@ -1,11 +1,90 @@
 ---
+{% if use_content_provider is defined and use_content_provider | bool -%}
+- name: Configure CRC for insecure registry
+  hosts: crcs
```
pablintino, line 4: Not 100% related to this PR, but we are starting using groups and I'm not sure we are documenting it

cjeanner: soooo that group is 100% managed by the reproducer. Still, good point regarding the group in zuul job definition.

docs/source/reproducers/03-zuul.md
```diff
@@ -137,6 +145,35 @@ $ tail -f ~/ansible.log
...
+
+### Ansible tags of interest
+You may want to re-run some parts of the reproducer only, and avoid re-running the whole
+virtual machine creation.
+
+In order to do so, the reproducer exposes two tags:
+* bootstrap_layout
+* bootstrap
+
+#### Tag: bootstrap
```
dasm, line 163: I'm not following the difference between `bootstrap` and `bootstrap_layout`.
IIUC, `bootstrap_layout` can be skipped, but `bootstrap` not?

docs/source/reproducers/03-zuul.md
```diff
@@ -137,6 +145,35 @@ $ tail -f ~/ansible.log
...
+This covers everything from the creation of the virtual layout (networks and machines) to
+the push of generated plays. With that, you'll only get the repositories, and some of the
+zuul data, mostly.
+
+Use with caution, since it may end with a broken environment, where some files may miss an update,
```
dasm, line 168: Do you mean `Skip with caution`? Does it mean `bootstrap` is required to download updates?

docs/source/reproducers/03-zuul.md
```diff
@@ -137,6 +145,35 @@ $ tail -f ~/ansible.log
...
+for instance the playbook running the job in case you wanted to just update the repository against
+a newer patch set.
+
+#### Tag: bootstrap_layout
+This covers everything related to "create and manage virtual machines". This is probably the
```
dasm, line 173: Can we change the description here to something like:
> This tag is responsible for creating and managing virtual machines.
> It can be skipped when used on already deployed layout.

However I don't understand last bit of information. If the tag is responsible for updating repositories, does it mean we don't need to use this tag?

ci_framework/roles/reproducer/templates/play.yml.j2
```diff
@@ -1,11 +1,90 @@
 ---
+{% if use_content_provider is defined and use_content_provider | bool -%}
```
marios, line 2: can't we get away with 

```
{% if use_content_provider | default(false) | bool %}
```

ci_framework/roles/reproducer/templates/play.yml.j2
```diff
@@ -1,11 +1,90 @@
...
+    - name: Set var for cifmw_operator_build_operators var
+      when:
+        - zuul is defined
+        - "'project' in zuul"
+        - "'short_name' in zuul.project"
+      ansible.builtin.set_fact:
+        cifmw_operator_build_operators:
+          - name: "openstack-operator"
+            src: "~/src/github.com/{{ cifmw_operator_build_org }}/openstack-operator"
+            image_base: >-
```
marios, line 47: this seems fragile but i am also not sure what we're trying to construct below with the reject and join

i see image_base via grep at 

```
./ci/playbooks/content_provider/content_provider.yml:28:      # It handles the case of setting image_base for
./ci/playbooks/content_provider/content_provider.yml-29-      # openstack-ansibleee-operator and openstack-operator project
./ci/playbooks/content_provider/content_provider.yml-30-      # for openstack-ansibleee-operator, it will return openstack-ansibleee
./ci/playbooks/content_provider/content_provider.yml-31-      # and for openstack-operator, openstack will be returned
./ci/playbooks/content_provider/content_provider.yml-32-      when:
./ci/playbooks/content_provider/content_provider.yml-33-        - zuul is defined
--
./ci/playbooks/content_provider/content_provider.yml-35-        - "'short_name' in zuul.project"
./ci/playbooks/content_provider/content_provider.yml-36-      ansible.builtin.set_fact:
./ci/playbooks/content_provider/content_provider.yml-37-        cifmw_operator_build_operators:
./ci/playbooks/content_provider/content_provider.yml-38-          - name: "openstack-operator"
./ci/playbooks/content_provider/content_provider.yml-39-            src: "~/src/github.com/{{ cifmw_operator_build_org }}/openstack-operator"
./ci/playbooks/content_provider/content_provider.yml:40:            image_base: >-
./ci/playbooks/content_provider/content_provider.yml-41-              {{ zuul.project.short_name | split('-') | reject('search','operator') | join('-') }}


```

docs/source/reproducers/03-zuul.md
```diff
@@ -137,6 +145,35 @@ $ tail -f ~/ansible.log
...
+more useful tag to skip once you want to iterate on a deployed layout, since it will
```
marios, line 174: i think you meant "run" and not skip? i.e. "more useful tag to not skip"

docs/source/reproducers/03-zuul.md
cjeanner: Both are needed (i.e. not --skip-tags) in order to get the deploy. Then, in case you want to iterate, you may skip them `bootstrap_layout` - this will jump over all of the VM bootstrap, while allowing to fetch and generate the various files of the PR you're re-running.
This is useful in case you're iterating a code change and see a failure in Zuul for each iteration - with that possibility to skip the VM bootstrap (`--skip-tags bootstrap_layout`) you may re-use the same env.
The `bootstrap` is a bit more aggressive, and would prevent most of the file generation, in addition to skipping most of the `layout` part (i.e. VM creation).

Both tags have an overlap, but don't cover exactly the same areas.

docs/source/reproducers/03-zuul.md
cjeanner: hmm, a bit clumsy indeed. Skipping `bootstrap_layout` will let the repositories and needed files being updated. Lemme rephrase it a bit.

docs/source/reproducers/03-zuul.md
cjeanner: nope, that's the one we may want to skip in case of iteration over an existing env

ci_framework/roles/reproducer/templates/play.yml.j2
cjeanner: so it's mostly a copy'n'paste from that ci playbook

ci_framework/roles/reproducer/templates/play.yml.j2
cjeanner: maybe - though I tend to dislike using |default() in these kind of cases

cjeanner: (updated locally, will push once I address the other points)

docs/source/reproducers/03-zuul.md
cjeanner: reworded a bit, added some words of caution, and reorganized the tag list order.

docs/source/reproducers/03-zuul.md
cjeanner: reworded

docs/source/reproducers/03-zuul.md
cjeanner: reworded

docs/source/reproducers/03-zuul.md
cjeanner: reworded a bit

cjeanner:
recheck

docs/source/reproducers/03-zuul.md
```diff
@@ -137,6 +145,72 @@ $ tail -f ~/ansible.log
...
+# Iterate with the second iteration of your PR that failed on Zuul
+$ ansible-playbook -i custom/inventory.yml reproducer.yml \
+    -e cifmw_target_host=builder1 \
+    -e @scenarios/reproducers/3-nodes.yml \
+    -e cifmw_job_uri="YOUR_UPDATE_URI" \
+    --skip-tags bootstrap_layout
+```
+
+#### Tag: bootstrap
+This covers everything from the creation of the virtual layout (networks and machines) to
```
dasm, line 187: Hmm... I'm confused by lines 164 and 187.
IIUC `bootstrap_layout` should give me working environment for VMs: setup of network, create VM, start VM. `bootstrap` on the other hand should download me repositories, zuul config (data), etc.
However line 187 mentions: "everything from [..] virtual layout to [..] plays".
Should it be: "This covers everything *after* virtual layout"?

So:
`bootstrap_layout` + `bootstrap` == first config
`bootstrap` == subsequent runs

ci_framework/roles/reproducer/templates/play.yml.j2
```diff
@@ -1,11 +1,93 @@
...
+              {{ zuul.project.short_name | split('-') | reject('search','operator') | join('-') }}
+
+    - name: Build Operators
+      ansible.builtin.include_role:
+        name: operator_build
+
+    - name: Get the containers list from container registry
+      ansible.builtin.uri:
+        url: "http://{{ cifmw_rp_registry_ip }}:5001/v2/_catalog"
```
marios, line 59: wondering why we hard-code to 5001 ... it seems to be the case in many places, but grep tells me there is a var for that
https://github.com/openstack-k8s-operators/ci-framework/blob/d071087d8b9c12b00ac3dfaf68dace34d8df30ba/ci_framework/roles/registry_deploy/defaults/main.yml#L24

marios (commented):
lgtm, inline comment/question about the registry port (but not really just confined to this pull request seems to be a more general thing)

docs/source/reproducers/03-zuul.md
cjeanner: if you check the actual tags associated to the tasks, you will see a fair overlap of both `bootstrap` and `bootstrap_layout`. The `bootstrap` is a pre-existing tag, reused for some of the templates, repositories and related tasks in this context. The `bootstrap_layout` has been introduced in order to select only the VM parts.
You may see that `bootstrap` *also* covers some of the VM parts, in addition to the repositories/templates management.

Also keep in mind those tags are supposed to be *skipped*, not targeted using `--tags` - they represent tasks to exclude from a run.

ci_framework/roles/reproducer/templates/play.yml.j2
cjeanner: that part was also copied from the pre-existing playbook.

Unfortunately, I can't just run that playbook since it's doing far more than deploying the registry + build content, it also manages logs and related things that I do not want to get at that stage of the run in the reproducer. Maybe refactoring that ci playbook in order to allow a better inclusion is something to consider, in a follow-up.

pablintino:
/approve
@dasm @cjeanner needs this PR, as your comment is about documentation, if you consider this needs a change, he can do it in a followup

PR #639, opened 2023-10-05
No cert validation on keystone endpoint check

stuggi:
The task is only used to check when keystone is available. No need for now to have cert validation by fetching the CA.

  - [x] README in the role


ci_framework/roles/edpm_prepare/tasks/main.yml
```diff
@@ -220,6 +220,7 @@
       ansible.builtin.uri:
         url: "{{ _cifmw_edpm_prepare_keystone_endpoint_out.stdout | trim }}"
         status_code: "{{ _keystone_response_codes }}"  # noqa: args[module]
+        validate_certs: false
```
cjeanner, line 223: We probably don't want to hardcode that "false". Care to add a new parameter to that role? like `cifmw_edpm_prepare_verify_tls` or the like?

PR #640, opened 2023-10-06
No cert validation in os_net_setup

stuggi:
For now disable cert validation on os_net_setup role. In a follow up we fetch the CA from a CA bundle secret and can enable it. Disable it is a pre-req to allow integration of the TLS functionality for the public endpoints via the openstack-operator. The openstack-operator will create the required CA.

  - [X] README in the role


cjeanner:
/approve

Can't wait to toggle all of those to true :)

PR #644, opened 2023-10-09
Add missing interfaces for adoption base job

cescgina:
The controller node in the adoption base job nodeset does not have
interfaces for the storage and tenant networks. The storage network is
needed to use ceph.


cjeanner:
/lgtm

Looks OK - care to have a look, @pablintino ? I'll also need to double-check the reproducer to ensure I didn't make any weird filtering.

dasm:
/lgtm
/approve

@cjeanner I briefly checked for "filtering". I don't know "reproducer" code enough, but I didn't find any weirdness.

PR #645, opened 2023-10-09
Increasing default ip range

afazekas:
The current renge too small for parallel runs.
The default range might bo moved to different subnet in the
future, but for know only the range should be increased
in case of no overlapping known.

  - [x] README in the role


dasm:
It looks good.
@afazekas care to give examples?

/lgtm

It is applied correctly here [1]

[1]: https://review.rdoproject.org/zuul/build/e6eae361d71f4f3498d9e9441c420498/log/job-output.txt#5153

lewisdenny (approved):
lgtm, thanks Attila!

PR #647, opened 2023-10-09
Split edpm_baremetal into 2 tasks

steveb:
The documented process is to call devsetup/Makefile
edpm_baremetal_compute and then Makefile edpm_deploy_baremetal. This
change does that, and also aligns the task names with the compute
tasks in 06-deploy-edpm.yml.

Once this has landed, target edpm_baremetal can be deleted so there is
no coupling between the 2 Makefiles.

  - [x] README in the role

### DIFF
--- a/ci_framework/roles/edpm_deploy_baremetal/tasks/main.yml
+++ b/ci_framework/roles/edpm_deploy_baremetal/tasks/main.yml
@@ -44,14 +44,23 @@
   ansible.builtin.include_role:
     name: set_openstack_containers
 
-- name: Install Dataplane with baremetal
+- name: Create virtual baremetal
   vars:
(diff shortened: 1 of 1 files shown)


ci_framework/roles/edpm_deploy_baremetal/tasks/main.yml
```diff
@@ -44,14 +44,23 @@
   ansible.builtin.include_role:
     name: set_openstack_containers
 
-- name: Install Dataplane with baremetal
+- name: Create and provision virtual baremetal
```
rabi, line 47: `provision` here is probably misleading.

steveb: good point

rabi (approved):
/lgtm

raukadah (commented):
/approve

PR #652, opened 2023-10-10
Correct how we waited for VM SSH to be ready

cjeanner:
Running the reproducer against a multi-compute scenario raised an issue
in the established logic in said step.

We now wait for SSH using the proper IP, and from the hypervisor instead
of some fancy localhost that wasn't configured until... next step.

We also take the opportunity to leverage a better filter for IP address
extraction.

This patch also corrects how repositories were cloned, especially when it
comes to github repository and the tested PR. Since most of the
repositories work with forks instead of dev-branches, the PR code/change
isn't usually known by the main repository - and we have to switch to
some specific reference instead.

We also take the opportunity to ensure the controller-0 knows the public
ssh key of all the hosts, even itself.

### DIFF
--- a/ci_framework/roles/libvirt_manager/tasks/create_vms.yml
+++ b/ci_framework/roles/libvirt_manager/tasks/create_vms.yml
@@ -52,7 +52,7 @@
 - name: "(localhost) Inject ssh jumpers for {{ vm_type }}"
   delegate_to: localhost
   vars:
-    extracted_ip: "{{ (vm_ip.stdout.split())[3] |replace('/24','') }}"
+    extracted_ip: "{{ (vm_ip.stdout.split())[3] | ansible.utils.ipaddr('address') }}"
   ansible.builtin.blockinfile:
     create: true
     path: "~/.ssh/config"
@@ -71,7 +71,7 @@
 
 - name: "({{ inventory_hostname }}) Inject ssh jumpers for {{ vm_type }}"  # noqa: name[template]
   vars:
-    extracted_ip: "{{ (vm_ip.stdout.split())[3] |replace('/24','') }}"
+    extracted_ip: "{{ (vm_ip.stdout.split())[3] | ansible.utils.ipaddr('address') }}"
   ansible.builtin.blockinfile:
     create: true
     path: "~/.ssh/config"
@@ -96,15 +96,16 @@
     src: inventory.yml.j2
 
 - name: "Wait for SSH on VMs type {{ vm_type }}"
-  delegate_to: localhost
+  vars:
+    extracted_ip: "{{ (vm_ip.stdout.split())[3] | ansible.utils.ipaddr('address') }}"
   ansible.builtin.wait_for:
-    host: "{{ vm_ip.item }}"
+    host: "{{ extracted_ip }}"
     port: 22
     delay: 5
   loop: "{{ vm_ips.results }}"
   loop_control:
     loop_var: vm_ip
-    label: "{{ vm_ip.item }}"
+    label: "{{ extracted_ip }}"
 
 - name: "Configure ssh access on type {{ vm_type }}"
   when:
--- a/ci_framework/roles/reproducer/files/pre-ci-play.yml
+++ b/ci_framework/roles/reproducer/files/pre-ci-play.yml
@@ -1,6 +1,6 @@
 ---
 - name: Configure networking on all nodes
-  hosts: all
+  hosts: all,!localhost
   gather_facts: false
   become: true
   vars:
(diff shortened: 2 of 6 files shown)


ci_framework/roles/libvirt_manager/tasks/create_vms.yml
```diff
@@ -96,15 +96,16 @@
     src: inventory.yml.j2
 
 - name: "Wait for SSH on VMs type {{ vm_type }}"
-  delegate_to: localhost
+  vars:
+    extracted_ip: "{{ (vm_ip.stdout.split())[3] |replace('/24','') }}"
```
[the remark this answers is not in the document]
cjeanner: as a follow-up then, it's used in multiple places

cjeanner:
/hold

need some more work

ci_framework/roles/libvirt_manager/tasks/create_vms.yml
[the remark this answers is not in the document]
cjeanner: done.

cjeanner:
/unhold

That patch needed some reworking - and it ended to some extension hitting some more things. But at least, I'm able to reproduce this job locally:
https://review.rdoproject.org/zuul/build/15490c8379e843ab99ee569b47e467e4

Which is:
- not ci-framework repository
- multi-compute
So something brand new for the reproducer !!!

ci_framework/roles/libvirt_manager/tasks/create_vms.yml
```diff
@@ -52,7 +52,7 @@
 - name: "(localhost) Inject ssh jumpers for {{ vm_type }}"
   delegate_to: localhost
   vars:
-    extracted_ip: "{{ (vm_ip.stdout.split())[3] |replace('/24','') }}"
+    extracted_ip: "{{ (vm_ip.stdout.split())[3] | ansible.utils.ipaddr('address') }}"
```
marios, line 55: seems legit :) https://docs.ansible.com/ansible/latest/collections/ansible/utils/ip_address_test.html

ci_framework/roles/reproducer/tasks/ci_job.yml
```diff
@@ -177,29 +183,38 @@
...
+                dict2items |
+                rejectattr('key', 'equalto', 'cifmw_operator_build_output') |
+                rejectattr('key', 'equalto', 'content_provider_registry_ip:') |
+                items2dict
+              }}
+          ansible.builtin.copy:
+            dest: "ci-framework-data/artifacts/parameters/zuul-params.yml"
+            content: "{{ zuul_params_filtered | to_nice_yaml }}"
+
+    - name: Check for ansible logs file and rotate it
```
marios, line 209: this is a nice optimization using loop and a rotate_log task file 

as a general comment this review could have been split into smaller commits (eg bits like this) for better separation and review

ci_framework/roles/reproducer/files/pre-ci-play.yml
```diff
@@ -61,18 +61,33 @@
...
+            cmd: >-
+              ssh-keyscan {{ hostvars[item].ansible_host }}
+              >> ~/.ssh/known_hosts
+          loop: >-
+            {{
+              hostvars | dict2items |
+              selectattr("value.ansible_host", "defined") |
+              map(attribute="key")
+            }}
+          loop_control:
```
pablintino, line 88: Is this needed? The item is already a simple string

ci_framework/roles/reproducer/tasks/rotate_log.yml
```diff
@@ -0,0 +1,12 @@
+---
+- name: "Check {{ item }}"
+  register: log_file
+  ansible.builtin.stat:
+    path: "{{ item }}"
+
+- name: Rotate log if present
+  when:
+    - log_file.stat.exists
+  ansible.builtin.shell:
```
pablintino, line 10: Is `ansible.builtin.command` not working for this `mv`?

ci_framework/roles/reproducer/tasks/ci_job.yml
```diff
@@ -177,29 +183,38 @@
...
+            path: "{{ job_id }}-params/zuul-params.yml"
+
+        - name: Push extracted content
+          vars:
+            zuul_params_filtered: >-
+              {{
+                (zuul_params['content'] | b64decode | from_yaml) |
+                dict2items |
+                rejectattr('key', 'equalto', 'cifmw_operator_build_output') |
+                rejectattr('key', 'equalto', 'content_provider_registry_ip:') |
```
pablintino, line 202: That `:` shouldn't be there

ci_framework/roles/reproducer/files/pre-ci-play.yml
cjeanner: woops, right.... It was late yesterday -.-'

ci_framework/roles/reproducer/tasks/ci_job.yml
cjeanner: woops... it was *really* late ;)

ci_framework/roles/reproducer/tasks/ci_job.yml
cjeanner: Yeah... maybe. Well, I had to work my way through too many things at the same time, and I got pretty annoyed with the logs rotation and the copy-pasting thing :).

cjeanner:
/hold

seems there's an issue with the items clone.

ci_framework/roles/reproducer/tasks/rotate_log.yml
cjeanner: ok, there's actually a better way I didn't know.
Note that, apparently, there's a fact available in ansible, created during the fact gathering, named `ansible_date_time` as described [here](https://www.ansiblepilot.com/articles/using-date-time-and-timestamp-in-ansible-playbook-ansible-tip-and-tricks/).

But since this timestamp isn't updated, it may be better to consume a fresh one, using the method I introduced (setting an internal var calling a `lookup('pipe', 'date +%s')` ).

cjeanner:
recheck

cjeanner:
recheck

dasm:
recheck

PR #667, opened 2023-10-15
Add high level tags to deploy-edpm.yml playbook

fultonj:
If the deploy-edpm.yml playbook fails it saves time to be able to re-run it and pass tags so certain steps can be skipped.

Tags are named their plays. The control-plane tag is the same as the edpm tag for now since the 06-deploy-edpm.yml playbook also deploys the control-plane. Further tagging can be done within the edpm_prepare role to differentiate the control-plane and data-plane.

Tags were not added to the 01-bootstrap.yml playbook call because the bootstrap tag already exists and the playbook has some tasks with an always tag.

  - [x] README in the role

### DIFF
--- a/docs/source/usage/01_usage.md
+++ b/docs/source/usage/01_usage.md
@@ -125,6 +125,15 @@ or `--skip-tags`:
 * `bootstrap`: Run all of the package installation tasks as well as the potential system configuration depending on the options you set.
 * `packages`: Run all package installation tasks associated to the options you set.
 * `bootstrap_layout`: Run the [reproducer](../reproducers/01-considerations.md) bootstrap steps only.
+* `infra`: Denotes tasks to prepare host virtualization and Openshift Container Platform when deploy-edpm.yml playbook is run.
+* `build-packages`: Denotes tasks to call the role [pkg_build](../roles/pkg_build.md) when deploy-edpm.yml playbook is run.
+* `build-containers`: Denotes tasks to call the role [build_containers](../roles/build_containers.md) when deploy-edpm.yml playbook is run.
+* `build-operators`: Denotes tasks to call the role [operator_build](../roles/operator_build.md) when deploy-edpm.yml playbook is run.
+* `control-plane`: Deploys the control-plane on OpenShift by creating `OpenStackControlPlane` CRs when deploy-edpm.yml playbook is run.
+* `edpm`: Deploys the data-plane (External Data Plane Management) on RHEL nodes by creating `OpenStackDataPlane` CRs when deploy-edpm.yml playbook is run.
+* `admin-setup`: Denotes tasks to call the role [os_net_setup](../roles/os_net_setup.md) when deploy-edpm.yml playbook is run.
+* `run-tests`: Denotes tasks to call the role [tempest](../roles/tempest.md) when deploy-edpm.yml playbook is run.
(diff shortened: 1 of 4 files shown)


cjeanner:
New tags - loving it! Just two remarks:
- care to "converge" toward tag list, even if there's only one tag? It would then be easier in the future to add more if wanted/needed
- care to update the doc? There's already a (small) list here: https://ci-framework.readthedocs.io/en/latest/usage/01_usage.html#ansible-tags - source code is [here](https://github.com/openstack-k8s-operators/ci-framework/blob/main/docs/source/usage/01_usage.md)

Thank you John!

fultonj:
> New tags - loving it! Just two remarks:
> 
> * care to "converge" toward tag list, even if there's only one tag? It would then be easier in the future to add more if wanted/needed
> * care to update the doc? There's already a (small) list here: https://ci-framework.readthedocs.io/en/latest/usage/01_usage.html#ansible-tags - source code is [here](https://github.com/openstack-k8s-operators/ci-framework/blob/main/docs/source/usage/01_usage.md)
> 
> Thank you John!

No problem, I've made these updates in the latest patch set. Thanks for the review.

lewisdenny (approved):
Thanks John tags and docs[1] look good to me

[1] https://logserver.rdoproject.org/67/667/77f2b44f3ca71f4e07d7b7b3d66a6f8a3bbc363e/github-check/cifmw-doc/497a12c/doc_build/html/usage/01_usage.html#ansible-tags

fultonj:
Thanks, I want to add a few more tags

fultonj:
> Thanks, I want to add a few more tags

Or really just move the `control-plane` tag to the subset of tasks in the edpm_deploy role which deploy the control plane.

https://github.com/openstack-k8s-operators/ci-framework/compare/242c0288d1753788368a36d3e72e14c00b60cd42..87b92e5b9c6448ba3eb5e53a2e580093eac357db

docs/source/usage/01_usage.md
```diff
@@ -134,6 +143,14 @@ $ ansible-playbook deploy-edpm.yml -K --tags bootstrap,packages [-e @scenarios/c
 $ ansible-playbook deploy-edpm.yml -K --skip-tags bootstrap,packages [-e @scenarios/centos-9/some-environment -e <...>]
 ```
 
+If you've already deployed OpenStack but it failed
```
cjeanner, line 146: I'd push that new block just bellow the next one actually - else, the talk about "running twice with --tags and --skip-tags" doesn't make sense anymore at that point. Care to move it?

fultonj: I don't understand.

Block = lines 146 to 153 ?

If so, you want it at the "next one". Which lines is the next one?

"the talk about "running twice with --tags and --skip-tags" doesn't make sense anymore"

My example only uses --tags, it doesn't use --skip-tags. There's an example prior with --skip-tags but they are independent.

Move it where?

fultonj: I think I understand. I was parsing this as:

- example 1
- example 2
- general statement

but I think you were thinking that the "general statement" was a statement about example 1. With that in mind I believe I have made the requested change.

cjeanner: thank you!

PR #669, opened 2023-10-16
Ensure we build the plugin documentation

cjeanner:
It seems RTD doesn't consume the Makefile, meaning it didn't build the
plugin documentation.

This patch adds a jobs.pre_build entry in order to make RTD builder
aware of that step, according the official documentation[1] (using their
example for doxygen pre-build).

[1] https://docs.readthedocs.io/en/stable/build-customization.html#generate-documentation-from-annotated-sources-with-doxygen

- [X] no checkbox available for it, we must merge before seeing if it breaks or works

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -10,18 +10,16 @@ build:
   os: ubuntu-22.04
   tools:
     python: "3.11"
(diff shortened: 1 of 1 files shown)


cjeanner:
recheck

PR #671, opened 2023-10-16
Fix 06-deploy-edpm.yaml playbook

rabi:
This is currently confusing with incorrect task naming.

### DIFF
--- a/ci_framework/playbooks/06-deploy-edpm.yml
+++ b/ci_framework/playbooks/06-deploy-edpm.yml
@@ -4,7 +4,7 @@
     step: pre_deploy
   ansible.builtin.import_playbook: ./hooks.yml
 
-- name: Prepare EDPM baremetal playbook
+- name: Deploy podified control plane
   hosts: "{{ cifmw_target_host | default('localhost') }}"
   gather_facts: false
   tasks:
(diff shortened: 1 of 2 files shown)


ci_framework/playbooks/06-deploy-edpm.yml
```diff
@@ -4,30 +4,39 @@
...
       ansible.builtin.import_role:
         name: edpm_prepare
 
-    - name: Perform Podified and EDPM deployment on compute nodes with virtual baremetal
-      when: cifmw_edpm_deploy_baremetal | default('false') | bool
-      ansible.builtin.import_role:
-        name: edpm_deploy_baremetal
-
 - name: Run post_ctlplane_deploy hooks
-  when: not cifmw_edpm_deploy_baremetal | default('false') | bool
```
cjeanner, line 25: so this is to be run on all kind of env - baremetal and pre-provisionned?

rabi: Why not? That's been defined for multinode jobs too[1]. It should also run validate_podified_deployment.yml like other jobs.

[1] https://github.com/openstack-k8s-operators/ci-framework/blob/main/scenarios/centos-9/multinode-ci.yml#L13

PR #673, opened 2023-10-16
[reproducer] take ssh key from {{ ansible_user_dir}}/.crc

stuggi:
  - [x] README in the role

### DIFF
--- a/ci_framework/roles/reproducer/tasks/main.yml
+++ b/ci_framework/roles/reproducer/tasks/main.yml
@@ -59,7 +59,7 @@
     - bootstrap_layout
   register: crc_priv_key
   ansible.builtin.slurp:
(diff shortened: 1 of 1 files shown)


ci_framework/roles/reproducer/tasks/main.yml
```diff
@@ -59,7 +59,7 @@
     - bootstrap_layout
   register: crc_priv_key
   ansible.builtin.slurp:
-    path: ".crc/machines/crc/id_ecdsa"
+    path: "{{ ansible_user_dir}}/.crc/machines/crc/id_ecdsa"
```
tosky, line 62: Late to the party, but doesn't this path depend on cifmw_artifacts_crc_sshkey ?

PR #674, opened 2023-10-16
[libvirt_manager] make sure ssh keys get created in {{ ansible_user_d…

stuggi:
…ir }}/.ssh

- [X] No CI for the reproducer yet

### DIFF
--- a/ci_framework/roles/libvirt_manager/tasks/deploy_layout.yml
+++ b/ci_framework/roles/libvirt_manager/tasks/deploy_layout.yml
@@ -70,20 +70,23 @@
     label: "{{ item.key }}"
 
 - name: Create temporary ssh keypair
+  delegate_facts: true
   ansible.builtin.command:
     cmd: >-
-      ssh-keygen -t ed25519 -f .ssh/cifmw_reproducer_key -P '' -q
-    creates: .ssh/cifmw_reproducer_key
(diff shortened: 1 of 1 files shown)


ci_framework/roles/libvirt_manager/tasks/deploy_layout.yml
```diff
@@ -72,18 +72,18 @@
 - name: Create temporary ssh keypair
   ansible.builtin.command:
     cmd: >-
-      ssh-keygen -t ed25519 -f .ssh/cifmw_reproducer_key -P '' -q
-    creates: .ssh/cifmw_reproducer_key
+      ssh-keygen -t ed25519 -f {{ ansible_user_dir }}/.ssh/cifmw_reproducer_key -P '' -q
```
cjeanner, line 75: I got issue locally with that, because I'm juggling with 3 distinct users (local from laptop, remote on hypervisor, remote in VMs) and for some reasons, that one here was pointing to the one on the hypervisor... Wondering if we may need to add `delegate_facts: true` as well... ?

stuggi: could you give this a try in your env ?

cjeanner: can't really now, it's locked on some other tasks.... If it works for you like this, let's get it in - if really, we can iterate.

Since we don't have CI on that precise part for now (it's a bit complicated to get it yet), we have to take an iterative approach anyway.

PR #677, opened 2023-10-17
Checkout PR branch

cjeanner:
Until now, the cloned repository got the branch containing the pull
request, but it was still in the `main` branch.

This patch ensures we switch branch to the proper content.


stuggi:
with this the change the operator PR was correct built and deployed in my reproducer env

PR #679, opened 2023-10-17
Replace lookup() by slurp calls

cjeanner:
File lookup() is running locally on the ansible controller, preventing
remote execution of the role.
This change switches to `ansible.builtin.slurp` in order to ensure we
can run the role against a remote hypervisor.


psathyan (commented):
/lgtm

PR #686, opened 2023-10-18
Patch dev-scripts on the fly for virtproxyd.socket

cjeanner:
This patch was proposed on the upstream dev-scripts repository - but
some environment would need it now.
So let's apply it directly - we can remove that block once it's merged
upstream.

The task won't fail even if the patch is merged, so it's safe. Unless we
have to modify the upstream patch, of course.


psathyan (approved):
/lgtm

PR #687, opened 2023-10-18
Ensure "patch" is available on the system

cjeanner:
It seems this binary isn't in the base system - let's ensure it's
available before trying to patch dev-scripts.


raukadah (approved):
/approve

PR #692, opened 2023-10-18
Keep false default value under quotes

raukadah:
If we keep `cifmw_edpm_deploy_hci | default(false) | bool`. The conditional check fails and complains with following error:
```
"The conditional check 'cifmw_edpm_deploy_hci | default(false) | bool' failed.
The error was: unable to evaluate conditional: cifmw_edpm_deploy_hci | default(false) | bool
The error appears to be in '/src/ci-framework/ci_framework/playbooks/ceph.yml': line 24, column 7,
but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:
```

Adding it under quotes fixes the issue.

### DIFF
--- a/ci_framework/playbooks/06-deploy-edpm.yml
+++ b/ci_framework/playbooks/06-deploy-edpm.yml
@@ -59,7 +59,7 @@
             tasks_from: deploy_edpm_compute.yml
 
         - name: Prepare for HCI deploy phase 1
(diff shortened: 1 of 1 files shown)


viroel:
/lgtm

trivial fix, hope that fixes your issues.

Thanks

PR #678, opened 2023-10-17
[devscripts] Add file related vars for pull secrets and ci-token

raukadah:
Currently we are passing pull secrets and ci-token content as a string in the ansible var. In order to consume it in CI, we need to slurp the content of pull secret and ci-token in these vars.

If somehow the these vars are passed via var files through command line, it might expose the secrets on failures.

This pr adds two new vars to pass the pull secrets and ci-token as file also.

As a pull request owner and reviewers, I checked that:
- [x] Appropriate testing is done and actually running

PR #696, opened 2023-10-19
Copy ci_token_file file to the devscript directory

raukadah:
https://github.com/openstack-k8s-operators/ci-framework/pull/678 added the support for pull secret and ci-token var. It also introduces a typo in the ci_token file conditional.

This patch first copy the ci_token_file in the devscript (if devscript role is used via reproducer) and then cat the output of ci_token to avoid unwanted failure.

### DIFF
--- a/ci_framework/roles/devscripts/tasks/03_install.yml
+++ b/ci_framework/roles/devscripts/tasks/03_install.yml
@@ -53,6 +53,17 @@
     regexp: "^    address_v4(.+)$"
     replace: "    address_v4: {{ ansible_default_ipv4.address }}"
 
+- name: Copy the CI token file in devscript repo directory
+  when: cifmw_devscripts_ci_token_file is defined
+  tags:
+    - bootstrap
+  ansible.builtin.copy:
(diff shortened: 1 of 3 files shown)


ci_framework/roles/devscripts/tasks/03_install.yml
```diff
@@ -53,6 +53,17 @@
...
+- name: Copy the CI token file in devscript repo directory
+  when: cifmw_devscripts_ci_token_file is defined
+  tags:
+    - bootstrap
+  ansible.builtin.copy:
+    src: "{{ cifmw_devscripts_ci_token_file }}"
+    dest: "{{ cifmw_devscripts_repo_dir }}/ci_token"
+    owner: "{{ cifmw_devscripts_user }}"
+    group: "{{ cifmw_devscripts_user }}"
+    mode: "0644"
```
cjeanner, line 65: would probably be safer to set "0400" or, at least, "0600"

raukadah: Done

lewisdenny (approved):

PR #697, opened 2023-10-19
Preset cifmw_devscripts_config var before running cleanup

raukadah:
The cleanup task fails with cifmw_devscripts_config undefined var.
    It needs to be set before running the actual cleanup tasks.

### DIFF
--- a/ci_framework/roles/devscripts/tasks/cleanup.yml
+++ b/ci_framework/roles/devscripts/tasks/cleanup.yml
@@ -14,6 +14,15 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+- name: Gather the configurations to be passed to dev-scripts.
+  ansible.builtin.set_fact:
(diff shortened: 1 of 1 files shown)


ci_framework/roles/devscripts/tasks/cleanup.yml
```diff
@@ -31,7 +31,7 @@
 - name: Cleanup the devscripts working directory.
   become: true
   ansible.builtin.file:
-    path: "{{ cifmw_devscripts_config.working_dir }}"
```
cjeanner, line 34: nope, we want to `|combine` like it's done in the main.yml here: 
https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci_framework/roles/devscripts/tasks/main.yml#L18-L25

psathyan (commented):
/lgtm

psathyan (approved):
/lgtm

PR #698, opened 2023-10-19
Fix unversioned python for RHEL 8

rdiazcam:
python-unversioned-command package does not exist for RHEL 8 so it has been removed from bindep for it.

In order to have unversioned python in RHEL 8 run this: 
~~~
sudo alternatives --set python3 /usr/bin/python3.9 sudo alternatives --set python /usr/bin/python3
~~~

  - [ ] README in the role


cjeanner:
We won't support el8 (nor cs8) as hypervisor. This was discussed directly with @rdiazcam

PR #700, opened 2023-10-19
devscripts tweaking

cjeanner:
- Ensure cleanup.yml tasks know about cifmw_devscripts_config
- Ensure we do not do anything BM related if no BM
- Move cifmw_devscripts_config_defaults to vars/main.yml


ci_framework/roles/devscripts/tasks/sub_tasks/42_add_bmh.yml
```diff
@@ -45,27 +45,27 @@
...
-- name: Apply the baremetal host definitions
-  environment:
-    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
-  ansible.builtin.command:
-    cmd: "oc apply -f {{ cifmw_devscripts_artifacts_dir }}/bmh/"
+    - name: Apply the baremetal host definitions
+      environment:
+        KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+      ansible.builtin.command:
+        cmd: "oc apply -f {{ cifmw_devscripts_artifacts_dir }}/bmh/"
```
psathyan, line 71: This should not come inside the block... it is supposed to be outside

`cifmw_devscripts_osp_compute_nodes` is either provided by the user as a param or is set here when `num_extra_nodes` is set

what is missing code be a block that checks `cifmw_devscripts_osp_compute_nodes`

psathyan: If we move it into the block then it would break when physical computes are being provided.

psathyan (commented):
Thank you for doing this

psathyan (commented):
/lgtm

PR #701, opened 2023-10-19
    Adds cifmw-adoption-standalone-base for rhel 9.2 OSP17/CDN job

marios:
Adds cifmw-adoption-standalone-base for rhel 9.2 OSP17/CDN job
    
    This adds cifmw-adoption-standalone-base which is essentially the same
    as cifmw-adoption-base but with the addition of the standalone node
    and ci-framework required networking config. Using this new parent
    will let us minimize disruption while adding the new job.
    
    Needed by https://review.rdoproject.org/r/c/rdo-jobs/+/50590/
    
    https://issues.redhat.com/browse/OSP-29746

- [X] no need for actual checkbox

### DIFF
--- a/zuul.d/adoption.yaml
+++ b/zuul.d/adoption.yaml
@@ -16,17 +16,17 @@
       groups:
         - name: computes
           nodes: []
-    roles:
+    roles: &adoption_roles
       - zuul: github.com/openstack-k8s-operators/ci-framework
-    pre-run:
+    pre-run: &pre_run
       - ci/playbooks/multinode-customizations.yml
       - ci/playbooks/e2e-prepare.yml
       - ci/playbooks/dump_zuul_data.yml
-    post-run:
+    post-run: &post_run
       - ci/playbooks/e2e-collect-logs.yml
       - ci/playbooks/collect-logs.yml
       - ci/playbooks/multinode-autohold.yml
-    vars:
+    vars: &adoption_vars
       zuul_log_collection: true
       registry_login_enabled: true
       push_registry: quay.rdoproject.org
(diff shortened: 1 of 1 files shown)


zuul.d/adoption.yaml
```diff
@@ -13,6 +13,8 @@
           label: cloud-centos-9-stream-tripleo-vexxhost-xl
         - name: crc
           label: coreos-crc-extracted-xxl
```
marios, line 15: lets move these additions to a duplicate base job that we can merge for now and then when we are done with this work we can reconcile the two

zuul.d/adoption.yaml
```diff
@@ -77,6 +77,74 @@
...
+- job:
+    name: cifmw-adoption-standalone-base
+    parent: base-extracted-crc
+    abstract: true
+    timeout: 10800
+    attempts: 1
+    nodeset:
+      nodes:
+        - name: controller
+          label: cloud-centos-9-stream-tripleo-vexxhost-xl
```
cescgina, line 89: not a big deal, but we could use `cloud-centos-9-stream-tripleo-vexxhost` here instead, since we no longer have a nested vm and the requirements are much less

cescgina (commented):
/lgtm just noted a small potential optimisation

zuul.d/adoption.yaml
marios: ack, i will add a TODO(marios) comment here to revisit. having some issues in post-run unregister subscription which is suspect is related to config repo and inheritance. So i want to push to get as much as possible of the series merged and start to iterate on improvements.

cescgina: that's perfectly fine, no need to block on such a detail

cescgina (commented):
/lgtm

cjeanner:
/approve

makes sense for a smooth migration.

raukadah (commented):
/approve

PR #702, opened 2023-10-19
Add custom RHEL 8 dependency packages

rdiazcam:
Some package names have different names for RHEL 8 compared to RHEL/CentOS 9 (e.g. libguestfs-tools instead of guestfs-tools)

  - [ ] README in the role


cjeanner:
We won't support el8 (nor cs8) as hypervisor. This was discussed directly with @rdiazcam

PR #703, opened 2023-10-19
Removed unwanted debug task that was merged by mistake.

pablintino:


cjeanneret:
/approve

though... might be interesting for further debugging?

cjeanner:
/approve

using right account may help right? Still - wouldn't those ones be useful for deeper debugging?

PR #704, opened 2023-10-19
Let user pick ansible_hostname or ansible_fqdn in Ceph spec

fultonj:
Follow up patch to 56fea615371e3982347cacf9baf27601945e311c

It's valid for the Ceph spec to have either a long hostname (foo.bar.com) or short hostname (foo). In TripleO we supported either option via the boolean CephSpecFqdn (default false) and defaulted to using the short name. The ci-framework only offered the shortname when building the ceph-spec originally.

A recent patch switched from ansible_hostname to ansible_fqdn for a CI job and users deploying locally saw this error when their spec had the longname.

  ERROR: hostname \"edpm-compute-0\" does not match expected
  hostname \"edpm-compute-0.ctlplane.example.com\"

This patch adds the ceph_spec_fqdn option with the same default and updates the CI job to set ceph_spec_fqdn true, thus mirroring the behavior we had in tripleo-ansible and its CI.

Also, remove cifmw_cephadm_bootstrap_host variable as it is declared once but otherwise never used.

  - [x] README in the role


fultonj:
Precedent:
- https://github.com/openstack/tripleo-heat-templates/commit/44d1e2ddd6813e57d818c77a1646920c6a2d08f9
- https://github.com/openstack/tripleo-ansible/commit/f223beb240e4e267e232aaca5fd5d2046ada73a7

dasm (commented):
/lgtm

fultonj:
Testing in larger ci job with https://review.rdoproject.org/r/c/testproject/+/50595

fmount:
+1 thanks John!

PR #707, opened 2023-10-20
Path ansibleee runner image to use correct image

raukadah:
Currently ansiblee runner image is not using the proper image. We need to force patch it to use the latest image.

### DIFF
--- a/ci_framework/roles/edpm_prepare/tasks/main.yml
+++ b/ci_framework/roles/edpm_prepare/tasks/main.yml
@@ -115,6 +115,20 @@
   ansible.builtin.include_role:
     name: set_openstack_containers
 
+- name: Patch ansible runner image temporarily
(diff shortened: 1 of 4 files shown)


ci_framework/roles/edpm_prepare/tasks/main.yml
```diff
@@ -115,6 +115,18 @@
   ansible.builtin.include_role:
     name: set_openstack_containers
 
+- name: Patch ansible runner image temprorily
+  when: not cifmw_edpm_prepare_update_os_containers | bool
```
viroel, line 119: we need another condition to skip this, since in downstream we don't want use os containers and also don't want to use upstream ansiblee-runner. So I think that this need a especific skip flag

viroel (dismissed):

PR #690, opened 2023-10-18
Add support for dev-scripts and VA1 to reproducer

cjeanner:
This patch introduces support for dev-scripts, a way to deploy a 3-nodes
OCP, in the reproducer.
For now, we can *either* run the standard libvirt_manager to get the CI
layout, *or* run devscripts and end with a 3-OCP + 3-compute layout.

As a pull request owner and reviewers, I checked that:
- [ ] Appropriate testing is done and actually running
- [ ] Appropriate documentation exists and/or is up-to-date:
  - [ ] README in the role
  - [ ] Content of the docs/source is reflecting the changes

PR #709, opened 2023-10-23
Update host's authorized_keys before injecting into base image

raukadah:
In libvirt manager role, We create a base image for deploying compute vms based on layout. We use user's authorized keys while creating the image.

Devscripts role is used for deploy openshift. DevScripts itself generates two ssh keys[1] at user and root level.

All the users on the host does not ssh into the vms created from above base image as used authorized keys does not have all the pub keys added.

This pr finds all the authorized keys at user's level and update the authorized key. It will allow user's to ssh into the
machine. It will also allow all non-root users to ssh into the compute vms.

[1]. https://github.com/openshift-metal3/dev-scripts/blob/1de0dc213e9f23820a915cbe85ff8afa820f34d8/02_configure_host.sh#L13

### DIFF
--- a/ci_framework/roles/libvirt_manager/tasks/deploy_layout.yml
+++ b/ci_framework/roles/libvirt_manager/tasks/deploy_layout.yml
@@ -50,6 +50,20 @@
   loop_control:
     label: "{{ item.key }}"
 
+- name: Find all public key in the default user directory
+  ansible.builtin.find:
+    paths: "{{ ansible_user_dir }}/.ssh"
+    file_type: file
+    patterns: '*.pub'
+  register: users_pub_key
+
+- name: Update authorized keys for all users
+  ansible.posix.authorized_key:
+    user: zuul
+    state: present
+    key: '{{ item }}'
+  with_file: "{{ users_pub_key.files | map(attribute='path') }}"
+
 - name: Prepare base image with user and ssh accesses  # noqa: risky-shell-pipe
   when:
     - item.key is not match('^crc.*$')


ci_framework/roles/libvirt_manager/tasks/deploy_layout.yml
```diff
@@ -50,6 +50,37 @@
...
 
+- name: Find all public key in the default user directory
+  ansible.builtin.find:
+    paths: "{{ ansible_user_dir }}/.ssh"
+    file_type: file
+    patterns: '*.pub'
+  register: users_pub_key
+
+- name: Update authorized keys for existing user
+  ansible.builtin.shell: |
```
cjeanner, line 61: please use the ansible.posix.ssh_authorized_key: https://docs.ansible.com/ansible/latest/collections/ansible/posix/authorized_key_module.html

ci_framework/roles/libvirt_manager/tasks/deploy_layout.yml
```diff
@@ -50,6 +50,37 @@
...
+- name: Find all public key in root directory
+  become: true
+  ansible.builtin.find:
+    paths: "/root/.ssh"
+    file_type: file
+    patterns: '*.pub'
+  register: root_pub_key
+
+- name: Update authorized keys for root user
+  ansible.builtin.shell: |
```
cjeanner, line 77: ditto - https://docs.ansible.com/ansible/latest/collections/ansible/posix/authorized_key_module.html

ci_framework/roles/libvirt_manager/tasks/deploy_layout.yml
```diff
@@ -50,6 +50,37 @@
   loop_control:
     label: "{{ item.key }}"
 
+- name: Find all public key in the default user directory
+  ansible.builtin.find:
```
cjeanner, line 54: slurp

ci_framework/roles/libvirt_manager/tasks/deploy_layout.yml
```diff
@@ -50,6 +50,20 @@
...
+    file_type: file
+    patterns: '*.pub'
+  register: users_pub_key
+
+- name: Update authorized keys for all users
+  ansible.posix.authorized_key:
+    user: zuul
+    state: present
+    key: '{{ item }}'
+  with_file: "{{ users_pub_key.files | map(attribute='path') }}"
```
cjeanner, line 65: I just hope `with_file` isn't a wrapper around some `lookup()` - is it working remotely?

ci_framework/roles/libvirt_manager/tasks/deploy_layout.yml
```diff
@@ -50,6 +50,20 @@
...
+- name: Find all public key in the default user directory
+  ansible.builtin.find:
+    paths: "{{ ansible_user_dir }}/.ssh"
+    file_type: file
+    patterns: '*.pub'
+  register: users_pub_key
+
+- name: Update authorized keys for all users
+  ansible.posix.authorized_key:
+    user: zuul
```
cjeanner, line 62: that user is supposed to be on the hypervisor? It exists only in CI - others may have yet another username on the hypervisor.

Also, for dev case, they will probably get an ever growing authorized_keys at this point.

raukadah:
This pr is no loner needed. https://github.com/openstack-k8s-operators/ci-framework/pull/690 solves the issue i have faced related to authorized keys.

PR #718, opened 2023-10-24
Allow to NOT affect network on nodes

cjeanner:
Until now, it was mandatory to get network layout for every single node
we have in the inventory. This change allows to not be so strict, and
will just end the play on the hosts that don't show up in the provided
layout.

- [X] Successfully tested in the VA-1 patch

### DIFF
--- a/ci_framework/roles/ci_network/tasks/validate.yml
+++ b/ci_framework/roles/ci_network/tasks/validate.yml
@@ -11,6 +11,11 @@
     loop_var: "network"
     label: "{{ network.key }}"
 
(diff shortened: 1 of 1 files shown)


cjeanner:
recheck

ci_framework/roles/ci_network/tasks/validate.yml
```diff
@@ -11,6 +11,11 @@
     loop_var: "network"
     label: "{{ network.key }}"
 
+- name: Abord host if not in the provided net_layout
```
pablintino, line 14: Typo in `abord`?

PR #719, opened 2023-10-24
Ensure we get some minimal packages on RHEL based nodes

cjeanner:
- [X] Successfully tested on the VA-1 deploy


cjeanner:
recheck

PR #720, opened 2023-10-24
Correct libvirt_manager

cjeanner:
This patch corrects some major issues in the role.

- cifmw_reproducer_key wasn't used at all, and generated at the wrong
  time
  This lead to issues where, if a user wasn't forwarding the SSH Agent,
  accesses to the virtual machines weren't working in a remote
  deployment.
  We now generate the key at the right time, inject it in the VM, and
  ensure the hypervisor is consuming the private key for the accesses.

- localhost deployment support
  It may happen users want to deploy on localhost instead of a remote
  hypervisor. This wasn't properly supported, because we were adding two
  kind of SSH configuration, leading to hiccups.
  We now prevent the ProxyJump config to appear in the localhost case.

- cifmw_reproducer_key cleanup
  Since we're injecting the public key in the authorized_keys, we take
  care to clean it up during the cleanup step.

- inject nodes in the ansible inventory
  In order to make things a bit easier to handle on the ansible side, we
  ensure the new VMs are all added to the current inventory (in-memory).
  This ensures we're using the correct data, and the nodes are properly
  known by the current play. It's NOT mandatory, but just ensures things
  are cleaner.

- allow to not define any networks in the configuration
  Until now, we had to provide a `networks` entry in the configuration
  parameter. This lead to issues when we were re-using existing
  networks.

- allow to reuse existing virtual networks
  It may happen we just want to reuse existing networks. In such case,
  they will probably don't get the `cifmw-` prefix. This patch takes
  care of this specific case.

- stabilize libvirt accesses
  We hit a fair amount of failures with the libvirt sockets over the
  past iteration. After many tests, it seems that removing the timeout
  of the virtproxyd service and ensuring it's enabled and started
  corrects this issue. With this hack, we could successfuly deploy the
  whole VA-1 layout without any issue.

- ensure correct ACLs are applied to home directory
  Since we're usually consuming disk images from within the home
  directory, we have to ensure `qemu` user has proper access.

- new network-data template
  Since we may not reproduce a CI run, we have to generate some content
  instead when it comes to network data pushed on the nodes.

-

- [X] Successfully tested in the VA-1 deployment.

### DIFF
--- a/ci_framework/roles/libvirt_manager/tasks/create_vms.yml
+++ b/ci_framework/roles/libvirt_manager/tasks/create_vms.yml
@@ -50,6 +50,8 @@
     - vm_ips.stdout != ''
 
 - name: "(localhost) Inject ssh jumpers for {{ vm_type }}"
+  when:
+    - inventory_hostname != 'localhost'
   delegate_to: localhost
   vars:
     extracted_ip: "{{ (vm_ip.stdout.split())[3] | ansible.utils.ipaddr('address') }}"
@@ -80,8 +82,22 @@
       Host {{ vm_type }}-{{ vm_ip.item }} cifmw-{{ vm_type }}-{{ vm_ip.item }} {{ extracted_ip }}
         Hostname {{ extracted_ip }}
         User zuul
+        IdentityFile ~/.ssh/cifmw_reproducer_key
         StrictHostKeyChecking no
         UserKnownHostsFile /dev/null
+  loop: "{{ vm_ips.results }}"
+  loop_control:
+    loop_var: vm_ip
+    label: "{{ vm_ip.item }}"
+
+- name: Inject nodes in the ansible inventory
+  delegate_to: localhost
+  vars:
+    extracted_ip: "{{ (vm_ip.stdout.split())[3] | ansible.utils.ipaddr('address') }}"
+  ansible.builtin.add_host:
+    name: "{{ vm_type }}-{{ vm_ip.item }}"
+    groups: "{{ vm_type }}"
+    ansible_host: "{{ extracted_ip }}"
   loop: "{{ vm_ips.results }}"
   loop_control:
     loop_var: vm_ip
@@ -178,7 +194,11 @@
 - name: "Inject network configuration file on {{ vm_type }}"
   ansible.builtin.template:
     dest: "{{ cifmw_reproducer_basedir }}/reproducer-network-env/{{ vm_type }}-{{ vm_id }}.yml"
-    src: "network-data.yml.j2"
+    src: >-
+      {{
+        (ci_job_networking is defined)
+        | ternary('ci-network-data.yml.j2', 'network-data.yml.j2')
+      }}
     mode: "0644"
   loop: "{{ range(0, vm_data.value.amount | default(1) | int) }}"
   loop_control:
--- a/ci_framework/roles/libvirt_manager/tasks/virsh_checks.yml
+++ b/ci_framework/roles/libvirt_manager/tasks/virsh_checks.yml
@@ -21,6 +21,40 @@
     state: started
     enabled: true
 
+# stabilize libvirt accesses
+# We hit a fair amount of failures with the libvirt sockets over the
+# past iteration. After many tests, it seems that removing the timeout
+# of the virtproxyd service and ensuring it's enabled and started
+# corrects this issue. With this hack, we could successfuly deploy the
+# whole VA-1 layout without any issue.
+- name: Hack in the virtproxyd service
+  become: true
+  block:
+    - name: Create service override directory
+      ansible.builtin.file:
+        path: "/etc/systemd/system/virtproxyd.service.d"
+        state: directory
+        owner: root
+        group: root
+        mode: "0755"
+
+    - name: Create service override
+      ansible.builtin.copy:
+        dest: "/etc/systemd/system/virtproxyd.service.d/override.conf"
+        content: |
+          # Managed by the CI Framework
+          # Remove timeout to service to ensure it's persistent
+          [Service]
+          Environment=
+          Environment=VIRTPROXYD_ARGS=""
+
+    - name: Reload systemctl and start/enable virtproxyd.service
+      ansible.builtin.systemd_service:
+        daemon_reload: true
+        name: virtproxyd.service
+        state: restarted
+        enabled: true
+
 - name: Get libvirt group users
   ansible.builtin.getent:
     database: group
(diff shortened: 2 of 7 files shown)


ci_framework/roles/libvirt_manager/tasks/create_vms.yml
```diff
@@ -178,7 +194,11 @@
 - name: "Inject network configuration file on {{ vm_type }}"
   ansible.builtin.template:
     dest: "{{ cifmw_reproducer_basedir }}/reproducer-network-env/{{ vm_type }}-{{ vm_id }}.yml"
-    src: "network-data.yml.j2"
+    src: >-
+      {{
+        (ci_job_networking is defined)
+        | ternary('ci-network-data.yml.j2', 'network-data.yml.j2')
```
pablintino, line 200: What's the purpose of having two? Just asking, not a comment for a change

ci_framework/roles/libvirt_manager/tasks/virsh_checks.yml
```diff
@@ -21,6 +21,34 @@
     state: started
     enabled: true
 
+- name: Hack in the virtproxyd service
```
pablintino, line 24: A hack usually needs some deeper explanations for our future self

ci_framework/roles/libvirt_manager/tasks/create_vms.yml
cjeanner: I kept both of them because it was "faster" at that time - we have to consume different content, but indeed, we may consider merging them into one again - though we may want to get some comments in order to properly display the different usages.

ci_framework/roles/libvirt_manager/tasks/virsh_checks.yml
cjeanner: there is the commit message? I may add a comment (same as I wrote in the commit message)

cjeanner:
/hold

I'd like to get @psathyan views.

psathyan (approved):
/lgtm

cjeanner:
/unhold

Thanks you @psathyan :)

ci_framework/roles/libvirt_manager/templates/ci-network-data.yml.j2
```diff
@@ -0,0 +1,50 @@
...
+the CI job configuration passed down to zuul so that we can ensure we're
+using the same networks.
+#}
+{% set ns = namespace(
+    ip4='',
+    gw4='',
+    iface=cifmw_reproducer_private_nic,
+    config_nm=false, ip_suffix=0)
+-%}
+{% if vm_type == 'controller' -%}
```
dasm, line 12: Reading this config is very difficult. Some values are duplicated, other are overriden.
Can you rewrite it into `if`/`elif`/`else` containing full set of values?
Something like:
```
{% if vm_type == 'controller' -%}
{% set ns.ip4 = ... %}
{% set ns.gw4 = ... %}
{% set ns.iface = cifmw_reproducer %}
{% set ns.config_nm = false %}
{% set ns.ip_suffix = 0 %}
{% elif == ... %}
{% set ns.ip4 = ... %}
{% set ns.gw4 = ... %}
{% set ns.iface = cifmw_reproducer %}
{% set ns.config_nm = false %}
{% set ns.ip_suffix = 0 %}
{% else %}
{% set ns.ip4 = ... %}
{% set ns.gw4 = ... %}
{% set ns.iface = cifmw_reproducer %}
{% set ns.config_nm = false %}
{% set ns.ip_suffix = 0 %}
{% endif %}
```

Or _at least_ keep order of values the same for all conditions?
It's very hard to track all variables between different places.

ci_framework/roles/libvirt_manager/templates/ci-network-data.yml.j2
```diff
@@ -0,0 +1,50 @@
...
+{% set vlan_mtu = default_mtu - 4 -%}
+crc_ci_bootstrap_networks_out:
+  {{ vm_type }}-{{ vm_id }}:
+    default:
+      iface: {{ ns.iface }}
+      connection: ci-private-network
+      ip4: {{ ns.ip4 }}/24
+      gw4: {{ ns.gw4 }}
+      dns4: {{ ci_job_networking['instances']['crc']['networks']['default']['ip'] }}
+      mtu: {{ ci_job_networking.networks['default'].mtu | default(default_mtu) }}
```
dasm, line 36: You are repeating [line](https://github.com/openstack-k8s-operators/ci-framework/pull/720/files#diff-502d2f5bb49aa7163735ee7df5f5fde39b85c21f1d5f4cc562970b338692cc9dR26) here.

ci_framework/roles/libvirt_manager/templates/ci-network-data.yml.j2
```diff
@@ -0,0 +1,50 @@
...
+{%   set ns.iface = cifmw_reproducer_crc_private_nic -%}
+{%   set ns.gw4 = cifmw_reproducer_crc_gw4 -%}
+{%   set ns.config_nm = true -%}
+{%   set ns.ip4 =  ci_job_networking['instances']['crc']['networks']['default']['ip'] %}
+{% else -%}
+{%   set ns.ip_suffix = 100+vm_id -%}
+{%   set ns.ip4 = ci_job_networking['instances']['compute-' ~ vm_id]['networks']['default']['ip']%}
+{%   set ns.gw4 = cifmw_reproducer_ctl_gw4 -%}
+{% endif -%}
+{% set default_mtu = ci_job_networking.networks['default'].mtu | default('1500') -%}
```
dasm, line 26: Can you move all `set` unrelated to VM config at the top of the file?

ci_framework/roles/libvirt_manager/tasks/virsh_checks.yml
```diff
@@ -34,3 +68,19 @@
...
+- name: Allow QEMU on home directory for the storage access
+  become: true
+  ansible.posix.acl:
+    path: "{{ ansible_user_dir }}"
+    entity: "qemu"
+    etype: "user"
+    permissions: rx
+    state: present
+
+- name: Inject system connection parameters in bashrc
```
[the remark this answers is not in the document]
cjeanner: the most important part in the doc is that actually:
```
If virsh finds the environment variable VIRSH_DEFAULT_CONNECT_URI set, it will try this URI by default. Use of this environment variable is, however, deprecated now that libvirt supports LIBVIRT_DEFAULT_URI itself.
```
VIRSH one is deprecated. I'll remove it.

ci_framework/roles/libvirt_manager/templates/ci-network-data.yml.j2
cjeanner: some values don't need to be overridden - so I won't repeat them. Re-ordering, done.

ci_framework/roles/libvirt_manager/templates/ci-network-data.yml.j2
cjeanner: done

dasm:
Thanks @cjeanner! Nice job.

PR #659, opened 2023-10-12
WIP: Improvements to devscripts role

psathyan:
Below are some of the optimizations carried out

- Supports Hybrid test environment.
- No longer supports creating of user instead opts to use existing executor.
- Improved way of inserting the default ipv4 address
- Integrated with `ci_nmstate`
-- Support attaching of extra networks to worker nodes
-- Support for extending virtual network into the physical layer
- Extra networks managed outside of dev-scripts
- Firewall configuration for all extra networks
- Forwarding rule for all extra networks.
- The default number of extra worker nodes created is set to `0`

As a pull request owner and reviewers, I checked that:
- [x] Appropriate testing is done and actually running
- [x] Appropriate documentation exists and/or is up-to-date:
  - [x] README in the role
  - [x] Content of the docs/source is reflecting the changes

PR #721, opened 2023-10-24
Ensure we clean ssh key from the authorized_keys

cjeanner:
Until now, in case we were iterating deploy and cleanup over and over,
with the removal of the `cifmw_devscripts_basedir`, we ended up with an
ever-growing authorized_keys.

This patch ensures, during the cleanup tasks, that we remove the
generated key. It won't have impact on the deployed nodes, since, at
this point, there won't be any node anymore.

### DIFF
--- a/ci_framework/roles/devscripts/tasks/cleanup.yml
+++ b/ci_framework/roles/devscripts/tasks/cleanup.yml
@@ -55,3 +55,24 @@
   ansible.builtin.yum_repository:
     name: "crb"
     state: absent
+
+- name: Remove generated keypair and related accesses
+  block:
+    - name: Get public key
+      register: _pub_key
+      ansible.builtin.slurp:
+        path: >-
+          {{ cifmw_devscripts_artifacts_dir }}/{{ cifmw_devscripts_user }}_ed25519.pub
+
+    - name: Remove key from the authorized_keys
+      ansible.posix.authorized_key:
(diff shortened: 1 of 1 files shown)


pablintino:
@cjeanner in the PR you say `Correct libvirt_manager` but I don't see that, is missing? Or it's in another PR?

raukadah (approved):
/approve

cjeanner:
> @cjeanner in the PR you say `Correct libvirt_manager` but I don't see that, is missing? Or it's in another PR?

It messed up the patches yesterday and pushed two instead of one - updated title and initial comment.

ci_framework/roles/devscripts/tasks/cleanup.yml
```diff
@@ -55,3 +55,24 @@
...
+    - name: Get public key
+      register: _pub_key
+      ansible.builtin.slurp:
+        path: >-
+          {{ cifmw_devscripts_artifacts_dir }}/{{ cifmw_devscripts_user }}_ed25519.pub
+
+    - name: Remove key from the authorized_keys
+      ansible.posix.authorized_key:
+        user: "{{ cifmw_devscripts_user }}"
+        key: "{{ _pub_key['content'] | b64decode }}"
```
psathyan, line 70: Based on #659 ,  we would not need these tasks.

cjeanner:
This change will be integrated in @psathyan #659 patch.

PR #723, opened 2023-10-24
Improve reproducer and reproduer-clean playbooks

cjeanner:
Since we're introducing devscripts in the reproducer, we have to ensure
we're able to clean it up.

We also ensure some basic setup steps are done on the target host.


dasm:
Typo in the title: "reproduer-clean"
/lgtm

cjeanner:
recheck

PR #575, opened 2023-09-19
Implement sosreport role for log collection

dasm:
Replace `ci_script` action with a "sos" log collector. By using industry-provided standard, we ensure that collected logs adhere to standards.

As a pull request owner and reviewers, I checked that:
- [x] Appropriate testing is done and actually running
- [x] Appropriate documentation exists and/or is up-to-date:
  - [x] README in the role
  - [x] Content of the docs/source is reflecting the changes

PR #724, opened 2023-10-24
[artifacts]Gather the whole /etc

gibizer:
There are many places in /etc the EDPM ansible roles are modifying like tuned config or ssh_known_hosts file. To be able to troubleshoot such configuration in CI runs this PR switch to gather the whole /etc directory from the EDPM nodes.

### DIFF
--- a/ci_framework/roles/artifacts/tasks/edpm.yml
+++ b/ci_framework/roles/artifacts/tasks/edpm.yml
@@ -77,8 +77,7 @@
           sudo test -d /var/lib/openstack && sudo cp -a /var/lib/openstack /tmp/{{ host_ip }}
           sudo test -d /var/lib/config-data && sudo cp -a /var/lib/config-data /tmp/{{ host_ip }}
           sudo test -d /var/lib/cloud && sudo cp -a /var/lib/cloud /tmp/{{ host_ip }}
-          sudo test -d /etc/nftables && sudo cp -a /etc/nftables /tmp/{{ host_ip }}
-          sudo test -d /etc/os-net-config && sudo cp -a /etc/os-net-config /tmp/{{ host_ip }}
+          sudo test -d /etc && sudo cp -RP /etc /tmp/{{ host_ip }}
           sudo find /tmp/{{ host_ip }} -type d -exec chmod ugoa+rx '{}' \;
           sudo find /tmp/{{ host_ip }} -type f -exec chmod ugoa+r '{}' \;
           command -v ovs-vsctl && sudo ovs-vsctl list Open_vSwitch > /tmp/{{ host_ip }}/ovs_vsctl_list_openvswitch.txt


ci_framework/roles/artifacts/tasks/edpm.yml
```diff
@@ -77,8 +77,7 @@
           sudo test -d /var/lib/openstack && sudo cp -a /var/lib/openstack /tmp/{{ host_ip }}
           sudo test -d /var/lib/config-data && sudo cp -a /var/lib/config-data /tmp/{{ host_ip }}
           sudo test -d /var/lib/cloud && sudo cp -a /var/lib/cloud /tmp/{{ host_ip }}
-          sudo test -d /etc/nftables && sudo cp -a /etc/nftables /tmp/{{ host_ip }}
-          sudo test -d /etc/os-net-config && sudo cp -a /etc/os-net-config /tmp/{{ host_ip }}
+          sudo test -d /etc && sudo cp -a /etc /tmp/{{ host_ip }}
```
cjeanner, line 80: I see potential issues with the recursion depth, some directories there have symlink to other places and may have a really deep recursion. Such recursions are usually really bad when it comes to logserver sync, unfortunately.

Would it be possible to get a listing of directories of interest?

gibizer: I know what I need today, the /etc/ssh/ssh_know_hosts file, but I cannot predict what I will need in my next CI troubleshooting. I agree about the recursion issue though. What about `cp -RP /etc` ? That would follow non of the symlink

ci_framework/roles/artifacts/tasks/edpm.yml
```diff
@@ -77,8 +77,7 @@
           sudo test -d /var/lib/openstack && sudo cp -a /var/lib/openstack /tmp/{{ host_ip }}
           sudo test -d /var/lib/config-data && sudo cp -a /var/lib/config-data /tmp/{{ host_ip }}
           sudo test -d /var/lib/cloud && sudo cp -a /var/lib/cloud /tmp/{{ host_ip }}
-          sudo test -d /etc/nftables && sudo cp -a /etc/nftables /tmp/{{ host_ip }}
-          sudo test -d /etc/os-net-config && sudo cp -a /etc/os-net-config /tmp/{{ host_ip }}
+          sudo test -d /etc && sudo cp -RP /etc /tmp/{{ host_ip }}
```
dasm, line 80: It's very far fetched. Can you be more specific?
There is another PR https://github.com/openstack-k8s-operators/ci-framework/pull/575 which is planning to use `sosreport`.
It would be beneficial to know what the files are you looking for.

gibizer: As I wrote above for @cjeanner:
> I know what I need today, the /etc/ssh/ssh_know_hosts file, but I cannot predict what I will need in my next CI troubleshooting. I agree about the recursion issue though. What about cp -RP /etc ? That would follow none of the symlink

gibizer: Having a full sos report form the EDPM nodes would be nice

cjeanner:
/approve

While I don't really like getting the full of /etc, it seems to be needed. It's also planed to move to sosreport when dust settles a bit, so we may get some even better debugging capabilities.

At least, we won't get the symlink, hopefully this should lower a bit the amount of potentially useless data :).

raukadah (approved):

PR #727, opened 2023-10-25
[set_openstack_containers] filter os containers based on prefix

raukadah:
Openstack operators got the support for disconnected environment. It changes the openstack services container url to use sha instead of tag.

It breaks the set_openstack_containers role as we were filtering the containers based on tags.

With this pr, it removes the filter var and replace it with image prefix and also allow user to use the image prefix in the image name.


viroel:
testprooject results:
https://logserver.rdoproject.org/11/50511/3/check/periodic-podified-multinode-edpm-deployment-antelope-ocp-crc-cs9/cb68e32/controller/ci-framework-data/artifacts/operator_env.txt

viroel:
/lgtm 
thanks Chandan

PR #729, opened 2023-10-25
Enable `oc` completion for the user

cjeanner:

### DIFF
--- a/ci_framework/roles/ci_setup/tasks/packages.yml
+++ b/ci_framework/roles/ci_setup/tasks/packages.yml
@@ -48,3 +48,20 @@
         remote_src: true
         mode: "0755"
         creates: "{{ cifmw_ci_setup_openshift_client_install_path }}/oc"
+
+- name: Inject oc completion in local profile
+  block:
+    - name: Create completion file
+      ansible.builtin.shell:  # noqa: risky-shell-pipe
+        cmd: >-
+          {{ cifmw_ci_setup_openshift_client_install_path }} completion bash |
+          tee -a ~/.oc_completion
+        creates: "{{ ansible_user_dir }}/.oc_completion"
+
+    - name: Source completion from within .bashrc
+      ansible.builtin.blockinfile:
+        path: "{{ ansible_user_dir }}/.bashrc"
+        block: |-
+          if [ -f ~/.oc_completion ]; then
+            source ~/.oc_completion
+          fi
--- a/ci_framework/roles/ci_setup/vars/main.yml
+++ b/ci_framework/roles/ci_setup/vars/main.yml
@@ -19,6 +19,7 @@
 # All variables within this role should have a prefix of "cifmw_ci_setup"
 
 cifmw_ci_setup_packages:
+  - bash-completion
   - make
 
 # openshift client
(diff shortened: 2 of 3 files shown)


ci_framework/roles/ci_setup/tasks/packages.yml
```diff
@@ -48,3 +48,9 @@
         remote_src: true
         mode: "0755"
         creates: "{{ cifmw_ci_setup_openshift_client_install_path }}/oc"
+
+- name: Inject oc completion in local profile
+  ansible.builtin.shell:  # noqa: risky-shell-pipe
+    cmd: >-
+      {{ cifmw_ci_setup_openshift_client_install_path }} completion bash |
```
dasm, line 55: This one looks solid, however there is one concern.
What's gonna happen if an user is going to run `reproducer` on the localhost (or remote host for that matter) multiple times? Is it a part of "non-ci workflow"?

dasm: This is being executed [here](https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci_framework/playbooks/01-bootstrap.yml#L40-L44)
The reproducer uses `bootstrap` and _I think_ it might be used many times when trying to reproduce  jobs [description here](https://github.com/openstack-k8s-operators/ci-framework/blob/main/docs/source/reproducers/03-zuul.md?plain=1#L180-L183)

cjeanner: we may generate the completion in another file, and ensure with a simple lineinfile in .bash_profile that we do source it. shouldn't be too hard.

That said - this one will run twice indeed - once on the hypervisor (because we consume that role) and once on the controller-0. The hypervisor may therefore get a growing profile.
Of course, best thing would be to get `oc` deployed in global location, allowing to get a global completion in /etc/profile.d :).

cjeanner:
recheck

cjeanner:
recheck

ci_framework/roles/ci_setup/tasks/packages.yml
```diff
@@ -48,3 +48,20 @@
...
         creates: "{{ cifmw_ci_setup_openshift_client_install_path }}/oc"
+
+- name: Inject oc completion in local profile
+  block:
+    - name: Create completion file
+      ansible.builtin.shell:  # noqa: risky-shell-pipe
+        cmd: >-
+          {{ cifmw_ci_setup_openshift_client_install_path }} completion bash |
+          tee -a ~/.oc_completion
+        creates: "{{ ansible_user_dir }}/.oc_completion"
```
dasm, line 59: Nice!
>  creates path | A filename, when it already exists, this step will not be run.
[src](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/shell_module.html#parameter-creates)

There is still a risk of running outdated `oc completion` with newer `oc`.
What do you think?

dasm (commented):
Minor thing, but overall LGTM

/lgtm

ci_framework/roles/ci_setup/tasks/packages.yml
cjeanner: this completion is done on the controller-0 - so it's a temporary host. If ppl update their openshift, they will redeploy it. And if really, they can regenerate the completion.

dasm:
recheck

/lgtm
/approve

PR #731, opened 2023-10-26
Fix include_role task

dasm:
include_role uses "name: [name of role]" as a way to import role.


cjeanner:
/approve

note: both work, the `role` param was probably exchange without proper deprecation, and is still valid.

ci_framework/roles/edpm_prepare/tasks/main.yml
```diff
@@ -250,7 +250,7 @@
       vars:
         cifmw_install_ca_bundle_src: "{{ cifmw_edpm_prepare_basedir }}/tls-ca-bundle.pem"
       ansible.builtin.include_role:
-        role: install_ca
```
cjeanner, line 253: note that this will probably conflict with #737

PR #732, opened 2023-10-27
Fix: DNS forwarders condition is always resulting in true

psathyan:
In this patch, the checks done for checking whether to update the local DNS forwarders is corrected. The fix enables us to update the forwarder records only when required.


dasm:
recheck

/lgtm
/approve

PR #734, opened 2023-10-27
Support reserving of IP address

psathyan:
This PR adds support of reserving IP addresses against a virtual machines MAC address. It enables the system to have a static IP address.

This feature is supported for those virtual networks managed by libvirt and has DHCP enabled.

### DIFF
--- a/ci_framework/roles/libvirt_manager/tasks/attach_interface.yml
+++ b/ci_framework/roles/libvirt_manager/tasks/attach_interface.yml
@@ -14,21 +14,39 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
-
 # This task requires
 #   vm_name         Domain name to which the interface needs to be attached.
-#   interface_name  source of the device
-#   interface_type  bridge or network
+#   network
+#     name          Name of the network bridge
+#     cidr          network notation
+#     mtu           MTU applied on the interface.
+#     static_ip     boolean if static address is required.
+
+
+- name: Generate a random MAC address
+  ansible.builtin.set_fact:
+    _lm_mac_address: "{{ '0A:02' | community.general.random_mac }}"
+
+- name: Reserve an IP address
+  when: network.static_ip | bool
+  vars:
+    mac_address: "{{ _lm_mac_address }}"
+  ansible.builtin.import_tasks: reserve_ip.yml
 
 - name: Attach interface to the virtual machine.
-  ci_script:
-    output_dir: "{{ cifmw_libvirt_manager_basedir }}/artifacts"
-    script: >-
+  vars:
+    _net_name: >
+      {{
+        (network.static_ip | bool) |
+        ternary(network.name, 'cifmw-net_' + network.name)
+      }}
+  ansible.builtin.command:
+    cmd: >-
       virsh -c qemu:///system
       attach-interface "{{ vm_name }}"
-      --source "{{ interface_name }}"
-      --type "{{ interface_type }}"
-      --mac "{{ '0A:02' | community.general.random_mac }}"
+      --source "{{ _net_name }}"
+      --type bridge
+      --mac "{{ _lm_mac_address }}"
       --model virtio
       --config
       --persistent
--- a/ci_framework/roles/libvirt_manager/tasks/reserve_ip.yml
+++ b/ci_framework/roles/libvirt_manager/tasks/reserve_ip.yml
@@ -1,0 +1,53 @@
+---
+# Copyright Red Hat, Inc.
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+# The below tasks reserves an IP from the dhcp range and assigns it to the
+# specified mac address and vm
+
+# The below information are required
+#   vm_name       Name of the virtual machine/domain
+#   mac_address   Associated MAC address
+#   network
+#     name        virtual network name
+#     cidr        network notation
+
+# The nth host is determined by looking to the vm_name
+#   master          20
+#   worker          23
+#   extraworker     26
+#   compute         26      # Has the same index of extraworker
+
+
+- name: Generate the IP address
+  vars:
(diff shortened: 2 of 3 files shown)


ci_framework/roles/libvirt_manager/tasks/reserve_ip.yml
```diff
@@ -0,0 +1,51 @@
...
+  vars:
+    _vm_name: "{{ vm_name | replace('-', '_') }}"
+    _split_vm_name: "{{ vm_replace | split('_') }}"
+    _ip_index: >-
+      {{
+        cifmw_libvirt_manager_vm_net_ip_set[_split_vm_name[1]] +
+        _split_vm_name[2] | int
+      }}
+  ansible.builtin.set_fact:
+    _lm_ip_address: "{{ network.cidr | ansible.utils.nthhost(_ip_index) }}"
```
cjeanner, line 44: never used that filter - what's the kind of IP it will output then? i.e. what's the compute-0 IP?

ci_framework/roles/libvirt_manager/tasks/attach_interface.yml
```diff
@@ -14,21 +14,41 @@
...
+  ansible.builtin.import_tasks: reserve_ip.yml
 
 - name: Attach interface to the virtual machine.
-  ci_script:
-    output_dir: "{{ cifmw_libvirt_manager_basedir }}/artifacts"
-    script: >-
+  vars:
+    _net_name: >
+      {{
+        'cifmw-net_' + network.name
```
cjeanner, line 40: networks as managed by libvirt_manager use `cifmw-` prefix, not `cifmw-net_`

psathyan: I had switch to using `cifmw-net_` for those virtual networks that would be created from Hybrid side. This line tends to replace the name of bridge https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci_framework/roles/libvirt_manager/tasks/deploy_layout.yml#L24 

Hence the new prefix. When the network name comes from the reproducer I believe it would automatically have it

In the case of hybrid, there is no DHCP supported in the libvirt network.

ci_framework/roles/libvirt_manager/tasks/reserve_ip.yml
```diff
@@ -0,0 +1,51 @@
...
+#   master          20
+#   worker          23
+#   extraworker     26
+#   compute         26      # Has the same index of extraworker
+
+
+- name: Generate the IP address
+  vars:
+    _vm_name: "{{ vm_name | replace('-', '_') }}"
+    _split_vm_name: "{{ vm_replace | split('_') }}"
```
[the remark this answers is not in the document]
psathyan: Thank you. I am still rusty using ansible filters. Will have the change incorporated.

ci_framework/roles/libvirt_manager/tasks/reserve_ip.yml
psathyan: This filter results the resulting IP address in that index

For example
network 192.168.10.0/24

nthhost(2) -> 192.168.10.2
nthhost(220) -> 192.168.10.220

nthhost(300) # would result in failure.

[Reference](https://docs.ansible.com/ansible/latest/collections/ansible/utils/nthhost_filter.html)

psathyan: compute-0 IP would be `26 + 0` i.e. 192.168.122.26

cjeanner: hmmm so it should be 100 - because computes are usually 100, 101, 102, ... :)

ci_framework/roles/libvirt_manager/tasks/attach_interface.yml
cjeanner: oh, ok, so that's NOT related to the layout created by the libvirt_manager. OK, I get it.

ci_framework/roles/libvirt_manager/tasks/reserve_ip.yml
psathyan: So we should have 
`controller: 10`
`worker: 13`
`compute: `00`

In the latest commit.

ci_framework/roles/libvirt_manager/tasks/attach_interface.yml
```diff
@@ -14,21 +14,39 @@
...
+#   network
+#     name          Name of the network bridge
+#     cidr          network notation
+#     mtu           MTU applied on the interface.
+#     static_ip     boolean if static address is required.
+
+
+- name: Generate a random MAC address
+  ansible.builtin.set_fact:
+    _lm_mac_address: "{{ '0A:02' | community.general.random_mac }}"
```
marios, line 28: is OA:02 a special OUI or did you just "choose" something. 
I would have expected it to be 3 octets though (eg OA:02:01)

marios (commented):
lgtm but question about the generated mac

ci_framework/roles/libvirt_manager/tasks/reserve_ip.yml
```diff
@@ -0,0 +1,53 @@
...
+    _lm_ip_address: "{{ network.cidr | ansible.utils.nthhost(_ip_index) }}"
+
+- name: Reserve static IP address for the node.
+  community.libvirt.virt_net:
+    command: modify
+    name: "{{ network.name }}"
+    uri: "qemu:///system"
+    xml: |
+      <host mac='{{ mac_address }}' name='{{ vm_name }}' ip='{{ _lm_ip_address }}'>
+        <lease expiry='60' unit='minutes'/>
```
cjeanner, line 52: the lease part isn't mandatory

ci_framework/roles/libvirt_manager/tasks/attach_interface.yml
```diff
@@ -14,21 +14,39 @@
...
+
+- name: Reserve an IP address
+  when: network.static_ip | bool
```
cjeanner, line 31: ah, nice catch. Will see to support that one as well. I thought about it, then just went for "meh, let's fix all IPs".

The fact I'm doing the reverse from within libvirt_manager (i.e. create the VM with associated NICs, then get the NICs, then fix IP, ...) is making things a liiiitle bit more complicated, but I'll figure out how to filter and pass the right data to that tasks file.

cjeanner:
/approve

I'll probably have to change a bit some things in there, in order to make it easier to reserve/fix IPs from another part of the code tree, but it should be easy enough.

Thank you @psathyan for that proposal!

ci_framework/roles/libvirt_manager/tasks/attach_interface.yml
psathyan: `xA` could enable us identify that it is a private MAC address range. Did

https://www.oasys.net/fragments/identify-private-macs/

PR #735, opened 2023-10-27
Enable patching of virtproxyd optional

psathyan:
As part of this patch, a new flag is being added so that users can opt out of applying the virtproxyd patch. This is introduced as in some cases a system reboot is required after applying the patch.

  - [x] README in the role


dasm:
recheck

/lgtm
/approve

PR #736, opened 2023-10-27
Move create network set of tasks out

psathyan:
In this PR, we are moving the tasks related to virtual network creation to it's own module. This helps in reusing the new module by other roles.


dasm:
recheck

/lgtm
/approve

PR #739, opened 2023-10-27
Fix kustomization input encoding

pablintino:
In certain situations, and only for plain list of dicts, Ansible passes the kustomization argument to the action plugin as a dict that may contain AnsibleUnicode for string fields that may end dumped into the final kustomization file wrongly, as there string representation is not it's string literal value. Safely enconding those dicts using AnsibleDumper to dump them to a safe yaml string and back.


cjeanner:
/approve
/lgtm

ahhh, encoding. Old daemons from older times biting back

PR #741, opened 2023-10-28
[devscripts] Minor improvements that fix doc and support optionally applying patch

psathyan:
As part of this PR, there are two changes

- minor doc fixes
- support applying of patch optional.
- switching the default OpenShift version to `4.14.0`

  - [x] README in the role

### DIFF
--- a/ci_framework/roles/devscripts/molecule/default/converge.yml
+++ b/ci_framework/roles/devscripts/molecule/default/converge.yml
@@ -19,7 +19,7 @@
   hosts: all
 
   vars:
-    ansible_user_dir: "{{ lookup('env', 'HOME') }}"
+    ansible_user_dir: "{{ ansible_env.HOME }}"
     cifmw_path: "{{ ansible_user_dir }}/.crc/bin:{{ ansible_user_dir }}/.crc/bin/oc:{{ ansible_user_dir }}/bin:{{ ansible_env.PATH }}"
     cifmw_use_devscripts: true
     cifmw_network_local_dns:
@@ -29,7 +29,7 @@
     cifmw_devscripts_ci_token: "random value"
     cifmw_devscripts_pull_secret: "should be a json"
     cifmw_devscripts_ocp_version: "4.13.12"
-    cifmw_devscripts_user: "{{ ansible_user | default(lookup('env', 'USER')) }}"
+    cifmw_devscripts_user: "{{ ansible_user_id }}"
     cifmw_devscripts_repo_dir: "{{ (ansible_user_dir, 'src', 'dev-scripts') | path_join }}"
     cifmw_devscripts_data_dir: "{{ (ansible_user_dir, 'ci-framework-data') | path_join }}"
     cifmw_devscripts_artifacts_dir: "{{ (cifmw_devscripts_data_dir, 'devscripts', 'artifacts') | path_join }}"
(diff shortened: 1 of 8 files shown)


ci_framework/roles/devscripts/tasks/sub_tasks/16_user.yml
```diff
@@ -14,11 +14,9 @@
...
 # under the License.
 
-- name: Create the specified user.
-  become: true
-  ansible.builtin.user:
-    name: "{{ cifmw_devscripts_user }}"
-    state: present
+- name: Gather the user
+  ansible.builtin.set_fact:
+    cifmw_devscripts_user: "{{ ansible_user_id }}"
```
pablintino, line 19: Do we really need this fact? We can use ansible_user_id straightforward

cjeanner: there's also a non-null risk of ppl overriding this via a parameter (while using an "old" parameter file for instance). Might be better to drop it for good.

ci_framework/roles/devscripts/molecule/default/converge.yml
```diff
@@ -19,7 +19,7 @@
   hosts: all
 
   vars:
-    ansible_user_dir: "{{ lookup('env', 'HOME') }}"
+    ansible_user_dir: "{{ ansible_env.HOME }}"
```
dasm, line 22: Is it being moved from controller over to target?

[docs](https://docs.ansible.com/ansible/latest/reference_appendices/faq.html#how-do-i-access-shell-environment-variables)

ci_framework/roles/devscripts/vars/main.yml
```diff
@@ -47,7 +47,7 @@ cifmw_devscripts_config_defaults:
   master_disk: 100
   master_vcpu: 10
   num_workers: 0
-  num_extra_workers: 3
+  num_extra_workers: 0
```
dasm, line 50: Can you remind why this change?

psathyan: The idea is to use `libvirt_manager` for creating the OSP computes and use `devscripts` for deploying only the OCP control plane. 

Setting this value to `0` prevents `dev-scripts` from creating additional nodes that could be used for compute. Hence changing the default behavior.

ci_framework/roles/devscripts/molecule/default/converge.yml
psathyan: Initial design was to run this on the controller however it has changed for `reproducer` hence replacing all `lookup` to enable supporting of remote execution also.

cjeanner:
/approve
allowing the execution against a remote node has been in the plans for some time already - so yep, replacing lookup() is a nice move, thank you!
Improving doc is also good :).

raukadah (approved):

ci_framework/roles/devscripts/molecule/default/converge.yml
dasm: ack, thx

ci_framework/roles/devscripts/vars/main.yml
dasm: thanks for the explanation

PR #744, opened 2023-10-28
Fix kube:admin for OAuth logins

pablintino:
In OCP kube:admin user, the one in the cluster, is mapped to the OAuth enabled user kubeadmin. whoami, the way we use to return the user always returns kube:admin for that user, but that username cannot be used to login using OAuth. This behaviour is hardcoded inside OCP itself.


raukadah:
recheck post_failure during delete network

raukadah (approved):
/approve

PR #746, opened 2023-10-31
Add some tags in the devscripts role

cjeanner:
This should allow people to re-start a failed run without needing to
re-run the whole role.

   [X] README in the role


raukadah (commented):
/approve

PR #747, opened 2023-10-31
Remove Devscript repo dir only if it exists

raukadah:
It will help to avoid these errors:
```
No rule to make target 'clean'.
```


psathyan (approved):
/lgtm

PR #748, opened 2023-10-31
Ensure we don't add network port multiple times

cjeanner:
Until now, we were at risk of adding the same port multiple times in
case we were re-running the playbook.

We now check the attached ports and, if we see one matching the network
we want to attach to the domain, we just don't re-add it.


dasm:
/approve

@psathyan can you check and LGTM it if you agree?
Thanks!

psathyan (commented):
/lgtm

PR #749, opened 2023-10-31
Allow to reserve/fix IPs for VM managed by libvirt_manager

cjeanner:
There are some assumptions we need to properly work, especially
regarding network name, since anything managed by this role will get a
fixed `cifmw-` prefix.


dasm:
/approve

@psathyan if you could review this one, that would be great! Thanks!

psathyan (commented):
/lgtm

PR #752, opened 2023-11-01
Support hybrid test environment

psathyan:
This PR adds support for using a virtual OCP environment and having OpenStack computes on physical environments. The changes that are done here

- Supports extra networks created by `ci_nmstate`
- Performances all the necessary configurations to enable using of the extra networks
- Enables of reading of physical nodes


ci_framework/roles/devscripts/tasks/cleanup.yml
```diff
@@ -54,7 +54,11 @@
 - name: Cleanup the devscripts working directory.
   become: true
   ansible.builtin.file:
-    path: "{{ cifmw_devscripts_config.working_dir }}"
+    path: >-
+      {{
+        cifmw_devscripts_config.working_dir |
+        default('/home/dev-scripts')
```
cjeanner, line 60: I thought it was set in the defaults already?

psathyan: Yes, it is configured as `defaults`... thanks looks like I was cautious

PR #755, opened 2023-11-01
Disable `host_key_checking` in ansible.cfg

rebtoor:
  - [x] README in the role


rebtoor:
as agreed with @cjeanner , it's better to not disable that behaviour by default.

PR #756, opened 2023-11-01
Correct Layout reproducer

cjeanner:
When the CI Job reproducer was introduced, we missed the need of a new
parameter in order to inject the actual network layout we're consuming
in a non-CI job case.

This patch also allows to bootstrap CRC directly during the reproducer
run, provided we define it in the layout (it's usually the case).

As a follow-up, we may want to properly manage pull-secret copy from the
host running the playbook onto the hypervisor, in case those are two
distinct nodes. Since this is a feature shared with devscripts, we may
want to get a common role for that.

We also update the doc to reflect this change, since we don't need to
have a CRC pre-provisionned anymore. You still may have it ready, but
it's not mandatory anymore.

### DIFF
--- a/ci_framework/roles/reproducer/tasks/main.yml
+++ b/ci_framework/roles/reproducer/tasks/main.yml
@@ -36,6 +36,35 @@
     - cifmw_job_uri is defined
   ansible.builtin.import_tasks: ci_data.yml
 
+- name: Deploy CRC if needed
+  when:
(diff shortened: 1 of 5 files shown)


cjeanner:
/hold

made a mistake - missing the `uri: qemu:///system` for virt commands.

cjeanner:
recheck

raukadah (approved):
/approve

PR #762, opened 2023-11-02
[ci_script] Create logs directory if missing

karelyatin:
Since the directory is relative to the output_dir
which is ensured to exist, should be fine to create the logs directory.

Faced issue with missing logs directory while running Token fetch task during the tempest run.

  - [ ] README in the role

### DIFF
--- a/ci_framework/plugins/action/ci_script.py
+++ b/ci_framework/plugins/action/ci_script.py
@@ -145,7 +145,7 @@ class ActionModule(ActionBase):
 
         logs_dir = output_dir.parent.joinpath("logs")
         if not logs_dir.is_dir():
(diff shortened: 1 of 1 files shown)


cjeanner:
Might be OK - but it would be good to get the directory created by ansible: it would ensure we don't have wild things going everywhere on the filesystem imho.
Also, ci_script being an action, it's executed locally - so it will create things on the host from where the ansible-playbook is launched, not the final destination. This is part of the "properly support remote execution" topic we have in the framework.

cjeanner:
Closing this one:
- directory should be remotely created
- a recent correction to ci_script prevents its crash when running against a remote node

PR #765, opened 2023-11-02
[cephadm] Switch HA containers if a custom container registry is used

rebtoor:
- fix typo in a task name

  - [x] README in the role

### DIFF
--- a/ci_framework/roles/cifmw_cephadm/tasks/bootstrap.yml
+++ b/ci_framework/roles/cifmw_cephadm/tasks/bootstrap.yml
@@ -32,7 +32,7 @@
   loop: "{{ cifmw_cephadm_bootstrap_files_stat.results | list }}"
   when: not item.stat.exists
 
-- name: Stat pre ceph conf file in case we should bootrap with it
+- name: Stat pre ceph conf file in case we should bootstrap with it
   ansible.builtin.stat:
     path: "{{ cifmw_cephadm_assimilate_conf }}"
   register: cifmw_cephadm_assimilate_conf_stat
--- a/ci_framework/roles/cifmw_cephadm/tasks/cephadm_config_set.yml
+++ b/ci_framework/roles/cifmw_cephadm/tasks/cephadm_config_set.yml
@@ -47,6 +47,25 @@
     {{ cifmw_cephadm_container_ns }}/{{ cifmw_cephadm_container_image }}
   changed_when: false
 
+- name: Set haproxy container image in ceph configuration
+  when:
+    - cifmw_cephadm_haproxy_container_image is defined
+    - not (cifmw_cephadm_default_container | default(false) | bool)
+  become: true
+  ansible.builtin.command: |
+    {{ cifmw_cephadm_ceph_cli }} config set mgr mgr/cephadm/container_image_haproxy \
+    {{ cifmw_cephadm_haproxy_container_image }}
+  changed_when: false
+
+- name: Set keepalived container image in ceph configuration
+  when:
+    - cifmw_cephadm_keepalived_container_image is defined
+    - not (cifmw_cephadm_default_container | default(false) | bool)
+  ansible.builtin.command: |
+    {{ cifmw_cephadm_ceph_cli }} config set mgr mgr/cephadm/container_image_keepalived \
+    {{ cifmw_cephadm_keepalived_container_image }}
+  changed_when: false
+
 - name: Set dashboard container image in ceph mgr configuration
   when:
     - cifmw_cephadm_dashboard_enabled | bool


ci_framework/roles/cifmw_cephadm/tasks/bootstrap.yml
```diff
@@ -98,6 +98,18 @@
...
+      become: true
+      when:
+        - cifmw_cephadm_registry_url | length > 0
+        - cifmw_cephadm_container_ns != "quay.io/ceph"
+      ansible.builtin.shell: |
+        {{ cifmw_cephadm_ceph_cli }} config set mgr mgr \
+          mgr/cephadm/container_image_keepalived {{ cifmw_cephadm_container_ns + '/keepalived:latest' }}
+        {{ cifmw_cephadm_ceph_cli }} config set mgr mgr \
+          mgr/cephadm/container_image_haproxy {{ cifmw_cephadm_container_ns + '/haproxy:latest' }}
+        {{ cifmw_cephadm_ceph_cli }} orch redeploy ingress.rgw.default
```
fultonj, line 111: We should not need redeploy. Maybe you had it because you wanted to update it after it was deployed, but once you get a fresh deployment, the redeploy won't be needed. Would you please delete line 111?

ci_framework/roles/cifmw_cephadm/tasks/bootstrap.yml
```diff
@@ -98,6 +98,18 @@
         {{ cifmw_cephadm_container_ns + '/' + cifmw_cephadm_container_image + ':' + cifmw_cephadm_container_tag|string }}
       changed_when: false
 
+    - name: Switch HA container images if a custom container registry is used
```
fultonj, line 101: The cephadm role already has a tasks file for doing these config sets. Please see:

https://github.com/openstack-k8s-operators/ci-framework/blob/reproducer/va1-layout/ci_framework/roles/cifmw_cephadm/tasks/cephadm_config_set.yml

tripleo-ansible already has code you can borrow in cephadm_config_set.yaml to set `mgr/cephadm/container_image_keepalived` and `mgr/cephadm/container_image_haproxy`.

https://github.com/openstack/tripleo-ansible/blob/master/tripleo_ansible/roles/tripleo_cephadm/tasks/cephadm_config_set.yaml#L87-L104

Could you use something like the above in `ci_framework/roles/cifmw_cephadm/tasks/cephadm_config_set.yml` so that the tasks for config setting is more consistent?

ci_framework/roles/cifmw_cephadm/tasks/bootstrap.yml
```diff
@@ -98,6 +98,18 @@
...
+      become: true
+      when:
+        - cifmw_cephadm_registry_url | length > 0
+        - cifmw_cephadm_container_ns != "quay.io/ceph"
+      ansible.builtin.shell: |
+        {{ cifmw_cephadm_ceph_cli }} config set mgr mgr \
+          mgr/cephadm/container_image_keepalived {{ cifmw_cephadm_container_ns + '/keepalived:latest' }}
+        {{ cifmw_cephadm_ceph_cli }} config set mgr mgr \
+          mgr/cephadm/container_image_haproxy {{ cifmw_cephadm_container_ns + '/haproxy:latest' }}
```
fultonj, line 110: TripleO used `{{ tripleo_cephadm_haproxy_container_image }}` and `{{ tripleo_cephadm_keepalived_container_image }}`. You could do that too (`s/tripleo_/cifmw_/g`) or I'm fine with what you have with `{{ cifmw_cephadm_container_ns + '/haproxy:latest' }}` and `{{ cifmw_cephadm_container_ns + '/keepalived:latest' }}` too.

fultonj (commented):
Thanks Roberto. Would you mind making a few small changes?

raukadah:
@rebtoor Thank you for proposing the patch. I think this change needs to be proposed against ci-framework main branch. During va1 testing, we run ceph playbook standalone. So I think getting it in main would be good.

ci_framework/roles/cifmw_cephadm/tasks/bootstrap.yml
fmount: Yeah, it makes sense to add the task mentioned by John in [cephadm_config_set.yaml](https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci_framework/roles/cifmw_cephadm/tasks/cephadm_config_set.yml): this allows to not have to `redeploy` rgw (or any other component that requires the ingress daemon) because you should have the right images in place before applying the spec via the `orchestrator`.

fmount:
+1 to have this change applied to the main branch and backported where needed.

ci_framework/roles/cifmw_cephadm/tasks/bootstrap.yml
fmount: Right, basically we have three phases during the ceph deployment:
1. bootstrap
2. apply spec
3. deploy additional daemons, pools, keys, etc

During the second phase, [ceph_config_set](https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci_framework/roles/cifmw_cephadm/tasks/apply_spec.yml#L48) is called before applying the spec via the orchestrator, and if we specify the container images (`ceph config set mgr mgr mgr/cephadm/container_image_{haproxy, keepalived} ....`) using that set of tasks, when RGW (phase 3) is deployed, we should be able to get the right container images/tags.

rebtoor:
I'll close this PR and open new one against main branch.

PR #766, opened 2023-11-03
Allow overlay support in devscripts role

cjeanner:
This patch allows to instruct devscripts to stop the cluster right after
it has been created, and so some small tasks such as dumping the domains
XML and so on.

These tasks will allow to describe the OCP cluster via the
libvirt_manager configuration data structure. We'll end up with a "slow"
first deploy, since we're deploying the full OCP, but future iteration
will be really faster since we'll just create an overlay instead of
re-deploying.

Further patches are needed to get full support for that feature.

  - [X] README in the role


ci_framework/roles/devscripts/tasks/04_overlays.yml
```diff
@@ -0,0 +1,57 @@
+---
+- name: Get OCP related nodes
+  register: _current_vms
+  community.libvirt.virt:
+    command: "list_vms"
+    uri: "qemu:///system"
+
+- name: Stop OCP Masters
```
psathyan, line 8: nit: Since we are also looking at workers ... can we state `nodes` instead of `masters`

ci_framework/roles/devscripts/tasks/04_overlays.yml
```diff
@@ -0,0 +1,57 @@
...
+  failed_when: false
+  vars:
+    matcher: "{{ cifmw_devscripts_config.cluster_name }}_(master|worker).*"
+  community.libvirt.virt:
+    command: "destroy"
+    name: "{{ item }}"
+    uri: "qemu:///system"
+  loop: "{{ _current_vms.list_vms | select('match', matcher) }}"
+
+- name: Get OCP Masters XMLs
```
psathyan, line 18: likewise nodes instead of `masters`

ci_framework/roles/devscripts/tasks/04_overlays.yml
```diff
@@ -0,0 +1,57 @@
+---
+- name: Get OCP related nodes
+  register: _current_vms
+  community.libvirt.virt:
+    command: "list_vms"
+    uri: "qemu:///system"
```
psathyan, line 6: I think it is a fair assumption to make that only the virtual OCP nodes are running at this point of time. Hence we could add `state: running`

wdyt?

psathyan: Allows for co-existence... I think it is good to check than assume.

PR #767, opened 2023-11-03
Support custom XML and multiple base image for overlays

cjeanner:
Following #766, we have now to add a feature in libvirt_manager to
support:
- user provided XML (note: those are *remote* - on the hypervisor)
- multiple base image for the same VM family

With these features, we're getting closer to a good and nice support of
overlays for the OCP cluster, allowing faster re-deploy.

  - [X] README in the role


raukadah (commented):
/approve

raukadah (approved):

PR #768, opened 2023-11-03
Avoid warning the user about oc not found

pablintino:
The approach we took for choose if oc needs to be installed or not is to call it, and if it fails, rescue the task and install it. That yields an error that a user is usually interpreting as a real one. We should not pass that error to the user, and ignore it.

At the same time, it seems that under certain circumstances, the fact catching mechanism is capturing the variable that holds where oc should be installed, what's a problem if the previous run was done with a different user, like root, as the extraction of oc will fail because of lack of permission.

  - [x] README in the role

### DIFF
--- a/ci_framework/roles/ci_setup/defaults/main.yml
+++ b/ci_framework/roles/ci_setup/defaults/main.yml
@@ -20,3 +20,4 @@
 
 cifmw_ci_setup_basedir: "{{ cifmw_basedir | default(ansible_user_dir ~ '/ci-framework-data') }}"
 cifmw_ci_setup_openshift_client_version: "stable"
+cifmw_ci_setup_oc_install_path: "{{ ansible_user_dir ~ '/bin' }}"
--- a/ci_framework/roles/ci_setup/tasks/packages.yml
+++ b/ci_framework/roles/ci_setup/tasks/packages.yml
@@ -15,46 +15,62 @@
     name: "{{ cifmw_ci_setup_packages }}"
     state: latest
 
+- name: Gather version of openshift client
+  register: _oc_version
+  environment: >-
+    {{
+      {'PATH': cifmw_path}
+      if cifmw_path is defined else
+      {}
+    }}
+  ansible.builtin.command:
+    cmd: "oc version --client -o yaml"
+  changed_when: false
+  failed_when: false
+
 - name: Install openshift client
   tags:
     - bootstrap
     - packages
+  when: >-
+    _oc_version.rc != 0 or
+    (
+      ((_oc_version.stdout | from_yaml).releaseClientVersion | split('.') | first | int) >=
+      cifmw_ci_setup_openshift_minimum_version
+    )
   block:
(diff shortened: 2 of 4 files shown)


ci_framework/roles/ci_setup/tasks/packages.yml
```diff
@@ -15,46 +15,50 @@
...
+
+  register: _oc_version
+  environment: >-
+    {{
+      {'PATH': cifmw_path}
+      if cifmw_path is defined else
+      {}
+    }}
+  ansible.builtin.command:
+    cmd: "oc version -o yaml"
```
cjeanner, line 28: in fact, we want the client, really, meaning we want:
`oc version --client -o yaml`
Else, it will try to query any known/configured remote server for its APIs versions, and will fail in case of unresponsive server. It will also wait until a timeout hits, making the whole thing unstable and, usually, throwing a wrong answer in such case (i.e. client installed, but no server answers -> crash).

ci_framework/roles/ci_setup/tasks/packages.yml
```diff
@@ -15,46 +15,50 @@
...
+  failed_when: false
+
 - name: Install openshift client
   tags:
     - bootstrap
     - packages
+  when: >-
+     _oc_version.rc != 0 or
+     (
+      _oc_version | int >= cifmw_ci_setup_openshift_minimum_version
```
cjeanner, line 39: hmmm that will most probably fail, the YAML answer is, at least with `--client`:
```YAML
clientVersion:
  buildDate: "2023-06-01T00:58:07Z"
  compiler: gc
  gitCommit: 05d83eff7e17160e679898a2a5cd6019ec252c49
  gitTreeState: clean
  gitVersion: 4.13.0-202305312300.p0.g05d83ef.assembly.stream-05d83ef
  goVersion: go1.19.9
  major: ""
  minor: ""
  platform: linux/amd64
kustomizeVersion: v4.5.7
releaseClientVersion: 4.13.3
```

ci_framework/roles/ci_setup/defaults/main.yml
```diff
@@ -20,3 +20,4 @@
 
 cifmw_ci_setup_basedir: "{{ cifmw_basedir | default(ansible_user_dir ~ '/ci-framework-data') }}"
 cifmw_ci_setup_openshift_client_version: "stable"
+cifmw_ci_setup_oc_install_path: "{{ ansible_user_dir ~ '/bin' }}"
```
cjeanner, line 23: shall we find a way to inject this in the `cifmw_path` ? FTR, this parameter is set in 01_bootstrap.yml playbook (and, sadly, in some scenario files)

cjeanner (changes requested):

ci_framework/roles/ci_setup/defaults/main.yml
pablintino: Should be nice, I'll give it a thought

PR #771, opened 2023-11-06
Install tcib from RPM

raukadah:
tcib is used in periodic line to build openstack services containers.

Periodic line should be instaling the tcib package itself from podified-ci-testing via rpm. But it is installed via github source. The tcib content in podified-ci-testing is different from source (New commit might merged and dlrn have not picked it up.)

So, it diverges the testing and causes unwanted issues.

This pr adds the support for installing the tcib content from rpm. It will help to install the correct content.

https://issues.redhat.com/browse/OSPCIX-101

### DIFF
--- a/ci/playbooks/tcib/tcib.yml
+++ b/ci/playbooks/tcib/tcib.yml
@@ -15,6 +15,15 @@
   hosts: "{{ cifmw_target_host | default('localhost') }}"
   gather_facts: true
   tasks:
+    - name: Copy repo-setup generated repos to yum.repos.d directory
+      tags:
+        - container_img_build
+      become: true
+      ansible.builtin.copy:
+        remote_src: true
+        src: "{{ cifmw_build_containers_repo_dir }}/"
+        dest: "/etc/yum.repos.d/"
+
     - name: Build tcib package
       vars:
         cifmw_bop_dlrn_target: centos9-stream
--- a/ci_framework/roles/build_containers/tasks/install.yml
+++ b/ci_framework/roles/build_containers/tasks/install.yml
@@ -11,17 +11,26 @@
       - buildah
     state: latest  # noqa: package-latest
 
-- name: Get tcib from repository  # noqa: latest[git]
-  ansible.builtin.git:
-    accept_hostkey: true
-    dest: "{{ cifmw_build_containers_basedir }}/tmp/tcib"
-    repo: "{{ cifmw_build_containers_tcib_src }}"
+- name: Install tcib from source
+  when: cifmw_build_containers_install_from_source | bool
+  block:
+    - name: Get tcib from repository  # noqa: latest[git]
+      ansible.builtin.git:
+        accept_hostkey: true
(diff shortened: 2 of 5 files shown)


rlandy:
looks ok - except for test failure: cifmw-tcib

pablintino:
Seems ok. I'll approve as soon as the issue in the test is fixed.

ci/playbooks/tcib/tcib.yml
```diff
@@ -15,6 +15,15 @@
   hosts: "{{ cifmw_target_host | default('localhost') }}"
   gather_facts: true
   tasks:
+    - name: Copy repo-setup generated repos to yum.repos.d directory
```
cjeanner, line 18: actually you may just push a new parameter to the job, `cifmw_repo_setup_output`, to get the repos in the right location.

ci_framework/roles/build_containers/tasks/install.yml
```diff
@@ -11,17 +11,26 @@
...
+- name: Install tcib from RPM
+  when: not cifmw_build_containers_install_from_source | bool
   become: true
-  ansible.builtin.shell: |
-    pip install -r requirements.txt
-    python setup.py install
-  args:
-    chdir: "{{ cifmw_build_containers_basedir }}/tmp/tcib"
+  ansible.builtin.package:
+    name: python3-tcib
```
marios, line 36: in fact, this ^ is the new thing here, installing python3-tcib from package.
we were already installing from source, looking at the diff right?

marios (commented):
looks right but there is an unresolved comment about the repo dir

ci/playbooks/tcib/tcib.yml
raukadah: We already have a repo-setup output directory var [`cifmw_repo_setup_output`](https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci_framework/roles/repo_setup/defaults/main.yml#L31) which is defaulted to [`cifmw_build_containers_repo_dir`](https://github.com/openstack-k8s-operators/ci-framework/blob/main/scenarios/centos-9/tcib.yml#L7). So we do not need that.

These repos are mounted with in the container during container build process. On the host system, there might be other repos in /etc/yum.repos.d directory. If we generate repos directly in yum.repos.d directory and mount it with container build process. Then we might include additional repos. it will affect the testing. That's why i have added this task to specifically copy the repos to yum.repos.d directory on host system to install python3-tcib rpm.

ci_framework/roles/build_containers/tasks/install.yml
raukadah: If we want to install tcib from source, we need to set `cifmw_build_containers_install_from_source` to true now.

ci/playbooks/tcib/tcib.yml
cjeanner: fair enough

cjeanner:
/approve

I'll let @marios have the last say. The goal of this patch being to install tcib from RPM, it just does what it says, with a new parameter to toggle source/rpm install.

PR #779, opened 2023-11-07
Refactor reproducer role

cjeanner:
In order to accept some extension, we first need to refactor things a
bit. This patch is also linked to #766 and #767.

Next steps will include adding devscripts support to the reproducer
role.

### DIFF
--- a/ci_framework/roles/reproducer/tasks/libvirt_layout.yml
+++ b/ci_framework/roles/reproducer/tasks/libvirt_layout.yml
@@ -1,0 +1,213 @@
+---
+- name: Deploy layout on target host
+  tags:
+    - libvirt_layout
+  ansible.builtin.import_role:
+    name: libvirt_manager
+    tasks_from: deploy_layout
+
+- name: Push generated inventory from hypervisor
+  ansible.builtin.command:  # noqa: command-instead-of-module
+    cmd: >-
+      rsync -r {{ cifmw_reproducer_basedir }}/reproducer-inventory/
+      zuul@controller-0:reproducer-inventory
+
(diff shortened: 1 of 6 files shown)


cjeanner:
@raukadah I'll need that one in before hitting the next steps for devscripts inclusion + the controller-0 vs hypervisor we discussed (tldr; a deploy may not define a controller, meaning the hypervisor (`cifmw_target_host`) will play the controller role).

ci_framework/roles/reproducer/tasks/libvirt_layout.yml
```diff
@@ -0,0 +1,213 @@
...
+      block:
+        - name: Get rhos-release
+          ansible.builtin.package:
+            name: "{{ cifmw_repo_setup_rhos_release_rpm }}"
+            state: present
+            disable_gpg_check: true
+
+        - name: Enable RHEL repos
+          ansible.builtin.command:
+            cmd: "rhos-release rhel"
```
raukadah, line 189:
```suggestion
            cmd: "rhos-release {{ cifmw_repo_setup_rhos_release_args }}"
```

raukadah (commented):
/approve

ci_framework/roles/reproducer/tasks/libvirt_layout.yml
cjeanner: hmmmm I don't think we want to get everything OSP on the controller-0, do we?

raukadah: I think `rhos-release rhel` is enough for controller.

PR #780, opened 2023-11-07
Allow auth file to be null

sdatko:
Currently the `cifmw_build_containers_authfile_path` variable has a default string value and because of that it is impossible to unset it in Ansible. However, when deploying local registry with `registry_deploy` role, the authentication is not necessary while pushing the built containers images. Instead of supplying a dummy empty file as workaround, it would be nice to just set the authfile_path variable to Null/None.

Additionally, a minor typo in task name is corrected.

  - [x] README in the role

### DIFF
--- a/ci_framework/roles/build_containers/tasks/main.yml
+++ b/ci_framework/roles/build_containers/tasks/main.yml
@@ -49,7 +49,7 @@
 
 - name: Make sure authfile exists
   when:
-    - cifmw_build_containers_authfile_path is defined
+    - cifmw_build_containers_authfile_path != None
     - cifmw_build_containers_push_containers | bool
   block:
     - name: Check for authfile
@@ -57,7 +57,7 @@
         path: '{{ cifmw_build_containers_authfile_path }}'
       register: authfile_exist
 
-    - name: Make sure autfile exists
+    - name: Make sure authfile exists
       ansible.builtin.assert:
         that:
           - authfile_exist.stat.exists | bool
--- a/ci_framework/roles/build_containers/README.md
+++ b/ci_framework/roles/build_containers/README.md
@@ -8,7 +8,7 @@ become - Required to install and execute tcib
 
 * `cifmw_build_containers_tcib_src`: (String) Repository of tcib to be installed
 * `cifmw_build_containers_basedir`: (String) Directory where we will hold all tcib files. Default to `cifmw_basedir/artifacts/containers_build` which defaults to `~/ci-framework-data/`.
-* `cifmw_build_containers_authfile_path`: (String) Authorization file to push containers to registry. Default to `${XDG_RUNTIME_DIR}/containers/auth.json`.
+* `cifmw_build_containers_authfile_path`: (String or Null) Authorization file to push containers to registry. Default to `${XDG_RUNTIME_DIR}/containers/auth.json`.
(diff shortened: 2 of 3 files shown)


ci_framework/roles/build_containers/tasks/main.yml
```diff
@@ -49,15 +49,15 @@
 
 - name: Make sure authfile exists
   when:
-    - cifmw_build_containers_authfile_path is defined
+    - cifmw_build_containers_authfile_path | default(None) != None
```
cjeanner, line 52: Can't we just pass this in a custom param file?
```YAML
cifmw_build_containers_authfile_path: null
```
meaning it will be "undefined", so it shouldn't hit that block anyway?

sdatko: hej, like what exactly you want to do that?
A variable with null value in Ansible is still a defined variable apprently (when overrriding within a Playbook) – and when supplying `cifmw_build_containers_authfile_path: ~` on Zuul job vars it also had no effect, but now I am unable to reproduce (`Run container_build.sh script` task fails)

arxcruz: I think the problem is that the default is set to a file https://github.com/openstack-k8s-operators/ci-framework/blob/6efd05125bc97d371853d71688c25c51b214ba19/ci_framework/roles/build_containers/defaults/main.yml#L22 so when we try to bring the content-provider, we do need to pass it to none. A better approach would be check also check only if push container is set to true.

sdatko: This is what I wrote in the commit message...
I need a `push_container` to be `true`, because I want the images to land in registry deployed with `registry_deploy` role.

arxcruz: Hmmm IIUC then this is wrong, we need to have a push_to_local_container when we want to push to the content-provider and push_container when we want to push to upstream and/or downstream... wdyt @cedric?

sdatko: @arxcruz IMO that overcomplicates the role with additional parameter, while it already has too many parameters...

cjeanner: now that I think about it.... in the job reproducer, for the content-provider support, we use the local registry as deployed by that `registry_deploy` role and I didn't have any issue when it came to push content to it.... Fun.

cjeanner:
/approve

it's not following the same path as the content-provider, using another role (`operator_build` is used in the content-provider, here we're talking about `container_build`)

ci_framework/roles/build_containers/tasks/main.yml
arxcruz: probably you already had the file locally in your computer for some other reason, so it passes, but in the clean vm it is failing :)

sdatko: According to https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci_framework/roles/reproducer/templates/play.yml.j2#L2-L78 the reproducer uses `operator_build`, not `build_containers` – I am not sure about the difference thought...

sdatko:
> sdatko force-pushed the main branch from 005f629 to 7a1c63d 
To clarify: just rebased it...

Issue #760, opened 2023-11-02
Please don't delete the defaiult libvirt network


hjensas:
I'm sorry, but I am using the default network on my lab machine for other things.
I don't see any reason why we need to delete this network? The ci-framework should instead create it's own network, and leave this network alone.

- name: Ensure the default network is removed.
  vars:
    net_name: "default"
  ansible.builtin.include_role:
    name: libvirt_manager
    tasks_from: delete_network.yml

PR #782, opened 2023-11-08
Remove virt-net default only when requested

psathyan:
As part of this PR, the default virtual network is not removed by default. This network is removed only when the user passes `cifmw_devscripts_remove_default_net: true`.

Fixes: #760

  - [x] README in the role


cjeanner:
/approve
it will indeed help fixing that - but iirc it's also removed from other places. Need to investigate a bit more

Issue #751, opened 2023-11-01
For RHEL9, libvirtd service should be disabled and start virt<service-name>d{,-ro,-admin}.socket instead.


knoha-rh:
When ci-framework steup runs on RHEL9 system, virt-install always failed.
It looks that virt-install requires individual virt<service-name>d{,-ro,admin}.socket. However, the related daemons conflicts with libvirt.service.

To avoid this situation, we should add something like below.

- libvirt_manager/tasks/virsh_checks.yml
~~~
- name: Ensure virt realted sockets are started and enabled
  become: true
  when: ansible_distribution in ['CentOS', 'RedHat'] and ansible_distribution_major_version == '9'
  block:
    - name: Ensure virt<<service name>>d.socket are started and enbled
      ansible.builtin.service:
        name: "virt{{ item }}d.socket"
        state: started
        enabled: true
      loop:
        - qemu
        - network
        - nodedev
        - nwfilter
        - secret
        - storage
        - interface
    - name: Ensure virt<<service name>>d-ro.socket are started and enbled
      ansible.builtin.service:
        name: "virt{{ item }}d-ro.socket"
        state: started
        enabled: true
      loop:
        - qemu
        - network
        - nodedev
        - nwfilter
        - secret
        - storage
        - interface
    - name: Ensure virt<<service name>>d-admin.socket are started and enbled
      ansible.builtin.service:
        name: "virt{{ item }}d-admin.socket"
        state: started
        enabled: true
      loop:
        - qemu
        - network
        - nodedev
        - nwfilter
        - secret
        - storage
        - interface

- name: Enable libvirt and configure  virtproxyd service
  when: "ansible_distribution in ['CentOS', 'RedHat'] and ansible_distribution_major_version == '8'"
  become: true
  ansible.builtin.service:
    name: "{{ cifmw_libvirt_manager_daemon }}"
    state: started
    enabled: true
~~~

cjeanner:
Unfortunately this isn't possible: it seems dev-scripts on its own is (re)starting the libvirtd.service. The only way to get out of that hell of socket was to remove the timeout set on the virtproxyd.service, allowing it to be long-running. If dev-scripts on its own and the underlying metal3-dev-env could NOT manage libvirtd service, it may be possible... But there's no option to disable that (wrong) management :(.

knoha-rh:
Hi Cedric,

[...]

PR #784, opened 2023-11-08
Better service management for libvirt components

cjeanner:
With EL-9 and CS-9, libvirt has moved to modular, socket activated
services.
We therefore must ensure proper services are enabled depending on the
release - and with the modular approach, we now have multiple sockets to
control.

Fixes: #751

Co-Authored-By: @knoha-rh

### DIFF
--- a/ci_framework/roles/libvirt_manager/tasks/virsh_checks.yml
+++ b/ci_framework/roles/libvirt_manager/tasks/virsh_checks.yml
@@ -14,47 +14,64 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
-- name: Start and enable libvirt service
-  become: true
-  ansible.builtin.service:
-    name: "{{ cifmw_libvirt_manager_daemon }}"
-    state: started
-    enabled: true
 
-# stabilize libvirt accesses
-# We hit a fair amount of failures with the libvirt sockets over the
-# past iteration. After many tests, it seems that removing the timeout
-# of the virtproxyd service and ensuring it's enabled and started
-# corrects this issue. With this hack, we could successfully deploy the
-# whole VA-1 layout without any issue.
-- name: Hack in the virtproxyd service
+- name: Clean hacky service override
   become: true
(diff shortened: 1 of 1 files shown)


cescgina (commented):
/lgtm

ci_framework/roles/libvirt_manager/tasks/virsh_checks.yml
```diff
@@ -14,47 +14,67 @@
...
-- name: Hack in the virtproxyd service
+- name: Remove service override
   become: true
-  when: cifmw_libvirt_manager_apply_virtproxy_patch | bool
   block:
     - name: Create service override directory
       ansible.builtin.file:
         path: "/etc/systemd/system/virtproxyd.service.d"
-        state: directory
+        state: absent
```
pablintino, line 24: If we are removing the dir:

- Task name makes no sense: Create service override directory
- Are owner and permissions needed?

ci_framework/roles/libvirt_manager/tasks/virsh_checks.yml
```diff
@@ -14,47 +14,67 @@
...
+      loop: "{{ _services }}"
+
+    - name: Ensure admin sockets are enabled
+      ansible.builtin.service:
+        name: "virt{{ item }}d-admin.socket"
+        state: started
+        enabled: true
+      loop: "{{ _services }}"
+
+- name: Manage service for older releases
```
pablintino, line 69: Why cannot we merge this task in the previous block that is already using become?

cjeanner: because of the block `when:` condition being in direct conflict

ci_framework/roles/libvirt_manager/tasks/virsh_checks.yml
cjeanner: right - corrected that.

PR #791, opened 2023-11-08
Fail CI FMW execution when running as root

dasm:
  - [x] README in the role

### DIFF
--- a/ci_framework/playbooks/01-bootstrap.yml
+++ b/ci_framework/playbooks/01-bootstrap.yml
@@ -2,6 +2,11 @@
   hosts: "{{ cifmw_target_host | default('localhost') }}"
   gather_facts: true
   tasks:
+    - name: Fail task execution when playbook is running as root.
+      ansible.builtin.fail:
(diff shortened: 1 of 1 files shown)


dasm:
The documentation is already there: https://github.com/openstack-k8s-operators/ci-framework/blob/main/docs/source/quickstart/01_requirements.md?plain=1#L19

ci_framework/playbooks/01-bootstrap.yml
```diff
@@ -2,6 +2,11 @@
   hosts: "{{ cifmw_target_host | default('localhost') }}"
   gather_facts: true
   tasks:
+    - name: Fail task execution when playbook is running as root.
```
cjeanner, line 5: I'd rather get a reusable role for those kind of validations. We also have to ensure the reproducer.yml (and subsequent clean plays) aren't called with root. Having a proper role, allowing to extend it in an easy way without having to edit 4 places, would be far easier to maintain.

cjeanner (changes requested):

PR #792, opened 2023-11-09
Use kustomize add operation for setting insecure registry

raukadah:
https://github.com/openstack-k8s-operators/ci-framework/pull/785 adds the support for setting insecure registry on EDPM nodes.

It works for EDPM. In BM job, we used replace operation which does not work as the key `edpm_container_registry_insecure_registries` does not exists in CR.

We need to use `add` as used in EDPM in order to make it working.

Note: It also bumps the timeout for tcib jobs as new containers got added and we need more then an hour to build all containers.

### DIFF
--- a/ci_framework/roles/edpm_deploy_baremetal/tasks/main.yml
+++ b/ci_framework/roles/edpm_deploy_baremetal/tasks/main.yml
@@ -192,9 +192,9 @@
       oc patch openstackdataplanenodeset openstack-edpm-ipam
       -n {{ cifmw_install_yamls_defaults['NAMESPACE'] }}
       --type json
-      -p '[{"op": "replace",
+      -p '[{"op": "add",
             "path": "/spec/nodeTemplate/ansible/ansibleVars/edpm_container_registry_insecure_registries",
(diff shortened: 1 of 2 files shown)


cjeanner:
@pablintino care to check that out once you're back? Thanks :)

ci_framework/roles/edpm_deploy_baremetal/tasks/main.yml
```diff
@@ -192,9 +192,9 @@
       oc patch openstackdataplanenodeset openstack-edpm-ipam
       -n {{ cifmw_install_yamls_defaults['NAMESPACE'] }}
       --type json
-      -p '[{"op": "replace",
+      -p '[{"op": "add",
             "path": "/spec/nodeTemplate/ansible/ansibleVars/edpm_container_registry_insecure_registries",
-            "value": "['{{ content_provider_registry_ip }}:5001']"}]'
+            "value": [{{ content_provider_registry_ip }}:5001]}]'
```
pablintino, line 197: @raukadah If `}]'` at this line where placed at the next line I'd be really really easier to follow the closing of each quote and bracket. It's fine, just a comment.

PR #793, opened 2023-11-09
Fix Baremetal EDPM logging

raukadah:
Currently `Check for virtualized compute` task is getting skipped
    due to wrong conditional.
    
    This patch fixes the conditional for libvirt services checking.
    
    In Baremetal Job, the default user with ssh access is cloud-admin
    not root user. It fixes the user for the same.

### DIFF
--- a/ci_framework/roles/artifacts/tasks/edpm.yml
+++ b/ci_framework/roles/artifacts/tasks/edpm.yml
@@ -4,9 +4,9 @@
     - not ansible_facts.services is defined
   ansible.builtin.service_facts:
 
-- name: Check for virtualized compute
+- name: Check for virtualized compute in Baremetal jobs
   when:
-    - "'libvirtd' in ansible_facts.services"
+    - "'libvirtd.service' in ansible_facts.services"
   block:
     - name: List all of the existing virtual machines
       register: vms_list
@@ -17,7 +17,7 @@
     - name: Filter out edpm vm
       ansible.builtin.set_fact:
         ssh_key_file: "{{ cifmw_artifacts_basedir }}/artifacts/edpm/ansibleee-ssh-key-id_rsa"
-        ssh_user: root
+        ssh_user: cloud-admin
         edpm_vms: >-
           {%- set listing=vms_list.list_vms | select('match', '^edpm-.*$') -%}
           {%- if listing | length == 1 -%}


ci_framework/roles/artifacts/tasks/edpm.yml
```diff
@@ -4,20 +4,21 @@
...
   ansible.builtin.service_facts:
 
-- name: Check for virtualized compute
+- name: Check for virtualized compute in Baremetal jobs
   when:
     - "'libvirtd' in ansible_facts.services"
   block:
     - name: List all of the existing virtual machines
       register: vms_list
+      become: true
```
cjeanner, line 13: as discussed, `become: true` isn't needed since we pass the `uri: qemu:///system`.
The actual issue is the condition on L9, it should check for `libvirtd.service` instead. We may want to also ensure it's running, but we may hit some issues with the modular libvirt sooner or later - especially since the ci-framework has full support for that. Up to you - but that's the actual issue.

ci_framework/roles/artifacts/tasks/edpm.yml
```diff
@@ -4,20 +4,21 @@
...
       community.libvirt.virt:
         command: list_vms
         uri: "qemu:///system"
 
     - name: Filter out edpm vm
       ansible.builtin.set_fact:
         ssh_key_file: "{{ cifmw_artifacts_basedir }}/artifacts/edpm/ansibleee-ssh-key-id_rsa"
-        ssh_user: root
+        ssh_user: cloud-admin
```
cjeanner, line 21: fine for me - since we're consuming the ansibleee-ssh-key, it's associated to that cloud-admin user.

ci_framework/roles/artifacts/tasks/edpm.yml
```diff
@@ -4,20 +4,21 @@
...
         edpm_vms: >-
```
cjeanner, line 22: eventually we should get the IP instead of the amount of hosts... That would allow to get more than one compute. For later.

cjeanner (changes requested):

ci_framework/roles/artifacts/tasks/edpm.yml
raukadah: Done.

ci_framework/roles/artifacts/tasks/edpm.yml
raukadah: Done

ci_framework/roles/artifacts/tasks/edpm.yml
cjeanner: please remove that  `become: true`

cjeanner (changes requested):

cjeanner (approved):

PR #795, opened 2023-11-09
Get actual test for reproducer role

cjeanner:
This patch activate a test for one of the reproducer features:
layout reproducer.

Since it's the base feature of the reproducer role, it's a good idea to
ensure it's working properly.


cjeanner:
Need some actual tests/validations, of course.

ci_framework/roles/libvirt_manager/tasks/deploy_layout.yml
```diff
@@ -119,14 +120,19 @@
...
+    crc_private_key: >-
+      {{
+        (_layout.vms.crc.image_local_dir, "id_ecdsa") |
+        path_join
+      }}
   block:
     - name: Copy authorized_keys
-      ansible.builtin.command:
+      ansible.builtin.shell:
         cmd: >-
```
pablintino, line 131: Hard-coding the user here is making my eye twitch. But, I understand that is OK for now

cjeanner: In this case I don't think we hit any issue. crc (and OCP actually) have that admin user.

PR #796, opened 2023-11-10
 Integrate devscripts layout support in reproducer role

cjeanner:
This addition also leads to some refactoring - it should make things
easier to maintain, by splitting the "main.yml" file into smaller,
dedicated pieces.

- [X] For now, testing this new feature isn't possible
  (resources, secrets management)


raukadah:
Overall looks good.

PR #797, opened 2023-11-10
Remove the namespace in the NMstate resources as is cluster wide

pablintino:


psathyan (approved):
/lgtm

PR #798, opened 2023-11-10
Fix the SSH proxy hostname used in the localhost controller

pablintino:
Right now we are using the inventory_hostname, that may not be a valid hostname, that's why now we default to the ansible_host.


cjeanner:
/approve
/lgtm

good catch!

cjeanner:
recheck

PR #801, opened 2023-11-14
Do not code any NIC related data in domain XML

cjeanner:
This change allows to use attach_network tasks in all cases instead of
trying to be too smart.

Converging like this will ensure a far happier path for OCP cluster
integration.

### DIFF
--- a/ci_framework/roles/libvirt_manager/tasks/create_vms.yml
+++ b/ci_framework/roles/libvirt_manager/tasks/create_vms.yml
@@ -105,6 +105,16 @@
     index_var: vm_id
     label: "{{ vm_type }}-{{ vm_id }}"
 
+- name: Attach listed networks to the VMs
+  vars:
+    vm_item: "{{ vm_type }}-{{ vm_id }}"
+    networks: "{{ vm_data.value.nets }}"
+  ansible.builtin.include_tasks: net_to_vms.yml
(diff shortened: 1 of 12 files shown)


ci_framework/roles/libvirt_manager/tasks/create_vms.yml
```diff
@@ -105,6 +105,16 @@
     index_var: vm_id
     label: "{{ vm_type }}-{{ vm_id }}"
 
+- name: Attach listed networks to the VMs
+  vars:
+    vm_item: "{{ vm_type }}-{{ vm_id }}"
+    networks: "{{ vm_data.value.nets }}"
+  ansible.builtin.include_tasks: net_to_vms.yml
+  loop: "{{ range(0, vm_data.value.amount | default(1) | int) }}"
+  loop_control:
```
pablintino, line 114: How nice is ansible to just loop over a loop...

pablintino:
/approve
Looks good, if no one lgtms it shortly I'll do it myself

ci_framework/roles/libvirt_manager/tasks/create_vms.yml
cjeanner: :) at least it's possible. Imagine if it wasn't....

PR #803, opened 2023-11-14
Reproducer: set job_id default to main in push_code tasks

raukadah:
In crc reproducer ci workflow, we are not passing any job url leading to job_id undefined variable in push_code task file.

Let's set the default value to main for job_id var to avoid the issue.


ci_framework/roles/reproducer/tasks/push_code.yml
```diff
@@ -11,8 +11,8 @@
       ansible.builtin.git:
         dest: "{{ repo.project.src_dir }}"
         repo: "https://{{ repo.project.canonical_name }}"
-        refspec: "pull/{{ repo.change }}/head:{{ job_id }}"
-        version: "{{ job_id }}"
+        refspec: "pull/{{ repo.change }}/head:{{ job_id | default('main') }}"
```
cjeanner, line 14: I'd rather go with `omit` for refspec in case of no job_id. This can be done with a ternary()

ci_framework/roles/reproducer/tasks/push_code.yml
```diff
@@ -11,8 +11,8 @@
...
+        version: "{{ job_id | default('main') }}"
```
cjeanner, line 15: we may go with `omit` as well in case of no job_id, it will get the default branch HEAD

PR #807, opened 2023-11-14
Fix small nits in run_ansible_test

pablintino:
The current run_ansible_test scripts doesn't point to the proper unitest directory, that is 'unit', not 'units'.

Some dependencies, needed by ansible-test to run pytest, like pytest-forked need to be installed. To ensure the needed dependencies are installed, we should let ansible-test install the required ones.

### DIFF
--- a/scripts/run_ansible_test
+++ b/scripts/run_ansible_test
@@ -39,7 +39,7 @@ case ${USE_VENV} in
 esac
 
 mkdir -p ${HOME}/code/ansible_collections/cifmw/general
-cp -ra ${PROJECT_DIR}ci_framework/* ${HOME}/code/ansible_collections/cifmw/general/
+rsync -a --delete ${PROJECT_DIR}ci_framework/* ${HOME}/code/ansible_collections/cifmw/general/
 ln -fs ${collection_path}/* ${HOME}/code/ansible_collections/
 
 # Create/append the sanity exceptions for the current ansible version
@@ -52,6 +52,6 @@ ${ansible_test} sanity --test validate-modules
 if [ -d tests/integration ]; then
     ${ansible_test} integration --color=yes
 fi
-if [ -d tests/units ]; then
-    ${ansible_test} units --color=yes
+if [ -d tests/unit ]; then
+    ${ansible_test} units --color=yes --requirements
 fi


scripts/run_ansible_test
```diff
@@ -52,6 +52,6 @@ ${ansible_test} sanity --test validate-modules
 if [ -d tests/integration ]; then
     ${ansible_test} integration --color=yes
 fi
-if [ -d tests/units ]; then
-    ${ansible_test} units --color=yes
+if [ -d tests/unit ]; then
```
cjeanner, line 55: actually, according to the doc, it should indeed be tests/units:
https://docs.ansible.com/ansible/latest/dev_guide/testing_units.html

also shown in the ansible repository:
https://github.com/ansible/ansible/tree/devel/test/units

fun, because it used to run anyway...

cjeanner (changes requested):

scripts/run_ansible_test
pablintino: @cjeanner are you sure? https://github.com/ansible/ansible/blob/devel/test/lib/ansible_test/_internal/provider/layout/collection.py#L123

This is the ansible-test output when I use tests/units instead of tests/unit:
`❯ ansible-test units
WARNING: Rename "tests/units/" to "tests/unit/" to run unit tests.
WARNING: All targets skipped.
`

cjeanner: consistency.. :rofl:

scripts/run_ansible_test
```diff
@@ -48,10 +48,9 @@ cat  "${HOME}/code/ansible_collections/cifmw/general/tests/sanity/ignore.txt" >>
     "${HOME}/code/ansible_collections/cifmw/general/tests/sanity/ignore-${ansible_version}.txt"
 
 pushd ${HOME}/code/ansible_collections/cifmw/general
-${ansible_test} sanity --test validate-modules
 if [ -d tests/integration ]; then
     ${ansible_test} integration --color=yes
 fi
```
marios, line 53: so no longer running the sanity test?

scripts/run_ansible_test
```diff
@@ -39,7 +39,7 @@ case ${USE_VENV} in
 esac
 
 mkdir -p ${HOME}/code/ansible_collections/cifmw/general
-cp -ra ${PROJECT_DIR}ci_framework/* ${HOME}/code/ansible_collections/cifmw/general/
+rsync -a ${PROJECT_DIR}ci_framework/* ${HOME}/code/ansible_collections/cifmw/general/ --delete
```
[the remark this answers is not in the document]
cjeanner: :+1: to that

cjeanner:
/approve cancel

there are some legitimate concerns.

scripts/run_ansible_test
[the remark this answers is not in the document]
pablintino: Done

scripts/run_ansible_test
pablintino: Ouch, restored, my fault with my local, cause I removed it locally to skip them

raukadah (approved):

PR #810, opened 2023-11-15
Expose a userfriendly container env

cjeanner:
ALlowing people to consume the framework via a container seems to be
mandatory.

This change allows to get rid of `buildah` call (not available on MacOS
for instance), while allowing some more friendly usage of the
ci-framework.

The embedded documentation will be added to the readthedocs one once we
finish the deep refacroring.

- [X] Appropriate documentation exists and/or is up-to-date

### DIFF
--- a/containerfiles/README.md
+++ b/containerfiles/README.md
@@ -1,0 +1,72 @@
+# Available containers
+
+The project provides two types of containers:
+
+- Containerfile.ci: builds a really basic container, mostly used in Prow, exposing needed tools
+- Containerfile.tests: based upon the previous one, allows to run some tests locally, or run the framework
+
+## Build containers
+
+Leveraging make, you can build the container on your own:
+```Bash
+$ make ci_ctx
+```
+
+## Using the container
+
+In the examples bellow, the options will do:
+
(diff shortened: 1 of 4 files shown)


raukadah (approved):

containerfiles/README.md
```diff
@@ -0,0 +1,72 @@
...
+    -w /home/prow \
+    -e HOME=/home/prow \
+    -v ~/.ssh/id_ed25519:/home/prow/.ssh/id_ed25519 \
+    -v .:/home/prow/ci-framework \
+    -v ~/inventory.yml:/home/prow/ci-framework/inventory.yml \
+    --security-opt label=disable \
+    localhost/cifmw:latest bash
+```
+
+##### On non-SELinux systems (Debian, Ubuntu, Mint, MacOS, ...):
```
tosky, line 60: This command is currently the same as the previous one for RHEL-based systems (technically, Fedora-based systems :)
Should they differ, or is it to prepare for future differences?

eshulman2 (commented):
I would also mention supported versions for contributing, I believe it is missing in the docs currently.

PR #812, opened 2023-11-16
Allow to inject DNF configuration tweaks

cjeanner:
This may help in case you have a proxy or any other things to configure
in dnf.

This will affect the hypervisor and controller-0 configuration files.

  - [X] README in the role


raukadah (approved):
/approve

PR #815, opened 2023-11-16
WIP: Adding a sample job to use hive role to claim clusters

frenzyfriday:
  - [ ] README in the role

### DIFF
--- a/ci/playbooks/hive/claim.yml
+++ b/ci/playbooks/hive/claim.yml
@@ -1,0 +1,6 @@
+- name: Install deps and prepare for KUTTL run
+  hosts: ""
+  tasks:
+    - name: Claim cluster through hive
+      ansible.builtin.include_role:
+        name: hive
--- a/scenarios/centos-9/hive.yml
+++ b/scenarios/centos-9/hive.yml
@@ -1,0 +1,8 @@
+---
+cifmw_hive_platform: "openstack"
+cifmw_hive_action: "claim_cluster"
+cifmw_hive_dry_run: false
+cifmw_hive_openstack_pool_name:
+cifmw_hive_namespace:
+claim_name:
+cifmw_hive_kubeconfig:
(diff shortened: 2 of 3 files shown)


ci/playbooks/hive/claim.yml
```diff
@@ -0,0 +1,6 @@
+- name: Install deps and prepare for KUTTL run
+  hosts: ""
```
frenzyfriday, line 2: What should be the hosts here?

scenarios/centos-9/hive.yml
```diff
@@ -0,0 +1,8 @@
+---
+cifmw_hive_platform: "openstack"
+cifmw_hive_action: "claim_cluster"
+cifmw_hive_dry_run: false
+cifmw_hive_openstack_pool_name:
+cifmw_hive_namespace:
+claim_name:
+cifmw_hive_kubeconfig:
```
frenzyfriday, line 8: How should I pass these?

ci/playbooks/hive/claim.yml
cjeanner: I think it should be `localhost` ? i.e. the host running ansible would request the cluster. Would that make sense?

scenarios/centos-9/hive.yml
cjeanner: are those private parameters? If so, they should be in zuul's secrets I guess.

zuul.d/end-to-end.yaml
```diff
@@ -18,6 +18,25 @@
       - github.com/openstack-k8s-operators/install_yamls
     timeout: 5400
 
+- job:
+    name: cifmw-end-to-end-base-hive
+    nodeset: centos-9-crc-3xl
```
cjeanner, line 23: hmmm if we're getting a cluster via hive, we don't need to get crc... do we?

PR #830, opened 2023-11-21
Correct how we detect nested virtualization layout

cjeanner:
Since we have to handle two kinds of layout, multinode and nested
virtualization, we better have to get a strong way to differenciate
them.

Until now, we were trying to be smart, by looking up libvirtd.service
availability, and listing VMs therein. But it wasn't strong enough and
lead to some issues with multinode having libvirtd available.

This patch introduces a new, CI dedicated parameter, allowing to toggle
the right "path" in the log gathering.

### DIFF
--- a/ci_framework/roles/artifacts/tasks/edpm.yml
+++ b/ci_framework/roles/artifacts/tasks/edpm.yml
@@ -1,12 +1,7 @@
 ---
-- name: Gather services
-  when:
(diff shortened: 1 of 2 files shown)


bogdando (commented):
so, this should improve the dataplane adoption CI .. um, logic?
LGTM

raukadah (approved):
it is much better for nested virt check

PR #831, opened 2023-11-21
Disable cloud-init override of NM DNS

pablintino:

### DIFF
--- a/ci/playbooks/multinode-customizations.yml
+++ b/ci/playbooks/multinode-customizations.yml
@@ -128,6 +128,28 @@
       ansible.builtin.set_fact:
         persistent_ssh_key: "{{ pub_key_slurp['content'] | b64decode }}"
 
+    - name: Check if cloud-init is overriding NM settings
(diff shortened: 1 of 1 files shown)


ci/playbooks/multinode-customizations.yml
```diff
@@ -128,6 +128,28 @@
       ansible.builtin.set_fact:
         persistent_ssh_key: "{{ pub_key_slurp['content'] | b64decode }}"
 
+    - name: Check if cloud-init is overriding NM settings
+      become: true
+      ansible.builtin.stat:
+        path: "/etc/NetworkManager/conf.d/99-cloud-init.conf"
+      register: _cloud_init_nm_override_stat
+
```
cjeanner, line 136: we may want to get handlers for this kind of things. We'd then be able to just get one task, triggering the handler when it changes. But not for now - it's not the only place either.

rlandy (commented):
/lgtm

PR #832, opened 2023-11-22
Fix the conditions when to use the default network as OCP external subnet

psathyan:
In this PR, the conditions when to use the default subnet as external subnet is correct. It is fixed by checking if the user has provided external_subnet_ipv[46] or not.

If provided the values are not replaced.


raukadah (approved):

psathyan:
recheck

PR #835, opened 2023-11-22
Consolidate documentation management

cjeanner:
Simplify the Makefile, remove useless Makefile_ori.

Rework the spellcheck workflow so that we first build the doc, then run
the spellchecking against the generated content.
This allow to ignore all of the code blocks and some other things,
ensuring we shouldn't get too many issues with fancy words such as RSA
key content, hashed output and so on.


raukadah:
recheck

cjeanner:
recheck

cjeanner:
recheck

cjeanner:
recheck

cjeanner:
recheck

PR #841, opened 2023-11-24
Networking mapping/ip pools

pablintino:
  - [ ] README in the role


pablintino:
> Wondering if we shouldn't get proper documentation - we may then generated some python doc, and expose it in the sphinx doc output?

I'll create it, for sure, with yaml examples, but I'd prefer to do it when all the pieces (referring to the role only) are glued together, just to avoid writing doc that may change a lot till the final picture.

ci_framework/plugins/module_utils/networking_mapping/ip_pools.py
```diff
@@ -0,0 +1,129 @@
...
+    def add_reservation(
+        self,
+        reservation: typing.Union[ipaddress.IPv4Address, ipaddress.IPv6Address, str],
+    ):
+        reservation_ip = ipaddress.ip_address(reservation)
+        if reservation_ip not in self.__range:
+            raise exceptions.NetworkMappingError(
+                f"Reservation {reservation} is out of range {self.__range}"
+            )
+        self.__reservations.add(reservation_ip)
```
viroel, line 33: Ip reservation would be a typical case where we need to worry about concurrency and add locks, but AFAIU we should not have that scenario while doing a simple mapping, right?

viroel (commented):
Just some points to discuss, but ingeneral code looks good

ci_framework/plugins/module_utils/networking_mapping/ip_pools.py
pablintino: The code will run as a module/action plugins, in one single thread, so no idea why we need to think about concurrency. This functionality is used to avoid the pool assigning IPs for instances that the uses has explicitly set, nothing related to the OS, libvirt or whatever low level tool that manages networking.

ci_framework/plugins/module_utils/networking_mapping/ip_pools.py
```diff
@@ -0,0 +1,129 @@
...
+    networking_definition,
+)
+
+
+class IPPool:
+    def __init__(
+        self,
+        ip_range: networking_definition.HostNetworkRange,
+        reservations: typing.List[
+            typing.Union[ipaddress.IPv4Address, ipaddress.IPv6Address, str]
```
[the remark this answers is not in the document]
pablintino: I'll add IPv6 test cases

PR #843, opened 2023-11-26
Gather additional namespaces when must-gather is executed

fmount:
This change aligns the `os_must_gather` role to pass additional `namespaces` that are relevant for CI.
By default all the ns matching `*kuttl*` and `sushy-emulator` are added, but the env variable can be used as an override to the default behavior.

- [x] README in the role

Depends-On: https://github.com/openstack-k8s-operators/openstack-must-gather/pull/31

### DIFF
--- a/roles/os_must_gather/README.md
+++ b/roles/os_must_gather/README.md
@@ -13,6 +13,8 @@ testing the new changes.
 * `cifmw_os_must_gather_repo_path`: (string) Path to local clone of openstack-must-gather git repo
 * `cifmw_os_must_gather_timeout`: (String) Timeout for must-gather command
 * `cifmw_os_must_gather_host_network`: (Bool) Flag to gather host network data
+* `cifmw_os_must_gather_namespaces`: (List) List of namespaces required by the gather task in case of failure
+* `cifmw_os_must_gather_additional_namespaces`: (String) List of comma separated additional namespaces
 
 ## Examples
 ```
--- a/roles/os_must_gather/defaults/main.yml
+++ b/roles/os_must_gather/defaults/main.yml
@@ -23,4 +23,13 @@ cifmw_os_must_gather_image_registry: "quay.rdoproject.org/openstack-k8s-operator
 cifmw_os_must_gather_output_dir: "{{ cifmw_basedir | default(ansible_user_dir ~ '/ci-framework-data') }}"
 cifmw_os_must_gather_repo_path: "{{ ansible_user_dir }}/src/github.com/openstack-k8s-operators/openstack-must-gather"
 cifmw_os_must_gather_timeout: "10m"
+cifmw_os_must_gather_additional_namespaces: "kuttl,sushy-emulator"
+cifmw_os_must_gather_namespaces:
+  - openstack-operators
+  - openstack
+  - baremetal-operator-system
+  - openshift-machine-api
+  - cert-manager
+  - openshift-nmstate
+  - metallb-system
 cifmw_os_must_gather_host_network: false
--- a/roles/os_must_gather/tasks/main.yml
+++ b/roles/os_must_gather/tasks/main.yml
@@ -61,8 +61,16 @@
           oc adm must-gather --image {{ cifmw_os_must_gather_image }}
           --timeout {{ cifmw_os_must_gather_timeout }}
           --host-network={{ cifmw_os_must_gather_host_network }}
-          --dest-dir {{ cifmw_os_must_gather_output_dir }}/logs > {{ cifmw_os_must_gather_output_dir }}/logs/os_must_gather.log
+          --dest-dir {{ cifmw_os_must_gather_output_dir }}/logs
+          -- ADDITIONAL_NAMESPACES={{ cifmw_os_must_gather_additional_namespaces }} gather &> {{ cifmw_os_must_gather_output_dir }}/logs/os_must_gather.log
+
   rescue:
+    - name: Create oc_inspect log directory
+      ansible.builtin.file:
+        path: "{{ cifmw_os_must_gather_output_dir }}/logs/oc_inspect"
+        state: directory
+        mode: "0755"
+
     - name: Inspect the cluster after must-gather failure
       ignore_errors: true # noqa: ignore-errors
       environment:
@@ -71,7 +79,13 @@
       cifmw.general.ci_script:
         output_dir: "{{ cifmw_os_must_gather_output_dir }}/artifacts"
         script: |
-          mkdir -p {{ cifmw_os_must_gather_output_dir }}/logs/oc_inspect
-          for ns in openstack-operators openstack baremetal-operator-system openshift-machine-api cert-manager openshift-nmstate metallb-system; do
-            oc adm inspect namespace/"${ns}" --dest-dir={{ cifmw_os_must_gather_output_dir }}/logs/oc_inspect || echo "Error getting logs from ${ns}"
-          done
+            oc adm inspect namespace/{{ item }} --dest-dir={{ cifmw_os_must_gather_output_dir }}/logs/oc_inspect
+      loop: >-
+        {{
+          (
+            cifmw_os_must_gather_namespaces | default([]) +
+            (
+              cifmw_os_must_gather_additional_namespaces | split(',') | list
+            )
+          ) | unique
+        }}


raukadah:
recheck

raukadah (commented):
/approve

ci_framework/roles/os_must_gather/tasks/main.yml
```diff
@@ -46,7 +46,8 @@
           oc adm must-gather --image {{ cifmw_os_must_gather_image }}
           --timeout {{ cifmw_os_must_gather_timeout }}
           --host-network={{ cifmw_os_must_gather_host_network }}
-          --dest-dir {{ cifmw_os_must_gather_output_dir }}/logs > {{ cifmw_os_must_gather_output_dir }}/logs/os_must_gather.log
+          --dest-dir {{ cifmw_os_must_gather_output_dir }}/logs
+          -- SOS_SERVICES= gather > {{ cifmw_os_must_gather_output_dir }}/logs/os_must_gather.log
```
raukadah, line 50: Thank you @fmount for improving the must_gather role. Can we add a Ansible var to pass the list of sos_services (default should be empty) so that end user can use it while consuming the role?

lewisdenny, line 50: Hi @fmount, I think there is a typo here and it should be `--SOS_SERVICES=gather` rather than `-- SOS_SERVICES= gather`

raukadah: Here https://paste.centos.org/view/raw/391e519e is the output and README command: https://github.com/openstack-k8s-operators/openstack-must-gather/blob/main/README.md?plain=1#L36

fmount: makes sense in case we are in a troubleshooting session or we need to gather sos reports as well.
Added in my last PS.

fmount: no it's good :D
`SOS_SERVICES` it's not a cli variable. It's an environment variable passed to the must-gather command, so it's good to have the space to properly invoke the `gather` script.

ci_framework/roles/os_must_gather/defaults/main.yml
```diff
@@ -20,4 +20,5 @@
 cifmw_os_must_gather_image: "quay.io/openstack-k8s-operators/openstack-must-gather:latest"
 cifmw_os_must_gather_output_dir: "{{ cifmw_basedir | default(ansible_user_dir ~ '/ci-framework-data') }}"
 cifmw_os_must_gather_timeout: "10m"
+cifmw_os_must_gather_additional_namespaces: "kuttl,sushy-emulator"
```
raukadah, line 23: Can we add additional namespaces here https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci_framework/roles/os_must_gather/tasks/main.yml#L60 also under rescue?

fmount: of course, doing that in my next iteration.

fmount: Done

fmount:
I just realized that this can't work with a simple `Depends-on` because we are not able to build the must-gather image in ci-framework on the fly (@lewisdenny is working on it), so putting this on hold until the dependency merges.

fmount:
I see this patch acts as I expected, and from the CI execution I can see `ADDITIONAL_NAMESPACES=kuttl,sushy-emulator` being added [1].
@raukadah I have to figure out what's going on with molecule (looks like it's failing on `crc start`), otherwise I think we can land this.
 
 [1] https://logserver.rdoproject.org/43/843/d60786fc46a3561c65759400371170e23fbf8bbc/github-check/podified-multinode-edpm-deployment-crc/00acf2a/controller/ci-framework-data/artifacts/ci_script_005_run_openstack-must-gather_command.sh

ci_framework/roles/os_must_gather/tasks/main.yml
```diff
@@ -61,7 +61,9 @@
           oc adm must-gather --image {{ cifmw_os_must_gather_image }}
           --timeout {{ cifmw_os_must_gather_timeout }}
           --host-network={{ cifmw_os_must_gather_host_network }}
-          --dest-dir {{ cifmw_os_must_gather_output_dir }}/logs > {{ cifmw_os_must_gather_output_dir }}/logs/os_must_gather.log
+          --dest-dir {{ cifmw_os_must_gather_output_dir }}/logs
+          -- ADDITIONAL_NAMESPACES={{ cifmw_os_must_gather_additional_namespaces }} gather &> {{ cifmw_os_must_gather_output_dir }}/logs/os_must_gather.log
+
   rescue:
     - name: Inspect the cluster after must-gather failure
```
pablintino, line 68: @fmount I know this was already kind of meh and this is not you fault, but can you please make this task ansible friendly? I mean, split it two:

1.  One task that creates the `{{ cifmw_os_must_gather_output_dir }}/logs/oc_inspect` directory
2. Another one, that using a loop over `cifmw_os_must_gather_additional_namespaces | split(',')` runs `oc adm inspect namespace/"${ns}" --dest-dir={{ cifmw_os_must_gather_output_dir }}/logs/oc_inspect` You can set the ignore errors to true, that way if it fails, we can get the entire output of the failure, instead of relying on a dumb message like `echo "Error getting logs from ${ns}"`.

pablintino:
@fmount In my opinion is more than OK, thanks for your work. If you can/want to address the comment I made I can wait to lgtm/approve, if not can you do it on a followup pr? Thanks.

ci_framework/roles/os_must_gather/tasks/main.yml
fmount: Doing that and refining in a couple of iterations

fmount:
> @fmount In my opinion is more than OK, thanks for your work. If you can/want to address the comment I made I can wait to lgtm/approve, if not can you do it on a followup pr? Thanks.

No reason to have a follow-up PR, I think it's ok addressing this issue here.

ci_framework/roles/os_must_gather/tasks/main.yml
```diff
@@ -61,8 +61,23 @@
...
   rescue:
+    - name: Create oc_inspect log directory
+      ansible.builtin.file:
+        path: "{{ cifmw_os_must_gather_output_dir }}/logs/oc_inspect"
+        state: directory
+        mode: "0755"
+
+    - name: Append additional namespaces to the default list
+      ansible.builtin.set_fact:
+        cifmw_os_must_gather_namespaces: "{{ cifmw_os_must_gather_namespaces | default([]) + [ item ] }}"
```
pablintino, line 76: Just one comment, based on awful experiences we have. Keep in mind that if the user passes `cifmw_os_must_gather_namespaces` as a variable the set_fact is not going to work and the value that will be used in the following task will be the one set in the variable and not in the fact. That's because the precedence of variables and facts...

pablintino: If you do something like this you may be able to get rid off what I'm saying at the same time we do this in a single task:
```yaml
- name: Inspect the cluster after must-gather failure
  ignore_errors: true # noqa: ignore-errors
  environment:
    PATH: "{{ cifmw_path }}"
  ci_script:
    output_dir: "{{ cifmw_os_must_gather_output_dir }}/artifacts"
    script: |
        oc adm inspect namespace/{{ item }} --dest-dir={{ cifmw_os_must_gather_output_dir }}/logs/oc_inspect
  loop: >-
    {{
      (
        cifmw_os_must_gather_namespaces | default([]) +
        (
          cifmw_os_must_gather_additional_namespaces |
          map('split', ',') |
          list
        )
      ) | unique
    }}
  

```

pablintino:
@fmount I left a couple of comments after checking what was pending here. We have moved the framework to a proper collection, so you may need to rebase the PR.

ci_framework/roles/os_must_gather/tasks/main.yml
fmount: ack, it makes sense and we do not want to expose such variable to external users because we know we need it to get the namespaces as a result of combining default + additional.
Thanks for the hint here, I'll check the version you're proposing here and propose something that fit what you're describing (even the same if it works, but I feel like it's cleaner see if can compute that variable outside the `loop` keywork:  I like this solution though!)

fmount:
recheck

lewisdenny:
Thank you @fmount, this patch is looking really nice. Are we able to get it merged now? We have approval, I just note your active `hold`.

Thanks!

fmount:
> Thank you @fmount, this patch is looking really nice. Are we able to get it merged now? We have approval, I just note your active `hold`.
> 
> Thanks!

Sure, this should be ready at this point, I was waiting for @pablintino to approve

lewisdenny:
The man himself has spoken! @pablintino @fmount ++

PR #846, opened 2023-11-27
Ensure we consume valid CA bundle

cjeanner:
Since we may inject custom CA in the system, we have to ensure we're
consuming the right bundle.


cjeanner:
Via a depends-on, I could successfully validate it's working fine :)
```
2023-11-27 10:00:28.501654 | controller | TASK [os_net_setup : Get network information auth={{ openstack_auth }}, validate_certs={{ cifmw_os_net_setup_verify_tls }}, ca_cert=/etc/pki/tls/certs/ca-bundle.crt] ***
2023-11-27 10:00:28.501665 | controller | Monday 27 November 2023  10:00:28 -0500 (0:00:00.045)       0:32:45.563 *******
2023-11-27 10:00:30.462016 | controller | ok: [localhost]
```

stuggi (approved):
/lgtm

PR #849, opened 2023-11-27
Add missing uri parameter

cjeanner:
While doc states default value is `qemu:///system`, it is not.

Also remove potentially problematic `cached: true`, and add some
debugging tasks, as well as removing the "filter" that may prevent
re-generating the fact if it's already set with some previous run. This
should ensure we're consuming up-to-date list.

References:
- https://docs.ansible.com/ansible/latest/collections/community/libvirt/virt_module.html#parameter-uri
- https://docs.ansible.com/ansible/latest/collections/community/libvirt/virt_net_module.html#parameter-uri


cjeanner:
/approve

Tested with multiple runs in a job that was failing on a steady trend, and it did succeed every single run. Auto-approving, so that any other reviewer will be able to merge.

PR #858, opened 2023-11-29
Added krb dlrn reporting support

raukadah:
In order to reporting zuul job status using DLRN kerberous auth, We need to do kinit with proper keytab and realm before calling dlrnapi-client.

This pr adds the same.

### DIFF
--- a/docs/dictionary/en-custom.txt
+++ b/docs/dictionary/en-custom.txt
@@ -192,6 +192,8 @@ keepalived
 kerberos
 keypair
 keyring
+keytab
+kinit
 kni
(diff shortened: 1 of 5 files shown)


docs/dictionary/en-custom.txt
```diff
@@ -190,8 +190,11 @@ jzxbol
 kcgpby
 keepalived
 kerberos
+kerberous
```
rlandy, line 193: kerberos - then you should not need this?

cjeanner:
recheck

PR #861, opened 2023-11-29
Integrate manage_secrets in devscripts role

cjeanner:
This will allow to converge secret management through the framework.

We also have a hard failure in case a user is passing the old parameter.
This may sound hard, but that's the best way to not have to maintain old
variables.
The error message is clear enough to help users moving to the new
schema.

  - [X] README in the role


psathyan (approved):
/lgtm

PR #863, opened 2023-11-29
Avoid running edpm jobs for a couple of roles

cjeanner:
devscripts and libvirt_manager role shouldn't trigger any edpm related
job: they are providing features that aren't used in CI.

- [x] no test nor doc


raukadah (commented):
/approve

PR #870, opened 2023-11-30
Adds config_nm false for data plane adoption standalone node

marios:
Then we can also remove the explicit deletion we have in [1]

[1] https://review.rdoproject.org/r/c/rdo-jobs/+/50590/48/playbooks/data_plane_adoption/setup_standalone_os_net_config.yaml


cjeanner:
You want to rebase against `main` - the pre-commit is fixed.

PR #871, opened 2023-11-30
[edpm_prepare]Pass controlplane cr via ansible var

raukadah:
We are working on removing the dependency from install_yamls reliance. In order to achieve that, we will be setting NETWORK_ISOLATION to false in Install yamls vars.

In Install yamls, control plane crs are tied with
NETWORK_ISOLATION flag. Since we are going to set it to false then it will pick different cr leading to breakage in job.

This pr drops openstack_make_deploy_prep target and allow users to pass control plane cr via var.

We can use ci_kustomize role to do additional
customization.


cjeanner:
You want to rebase against `main` - the pre-commit is fixed.

cjeanner:
/lgtm
I'll defer the approval of this patch to @pablintino - I want to ensure it won't make the next steps more difficult when he's getting to kustomize the whole deploy.

cjeanner:
/lgtm cancel
install_yaml will be consumed to generate CRs - we'll handle the kustomization + deploy on our own.

raukadah:
Closing currently, will revisit later.

PR #873, opened 2023-11-30
Add needed dependency in CI container

cjeanner:
netaddr is needed for ansible-lint to properly work.


raukadah (approved):
/approve

PR #874, opened 2023-11-30
Remove check on deprecated parameters

cjeanner:


psathyan (commented):
/lgtm

cjeanner:
Merging. We need to correct the reproducer.

I'll try to find out why the pre-commit is now crashing on the  `netaddr` library... doesn't make sens at all, until now it was just fine.

PR #876, opened 2023-11-30
Remove local_env capability

cjeanner:
Nested virtualization has bitten us far too many times. With the
reproducer and associated new layout, plus the containerized
ci-framework, there is no actual use for the local_env.

- [x] Removing a deprecated feature and related code.


raukadah (commented):
@cjeanner We also need to drop it from https://github.com/openstack-k8s-operators/ci-framework/blob/f02b07ab5e37f61e6ce9d7cc6757e295af8803c8/ci/playbooks/dev-workflow.yml#L25 .

raukadah (approved):
/approve

PR #880, opened 2023-12-01
Pull tempest from content provider

raukadah:
This pr adds more zuul return vars to openstack content provider job so that we can fetch tag and registry url for pulling tempest container.

It will help to test proper tempest container in content provider job.


cjeanner:
/approve

wondering if our ci playbooks may leverage the "run_hook" role.

raukadah:
> /approve
> 
> wondering if our ci playbooks may leverage the "run_hook" role.

Not sure about that, I am thinking to use run_test role (which I was working on in the past) to run these ci playbooks and keep only one run.yml playbook to execute the nested playbook from there.

raukadah:
Closing currently, will revisit later.

PR #882, opened 2023-12-01
Set default network vals on the base job

rlandy:
In order to be able to run the same job
on any cloud provider, the values required to
use that provider must be able to be vary.

This PR allows for setting the cloud name,
mtus and router_net as specified in cloud
secrets.

  - [x] README in the role

### DIFF
--- a/zuul.d/adoption.yaml
+++ b/zuul.d/adoption.yaml
@@ -41,10 +41,12 @@
       cifmw_use_libvirt: true
       cifmw_use_crc: false
       cifmw_zuul_target_host: controller
+      crc_ci_bootstrap_cloud_name: "{{ nodepool.cloud | replace('-nodepool-tripleo','') }}"
       crc_ci_bootstrap_networking:
         networks:
           default:
-            mtu: 1500
+            mtu: "{{ ('ibm' in nodepool.cloud) | ternary('1440', '1500') }}"
+            router_net: "{{ ('ibm' in nodepool.cloud) | ternary('hostonly', 'public') }}"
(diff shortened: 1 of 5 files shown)


viroel:
ok, this looks a good solution
as mentioned in review call, we are missing the same for hci (3compute) jobs:
https://github.com/openstack-k8s-operators/ci-framework/blob/6a0e021a955332078412d2d593585038839c1e2c/zuul.d/edpm_multinode.yaml#L25

So you may want to add it in another PR or update this one? 
Otherwise, I think that hci jobs will fail if get scheduled to IBM?

rlandy:
@viroel will update with the HCI job

cescgina:
lgtm

marios (commented):
looks good but why are there 2 commits

rebtoor:
> looks good but why are there 2 commits

The second commit is a merge commit, let's see if I can get rid of it so the PR can be merged safely :)

cescgina:
Adoption failure seems unrelated:
`"stderr": "+ oc login -u kubeadmin -p 123456789\nError from server (InternalError): Internal error occurred: unexpected response: 400"`
https://review.rdoproject.org/zuul/build/d4787177796848d0a8ab074bf7d094a5/log/controller/data-plane-adoption-tests-repo/data-plane-adoption/tests/logs/test_with_ceph_out_2023-12-22T05:14:26EST.log#443

pablintino:
recheck

pablintino:
We have clean greens for both of each failed Zuul jobs, adoption and kuttl in different runs

PR #883, opened 2023-12-04
Add cifmw-base-minimal base job for running nested ansible calls

raukadah:
This job will be helpful to run hive standalone playbook in downstream. It does not need to clone multiple projects. We just need to setup the ci-framework environment, populate zuul vars at one place and run the standalone playbook.


cjeanner:
Do we intend to inherits from this base job upstream as well?

raukadah:
recheck

raukadah:
> Do we intend to inherits from this base job upstream as well?

Yes, I have updated the pr to use the same job in content provider.

viroel:
/lgtm

agree, that's the minimum to run nested playbook jobs. Should be good to merge.

PR #886, opened 2023-12-04
[doc] Fixed a missing `\` in a shell command

rebtoor:
Also, renamed inventory snippet to `custom-inventory` to match file passed to ansible-playbook command

  - [ ] README in the role


docs/source/quickstart/03_deploy_infra.md
```diff
@@ -28,7 +28,7 @@ It must expose at least two hosts:
 You can take this as a template:
 
 ~~~{code-block} YAML
-:caption: inventory.yml
+:caption: custom-inventory.yml
```
cjeanner, line 31: ah in fact it should be `custom/inventory.yml` (and would need an update to the later command on L65) - that way everything would be in that `custom` location, which is ignored from git tree.

rebtoor: ack!

Issue #761, opened 2023-11-02
Consider using --system-site-packages during venv


gibizer:
Adding --system-site-packages to the venv command allows to use the python packages from the host if they are not installed in the venv (venv packages still take preference). For example python3-firewall is installed on the hypervisor but at the moment I did not found a way to install it in the venv (pypi firewall package is a different package!)

This is just a WA for deploying locally on the hypervisor. In general --system-site-packages should be avoided as it makes the system dependent on python packages installed on the host.

https://github.com/openstack-k8s-operators/ci-framework/blob/9a0f52d3b0781eeeb842073b3d0dd693ce12d78e/scripts/setup_env#L74-L76

PR #889, opened 2023-12-04
[WIP] debuging venv command seams to behave diffently on fedora 39

SeanMooney:


ci_framework/roles/repo_setup/tasks/install.yml
```diff
@@ -26,7 +26,7 @@
   ansible.builtin.pip:
     virtualenv: "{{ cifmw_repo_setup_basedir }}/venv"
     requirements: "{{ cifmw_repo_setup_basedir }}/tmp/repo-setup/requirements.txt"
-    virtualenv_command: "python3 -m venv"
```
SeanMooney, line 29: with pip 23.3 from /usr/lib/python3/dist-packages/pip (python 3.11)
"python3 -m venv" results in a venv with no packages installed.

so on fedora 39 this venv does not have setuptools installed 

so cmd: "{{ cifmw_repo_setup_basedir }}/venv/bin/python setup.py install" below fails.

i am currently hitting this on a downstream system that currently cant boot with rhel 9 or centos 9 stream. it looks like a issue with either the relevant kernel or the relvent internal images so I'm currently using fedora 39 as a workaround.

SeanMooney: oh i didnt know gibi had filed that yes ill mention that here.
this should also work with centos 9 stream or rhel 9.

the delta seams to be that on newer systems venv just does not include setuptools by default so that is why we only see this failing there.

cjeanner:
/approve

fixes #761

PR #890, opened 2023-12-05
Fix the path for cifmw_openshift_kubeconfig

raukadah:
In extracted crc jobs, the location of kubeconfig file is `~/.kube/config` on the controller. Let's use this one.

The crc provided kubeconfig does not exists on controller. So let's use the ~/.kube/config to avoid any issue.


cjeanner:
/approve

I'm all for converging - that said, we may even set that as the default value, at least for all multinode jobs. As a follow up maybe?

PR #891, opened 2023-12-05
Temporarly pin ansible-core

cjeanner:
ansible-core 2.15.7 is out, and apparently there was a change in the way
it passes down some parameters to the custom plugins.

This is mostly a workaround in order to get the time to fix ci_script
and, probably, ci_make plugins.

### DIFF
--- a/ansible-requirements.txt
+++ b/ansible-requirements.txt
@@ -1,4 +1,5 @@
 ansible>=7.0.0
+ansible-core<2.15.6
 importlib-metadata
 jsonschema  # MIT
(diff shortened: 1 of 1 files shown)


raukadah (approved):
/approve

cjeanner:
The patches triggers all of molecules and deploys. All of the finished jobs were green (molecule), and prow is back on green as well. In addition, we got a proof of work on a broken env. This is, again, a temporary fix, while we're working on the actual fix in the modules/plugins.

PR #896, opened 2023-12-05
Add sudo into nft

arxcruz:
For some reason, even using become: true, nft command fails because it's not using sudo. Adding sudo here to fix it. Probably some update in the nft package or in ansible that breaks it.

  - [ ] README in the role


ci_framework/roles/registry_deploy/tasks/main.yml
```diff
@@ -51,7 +51,7 @@
   when:
     - cifmw_rp_registry_firewall | bool
   ansible.builtin.command: >-
-    nft insert rule ip filter INPUT tcp dport {{ cifmw_rp_registry_port }} counter accept
+    sudo nft insert rule ip filter INPUT tcp dport {{ cifmw_rp_registry_port }} counter accept
```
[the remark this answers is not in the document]
bshewale: https://sf.hosted.upshift.rdu2.redhat.com/logs/83/448483/10/check/container-tcib-build-rhel-9-rhos-18-local-component-cinder/95f235f/job-output.txt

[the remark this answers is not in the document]
bshewale: 2023-12-05 09:39:29.142314 | TASK [registry_deploy : Open port for local registry]
2023-12-05 04:39:29.286888 | primary | Error: Could not process rule: No such file or directory
2023-12-05 04:39:29.286965 | primary | insert rule ip filter INPUT tcp dport 5001 counter accept

ci_framework/roles/registry_deploy/tasks/main.yml
```diff
@@ -50,8 +50,11 @@
     - bootstrap
   when:
     - cifmw_rp_registry_firewall | bool
-  ansible.builtin.command: >-
-    nft insert rule ip filter INPUT tcp dport {{ cifmw_rp_registry_port }} counter accept
+  ansible.builtin.command: "{{ item }}"
+  with_items:
+    - "nft add table ip filter"
+    - "nft add chain ip filter"
```
raukadah, line 56:
```suggestion
    - "nft add chain ip filter INPUT { type filter hook input priority 0 \\; }"
```

ci_framework/roles/registry_deploy/tasks/main.yml
cjeanner, line 1: as an FYI, edpm-ansible has a bootstrap ruleset available:
https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/roles/edpm_nftables/files/iptables.nft

maybe something to consider?

It creates the same layout as the iptables one so that we may even leverage iptables CLI if needed later to check rules and related.

raukadah: @cjeanner sounds good. We can add something similar in future as a follow up. Currently downstream component pipeline is broken. We need this to fix it.

ci_framework/roles/registry_deploy/tasks/main.yml
```diff
@@ -50,8 +50,11 @@
     - bootstrap
   when:
     - cifmw_rp_registry_firewall | bool
-  ansible.builtin.command: >-
-    nft insert rule ip filter INPUT tcp dport {{ cifmw_rp_registry_port }} counter accept
+  ansible.builtin.command: "{{ item }}"
+  with_items:
```
cjeanner, line 54: `loop` would be better, but won't block on that.

cjeanner (commented):
/approve

PR #897, opened 2023-12-05
fix: correctly print the excluded tests, if any

tosky:
When setting the list of excluded tests, The debug instruction was printing the list of allowed tests, not the excluded ones.

- [x] No testing executed, just code inspection, but the change looks trivial enough.

### DIFF
--- a/ci_framework/roles/tempest/tasks/tempest-tests.yml
+++ b/ci_framework/roles/tempest/tasks/tempest-tests.yml
@@ -92,6 +92,6 @@
         dest: "{{ cifmw_tempest_artifacts_basedir }}/exclude.txt"
         content: "{% for test in cifmw_tempest_tests_skipped %}{{ test }}\n{% endfor %}"
 
-    - name: Show tests to be executed
+    - name: Show tests to be excluded
       ansible.builtin.debug:
-        msg: "{{ cifmw_tempest_tests_allowed }}"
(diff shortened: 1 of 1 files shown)


tosky:
@arxcruz

ci_framework/roles/tempest/tasks/tempest-tests.yml
```diff
@@ -92,6 +92,6 @@
         dest: "{{ cifmw_tempest_artifacts_basedir }}/exclude.txt"
         content: "{% for test in cifmw_tempest_tests_skipped %}{{ test }}\n{% endfor %}"
 
-    - name: Show tests to be executed
```
cjeanner, line 95: so you'd rather want to print only the excluded ones? Maybe printing both listing would be better?

tosky: Please check the previous block, or rather the whole file. This file contains a set of 4 blocks, the first two handle the allowed and skipped tests when a skiplist is specified, the second two handle the allowed and skipped tests when a direct list of tests is specified. The last block printed the wrong information before this patch.

cjeanner: heh, probably a wild copy/paste - fine!

cescgina (commented):
/lgtm

PR #903, opened 2023-12-06
Update default value for the layout config

cjeanner:
While it would make more sense to remove any default value, we can't for
now due to jobs that may depend on it.

So this patch just ensure we don't show any unsupported parameter.

### DIFF
--- a/ci_framework/roles/libvirt_manager/defaults/main.yml
+++ b/ci_framework/roles/libvirt_manager/defaults/main.yml
@@ -42,8 +42,6 @@ cifmw_libvirt_manager_configuration:
       # GB
       memory: "{{ cifmw_libvirt_manager_compute_memory }}"
       cpus: "{{ cifmw_libvirt_manager_compute_cpus }}"
-      start_ip: 192.168.122.100
-      subnet_mask: 255.255.255.0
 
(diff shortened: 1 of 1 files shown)


cjeanner:
recheck

ci_framework/roles/libvirt_manager/defaults/main.yml
```diff
@@ -42,8 +42,6 @@ cifmw_libvirt_manager_configuration:
       # GB
       memory: "{{ cifmw_libvirt_manager_compute_memory }}"
       cpus: "{{ cifmw_libvirt_manager_compute_cpus }}"
-      start_ip: 192.168.122.100
-      subnet_mask: 255.255.255.0
```
marios, line 46: would be nice to have more context about the vars you are removing in the commit message e.g. "vars were deprecated after [1]"

marios (commented):
i have no context about the vars being removed so kind of hard to say if this is a good idea or not :)

but yes it removes two vars start_ip and subnet_mask

ci_framework/roles/libvirt_manager/defaults/main.yml
cjeanner: I don't even think it was used at some point actually... I just happen to have eyed in that file while doing the README update for the proper structure and saw those 2 params :/.

cjeanner:
recheck

raukadah:
recheck

PR #915, opened 2023-12-07
Corrections to the reproducer role and doc

cjeanner:
We need to get the kubeconfig on the crc node in order to edit its
configuration during the bootstrap step of the job.

Also, the inventory pointed in the doc was wrong.


raukadah (commented):
/approve

raukadah:
recheck error: unable to default to a user name: the server is currently unable to handle the request (get users.user.openshift.io ~)

PR #917, opened 2023-12-08
Kustomize hostName to align it with hostnames set

rabi:
NodeSet spec hostName should be aligned with the actual hostnames set on nodes.


gibizer (approved):
Looks good to me

PR #918, opened 2023-12-08
[devscripts] Switch to OVNKubernetes by default

karelyatin:
OVNKubernetes is the default since OpenShift 4.12+, and it supports both v4 and v6 ip_stack. Let's setup it by default in devscripts deployments.

  - [ ] README in the role


karelyatin:
Tested it and the Openshift deployment works fine.
But faced an issue with OpenStack setup where dnsmasq pods can't connect to dns at 192.168.122.1 network. So that needs to be checked.

karelyatin:
> Tested it and the Openshift deployment works fine. But faced an issue with OpenStack setup where dnsmasq pods can't connect to dns at 192.168.122.1 network. So that needs to be checked.

Got it working with below changes:-
oc patch network.operator cluster -p '{"spec":{"defaultNetwork":{"ovnKubernetesConfig":{"gatewayConfig":{"routingViaHost": true}}}}}' --type=merge

pablintino:
@psathyan @karelyatin we need this one in soon: https://github.com/openshift-metal3/dev-scripts/issues/1613

psathyan:
if end to end is working, let's get this PR merged

karelyatin:
recheck

karelyatin:
@pablintino @psathyan this should be ready now !!

ci_framework/roles/devscripts/vars/main.yml
```diff
@@ -39,7 +39,8 @@ cifmw_devscripts_config_defaults:
   base_domain: "openstack.lab"
   ntp_servers: "clock.corp.redhat.com"
   ip_stack: "v4"
-  network_type: "OpenShiftSDN"
+  network_type: "OVNKubernetes"
+  ovn_local_gateway_mode: "false"
```
psathyan, line 43: query: shouldn't this be `true` by default.

karelyatin: I kept same as Openshift default, false i.e shared mode and it works for most of the use cases. and enabled only on the scenario where we manage libvirt networks. My idea was to enable it only when really needed else not touch it.

psathyan: Thanks @karelyatin for the explanation.

psathyan (approved):
/lgtm

slawqo (approved):

pablintino:
Thanks to all the ones that reviewed this one @psathyan @slawqo. Thanks for the PR @karelyatin 
/approve

PR #919, opened 2023-12-11
Fix the typo in create_new_role target help message

raukadah:
It should be `ROLE_NAME` not ROLENAME.


cjeanner (approved):
/approve

PR #906, opened 2023-12-06
Wire up new data-plane-adoption-OSP-17-to-extracted-crc for zuul check

marios:
The new job was recently added with [1]

[1] https://review.rdoproject.org/r/c/rdo-jobs/+/50590

As a pull request owner and reviewers, I checked that:
- [ ] Appropriate testing is done and actually running
- [ ] Appropriate documentation exists and/or is up-to-date:
  - [ ] README in the role
  - [ ] Content of the docs/source is reflecting the changes

PR #921, opened 2023-12-11
Add go.mod for files: match in data plane adoption job for meta operator

marios:
```
    Add go.mod for files: match in data plane adoption job for meta operator
    
    We have data-plane-adoption-pipeline wired up in the openstack operator
    but it never runs because of the current files: match for the job which
    is specific to ci-framework. This add go.mod to files: so it will run
    when we bump any of the operators. See [1] related recent patch.
    
    [1] https://github.com/openstack-k8s-operators/ci-framework/pull/906
```

[1] https://github.com/openstack-k8s-operators/ci-framework/pull/906


cescgina (commented):
/lgtm

PR #922, opened 2023-12-11
Add ci_local_storage role

raukadah:
This role is similar to `make crc_storage` and `make crc_storage_cleanup` install_yamls target.
    
It will create Persistant volumes needed during openstack services deployment.

  - [x] README in the role

### DIFF
--- a/roles/ci_local_storage/README.md
+++ b/roles/ci_local_storage/README.md
@@ -1,0 +1,34 @@
+# ci_local_storage
+
+Ansible role to create persistent volumes on Openshift cluster needed
+by openstack services.
+
+## Privilege escalation
+If apply, please explain the privilege escalation done in this role.
+
+## Parameters
+* `cifmw_cls_basedir`:  (String) Base directory. Defaults to `"{{ cifmw_basedir | default(ansible_user_dir ~ '/ci-framework-data') }}"`
+* `cifmw_cls_manifests_dir`: (String) Directory in where OCP manifests will be placed. Defaults to `"{{ cifmw_manifests | default(cifmw_cls_basedir ~ '/artifacts/manifests') }}/storage"`
+* `cifmw_cls_storage_class`: (String) The name of the storage class. Defaults to `local-storage`.
+* `cifmw_cls_storage_capacity`: (String) Storage capacity in GB. Defaults to `10Gi`.
+* `cifmw_cls_local_storage_name`: (String) Name of the local storage name. Defaults to `/mnt/openstack`.
+* `cifmw_cls_pv_count`: (Int) Numbers of pvs to create. Defaults to `12`.
+* `cifmw_cls_storage_provisioner`: (String) Name of the storage provisioner. Defaults to `cifmw`.
+* `cifmw_cls_create_ee_storage`: (Bool) Param to create ee_storage. Defaults to `false`.
+* `cifmw_cls_namespace`: (String) The namespace where OCP resources will be installed. Defaults to `openstack`.
+* `cifmw_cls_action`: (String) Command to creating/deleting directories on worker node. Defaults to `mkdir`.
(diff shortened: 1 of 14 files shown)


roles/ci_local_storage/molecule/default/converge.yml
```diff
@@ -0,0 +1,72 @@
...
+        name: ci_local_storage
+        tasks_from: cleanup.yml
+
+    - name: Get all namespaces
+      kubernetes.core.k8s_info:
+        kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
+        kind: Namespace
+      register: _ns
+
+    - name: Assert that the openstack ns is deleted
```
cescgina, line 63: minor thing, but in the task name `openstack` should be  `cifmw_cl_namespace` right?

roles/ci_local_storage/tasks/main.yml
```diff
@@ -0,0 +1,87 @@
...
+          map(attribute="metadata.name")
+      }}
+
+- name: Generate node_storage script
+  ansible.builtin.template:
+    src: node_storage.sh.j2
+    dest: "{{ cifmw_cls_manifests_dir }}/node_storage.sh"
+    mode: "a+x"
+
+- name: Run node_storage script to create directories on worker node
```
pablintino, line 68: Do we really need a bash script for running an `oc` command in a loop?

rebtoor: yeah, I'd switch to an `ansible.builtin.shell` task with the `oc` loop. I guess we can also get rid of the messages (since we don't register the execution of the task) and just create or remove folders.

raukadah: Make sense, Done!

roles/ci_local_storage/molecule/default/converge.yml
raukadah: Done

roles/ci_local_storage/tasks/main.yml
```diff
@@ -0,0 +1,79 @@
...
+    api_key: "{{ cifmw_openshift_token | default(omit)}}"
+    context: "{{ cifmw_openshift_context | default(omit)}}"
+    kind: Node
+    label_selectors:
+      - node-role.kubernetes.io/worker
+  register: _node_info
+
+- name: Get all the worker nodes
+  ansible.builtin.set_fact:
+    cifmw_cls_nodes: >-
```
pablintino, line 56: Is this one internal? If it's only consumed here I'd say to prepend a _ as a prefix.

rebtoor: ^ with this fix in i'll give my lgtm.

raukadah: Done!

rebtoor (approved):

raukadah:
recheck

roles/ci_local_storage/README.md
```diff
@@ -0,0 +1,32 @@
...
+- hosts: localhost
+  vars:
+    cifmw_openshift_user: "kubeadmin"
+    cifmw_openshift_password: "12345678"
+    cifmw_openshift_kubeconfig: "{{ ansible_user_dir }}/.crc/machines/crc/kubeconfig"
+    ansible_user_dir: "{{ lookup('env', 'HOME') }}"
+    cifmw_path: "{{ ansible_user_dir }}/.crc/bin:{{ ansible_user_dir }}/.crc/bin/oc:{{ ansible_user_dir }}/bin:{{ ansible_env.PATH }}"
+  tasks:
+    - ansible.builtin.include_role:
+        name: ci_local_storage
```
rebtoor, line 32: Can you use a block code to improve readability?

raukadah: Done

PR #923, opened 2023-12-11
Enable documentation build for python libs

cjeanner:
#### Rename networking_mapping to net_map

The name was so long we hit the line length limit far too often,
especially when we had to indent them to fit some try/except block.

Also, shorter name means less risk of typo.

This patch also includes some changes to match flake8 and black linters.

#### Documentation building

Since we're getting proper docstring in our python module nested in
module_utils, let's expose them in sphinx.

This requires some changes in the modules directly related to the
custom libs import, to the way we check spelling, and to the doc build
itself, since we have to inject a mid-step, like we did for the
ansible-doc build.

Co-Authored-By: @pablintino


cjeanner:
Generated doc: https://ci-framework--923.org.readthedocs.build/en/923/module_utils/networking_mapping.html

cjeanner:
recheck

ci_framework/plugins/module_utils/net_map/ip_pools.py
```diff
@@ -1,15 +1,18 @@
 import ipaddress
 import typing
 
-from ansible_collections.cifmw.general.plugins.module_utils.networking_mapping import (
-    exceptions,
-    networking_definition,
-)
+try:
```
pablintino, line 4: This import hack here is my only concern. I'm ok with the entire PR, but for me, touching the "product" code just because we need it to build docs is an smell like when people touch the code to make unit tests. Something is not properly done. I know that in this case the limitation is in the way ansible handles imports, that need to be done using that ansible_collections.<namespace>.<collection> pattern, but, why can't us play with the PYTHONPATH to make it work without touching the code itself?

cjeanner: seems really related to the PYTHONPATH in fact: unless we nudge the content to something matching the `ansible_collections.foo.bar.super.long.name` while building the doc, we're a bit stuck with that.

I can add a comment explaining why this is done.
We'd need to mock the actual module in sphinxdoc - or "proxy" it... AFAIK we can do it with python, so maybe we can do it in the conf.py?

docs/source/conf.py
```diff
@@ -109,3 +111,6 @@
 
 # Output file base name for HTML help builder.
 # htmlhelp_basename = 'doc'
+
+# Exposes our custom module_utils content for docstring
+sys.path.append("../../ci_framework/plugins/module_utils")
```
cjeanner, line 116: @pablintino maybe we can tweak something here so that it loads with the same ansible_collections namespace or something?

ci_framework/plugins/module_utils/net_map/ip_pools.py
pablintino: Can be done there, but I was thinking on a simpler way, create an directory structure that matches the `ansible_collections/cifmw/general` pattern, symlink that general dir to the `ci_framework` directory and add the parent dir of the created `ansible_collections` to the PYTHONPATH. It should do the job.

PR #930, opened 2023-12-12
Create collection for CI framework

cescgina:
Modify run_ansible_test and setup_molecule scripts to consume the new
collection.

### DIFF
--- a/galaxy.yml
+++ b/galaxy.yml
@@ -1,0 +1,107 @@
+### REQUIRED
+# The namespace of the collection. This can be a company/brand/organization or product namespace under which all
+# content lives. May only contain alphanumeric lowercase characters and underscores. Namespaces cannot start with
+# underscores or numbers and cannot contain consecutive underscores
+namespace: cifmw
+
+# The name of the collection. Has the same character restrictions as 'namespace'
+name: general
+
+# The version of the collection. Must be compatible with semantic versioning
+version: 1.0.0
+
+# The path to the Markdown (.md) readme file. This path is relative to the root of the collection
+readme: README.md
+
+# A list of the collection's content authors. Can be just the name or in the format 'Full Name <email> (url)
+# @nicks:irc/im.site#channel'
+authors:
+  - Sandeepyadav93 (https://github.com/Sandeepyadav93)
+  - adrianfusco (https://github.com/adrianfusco)
+  - afazekas (https://github.com/afazekas)
+  - arxcruz (https://github.com/arxcruz)
+  - bshephar (https://github.com/bshephar)
+  - bshewale (https://github.com/bshewale)
+  - cescgina (https://github.com/cescgina)
+  - cjeanner (https://github.com/cjeanner)
+  - dasm (https://github.com/dasm)
+  - dsariel (https://github.com/dsariel)
+  - fmount (https://github.com/fmount)
+  - frenzyfriday (https://github.com/frenzyfriday)
+  - fultonj (https://github.com/fultonj)
+  - jpodivin (https://github.com/jpodivin)
+  - lewisdenny (https://github.com/lewisdenny)
+  - pablintino (https://github.com/pablintino)
+  - psathyan (https://github.com/psathyan)
+  - rabi (https://github.com/rabi)
+  - raukadah (https://github.com/raukadah)
+  - rebtoor (https://github.com/rebtoor)
+  - rlandy (https://github.com/rlandy)
+  - slagle (https://github.com/slagle)
+  - son-vyas (https://github.com/son-vyas)
+  - steveb (https://github.com/steveb)
+  - stuggi (https://github.com/stuggi)
+  - tosky (https://github.com/tosky)
+  - viroel (https://github.com/viroel)
+
+### OPTIONAL but strongly recommended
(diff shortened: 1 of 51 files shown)


galaxy.yml
```diff
@@ -0,0 +1,107 @@
+### REQUIRED
+# The namespace of the collection. This can be a company/brand/organization or product namespace under which all
+# content lives. May only contain alphanumeric lowercase characters and underscores. Namespaces cannot start with
+# underscores or numbers and cannot contain consecutive underscores
+namespace: cifmw
```
cjeanner, line 5: beware: we may need to update all the roles "meta/main.yml" to reflect that namespace. And the skeleton for new roles.

scripts/run_ansible_test
```diff
@@ -34,20 +34,21 @@ collection_path='/usr/share/ansible/collections/ansible_collections'
...
-        collection_path="${HOME}.ansible/collections/ansible_collections"
+        collection_path="${HOME}/.ansible/collections/ansible_collections"
         ;;
 esac
 
-mkdir -p ${HOME}/code/ansible_collections/cifmw/general
-rsync -a --delete ${PROJECT_DIR}ci_framework/* ${HOME}/code/ansible_collections/cifmw/general/
-ln -fs ${collection_path}/* ${HOME}/code/ansible_collections/
+mkdir -p ${HOME}/code
+ansible-galaxy collection install git+file://${PROJECT_DIR}.git -p ${HOME}/code
```
cjeanner, line 42: I think you don't want to keep the `.git` in there.

scripts/setup_molecule
```diff
@@ -48,6 +48,4 @@ ${PIP} install \
         -r "${PROJECT_DIR}molecule-requirements.txt" ${@:-}
 ${GALAXY} collection install --timeout=120 \
         -p "${COLLECTION_PATH}" \
-        -r "${PROJECT_DIR}requirements.yml"
-mkdir -p ${HOME}/.ansible/plugins
-cp -a ${PROJECT_DIR}plugins/* ${HOME}/.ansible/plugins/
+        git+file:${PROJECT_DIR}.git
```
cjeanner, line 51: ditto - removing the `.git` should help

galaxy.yml
cescgina: right, I see that roles use the `edpm`  namespace, should I changed it here to `edpm` or change the roles one to `cifmw`?

scripts/run_ansible_test
cescgina: thanks for the suggestion, testing it now

galaxy.yml
cjeanner: cifmw makes more sense. I chose the "edpm" more because... "why not converging in the same namespace", but we may want our own.

galaxy.yml
```diff
@@ -0,0 +1,107 @@
...
+
+# A list of tags you want to associate with the collection for indexing/searching. A tag name has the same character
+# requirements as 'namespace' and 'name'
+tags:
+  - cloud
+  - linux
+  - tools
+  - networking
+  - security
+  - RHOSP
```
cjeanner, line 67: RHOSO instead?

cescgina: true, will update in next round

galaxy.yml
cescgina: ack, I renamed all the roles namespaces to `cifmw` in a second commit, so it can be easily reverted

scripts/run_ansible_test
```diff
@@ -34,20 +34,21 @@ collection_path='/usr/share/ansible/collections/ansible_collections'
...
 
 # Create/append the sanity exceptions for the current ansible version
 ansible_version=$(python3 -c "import ansible; print(ansible.__version__)" | sed 's/\.[^.]*$//')
-cat  "${HOME}/code/ansible_collections/cifmw/general/tests/sanity/ignore.txt" >> \
-    "${HOME}/code/ansible_collections/cifmw/general/tests/sanity/ignore-${ansible_version}.txt"
+cat  "${HOME}/code/ansible_collections/cifmw/general/ci_framework/tests/sanity/ignore.txt" >> \
+    "${HOME}/code/ansible_collections/cifmw/general/ci_framework/tests/sanity/ignore-${ansible_version}.txt"
 
 pushd ${HOME}/code/ansible_collections/cifmw/general
+cp -r ci_framework/tests .
```
pablintino, line 50: With the collection structure I doubt we still need to separate the framework in the ci_framework dir and create symlinks... What's the point of that right now? @cjeanner

pablintino: PS: If we still need to have the ci_framework directory copying the tests directory is the only solution, as ansible-test doesn't follow symlinks.

cescgina: Can't confirm that symlinks don't work, I tried :) I kept the tests inside ci_framework for this first version because I was not sure whether it would break anything else, but I think it would be better to move them on level up to follow the conventional collection structure

cjeanner: We can do a follow-up in order to move tests in the repository.

cjeanner:
/lgtm

we got a proof of work via #937

PR #935, opened 2023-12-13
Use proper periodic repos in periodic jobs

raukadah:
Currently we use install_yamls dataplane_v1beta1_openstackdataplaneservice_reposetup.yaml service to lay down repos on edpm nodes. It will populate dlrn repos with current-podified antelope branch. We donot need these repos in periodic jobs. Otherwise we will issue mentioned in https://issues.redhat.com/browse/OSPCIX-132.

This pr fixes multiple things to address the above issue:
- We generate proper repos on controller, it needs to be copied on compute node in periodic EDPM jobs. Then we also need to disable repo-setup services coming from edpm_deploy_prep make target. It will make sure proper periodic repos are enabled on extracted crc edpm job.

- In BM job, we need to update the dataplane_v1beta1_openstackdataplaneservice_reposetup.yaml service file to pull proper repos based on branch and tag.

### DIFF
--- a/ci_framework/hooks/playbooks/fetch_compute_facts.yml
+++ b/ci_framework/hooks/playbooks/fetch_compute_facts.yml
@@ -1,4 +1,17 @@
 ---
+- name: Sync repos for controller to compute for periodic jobs
+  hosts: computes
+  gather_facts: true
+  tasks:
+    - name: Copy repositories from controller to computes
+      when:
+        - zuul is defined
+        - "'periodic' in zuul.job"
+      become: true
+      ansible.builtin.copy:
+        dest: "/etc/yum.repos.d/"
+        src: "{{ cifmw_basedir }}/artifacts/repositories/"
+
 - name: Build dataset hook
   hosts: localhost
   gather_facts: false
@@ -214,6 +227,12 @@
                   - op: add
                     path: /spec/nodeTemplate/ansible/ansibleVars/edpm_container_registry_insecure_registries
                     value: ["{{ content_provider_registry_ip }}:5001"]
+              {% endif %}
+
+              {% if zuul is defined and 'periodic' in zuul.job %}
+                  - op: remove
+                    path: /spec/services/0
+                    value: repo-setup
               {% endif %}
 
 


arxcruz (approved):

ci_framework/hooks/playbooks/fetch_compute_facts.yml
```diff
@@ -1,4 +1,17 @@
 ---
+- name: Sync repos for controller to compute for periodic jobs
```
SeanMooney, line 2: i dont think this is corerct either.

we should be running reposetup on all the edpm nodes either as its own service or
via its own service.

i thought install yamls was injecting this into the default nodeset already so this should not be requried.

SeanMooney (changes requested):

ci_framework/hooks/playbooks/fetch_compute_facts.yml
raukadah: Thank you @SeanMooney for reviewing this. [edpm_deploy_prep target](https://github.com/openstack-k8s-operators/install_yamls/blob/main/Makefile#L685) and [ci-framework edpm_deploy role](https://github.com/openstack-k8s-operators/ci-framework/commit/5cdd7276f523fdb082b33dc3fb1c9e836021fc34#diff-3a7b8227de258e968c75e9c48b1728ab96aace66d7cab681770b28caf08fba9aR42) install the [repo-setup service](https://github.com/openstack-k8s-operators/install_yamls/blob/main/devsetup/edpm/services/dataplane_v1beta1_openstackdataplaneservice_reposetup.yaml). But repo-setup service lay down current-podified repos which is good for check job. We donot need these repos for downstream component and RDO periodic integration line. That's why we are going via this approach.  It avoid the usage of maintaining multiple repo-setup service file at different promotion line.

cjeanner:
/approve

we have safeguards in order to target only periodic jobs, where we have to be able to specify more options to repo-setup.

The "right" thing would be to edit the reposetup service in order to be able to pass all of the needed data - maybe it will come? But for now, we can go with this way.

ci_framework/hooks/playbooks/fetch_compute_facts.yml
SeanMooney: the repo setup roles supprot configing product repos too not just the rdo ones and i belive they had support for compose in the past
so for downstream we would ideally just set ansible vars to state which we are enabling

SeanMooney: the docs for repo setup are a little lacking but they have some example of how you are meant to configure it

https://github.com/openstack-k8s-operators/repo-setup/tree/main/roles/repo_setup#example-playbook

it has support for RHSM to mange our OSP product version 
https://github.com/openstack-k8s-operators/repo-setup/tree/main/roles/repo_setup#rhsm
and i belive it should be possible to confire the rdo perodic and downstream component pipe like content in the same way.

it would be better not to have two ways to do this if we can avoid it.

SeanMooney: by the way i belive in infraread this functionality was implemented via rhos-release

https://github.com/redhat-openstack/infrared/tree/70758116ec087bdc540a713ee84c6305165776fd/infrared/common/roles/rhos-release

https://github.com/redhat-openstack/infrared/blob/70758116ec087bdc540a713ee84c6305165776fd/infrared/common/roles/rhos-release/defaults/main.yml


I'm not sure its a good thing for the ci framework to automated diffently unless customer wont use repo_setup 

looking at our dev preview documentation we seam to entirely skip how subscription manage and the yum repos are configured on the edpm nodes so that 

is a required part of the edpm deployment and its expected to be done as a openstackDataplaneService on each compute.

SeanMooney: ok in devpreview 1

there is an edpm_bootstrap_command that is defined in the ansibleVars

and you did

edpm_bootstrap_command: |
            subscription-manager register --username <subscription_manager_username> --password <subscription_manager_password>
            subscription-manager release --set=9.2
            subscription-manager repos --disable=*
            subscription-manager repos --enable=rhel-9-for-x86_64-baseos-eus-rpms --enable=rhel-9-for-x86_64-appstream-eus-rpms --enable=rhel-9-for-x86_64-highavailability-eus-rpms --enable=openstack-17.1-for-rhel-9-x86_64-rpms --enable=fast-datapath-for-rhel-9-x86_64-rpms --enable=openstack-dev-preview-for-rhel-9-x86_64-rpms

SeanMooney: it looks like repo_setup service is not required but customer will likely be expected to use edpm_bootstrap_command or a dataplaen service to do this so the ci frame should likely do the same.

@cjeanner what do you think ^

raukadah: @SeanMooney Thank you for input. 
repo-setup role calls [repo-setup](https://github.com/openstack-k8s-operators/repo-setup) to generate centos and dlrn repos for upstream and downstream. repo-setup role currently provides support for rhos-release, repo-setup tool and generating repos for component pipeline. That's why we are using this role heavily.

Thanks for the rhsm suggestion.
We can integrate subscription manager command in the repo-setup role. I can also extend this role to generate the `edpm_bootstrap_command` containing the content of repo files. This will help to avoid the use of additional repo-setup service file.

raukadah:
recheck

viroel:
Results from testproject:
periodic-podified-multinode-edpm-deployment-antelope-ocp-crc-cs9 https://review.rdoproject.org/zuul/build/f5049b9e4c0f4c678771903aba60eb00 : SUCCESS in 1h 09m 11s
periodic-podified-multinode-edpm-deployment-master-ocp-crc-cs9 https://review.rdoproject.org/zuul/build/12eafbb1ddb84245ace705a2b2d8b332 : SUCCESS in 1h 16m 30s

PR #936, opened 2023-12-13
Patch BM nodeset to update all packages during bootstrap phase

raukadah:
In baremetal EDPM job, we use EDPM image to create EDPM node. it has all the required packages pre-installed during image build process.

If an older version of package got included during the image build process. Then during edpm_deploy, bootstrap role is not going to update the package to latest.

It is not advisable to use latest package installation from edpm-ansible bootstrap ansible role.

For CI case, we use edpm_bootstrap_command to update all the packages.

### DIFF
--- a/ci_framework/roles/edpm_deploy_baremetal/tasks/main.yml
+++ b/ci_framework/roles/edpm_deploy_baremetal/tasks/main.yml
@@ -196,6 +196,20 @@
             "path": "/spec/nodeTemplate/ansible/ansibleVars/edpm_container_registry_insecure_registries",
             "value": [{{ content_provider_registry_ip }}:5001]}]'
 
+- name: Patch OpenStackDataPlaneNodeSet resource to update bootstrap packages
+  when: not cifmw_edpm_deploy_baremetal_dry_run
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  ansible.builtin.command:
+    cmd: >-
+      oc patch openstackdataplanenodeset openstack-edpm-ipam
+      -n {{ cifmw_install_yamls_defaults['NAMESPACE'] }}
+      --type json
+      -p '[{"op": "add",
+            "path": "/spec/nodeTemplate/ansible/ansibleVars/edpm_bootstrap_command",
+            "value": "sudo dnf -y update"}]'
+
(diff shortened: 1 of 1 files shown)


raukadah:
tested here: https://review.rdoproject.org/r/c/testproject/+/51049/1#message-e7e8012f7e79f7089ca47603f1b74b3c8bcbb528

ci_framework/roles/edpm_deploy_baremetal/tasks/main.yml
```diff
@@ -196,6 +196,20 @@
...
+    PATH: "{{ cifmw_path }}"
+  ansible.builtin.command:
+    cmd: >-
+      oc patch openstackdataplanenodeset openstack-edpm-ipam
+      -n {{ cifmw_install_yamls_defaults['NAMESPACE'] }}
+      --type json
+      -p '[{"op": "add",
+            "path": "/spec/nodeTemplate/ansible/ansibleVars/edpm_bootstrap_command",
+            "value": "sudo dnf -y update"}]'
+
```
SeanMooney, line 212: this might mask actrul reall issue with edpm-ansible.
so I'm not sure its correct to do this in the ci-framework

i think adding a dnf update to the bootstrap role in edpm-ansibele is actually the correct fix.

SeanMooney (changes requested):

ci_framework/roles/edpm_deploy_baremetal/tasks/main.yml
raukadah: thank you @SeanMooney , Sounds good. I will update the same in edpm-ansible.

raukadah: We will merge this one to unblock integration line and then we will revert back once edpm-ansible pr is ready.

cjeanner:
/approve

don't forget to revert once the proper fix in edpm-ansible is done.

ci_framework/roles/edpm_deploy_baremetal/tasks/main.yml
viroel: @raukadah please revert this one after you get edpm-ansible pr merged. Thanks

viroel:
/lgtm

Considering that the real fix will land at edpm-ansible and this can be reverted afterwards

PR #938, opened 2023-12-13
Add jq package the reproducer role to be installed in controller-0

dpinhas:
This is needed for crc_storage in stage3..4


psathyan (approved):
/lgtm

pablintino:
For now we relied on the download_tools play executed further to get it. Looks good to install it at the very beginning.
/lgtm

cjeanner:
/approve

Or maybe call that download_tools target sooner? It's called as a hook anyway.

PR #942, opened 2023-12-13
Quickstart

Akrog:
This PR fixes issues when running the lightweight deployment as described on the quickstart guide on host that have very few packages installed.

Issues found and fixed are:

- Hypervisor host needed `tar` and `rsync` to be present which was not mentioned in the docs.
- `setup_env` was failing when working on virtualenv because it was trying to use `pip` from the host
- A couple of roles were missing the `python-lxml` dependency because the ansible role documentation for the `community.libvirt` collection is incorrect.
- Change SSH key algorithm from `ed25519` to `ecdsa` to avoid failing with FIPS enabled.

This PR also changes the default user for the CRC VM so it uses `core`, which is the right user.

- [x] Appropriate testing is done and actually running: I deployed with the changed code
- [x] Appropriate documentation exists and/or is up-to-date: I added the dependencies for the hypervisor host

### DIFF
--- a/ci/playbooks/multinode-customizations.yml
+++ b/ci/playbooks/multinode-customizations.yml
@@ -129,12 +129,12 @@
 
     - name: Generate an ssh keypair
       ansible.builtin.command:
-        cmd: ssh-keygen -t ed25519 -f ~/.ssh/id_ed25519 -P ''
+        cmd: ssh-keygen -t ecdsa -f ~/.ssh/id_cifw -P ''
 
     - name: Get public key
       register: pub_key_slurp
       ansible.builtin.slurp:
-        path: "{{ ansible_user_dir }}/.ssh/id_ed25519.pub"
+        path: "{{ ansible_user_dir }}/.ssh/id_cifw.pub"
 
     - name: Register pub key as a fact
       ansible.builtin.set_fact:
--- a/docs/source/reproducers/03-zuul.md
+++ b/docs/source/reproducers/03-zuul.md
@@ -155,7 +155,7 @@ instance.
 
 #### Access to the virtual machines
 Your local ssh configuration file will be modified by the reproducer in order to offer
-an easy access to the nodes: `ssh controller-0`, `ssh core@crc-0`, `ssh compute-0` and
+an easy access to the nodes: `ssh controller-0`, `ssh crc-0`, `ssh compute-0` and
 so on will work right out of the box.
 
 ### Specific parameters
--- a/roles/devscripts/tasks/sub_tasks/14_user.yml
+++ b/roles/devscripts/tasks/sub_tasks/14_user.yml
@@ -32,7 +32,7 @@
     comment: "cifmw generated key for ocp access"
     path: >-
       {{ cifmw_devscripts_artifacts_dir }}/cifmw_ocp_access_key
-    type: "ed25519"
+    type: "ecdsa"
   register: cifmw_devscripts_ssh_key
 
 - name: Verify sudoers privileges
--- a/roles/reproducer/tasks/push_code.yml
+++ b/roles/reproducer/tasks/push_code.yml
@@ -7,6 +7,24 @@
   delegate_to: controller-0
   delegate_facts: true
   block:
+    - name: Check if repository directories already exist
+      when: job_id is defined
+      ansible.builtin.stat:
+        path: "{{ item.project.src_dir }}"
+      register: repos_dir_stats
+      with_items: "{{ zuul['items'] }}"
+
+    - name: Ensure we are not in the job_id branch
+      when:
+        - repos_dir_stats is defined
+        - item.1.stat.exists
+      ansible.builtin.git:
+        dest: "{{ item.0.project.src_dir }}"
+        repo: "https://{{ item.0.project.canonical_name }}"
+        version: "origin/main"
+        force: true
+      loop: "{{ zuul['items'] | zip(repos_dir_stats.results) | list }}"
+
     - name: Fetch zuul.items repositories
       ansible.builtin.git:
         dest: "{{ repo.project.src_dir }}"
--- a/hooks/playbooks/fetch_compute_facts.yml
+++ b/hooks/playbooks/fetch_compute_facts.yml
@@ -102,7 +102,7 @@
           vars:
             dns_servers: "{{ ((['192.168.122.10'] + ansible_facts['dns']['nameservers']) | unique)[0:2] }}"
             edpm_install_yamls_vars:
-              SSH_KEY_FILE: "{{ ansible_user_dir }}/.ssh/id_ed25519"
+              SSH_KEY_FILE: "{{ ansible_user_dir }}/.ssh/id_cifw"
               DATAPLANE_COMPUTE_IP: >-
                 {{
                   crc_ci_bootstrap_networks_out['compute-0'].default.ip4 |
--- a/roles/hive/README.md
+++ b/roles/hive/README.md
@@ -100,7 +100,7 @@ cifmw_hive_namespace: openstack
 cifmw_hive_baremetal:
   cluster_name: unittest-01
   install_config: "{{ ansible_user_dir }}/install_config.yml"
-  ssh_key: 'ssh-ed25519 ...REDACTED'
+  ssh_key: 'ssh-cifw ...REDACTED'
   ocp_image: "quay.io/openshift-release-dev/ocp-release:4.13.0-x86_64"
 
 cifmw_operator_build_push_registry: "default-route-openshift-image-registry.unittest-01.openstack.ccitredhat.com"
@@ -161,7 +161,7 @@ platform:
                 enabled: true
                 dhcp: true
 pullSecret: ''
-sshKey: 'ssh-ed25519 ...'
+sshKey: 'ssh-cifw ...'
 ```
 
 ## Reference
--- a/roles/hive/molecule/default/converge.yml
+++ b/roles/hive/molecule/default/converge.yml
@@ -56,11 +56,11 @@
             state: touch
             mode: "0644"
 
-        - name: Testing requirements - root ed25519 ssh key
-          register: root_ed25519_key
+        - name: Testing requirements - root ecdsa ssh key
+          register: root_cifw_key
           community.crypto.openssh_keypair:
-            path: "{{ cifmw_hive_artifacts_dir }}/id_ed25519_root"
-            type: ed25519
+            path: "{{ cifmw_hive_artifacts_dir }}/id_cifw_root"
+            type: ecdsa
 
         - name: Including the role with right values
           vars:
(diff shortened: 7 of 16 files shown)


fultonj:
This LGTM. What do you think @cjeanner @pablintino ?

pablintino:
@fultonj @Akrog I saw that error about python-lxml in Brno. Agree with adding rsync installation in the doc. Looks good.
I let @cjeanner nudge it tomorrow if he is ok

/approve

docs/source/quickstart/02_prepare_env.md
```diff
@@ -22,13 +22,19 @@ There are a few needed dependencies to install before starting consuming the fra
...
 
-On the hypervisor, please ensure you have:
+On the hypervisor host, please ensure you have:
 
 - a non-root user, with passwordless SSH access (use SSH keys)
 - `sudo` configuration allowing that non-root user to run any random command, with or without password
 - at least 400G of free space in /home
 - an up-to-date CentOS Stream 9 or RHEL-9.2 system
 
+Besides the `qemu-kvm` and the `libvirt` packages, you will also need to have `tar` and `rsync`:
```
cjeanner, line 32: qemu-kvm and libvirt should be installed via the libvirt_manager - there shouldn't be manual install at this point.

same goes for tar and rsync - imho we should inject those two in the ci_setup role - see https://github.com/openstack-k8s-operators/ci-framework/tree/main/ci_framework/roles/ci_setup/vars content for instance.

Akrog: I didn't automate its installation because sometimes setting the hypervisor requires extra manual steps to do it right.

I usually run `virt-host-validate` to check out the status after installing libvirt and qemu/kvm to see what's missing.

In my server I had to enable IOMMU and reboot before using ci-framework:
```
$ sudo grubby --update-kernel ALL --args intel_iommu=on
$ sudo grubby --update-kernel ALL --args iommu=pt
```

But I'm ok adding it to the automation.

cjeanner: we can also get "unzip" in the ci_setup package list. I happen to be redeploying a hypervisor from 0 right now and I'm seeing those missing packages as well.
Also, the "sudo" injected by the rhol_crc role isn't covering the whole setup anymore - guess we'll go with a full NOPASSWD:ALL - I'll set this up in the ci_setup...

Akrog: Now that I think about it, one of the steps talks about running `sudo virsh net-destroy default`, so if we are installing the packages, shouldn't we also run that command with ansible?

cjeanner: that `net-destroy default` is temporary. I have to get some time to check if we can re-use that default network as-is, or if we can change "our" custom net range to not overlap.

ci_framework/roles/ci_setup/vars/redhat.yml
```diff
@@ -4,6 +4,11 @@ cifmw_ci_setup_packages:
   - ca-certificates
   - git-core
   - make
+  - tar
+  - unzip
+  - rsync
+  - qemu-kvm
+  - libvirt
```
cjeanner, line 11: it should be in libvirt_manager role

ci_framework/roles/libvirt_manager/vars/main.yml
```diff
@@ -30,6 +30,7 @@ cifmw_libvirt_manager_dependency_packages:
   - qemu-kvm
   - libguestfs
   - guestfs-tools
+  - python-lxml
```
cjeanner, line 33: isn't it `python3-lxml` ?

ci_framework/roles/rhol_crc/vars/main.yml
```diff
@@ -24,6 +24,7 @@
 # https://access.redhat.com/documentation/en-us/red_hat_openshift_local/2.14/html/getting_started_guide/installation_gsg#required-software-packages_gsg
 cifmw_rhol_crc_dependency_packages:
   - NetworkManager
+  - python-lxml
```
cjeanner, line 27: ditto - `python3-lxml` ?

ci_framework/roles/ci_setup/vars/main.yml
```diff
@@ -22,6 +22,11 @@ cifmw_ci_setup_packages:
   - git-core
   - bash-completion
   - make
+  - tar
+  - unzip
+  - rsync
+  - qemu-kvm
+  - libvirt
```
cjeanner, line 29: ditto - should be in libvirt_manager role

ci_framework/roles/libvirt_manager/vars/main.yml
Akrog: `python-lxml` works fine because `python3-lxml` has a virtual provides, so this works: `dnf -y install python-lxml`.

But I'll change it for the sake of correctness.

ci_framework/roles/ci_setup/vars/redhat.yml
Akrog: OK

SeanMooney:
Instead of RSA keys we changed to ecdsa for the nova SSH key to work around the same issue since normal RSA key won't work in fips enforcing mode as far as I am aware. Ecdsa should work on all modes. We should probably use the same key type for all keys.

Akrog:
> Instead of RSA keys we changed to ecdsa for the nova SSH key to work around the same issue since normal RSA key won't work in fips enforcing mode as far as I am aware. Ecdsa should work on all modes. We should probably use the same key type for all keys.

As far as I know RSA (`rsa-sha2-256`) works just fine in FIPS mode, since the key length is 2048 bits which is the minimum for FIPS level 2 and 3.

Even if it works, consistency is a good enough reason to change it, thanks. :-)

ci/playbooks/multinode-customizations.yml
```diff
@@ -129,12 +129,12 @@
 
     - name: Generate an ssh keypair
       ansible.builtin.command:
-        cmd: ssh-keygen -t ed25519 -f ~/.ssh/id_ed25519 -P ''
```
cjeanner, line 132: why switch to ecdsa? ed25519 is usually flagged as better. Is it due to FIPS?
Otherwise, OK for a generic name, though it will prevent any "default behavior" in ssh.

cjeanner (changes requested):
I'm against changing the ssh key type in this patch. If really, we should create proper parameters so that ppl may change the key type, in the same way we can get global parameter consumed in the roles. You can check for the `cifmw_basedir` example - and mimic it with a `cifmw_ssh_keytype` for instance.

roles/reproducer/tasks/push_code.yml
```diff
@@ -7,6 +7,22 @@
...
   block:
+    - name: Check if repository directories already exist
+      when: job_id is defined
+      ansible.builtin.stat:
+        path: "{{ item.project.src_dir }}"
+      register: repos_dir_stats
+      with_items: "{{ zuul['items'] }}"
+
+    - name: Ensure we are not in the job_id branch
+      when: repos_dir_stats is defined and item.1.stat.exists
```
cjeanner, line 18: I'd rather see this syntax:
```YAML
- name: foobar
  when:
    - condition_1
    - condition_2
    - ....
```

docs/source/reproducers/03-zuul.md
```diff
@@ -155,7 +155,7 @@ instance.
 
 #### Access to the virtual machines
 Your local ssh configuration file will be modified by the reproducer in order to offer
-an easy access to the nodes: `ssh controller-0`, `ssh core@crc-0`, `ssh compute-0` and
```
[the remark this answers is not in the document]
cjeanner: ('k, not really obvious when you check the whole PR - but while checking patch par patch, the start_manage_vms.yml has an update to set the User in the ssh_config)

ci/playbooks/multinode-customizations.yml
Akrog: Exactly, ed25519 doesn't work with FIPS, and ecdsa is what nova is using.

docs/source/reproducers/03-zuul.md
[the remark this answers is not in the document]
Akrog: I'll update the PR message to reflect that we are fixing the user configuration in this commit https://github.com/openstack-k8s-operators/ci-framework/pull/942/commits/e5d098efbca7b6d021d08678b42aaea0cca54fef

roles/devscripts/tasks/sub_tasks/14_user.yml
```diff
@@ -32,7 +32,7 @@
     comment: "cifmw generated key for ocp access"
     path: >-
       {{ cifmw_devscripts_artifacts_dir }}/cifmw_ocp_access_key
-    type: "ed25519"
```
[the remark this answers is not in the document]
Akrog: Here is the dedicated patch https://github.com/openstack-k8s-operators/ci-framework/pull/942/commits/527b52aee5c5198058d7961df24e73048a740786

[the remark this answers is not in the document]
cjeanner: a follow-up would be to allow to chose the type imho.

Multiple patches in the same PR isn't that obvious unfortunately in the UI :(

roles/reproducer/tasks/push_code.yml
Akrog: I'll change it since I have to rebase the PR because is out-of-date with the base branch

containerfiles/README.md
```diff
@@ -18,7 +18,7 @@ In the examples bellow, the options will do:
 
 - `-w /home/prow`: change the working directory to /home/prow
 - `-e HOME=/home/prow`: switch *prow* user home directory location.
-- `-v ~/.ssh/id_ed25519:/home/prow/.ssh/id_ed25519`: exposes your private SSH key.
```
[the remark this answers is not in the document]
Akrog: I feel like that can be changed in another PR, I'm just trying to "make things" work with this one.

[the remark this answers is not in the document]
Akrog: If they just copy/paste things they could also have problems with missing `~/inventory.yml` and `~/custom_files`.

In any case, I don't really mind either way. What would you like to see here?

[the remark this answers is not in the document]
cjeanner: just don't touch that specific doc file? as said, this ssh key is to access the hypervisor, and isn't really used during the product deploy itself.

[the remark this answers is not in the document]
Akrog: OK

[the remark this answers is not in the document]
cjeanner: thanks :)

cjeanner:
/approve

As said, a nice follow-up would be to create a new param for the ssh key type, but since we're are least converging with Nova key type, it's mostly a nice to have at this point.

cjeanner:
It will need #998 - more specifically https://github.com/openstack-k8s-operators/ci-framework/pull/998/commits/cf3ae81cd2bea77193cffeef35024b56178407a0

Apparently the interface name changed, suspecting some update in CRC that weren't spotted until we trigger a job hitting the reproducer role.  I saw it while working on that other patch - sooo.... fix is on its way.

lewisdenny:
recheck

PR #945, opened 2023-12-14
Install cert-manager operator from community operators

raukadah:
Currently we are installing cert-manager from redhat-operators. Let's try using community-operators.


raukadah:
recheck

PR #949, opened 2023-12-14
Ensure we clean ssh config entries

cjeanner:
Until now, we were keeping the ~/.ssh/config entries, leading to
potential confusion and errors.

With this patch, we ensure we remove all of the entries, ensuring we are
back in a clean state.

In order to do so, we don't need to actual block content, using the
"mark" is sufficient.


rebtoor (approved):

raukadah (approved):
/approve

rebtoor (approved):

danpawlik:
recheck

cjeanner:
recheck

PR #950, opened 2023-12-14
Support TLS for RGW

katarimanojk:
With this patch, during cifmw ceph deployment, rgw deployment includes the TLS certificate if it is passed.

  - [x] README in the role

### DIFF
--- a/roles/cifmw_cephadm/templates/ceph_rgw.yml.j2
+++ b/roles/cifmw_cephadm/templates/ceph_rgw.yml.j2
@@ -13,6 +13,11 @@ spec:
   rgw_frontend_port: 8082
   rgw_realm: default
   rgw_zone: default
+{% if rgw_frontend_cert is defined %}
+  ssl: true
+  rgw_frontend_ssl_certificate: |
+    {{ rgw_frontend_cert | indent( width=4 ) }}
+{% endif %}
 ---
 service_type: ingress
 service_id: rgw.default
@@ -26,3 +31,7 @@ spec:
   virtual_ip: {{ cifmw_cephadm_rgw_vip }}
   virtual_interface_networks:
   - {{ cifmw_cephadm_rgw_network }}
+{% if rgw_frontend_cert is defined %}
+  ssl_cert: |
+    {{ rgw_frontend_cert | indent( width=4 ) }}
+{% endif %}
--- a/playbooks/ceph.yml
+++ b/playbooks/ceph.yml
@@ -253,6 +253,9 @@
         tasks_from: monitoring
       vars:
         cifmw_cephadm_monitoring_network: "{{ lookup('ansible.builtin.ini', 'public_network section=global file=' ~ cifmw_cephadm_bootstrap_conf) }}"
+        cifmw_cephadm_dashboard_crt: "{{ cifmw_cephadm_certificate }}"
+        cifmw_cephadm_dashboard_key: "{{ cifmw_cephadm_key }}"
+
 
     - name: Create cephfs volume
       ansible.builtin.import_role:
--- a/roles/cifmw_cephadm/README.md
+++ b/roles/cifmw_cephadm/README.md
@@ -57,15 +57,15 @@ need to be changed for a typical EDPM deployment.
  `cifmw_cephadm_certs`: The path on the ceph host where TLS/SSL certificates
    are located. It points to '/etc/pki/tls'
 
-* `cifmw_cephadm_dashboard_crt`: The SSL/TLS certificate signed by CA which is
-   an optional parameter. If it is provided, ceph dashboard will be configured
-   for SSL automatically. Certificate should be made available in
+* `cifmw_cephadm_certificate`: The SSL/TLS certificate signed by CA which is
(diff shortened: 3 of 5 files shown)


ci_framework/roles/cifmw_cephadm/tasks/rgw.yml
```diff
@@ -29,6 +29,20 @@
...
+  become: true
+  block:
+    - name : Get the certificate content
+      register: slurp_cert
+      ansible.builtin.slurp:
+        src: "{{ cifmw_cephadm_certificate }}"
+
+    - name : Set rgw_frontend_cert for rgw spec
+      set_fact:
+        rgw_frontend_cert: "{{ slurp_cert.get('content', '') | b64decode }}"
```
cjeanner, line 42: at this point we can go directly with `slurp_cert['content'] | b64decode` - there will be a content, unless previous task fails.

ci_framework/roles/cifmw_cephadm/tasks/rgw.yml
```diff
@@ -29,6 +29,20 @@
...
+  when: cifmw_cephadm_certificate | length > 0 and
```
cjeanner, line 43: care to push the `when` after the associated task `name` ? It will be easier to read.

ci_framework/roles/cifmw_cephadm/tasks/rgw.yml
```diff
@@ -29,6 +29,20 @@
     _hosts: "{{ _hosts|default([]) + [ hostvars[item][ceph_hostname_var] ] }}"
   loop: "{{ groups['edpm'] }}"
 
+- name: Get and save TLS certificate for rgw spec if provided
+  when: cifmw_cephadm_certificate | length > 0 and
+        cifmw_cephadm_key | length > 0
+  become: true
```
fmount, line 35: do we need `become: true` here?

ci_framework/roles/cifmw_cephadm/tasks/rgw.yml
```diff
@@ -29,6 +29,20 @@
...
+  block:
+    - name: Get the certificate content
+      register: slurp_cert
+      ansible.builtin.slurp:
```
fmount, line 39: `slurp` assumes you're reading a remote file: is that the case? I see the playbook is executed on `edpm[0]`, so we can simple use [1] to read the file content.
On the other side, I'm wondering what's happen when we have 3 different nodes w/ 3 diff certs: do you see any problem with this config?

[1] https://docs.ansible.com/ansible/latest/collections/ansible/builtin/file_lookup.html

ci_framework/roles/cifmw_cephadm/tasks/rgw.yml
katarimanojk: yeah it is not needed, i might have added during my experiments.
I will remove it in my next patchset.

roles/cifmw_cephadm/templates/ceph_rgw.yml.j2
```diff
@@ -13,6 +13,11 @@ spec:
   rgw_frontend_port: 8082
   rgw_realm: default
   rgw_zone: default
+{% if rgw_frontend_cert is defined %}
```
fmount, line 16: by default this is skipped if we don't pass/set a path to `cifmw_cephadm_certificate` || `cifmw_cephadm_key`

fmount (approved):
/lgtm

roles/cifmw_cephadm/templates/ceph_rgw.yml.j2
cjeanner: I'd double-ensure things by checking it's not empty either. I remember seeing some weird things with `set_fact` associated to a `when` (basically, the `block` instructs ansible to apply the same when/vars/environment to each of task nested in that block - it's evaluating them anyway).

fmount: ack, @katarimanojk do you want to test the same by having also `len(rgw_frontend_cert) > 0` ? This way we can avoid the situation described by Cedric.

katarimanojk: @fmount sure , i will consider the comment from cedric , test it and make changes accordingly.

cjeanner: just made a quick test locally:
```YAML
- hosts: localhost
  gather_facts: false
  tasks:
    - name: Maybe set fact
      when:
        - false
      ansible.builtin.set_fact:
        foo: 'bar'

    - name: Debug foo fact
      ansible.builtin.debug:
        var: foo
```

seems to work as it should:
```
TASK [Maybe set fact] *
skipping: [localhost]

TASK [Debug foo fact] *
ok: [localhost] => {
    "foo": "VARIABLE IS NOT DEFINED!"
}
```

same if we nest the task in a block. So my memory probably deserved me on that one.

cjeanner: (so., solving this one :) )

PR #952, opened 2023-12-14
libvirt manager/multi hypervisor

cjeanner:
#### Add support for multiple hypervisor in libvirt_manager

    This patch alone won't work on its own. But we need to get that feature
    in before we can work on the reproducer side in order to ensure we
    filter things properly.

#### Add support for multiple hypervisors to reproducer

    Here, we have to ensure we're running delegated tasks from the correct
    hypervisor. We also have to ensure we're deploying the OCP or CRC or
    controller-0 on the correct hypervisor.

#### Ensure we gather the various inventories

    Since the controller-0 is running on one specific hypervisor, we have to
    ensure we can collect and group all of the generated inventories.
    
    This patch ensures we're fetching other hypervisors inventories and copy
    them onto the hypervisor running controller-0 before we push everything
    onto the controller-0.
    
    That way, we should get all of the needed data for the future steps,
    meaning mostly "deploy".
    
    Note that networking may be an issue - but ensuring hypervisors are able
    to communicate with each others is not really in the scope of the
    ci-framework.

#### Add proper testing for multi-hypervisor

    We create a series of "fake" hosts in the molecule inventory in order to
    show we're able to deploy VM only on the targeted hypervisors.
    
    Doing so, we ensure we don't use too many resources, allowing to keep
    our CI footprint low.

  - [X] README in the role

### DIFF
--- a/ci_framework/roles/libvirt_manager/tasks/deploy_layout.yml
+++ b/ci_framework/roles/libvirt_manager/tasks/deploy_layout.yml
@@ -40,7 +40,6 @@
     label: "{{ item.key }}"
 
 - name: Create temporary ssh keypair
-  delegate_facts: true
   ansible.builtin.command:
     cmd: >-
       ssh-keygen -t ed25519
@@ -48,13 +47,11 @@
     creates: "{{ ansible_user_dir }}/.ssh/cifmw_reproducer_key"
 
 - name: Slurp public key for later use
-  delegate_facts: true
   register: pub_ssh_key
   ansible.builtin.slurp:
     path: "{{ ansible_user_dir }}/.ssh/cifmw_reproducer_key.pub"
 
 - name: Slurp private key for later use
-  delegate_facts: true
   register: priv_ssh_key
   ansible.builtin.slurp:
     path: "{{ ansible_user_dir }}/.ssh/cifmw_reproducer_key"
@@ -101,6 +98,12 @@
       }}
 
 - name: Create and run VMs
+  when:
+    - (
+        vm_data.value.target is defined and
+        vm_data.value.target == inventory_hostname
+      ) or
+      vm_data.value.target is undefined
   vars:
     vm_type: "{{ vm_data.key }}"
     pub_key: "{{ pub_ssh_key.content | b64decode }}"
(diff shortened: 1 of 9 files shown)


cjeanner:
/hold

I'd like to discuss it during the review call to clear any questions.

Note that proper doc with config example will be provided shortly - I need to get my local lab ready for that.

cjeanner:
Without the network, here's what I get now:
```
PLAY RECAP 
builder1                   : ok=437  changed=98   unreachable=0    failed=0    skipped=284  rescued=0    ignored=0   
builder2                   : ok=171  changed=34   unreachable=0    failed=0    skipped=253  rescued=0    ignored=0
```
with all the wanted things deployed at the right place. The main "issue" will be the networking, which is not in the scope of this review, nor really in the scope of ci-framework project itself - it will mostly be a documentation. And we'll probably go with network isolation based on actual NIC in the VMs instead of vlans. More on that in a follow-up *doc* patch once I have it right.

rebtoor (approved):
+2

ci_framework/roles/libvirt_manager/tasks/deploy_layout.yml
```diff
@@ -101,6 +98,12 @@
       }}
 
 - name: Create and run VMs
+  when:
```
pablintino, line 101: May make sense to add an assertion if the target is given but the host is not part of the inventory with a clean and clear failure?

ci_framework/roles/reproducer/tasks/configure_controller.yml
```diff
@@ -11,6 +11,12 @@
       }}
 
 - name: Configure controller-0
+  when:
```
pablintino, line 14: Same comment as before.

ci_framework/roles/libvirt_manager/tasks/deploy_layout.yml
cjeanner: for now it won't fail. Not sure we want to go that way - we may have one variable file shared across multiple env, and then rely on the inventory only to deploy what we need. Might be something to consider, but it may create more issues than wanted.

PR #953, opened 2023-12-14
Avoid too much verbosity in network creation

cjeanner:


raukadah (approved):
/approve

PR #955, opened 2023-12-15
Updating hosts so that the task runs on both controller and primary nodes

frenzyfriday:
Updating hosts so that the task runs on both controller and primary nodes. Using inclusion instead of exclusion as later when we change the nodesets to use "controller " instead of "primary" , we can simply remove primary from the hosts here.
  - [X] README in the role

### DIFF
--- a/ci/playbooks/dump_zuul_data.yml
+++ b/ci/playbooks/dump_zuul_data.yml
@@ -1,6 +1,6 @@
 ---
 - name: "Run ci/playbooks/dump_zuul_data.yml"
-  hosts: controller
+  hosts: controller, primary
   gather_facts: true
   tasks:
     - name: Create zuul-output directory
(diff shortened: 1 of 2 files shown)


frenzyfriday:
Context: We are trying to run a job that claims a hive cluster (in pre) runs some job and cleans up the claim in downstream. 
The job is parented off cifmw-base-minimal so that tasks from e2e-prepare.yml and dump_zuul_data.yml can be run utilized. 
The nodeset for this job does not have a controller (it has primary) . In this patch I am trying to get the tasks to run on both primary or controller.

Job def: https://code.engineering.redhat.com/gerrit/c/openstack/tripleo-ci-internal-jobs/+/448539
TP: https://code.engineering.redhat.com/gerrit/c/testproject/+/448536

lewisdenny (approved):
Sounds good and we can update in the future if needed to exclude other hosts.

frenzyfriday:
recheck

raukadah (approved):
/approve

PR #963, opened 2023-12-18
Ensure inventory supports multi-hypervisor

cjeanner:
With #952, we add the capability to get multiple hypervisor.

We therefore need to ensure the inventory is properly generated to allow
accesses from the controller-0.

Note that specific documentation about this feature will come shortly,
in a dedicated patch, with proper example for the hypervisor setup.

### DIFF
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -12,5 +12,6 @@ fact_caching_connection = ~/ansible_facts_cache
 fact_caching_timeout = 0
 inventory = inventory.yml
 pipelining = True
+any_errors_fatal = True
 [ssh_connection]
 ssh_args = -o ControlMaster=auto -o ControlPersist=60s
--- a/roles/libvirt_manager/tasks/clean_layout.yml
+++ b/roles/libvirt_manager/tasks/clean_layout.yml
@@ -108,6 +108,14 @@
         key: "{{ _pub_key['content'] | b64decode }}"
         state: absent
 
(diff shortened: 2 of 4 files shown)


rebtoor (approved):

rebtoor (approved):

ci_framework/roles/reproducer/tasks/configure_controller.yml
```diff
@@ -88,6 +87,49 @@
...
+        content: "{{ _ssh_key }}"
+        mode: "0600"
+        owner: "zuul"
+        group: "zuul"
+      loop: "{{ hostvars.keys() }}"
+      loop_control:
+        label: "{{ host }}"
+        loop_var: "host"
+
+    - name: Inject ProxyHost configuration
```
pablintino, line 111: Is this name/task code enough to explain why this task is needed? I know how the PR works cause you explain me the idea beforehand but I hardly doubt that I'm gonna be able to understand why is this needed and why is this mentioning ProxyHost without doing anything with/to it in the SSH config file.

cjeanner: right - might be worth a comment or two. Task name on its own is correct - but getting some context in comments may be good.

cjeanner:
recheck

raukadah (commented):
/approve

raukadah (approved):

PR #943, opened 2023-12-13
Fix the ci-framework collection

pablintino:
As a pull request owner and reviewers, I checked that:
- [x] Appropriate testing is done and actually running
- [x] Appropriate documentation exists and/or is up-to-date:
  - [x] README in the role
  - [x] Content of the docs/source is reflecting the changes

PR #970, opened 2023-12-20
Reconfig molecule tests include files

psathyan:
Post [#943](https://github.com/openstack-k8s-operators/ci-framework/pull/943) no molecule tests are triggered as the include file directory is a soft link.


cescgina (commented):
/lgtm I was working on the same thing, so let's get this in and I'll rebase

PR #981, opened 2023-12-22
Add retry & delay while creating dirs on worker node

raukadah:
While creating dirs on worker node, ci_local_storage role fails randomly. Re-running the same task fixes the issue.

As a temprory measure, we are adding retries to the task to fix the issue.


cescgina (commented):
/lgtm

rebtoor (approved):

PR #983, opened 2023-12-24
[devscripts] Enhance cleanup for CI token errors

dpinhas:
Improved cleanup to handle CI token issues by checking logs for errors.
In case CI token is invalid, deployment would fail and a cleanup
afterwards will try to clean non-existed directory.
This is ensuring `make clean` will be done only if the required
directory exists.

### DIFF
--- a/roles/devscripts/tasks/cleanup.yml
+++ b/roles/devscripts/tasks/cleanup.yml
@@ -29,10 +29,16 @@
     path: "{{ cifmw_devscripts_repo_dir }}"
   register: devscript_repodir
 
+- name: Gather the metal3-dev-env repo directory information.
+  ansible.builtin.stat:
+    path: "{{ cifmw_devscripts_config_defaults.working_dir }}/metal3-dev-env"
+  register: metal3_dev_env_repodir
+
 - name: Remove the deployed OpenShift platform.
   when:
     - not cifmw_devscripts_dry_run | bool
     - devscript_repodir.stat.exists
+    - metal3_dev_env_repodir.stat.exists
   community.general.make:
     chdir: "{{ cifmw_devscripts_repo_dir }}"
     target: clean


dpinhas:
recheck

roles/devscripts/tasks/03_install.yml
```diff
@@ -25,6 +25,17 @@
...
+- name: Gather the metal3-dev-env repo directory information
+  ansible.builtin.stat:
+    path: "{{ cifmw_devscripts_config_defaults.working_dir }}/metal3-dev-env"
+  register: metal3_dev_env_repodir
+
+- name: Overwrite cifmw_devscripts_make_target value
+  when:
+    - metal3_dev_env_repodir.stat.exists
+  ansible.builtin.set_fact:
+    cifmw_devscripts_make_target: redeploy
```
psathyan, line 37: There is very high probability that we would not hit this code because of the precheck occurring earlier.

https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/devscripts/tasks/00_pre_check.yml

If there exists a working cluster, we would skip this stage. If there is a non-working cluster deployment, cleanup is executed within the pre-check tasks.

dpinhas: redeploy method was ditched

roles/devscripts/tasks/cleanup.yml
```diff
@@ -29,6 +29,23 @@
     path: "{{ cifmw_devscripts_repo_dir }}"
   register: devscript_repodir
 
+- name: Check logs for token expiration
+  when:
+    - devscript_repodir.stat.exists
+  ansible.builtin.command: >-
+    egrep -r "The token provided is invalid or expired" {{ cifmw_devscripts_logs_dir }}
```
psathyan, line 36: The logs of `dev-scripts` wouldn't be available in this location `cifmw_devscripts_logs_dir`. It would be available in `(cifmw_devscripts_repo_dir, 'logs') | path_join)` . In the `01_install_requirements...log`

lewisdenny: It would be more "Ansibley" to use the find module[1] with the `contains` parameter. We don't have any example in the case sadly but if it works I think that would be cleaner. 

[1] https://docs.ansible.com/ansible/2.7/modules/find_module.html

dpinhas: Thanks for both comments, I've changed the proposed code to just check if directory exists without greping.

lewisdenny:
/lgtm

Thanks @dpinhas, the latest revision is much nicer!

psathyan (approved):
/lgtm

raukadah (approved):
/approve

PR #985, opened 2023-12-26
Add playbooks and job definition for ci-bootstrap jobs

viroel:
Adds a playbook to reconfigure kubelet, which lives in config repo[1], and defines a staging job that uses ci-bootstrap to configure networks between zuul nodes.
This PR converts crc-dnsmasq restart task to 2 tasks, stop and start. In this new job layout we are seeing many errors while restarting crc-dnsmasq, due to IP no yet available:

```
crc-74q6p-master-0 podman[3004]: Error: unable to start container
"7293b76c8f99e18623d437dfa941b6a82d984130612353c9a7b65a5100156ca1":
plugin type="bridge" failed (add): cni plugin bridge failed: failed
to allocate for range 0: requested IP address 10.88.0.8 is not
available in range set 10.88.0.1-10>
```

[1] https://github.com/rdo-infra/review.rdoproject.org-config/blob/master/playbooks/crc/reconfigure_kubelet.yaml

  - [x] README in the role

### DIFF
--- a/ci/playbooks/multinode-customizations.yml
+++ b/ci/playbooks/multinode-customizations.yml
@@ -64,10 +64,22 @@
         state: restarted
         name: crc-pre
 
-    - name: Restart dnsmasq
+    # Avoid 'state: restarted' due to issues with IP not available when crc-dnsmasq starts
+    - name: Stop dnsmasq
       become: true
       ansible.builtin.systemd:
-        state: restarted
+        state: stopped
+        name: crc-dnsmasq
+
+    - name: Make sure that crc-dnsmasq is not running
+      containers.podman.podman_container:
+        name: crc-dnsmasq
+        state: absent
+
+    - name: Start dnsmasq
+      become: true
+      ansible.builtin.systemd:
+        state: started
         name: crc-dnsmasq
 
     - name: Wait for CRC to be ready
--- a/zuul.d/base.yaml
+++ b/zuul.d/base.yaml
@@ -107,7 +107,7 @@
     parent: base-extracted-crc
     timeout: 10800
     attempts: 1
-    nodeset:
+    nodeset: &multinode_edpm_nodeset
       nodes:
         - name: controller
           label: cloud-centos-9-stream-tripleo-vexxhost-medium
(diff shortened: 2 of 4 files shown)


viroel:
/hold

testing here: https://review.rdoproject.org/r/c/testproject/+/48840

ci/playbooks/multinode-customizations.yml
```diff
@@ -64,10 +64,16 @@
         state: restarted
         name: crc-pre
 
-    - name: Restart dnsmasq
```
pablintino, line 67: If this is finally enough I'd drop a comment here, in the code, to avoid someone in the future to break this by changing this back to a restart.

zuul.d/base.yaml
```diff
@@ -197,6 +197,25 @@
...
+    name: cifmw-podified-multinode-edpm-ci-bootstrap-staging
+    parent: base-extracted-crc-ci-bootstrap-staging
+    timeout: 10800
+    attempts: 1
+    nodeset: *multinode_edpm_nodeset
+    irrelevant-files: *ir_files
+    required-projects: *multinode_edpm_rp
+    roles: *multinode_edpm_roles
+    pre-run:
+      - ci/playbooks/crc/reconfigure-kubelet.yml
```
pablintino, line 213: This thing was extracted from the bootstrap playbooks and nudged it here, isn't it? Just to confirm.

ci/playbooks/multinode-customizations.yml
viroel: Yeah, it seems to work better now. Added a comment. Tks

zuul.d/base.yaml
viroel: yes correct, and this one doesn' t need to be on trusted repos.
added that as a comment in commit message also.

ci/playbooks/multinode-customizations.yml
viroel: I had to add an additional check if crc-dnsmasq is absent, before starting again. Seems more stable now. Testproject ended with success 3x in a row:
https://review.rdoproject.org/r/c/testproject/+/48840

viroel:
/hold cancel

testproject results show 3x success: https://review.rdoproject.org/r/c/testproject/+/48840

raukadah (commented):
/approve

PR #986, opened 2023-12-27
[manage_secrets] Implement OSP secrets management

rebtoor:
  - [] README in the role

### DIFF
--- a/roles/manage_secrets/molecule/osp_secrets/converge.yml
+++ b/roles/manage_secrets/molecule/osp_secrets/converge.yml
@@ -1,0 +1,88 @@
+---
+# Copyright Red Hat, Inc.
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+
+- name: Converge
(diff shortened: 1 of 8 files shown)


roles/manage_secrets/molecule/osp_secrets/converge.yml
```diff
@@ -0,0 +1,87 @@
...
+# License for the specific language governing permissions and limitations
+# under the License.
+
+
+- name: Converge
+  hosts: all
+  vars:
+    cifmw_path: "{{ ansible_user_dir }}/.crc/bin:{{ ansible_user_dir }}/.crc/bin/oc:{{ ansible_user_dir }}/bin:{{ ansible_env.PATH }}"
+    cifmw_openshift_login_kubeconfig: "{{ ansible_user_dir }}/.crc/machines/crc/kubeconfig"
+    cifmw_manage_secrets_ospsecrets_list:
```
pablintino, line 23: Not sure how are we going to glue this with the rest? Thea idea is to generate this list in the new "edpm_prepare"?

roles/manage_secrets/templates/osp_secrets.yml.j2
```diff
@@ -0,0 +1,16 @@
+{% for secret in cifmw_manage_secrets_ospsecrets_list %}
+---
+
+kind: Secret
+apiVersion: v1
+data:
+  {% for username, password in secret.data.items() %}
```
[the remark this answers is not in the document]
rebtoor: yeah, `key`, `value` could be a safer choice here.

roles/manage_secrets/molecule/osp_secrets/converge.yml
rebtoor: yeah, exactly. For now we have only two elements in that list (the `osp-secret` which contains password for every controlplane services and the `octavia-ca-passhprase` which contains, surprise surprise, the passphrase for its CA).
In our next iteration we'll have a secret for each service so we can handle their lifecycle in a more granular way.

PR #990, opened 2023-12-29
Install gcc for dlrn reporting

bshewale:


raukadah (approved):
Need in downstream

PR #992, opened 2024-01-03
[networking_mapping] Add the Ansible wrapper role

pablintino:

### DIFF
--- a/plugins/action/ci_net_map.py
+++ b/plugins/action/ci_net_map.py
@@ -1,0 +1,325 @@
+#!/usr/bin/env python3
+
+# Copyright Red Hat, Inc.
+# Apache License Version 2.0 (see LICENSE)
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+import typing
+
+DOCUMENTATION = r"""
+---
+action: ci_net_map
+
+short_description: Maps the Networking Definition to a Networking Env. Definition
+
+description:
+- Takes the Networking Definition, and based on Ansible facts and a set of
+  rules, translates that info into a valid Networking Env. Definition.
+- This task plugin requires networking facts to be present for every
+  instance referenced in the Networking Definition.
+- Partial mappings are allowed to allow infrastructure bootstrapping based on the
+  mapping result even without already deployed instances.
+
+options:
+  networking_definition:
+    description:
+    - The Networking Definition as a dictionary.
+    type: dict
+    required: true
+  interfaces_info:
+    description:
+    - Dict that contains the list of MAC addresses of each instance to be mapped.
+    - Each dict entry is a list of dicts with at least a `mac` field.
+    - Each item in the list is an interface in a given infrastructure network.
+    - If not passed a partial map is done.
+    type: iterable
+    required: false
+  network_name:
+    description:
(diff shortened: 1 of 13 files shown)


roles/networking_mapper/README.md
cjeanner, line 1: please update the README :)

roles/networking_mapper/defaults/main.yml
```diff
@@ -0,0 +1,27 @@
...
+
+
+# All variables intended for modification should be placed in this file.
+# All variables within this role should have a prefix of "cifmw_networking_mapper"
+cifmw_networking_mapper_basedir: >-
+  {{
+    cifmw_basedir | default(ansible_user_dir ~ '/ci-framework-data')
+  }}
+cifmw_networking_mapper_infra_dir: "/etc/ci/env"
+cifmw_networking_mapper_networking_def_path: "{{ cifmw_networking_mapper_infra_dir }}/networking-definition.yml"
```
cjeanner, line 25: if you want to keep lines shorter, you may use the `path_join` filter. We should consider this (included myself) when we have long paths.

roles/networking_mapper/vars/main.yml
cjeanner, line 1: we can remove this file if empty

cjeanner (changes requested):

plugins/action/ci_net_map.py
```diff
@@ -0,0 +1,148 @@
...
+- This task plugin requires networking facts to be present for every 
+  instance referenced in the Networking Definition.
+- Partial mappings are allowed to allow infrastructure bootstrapping based on the
+  mapping result even without already deployed instances.
+
+options:
+  networking_definition:
+    description:
+    - The Networking Definition as a dictionary.
+    type: str
```
cescgina, line 30: Does this mean that `networking_definition` should be passed as a string containing a dictionary, or the type should be dict?

roles/networking_mapper/README.md
pablintino: Done.

viroel:
in a first pass, looks good

plugins/action/ci_net_map.py
```diff
@@ -0,0 +1,325 @@
...
+            dict(task_vars["hostvars"]),
+            task_vars["groups"],
+            options=networking_mapper.NetworkingMapperOptions.from_dict(task_args),
+        )
+
+        try:
+            mapping_result = (
+                mapper.map_complete(networking_definition, interfaces_info_dict)
+                if is_complete_map
+                else mapper.map_partial(networking_definition)
```
lewisdenny, line 317: If `interfaces_info` is not provided `else mapper.map_partial(networking_definition)` handles it? I see `interfaces_info` is optional, what is the use case for a partial map?

pablintino: Yes, it's handled internally properly. That use case is for scenarios where the bootstraping of resources like instances or the networking of the instance requires networking information such as MTU, network ranges, lack or presence of IPv6, etc. For those scenarios the mapper can be called twice, one first call to perform a best effort map without the instances (or part of them only) and a second call with the mac addresses of each instance that will output the entire mapping.

cjeanner (approved):

PR #998, opened 2024-01-04
Enable compute accesses from the controller-0

cjeanner:
Until now, we had to check the IP in the ansible inventory. While it
works "well" when it comes to mono-hypervisor, the ssh command is a bit
more complicated to tweak when it comes to multi-hypervisor.

Note: it also has another patch, fixing the default NIC for CRC node.


cjeanner:
recheck

cjeanner:
/hold

some templating issues - working on it right now

psathyan (approved):
/lgtm

lewisdenny (approved):

PR #995, opened 2024-01-04
Refactor: Simplify Ansible tasks for directory creation and variable inclusion

son-vyas:
Refactor: Simplify ansible tasks for directory creation and variable inclusion
- remove loop in directory creation task for 'kustomizations/controlplane'
- directly use 'cifmw_control_plane_ceph_backend_include_vars' in include_vars task without loop
- enhance playbook efficiency and readability by eliminating unnecessary loops

As a pull request owner and reviewers, I checked that:
- [x] Appropriate testing is done and actually running
- [x] Appropriate documentation exists and/or is up-to-date:
  - [x] README in the role
  - [x] Content of the docs/source is reflecting the changes

PR #1000, opened 2024-01-04
Fix used of cifmw_control_plane_ceph_backend_include_vars

viroel:
This PR is a follow up from #995, which converts some single element loops into a single var call, but ceph_backends still provides a list of a single element to it.

  - [X] README in the role

### DIFF
--- a/scenarios/centos-9/ceph_backends.yml
+++ b/scenarios/centos-9/ceph_backends.yml
@@ -1,6 +1,5 @@
 ---
-cifmw_control_plane_ceph_backend_include_vars:
-  - "{{ cifmw_basedir }}/artifacts/pre_deploy_ceph_deploy.yml"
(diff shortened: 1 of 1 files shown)


cjeanner:
/lgtm
wondering if there may be any usecase where we'd want to get a list? @fultonj or @fmount maybe?

viroel:
> /lgtm wondering if there may be any usecase where we'd want to get a list? @fultonj or @fmount maybe?

This is only used in a kustomization hook that we added, not used in HCI deploy. The idea is to include vars from previous hook (ceph-toy), when creating openstack storage backend kustomizations.[1]

[1] https://github.com/openstack-k8s-operators/ci-framework/blob/main/hooks/playbooks/control_plane_ceph_backends.yml#L9

fultonj (commented):
+2

lewisdenny (approved):
Thanks @viroel

raukadah (approved):
/approve

PR #1004, opened 2024-01-05
Remove unused adoption job

cescgina:
Remove data-plane adoption job that deploys standalone in a nested vm,
since we have a new job that supercedes it.

### DIFF
--- a/zuul.d/adoption.yaml
+++ b/zuul.d/adoption.yaml
@@ -152,11 +152,10 @@
                 ip: 172.19.0.100
                 config_nm: false
 
-# TODO(marios) remove this definition after new OSP17 job is running and stable
 - job:
-    name: cifmw-data-plane-adoption-github-rdo-centos-9-extracted-crc
-    parent: content-provider-data-plane-adoption-github-rdo-centos-9-extracted-crc
-    files: &dpa_job_files
+    name: cifmw-data-plane-adoption-OSP-17-to-extracted-crc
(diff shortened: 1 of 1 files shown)


cescgina:
recheck

zuul.d/adoption.yaml
```diff
@@ -152,11 +152,10 @@
                 ip: 172.19.0.100
                 config_nm: false
 
-# TODO(marios) remove this definition after new OSP17 job is running and stable
 - job:
-    name: cifmw-data-plane-adoption-github-rdo-centos-9-extracted-crc
-    parent: content-provider-data-plane-adoption-github-rdo-centos-9-extracted-crc
-    files: &dpa_job_files
+    name: cifmw-data-plane-adoption-OSP-17-to-extracted-crc
+    parent: content-provider-data-plane-adoption-OSP-17-to-extracted-crc
```
pablintino, line 157: Is not the firstime I hear complains (and I complain too) about using uppercase names here. Is really necessary?

cescgina: I have no strong opinion either way, but if we want to change it I would do it in a dedicated PR and duplicating it first, since there are other jobs that inherit from it, wdyt @marios?

marios: ack ... cleanup is definitely on the agenda for this quarter, including the job names and removing unnecessary duplicate definition (with only slightly different names)

marios (approved):

PR #1005, opened 2024-01-05
Add test-operator role

lpiwowar:
This patch introduced [test-operator](https://openstack-k8s-operators.github.io/test-operator/) role. The role first spawns a test-operator, then creates an instance of the Tempest CRD which initiates creation of tempest pod. Finally, it waits for the end of the tests execution and copies artifacts from the tempest pod to a predefined directory.

  - [x] README in the role

### DIFF
--- a/roles/test_operator/README.md
+++ b/roles/test_operator/README.md
@@ -1,0 +1,45 @@
+# Test operator
+
+Execute tests via the [test-operator](https://openstack-k8s-operators.github.io/test-operator/).
+
+## Parameters
+
+* `cifmw_test_operator_artifacts_basedir`: (String) Directory where we will have all test_operator related files. Default value: `{{ cifmw_basedir }}/tests/test_operator` which defaults to `~/ci-framework-data/tests/test_operator`
+* `cifmw_test_operator_namespace`: (String) Namespace inside which all the resources are created. Default value: `openstack`
+* `cifmw_test_operator_index`: (String) Full name of container image with index that contains the test_operator. Default value: `quay.io/openstack-k8s-operators/test-operator-index:latest`
+* `cifmw_test_operator_timeout`: (Integer) Timeout in seconds for the execution of the tests. Default value: `3600`
+* `cifmw_test_operator_logs_image`: (String) Image that should be used to collect logs from the pods spawned by the test-operator. Default value: `quay.io/quay/busybox`
+* `cifmw_test_operator_concurrency`: (Integer) Tempest concurrency value. Default value: `8`
+* `cifmw_test_operator_cleanup`: (Bool) Delete all resources created by the role at the end of the testing. Default value: `false`
+* `cifmw_test_operator_default_groups`: (List) List of groups in the include list to search for tests to be executed. Default value: `[ 'default' ]`
+* `cifmw_test_operator_default_jobs`: (List) List of jobs in the exclude list to search for tests to be excluded. Default value: `[ 'default' ]`
+* `cifmw_test_operator_dry_run`: (Boolean) Whether test-operator should run or not. Default value: `false`
+
+* `cifmw_test_operator_tempest_registry`: (String) The registry where to pull tempest container. Default value: `quay.io`
+* `cifmw_test_operator_tempest_namespace`: (String) Registry's namespace where to pull tempest container. Default value: `podified-antelope-centos9`
+* `cifmw_test_operator_tempest_container`: (String) Name of the tempest container. Default value: `openstack-tempest`
+* `cifmw_test_operator_tempest_image`: (String) Tempest image to be used. Default value: `{{ cifmw_test_operator_tempest_registry }}/{{ cifmw_test_operator_tempest_namespace }}/{{ cifmw_test_operator_tempest_container }}`
+* `cifmw_test_operator_tempest_image_tag`: (String) Tag for the `cifmw_test_operator_tempest_image`. Default value: `current-podified`
+* `cifmw_test_operator_tempest_include_list`: (String) List of tests to be executed. Setting this will not use the `list_allowed` plugin. Default value: `''`
+* `cifmw_test_operator_tempest_exclude_list`: (List) List of tests to be skipped. Setting this will not use the `list_skipped` plugin. Default value: `''`
+* `cifmw_test_operator_tempest_tests_include_override_scenario`: (Boolean) Whether to override the scenario `cifmw_tempest_tests_allowed` definition. Default value: `false`
+* `cifmw_test_operator_tempest_tests_exclude_override_scenario`: (Boolean) Whether to override the scenario `cifmw_tempest_tests_skipped` definition. Default value: `false`
+
+# Please refer to https://openstack-k8s-operators.github.io/test-operator/guide.html#executing-tempest-tests
+* `cifmw_test_operator_tempest_config`: (Object) Definition of Tempest CRD instance that is passed to the test-operator. Default value:
+```
+  apiVersion: test.openstack.org/v1beta1
+  kind: Tempest
+  metadata:
+    name: tempest-tests
+    namespace: "{{ cifmw_test_operator_namespace }}"
+  spec:
+    containerImage: "{{ cifmw_test_operator_tempest_image }}:{{ cifmw_test_operator_tempest_image_tag }}"
+    tempestRun:
+      includeList: |
+        {{ cifmw_test_operator_include_list | default('') }}
+      excludeList: |
+        {{ cifmw_test_operator_exclude_list | default('') }}
+      concurrency: "{{ cifmw_test_operator_concurrency }}"
+    tempestconfRun: "{{ cifmw_tempest_tempestconf_config | default(omit) }}"
+```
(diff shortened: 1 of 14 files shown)


roles/test_operator/README.md
```diff
@@ -0,0 +1,45 @@
+# Test operator
```
marios, line 1: forgot to ask... so is the plan that we wire this up in those jobs that are currently running tempest? i.e. they should start using this role instead?

lpiwowar: Yes, the goal would be to execute all the tests via the test-operator. We plan to have a demo for the test-operator at the end of the month.

marios (approved):
stared again lgtm

roles/test_operator/tasks/main.yml
```diff
@@ -0,0 +1,210 @@
...
+
+- name: Wait for tempest job to start
+  register: tempest_job
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  delay: 10
+  retries: 20
+  ansible.builtin.command: |
+    oc get job {{ cifmw_test_operator_tempest_name }}
```
viroel, line 111: ditto

viroel (commented):
Thanks for proposing this PR, some comments inline, it looks good at first pass

roles/test_operator/tasks/main.yml
```diff
@@ -0,0 +1,210 @@
...
+- name: Check that test-operator is running # noqa risky-shell-pipe
+  register: test_operator_status
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  until: test_operator_status.stdout is search("Running")
+  delay: 10
+  retries: 20
+  ansible.builtin.shell: |
+    oc get pods | grep {{ cifmw_test_operator_controller_name }}
```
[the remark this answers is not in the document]
pablintino: +1. We have a tone of examples for this. 
I'd suggest to pass the target namespace explicitly to avoid relying on the one set in kubeconfig by who knows which command (probably, the last `oc new-project`.

roles/test_operator/tasks/main.yml
```diff
@@ -0,0 +1,210 @@
...
+    PATH: "{{ cifmw_path }}"
+  ansible.builtin.command: |
+    oc get pod {{ tempest_pod_name.stdout }} -o=jsonpath='{.status.phase}'
+  when: not cifmw_test_operator_dry_run | bool
+
+- name: Delete all resources created by the role
+  ansible.builtin.include_tasks: cleanup.yml
+  when: cifmw_test_operator_cleanup | bool and not cifmw_test_operator_dry_run | bool
+
+- name: Fail if podman container did not succeed
```
pablintino, line 206: podman?

roles/test_operator/tasks/cleanup.yml
```diff
@@ -0,0 +1,111 @@
...
+
+- name: Delete ClusterServiceVersion
+  kubernetes.core.k8s:
+    kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
+    api_key: "{{ cifmw_openshift_token | default(omit)}}"
+    context: "{{ cifmw_openshift_context | default(omit)}}"
+    kind: ClusterServiceVersion
+    state: absent
+    api_version: operators.coreos.com/v1alpha1
+    name: test-operator.v0.0.1
```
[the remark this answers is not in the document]
lpiwowar: Yes, I updated the code. Thank you!

roles/test_operator/defaults/main.yml
```diff
@@ -0,0 +1,57 @@
...
+cifmw_test_operator_tempest_image: "{{ cifmw_test_operator_tempest_registry }}/{{ cifmw_test_operator_tempest_namespace }}/{{ cifmw_test_operator_tempest_container }}"
+cifmw_test_operator_tempest_image_tag: current-podified
+cifmw_test_operator_timeout: 60
+cifmw_test_operator_logs_image: quay.io/quay/busybox
+cifmw_test_operator_concurrency: 8
+cifmw_test_operator_cleanup: false
+cifmw_test_operator_dry_run: false
+
+# Please refer to https://openstack-k8s-operators.github.io/test-operator/guide.html#executing-tempest-tests
+cifmw_test_operator_config:
```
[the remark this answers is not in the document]
lpiwowar: >  Is expected that other tools will be added to test-operator?

Yes, it is expected. We are in the middle of adding tobiko to the test-operator. 

> If yes, would be good to rename these vars to include _tempest_ to avoid renaming in the future?

Thank you, good point. I updated the code.

roles/test_operator/tasks/main.yml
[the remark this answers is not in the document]
lpiwowar: > Isn' t possible to do this with kubernetes.core.k8s_info instead?

Yes, it is possible. I just couldn't figure out how to do a partial match via the k8s_info at first. 

> I'd suggest to pass the target namespace explicitly to avoid relying on the one set in kubeconfig by who knows which command (probably, the last oc new-project.

Done. Thank you!

roles/test_operator/tasks/main.yml
lpiwowar: Done

kopecmartin (approved):
looks good to me, thanks Lukas!

roles/test_operator/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+Execute tests via the [test-operator](https://openstack-k8s-operators.github.io/test-operator/).
+
+## Parameters
+
+* `cifmw_test_operator_artifacts_basedir`: (String) Directory where we will have all test_operator related files. Default value: `{{ cifmw_basedir }}/tests/test_operator` which defaults to `~/ci-framework-data/tests/test_operator`
+* `cifmw_test_operator_namespace`: (String) Namespace inside which all the resources are created. Default value: `openstack`
+* `cifmw_test_operator_index`: (String) Full name of container image with index that contains the test_operator. Default value: `quay.io/openstack-k8s-operators/test-operator-index:latest`
+* `cifmw_test_operator_timeout`: (Integer) Timeout in seconds for the execution of the tests. Default value: `3600`
+* `cifmw_test_operator_logs_image`: (String) Image that should be used to collect logs from the pods spawned by the test-operator. Default value: `quay.io/quay/busybox`
```
afazekas, line 11: NOTE: even tough it is small image, at this point you can expect our base image used by all openstack 
pod is already cached on all nodes.

roles/test_operator/tasks/main.yml
```diff
@@ -0,0 +1,248 @@
...
+      apiVersion: v1
+      kind: Pod
+      metadata:
+        name: test-operator-logs-pod
+        namespace: "{{ cifmw_test_operator_namespace }}"
+      spec:
+        containers:
+          - name: test-operator-logs-container
+            image: "{{ cifmw_test_operator_logs_image }}"
+            command: ["/bin/sh", "-c", "--"]
```
afazekas, line 170: in the command you typically specify only the arg0  "/bin/sh", and the rest of the args  go to args.

busybox sleep infinite # works too , no need for loop.

you might want to use oc run  with overrides:
https://gist.github.com/welshstew/4f9532fd8737de1a43277c9154caa95e

You can specify a container which  oc run --overrides tar -c mydir | tar -x  .
In this case you can run a container which runs tar and have the pv mount and just getting the output stream in single command.

roles/test_operator/tasks/main.yml
```diff
@@ -0,0 +1,248 @@
...
+- name: Get logs from test-operator-logs-pod
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  ansible.builtin.shell: >
+    oc cp -n {{ cifmw_test_operator_namespace }} openstack/test-operator-logs-pod:mnt/
+    {{ cifmw_test_operator_artifacts_basedir }}
+  when: not cifmw_test_operator_dry_run | bool
+
+- name: Get list of all pods
```
afazekas, line 204: filtering isn't allowed here ?

afazekas (commented):
Few notes inside.

roles/test_operator/tasks/main.yml
lpiwowar: I think it isn't. Or at least I wasn't able to figure out how. I need to find a partial match because the name of the pod is: `tempest-tests-[random_suffix]` and partial match is probably not supported.

roles/test_operator/tasks/main.yml
lpiwowar: I updated the code so that I use the sleep infinity instead and fixed the `command:` and `args:` sections.

I will take a look at the ` oc run --overrides tar -c mydir | tar -x`. At this point, I just want to get this role merged and this can be addressed without breaking the backward compatibility when I will be incorporating tobiko to the role. I hope it is ok.

roles/test_operator/README.md
lpiwowar: Do you know please where is the base image stored?

cjeanner: Keeping the public one by default is better, it can be overridden by a ci scenario file.

roles/test_operator/tasks/main.yml
afazekas: Isn't the pods has some sot of annotation or lable what you can use ?

PR #1007, opened 2024-01-08
Correct/update file location for generated content

cjeanner:
We have to be consistent with the content we create and their location.

We also take the opportunity to rename the "mac-mapping.yml" into a more
generic name, "interfaces-info.yml".
For the sake of consistency, we now use lowercase indexes (IP -> ip, MAC
-> mac)

Lastly, the `interfaces-info.yml` file drops its top-level parameter
name. This allows anyone to load the parameters into their own, custom
parameter name.


psathyan (approved):
/lgtm

PR #1009, opened 2024-01-08
Explain how to consume multiple hypervisors

cjeanner:
While there's close to no chance anyone will use this feature with CRC,
it's always good to show it.

  - [X] README in the role

### DIFF
--- a/docs/source/files/bootstrap-hypervisor.yml
+++ b/docs/source/files/bootstrap-hypervisor.yml
@@ -1,0 +1,87 @@
+---
+- hosts: "{{ cifmw_target_host | default('all') }}"
+  vars:
+    _user: "{{ remote_user | default('zuul') }}"
+  tasks:
+    - name: Create user on remote host
+      ansible.builtin.user:
+        name: "{{ _user }}"
+        state: present
+
+    - name: Look for local ssh pub key
+      delegate_to: localhost
+      block:
+        - name: Look for id_rsa.pub file
+          ansible.builtin.stat:
+            path: "{{ lookup('env', 'HOME') }}//.ssh/id_rsa.pub"
+          register: rsa_pub_key
+
+        - name: Look for id_ed25519.pub file
+          delegate_to: localhost
+          ansible.builtin.stat:
+            path: "{{ lookup('env', 'HOME') }}//.ssh/id_ed25519.pub"
+          register: ed_pub_key
+
+    - name: Assert at least one public key exists
+      ansible.builtin.assert:
+        that:
+          - rsa_pub_key.stat.exists or ed_pub_key.stat.exists
+        quiet: true
+        msg: |
+          FATAL: you need to have at least one public key in your ~/.ssh directory.
+          Please generate one using either ssh-keygen -t ed25519 -b 512 or
+          ssh-keygen -t rsa -b 4096.
+
+    - name: Add RSA pub key to authorized keys
+      when: rsa_pub_key.stat.exists
+      ansible.posix.authorized_key:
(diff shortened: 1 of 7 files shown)


docs/source/files/bootstrap-hypervisor.yml
```diff
@@ -0,0 +1,89 @@
+---
+- hosts: "{{ cifmw_target_host | default('all') }}"
+  vars:
+    _user: "{{ remote_user | default('zuul') }}"
+    _rhos_release: "{{ use_rhose_release | default(true) }}"
```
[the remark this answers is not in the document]
psathyan: `use_rhose_release` could we add some information about this variable.

docs/source/files/bootstrap-vxlan.yml
```diff
@@ -0,0 +1,34 @@
+---
+- name: Prepare hypervisors vxlan
+  hosts: "{{ cifmw_target_host | default('localhost') }}"
+  gather_facts: true
+  become: true
+  tasks:
+    - name: Open VXLAN port
```
psathyan, line 7: IMO, we could check if firewall services is running before opening the port.

docs/source/files/bootstrap-hypervisor.yml
```diff
@@ -0,0 +1,89 @@
...
+      ansible.posix.authorized_key:
+        user: "{{ _user }}"
+        state: present
+        key: "{{ lookup('file', lookup('env','HOME') + '/.ssh/id_ed25519.pub') }}"
+
+    - name: Grant sudo privileges to remote user
+      ansible.builtin.copy:
+        content: |
+          "{{ _user }}" ALL=(ALL) NOPASSWD:ALL
+        dest: /etc/sudoers.d/zuul
```
psathyan, line 55: I think it should be `"/etc/sudoers.d/{{ _user }}"

docs/source/files/bootstrap-vxlan.yml
cjeanner: hmm maybe. I didn't have to do this on a freshly installed Stream node though... That playbook is mostly an example of what ppl can do. To be fair, I'd rather see them get a second NIC on a physical, dedicated switch rather than this vxlan thingy....
Let's keep it simple and see how it goes?

docs/source/files/bootstrap-hypervisor.yml
cjeanner: indeed - also removed one last hard-coded zuul mention in a task name.

psathyan (approved):
/lgtm

docs/source/files/bootstrap-hypervisor.yml
```diff
@@ -0,0 +1,87 @@
...
+          - git
+          - make
+          - libvirt
+          - libvirt-client
+          - libvirt-daemon
+          - libvirt-daemon-kvm
+          - virt-install
+          - qemu-kvm
+          - libguestfs
+          - guestfs-tools
```
raukadah, line 70: In downstream, this part will fail. On Remote RHEL machines, we donot have repos configured. May be we can add a message to RHEL user to configure repo and re-run this playbook. What do you say?

cjeanner: this is upstream. Downstream has its own doc for that, aiming OCP cluster and internal rhos-release things.

cjeanner: (so mostly, upstream is aiming centos hosts)

raukadah (approved):
/approve

PR #1011, opened 2024-01-08
[networking_mapping] Fix group nets duplication check and test definition

pablintino:
[networking_mapping] Fix group nets duplication check and test definition


cjeanner:
/approve

mostly due to the fact we're exposing interface data as a list of networks, not a dict.

raukadah (approved):

PR #1013, opened 2024-01-09
Tobiko role

eduolivares:
This new role executes tobiko tests using the corresponding tobiko image created by tcib:
https://github.com/openstack-k8s-operators/tcib/tree/main/container-images/tcib/base/os/tobiko

  - [X] README in the role


eduolivares:
@slawqo, I am thinking the logic that implements tobiko workflows in legacy CI should be implemented in ci-fmw now. Hence, the tobiko role needs to accept a list of options like this:
```
ci_tobiko_workflow:
  - # step1
    ci_tobiko_testenv: scenario
  - # step 2
    ci_tobiko_testenv: faults
    ci_tobiko_pytest_addopts: "--maxfail 1"
  - # step 3
    ci_tobiko_testenv: scenario
    ci_tobiko_prevent_create: true
  ```

Maybe that could be added in a later PR.

File: roles/tobiko/tasks/main.yml [hunk 1]
```diff
@@ -0,0 +1,102 @@
...
+    network: host
+    volume:
+      - "{{ cifmw_tobiko_artifacts_basedir }}/:/var/lib/tobiko/external_files:Z"
+      - "{{ cifmw_tobiko_artifacts_basedir }}/tls-ca-bundle.pem:/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem:Z"
+    detach: false
+    dns: "{{ cifmw_tobiko_dns_servers }}"
+    env:
+      TOBIKO_DEBUG: "{{ ci_tobiko_debug }}"
+      TOBIKO_TESTENV: "{{ ci_tobiko_testenv }}"
+      TOBIKO_PREVENT_CREATE: "{{ ci_tobiko_prevent_create if ci_tobiko_prevent_create is not none else omit }}"
```
pablintino, line 70: Omit is not going to work here I think. The value of the variable will be the "omit_<HASH>" calculated randomly by Ansible when starting. omit works only for module arguments.

roles/tobiko/tasks/configure-tobiko.yml
```diff
@@ -0,0 +1,31 @@
...
+  register: keystone_data
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  ansible.builtin.command:
+    cmd: oc get keystoneapi keystone -n openstack -o json
+
+- name: Set keystone vars
+  ansible.builtin.set_fact:
+    keystone_secret_name: "{{ keystone_data.stdout | from_json | community.general.json_query('spec.secret') }}"
```
pablintino, line 11: json_query for this simple dict access is an overkill that will force the running host to have jmespath in place

roles/tobiko/tasks/cleanup.yml
pablintino, line 1: Uhmm this is a duplication of what we do in the os_net_setup role. Maybe is time to do the thing I suggested here https://issues.redhat.com/browse/OSPRH-1670

deploy-edpm.yml
```diff
@@ -47,9 +47,27 @@
   tags:
     - admin-setup
 
-- name: Import run test playbook
+- name: Import run tempest test playbook
```
cjeanner, line 50: I'm not that comfortable with importing twice the same playbook with different vars. Can't we get the logic directly in the 08-run-test.yml instead?

roles/tobiko/vars/main.yml
cjeanner, line 1: we can remove this file (and `vars` directory) if there's nothing in it.

roles/tobiko/tasks/cleanup.yml
cjeanner, line 1: this cleanup.yml file should be removed since it doesn't clean anything. Or shall it clean the tobiko container instead?

roles/tobiko/tasks/main.yml
eduolivares: @pablintino it seems it works.
When I access the container, only the TOBIKO_* that are not "omitted" have been set.
```
[tobiko@controller-0 ~]$ env | grep ^TOBIKO
TOBIKO_TESTENV=neutron
TOBIKO_DEBUG=True
TOBIKO_UBUNTU_MINIMAL_IMAGE_URL=https://cloud-images.ubuntu.com/minimal/releases/jammy/release/ubuntu-22.04-minimal-cloudimg-amd64.img

```

deploy-edpm.yml
eduolivares: @cjeanner as we checked, this is needed in case a job runs both tempest and tobiko

roles/tobiko/defaults/main.yml
```diff
@@ -0,0 +1,40 @@
...
+cifmw_tobiko_registry: quay.io
+cifmw_tobiko_namespace: podified-antelope-centos9
+cifmw_tobiko_container: openstack-tobiko
+cifmw_tobiko_image: "{{ cifmw_tobiko_registry }}/{{ cifmw_tobiko_namespace }}/{{ cifmw_tobiko_container }}"
+cifmw_tobiko_image_tag: current-podified
+cifmw_tobiko_dry_run: false
+cifmw_tobiko_remove_container: false
+cifmw_tobiko_dns_servers:
+  - "192.168.122.10"
+ci_tobiko_debug: true
```
cjeanner, line 30: `cifmw_` instead?

roles/tobiko/tasks/main.yml
```diff
@@ -0,0 +1,106 @@
...
+  delay: 5
+  until: _tobiko_fetch_img is success
+
+- name: Run tobiko
+  ignore_errors: true
+  containers.podman.podman_container:
+    name: tobiko
+    image: "{{ cifmw_tobiko_image }}:{{ cifmw_tobiko_image_tag }}"
+    state: started
+    auto_remove: "{{ cifmw_tobiko_remove_container }}"
```
cjeanner, line 64: default value seems to be `false` for that auto_remove. Meaning we'll have that dangling container.
Since it defaults to `false`, it may be valuable to get a `cleanup.yml` that will take care of the various things, such as:
- container
- cloud.yaml generated for that specific role
- any other files that aren't logs?

roles/tobiko/tasks/main.yml
```diff
@@ -0,0 +1,106 @@
...
+- name: Change tobiko directory permission back to original
+  become: true
+  ansible.builtin.file:
+    path: "{{ cifmw_tobiko_artifacts_basedir }}"
+    state: directory
+    recurse: true
+    owner: "{{ lookup('env', 'USER') }}"
+    group: "{{ lookup('env', 'USER') }}"
+
+- name: Save logs from podman
```
cjeanner, line 95: wondering if we couldn't use the internal podman logging facility to avoid loading the ansible process memory.
For instance, using [log_driver](https://docs.ansible.com/ansible/latest/collections/containers/podman/podman_container_module.html#parameter-log_driver) (and setting it to k8s iirc), in addition to [log_opt](https://docs.ansible.com/ansible/latest/collections/containers/podman/podman_container_module.html#parameter-log_opt) (and setting it to `path: "{{ cifmw_tobiko_artifacts_basedir }}/podman_tobiko.log"` for instance) may be better/faster.
We can also play with the [log_level](https://docs.ansible.com/ansible/latest/collections/containers/podman/podman_container_module.html#parameter-log_level) while at it.

Might be a follow-up, iirc Tempest role is doing the same.

roles/tobiko/tasks/main.yml
```diff
@@ -0,0 +1,106 @@
...
+  when: not cifmw_tobiko_dry_run | bool
+  register: tobiko_run_output
+
+- name: Change tobiko directory permission back to original
+  become: true
+  ansible.builtin.file:
+    path: "{{ cifmw_tobiko_artifacts_basedir }}"
+    state: directory
+    recurse: true
+    owner: "{{ lookup('env', 'USER') }}"
```
cjeanner, line 92: warning, `lookup()` will execute on the ansible controller. So, if you're running from your laptop against some remote node, it will take your laptop user, not the one on the remote.
You'd rather go with `ansible_user | default(ansible_user_id)` or some similar value, to ensure it will work remotely.

roles/tobiko/README.md
```diff
@@ -0,0 +1,26 @@
...
+## Privilege escalation
+become - Required to install required rpm packages
+
+## Parameters
+* `cifmw_tobiko_artifacts_basedir`: (String) Directory where we will have all tobiko files. Default to `cifmw_basedir/artifacts/tobiko` which defaults to `~/ci-framework-data/tests/tobiko`.
+* `cifmw_tobiko_image`: (String) Name of the tobiko image to be used. Default to `quay.io/podified-antelope-centos9/openstack-tobiko`
+* `cifmw_tobiko_image_tag`: (String) Tag for the `cifmw_tobiko_image`. Default to `current-podified`
+* `cifmw_tobiko_dry_run`: (Boolean) Whether tobiko should run or not. Default to `false`
+* `cifmw_tobiko_remove_container`: (Boolean) Cleanup tobiko container after it is done. Default to `false`
+* `ci_tobiko_debug`: (Boolean) Whether tobiko log level is debug or not
```
cjeanner, line 13: `cifmw_` ?

roles/tobiko/tasks/main.yml
eduolivares: My impression is that saving the podman logs was implemented this way on the tempest role too because of problems with the permissions.
The `cifmw_tobiko_artifacts_basedir` owner is changed in line 47.
Original ownership is restored in line 92.

If the task `Run tobiko` tries to create a file under that directory, it will fail.

roles/tobiko/tasks/main.yml
eduolivares: I am going to set both `cifmw_tobiko_remove_container` and `cifmw_tempest_remove_container` default values to true.
Regarding other cleanups, all the files that are created under `cifmw_tobiko_artifacts_basedir` look relevant to me:
- `clouds.yaml`
- `tobiko.conf`
- `podman_tobiko.log`
- and the tobiko results

roles/tobiko/tasks/main.yml
cjeanner: possible. I'll try to get some time to some tests. Maybe letting podman write to its standard log location and copying it is better - as said, potential follow-up.

roles/tobiko/tasks/main.yml
cjeanner: sounds good.

cjeanner (approved):

roles/tempest/defaults/main.yml
```diff
@@ -28,7 +28,7 @@ cifmw_tempest_container: openstack-tempest
 cifmw_tempest_image: "{{ cifmw_tempest_registry }}/{{ cifmw_tempest_namespace }}/{{ cifmw_tempest_container }}"
 cifmw_tempest_image_tag: current-podified
 cifmw_tempest_dry_run: false
-cifmw_tempest_remove_container: false
+cifmw_tempest_remove_container: true
```
slawqo, line 31: why this change?

eduolivares: @slawqo please go to this conversation:
https://github.com/openstack-k8s-operators/ci-framework/pull/1013/files#r1448599697

Cc @cjeanner

pablintino:
/approve

@cjeanner care to lgtm if everything is ok?

cjeanner:
/lgtm

all good - might need some follow-up but things are OK as-is. Mostly the podman logging thing being a bit annoying, but we can come back on this later.

PR #1014, opened 2024-01-10
Improve cert manager cleanup playbook

raukadah:
It now includes:
- Deleting the deployment
- Deleting the crds.

  - [x] README in the role

### DIFF
--- a/roles/cert_manager/tasks/cleanup.yml
+++ b/roles/cert_manager/tasks/cleanup.yml
@@ -14,6 +14,28 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+- name: Delete deployment of cert-manager
+  kubernetes.core.k8s:
+    state: absent
+    kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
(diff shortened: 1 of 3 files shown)


roles/cert_manager/tasks/cleanup.yml
```diff
@@ -14,6 +14,33 @@
...
+- name: Delete CRDs related to cert-manager
+  kubernetes.core.k8s:
+    state: absent
+    kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
+    api_key: "{{ cifmw_openshift_token | default(omit)}}"
+    context: "{{ cifmw_openshift_context | default(omit)}}"
+    kind: CustomResourceDefinition
+    NameSpace: "{{ cifmw_cert_manager_operator_namespace }}"
+    name: "{{ item }}"
+  loop:
```
cjeanner, line 37: is that loop static and always the same? Should we consider having that as a parameter to the role?

raukadah: Nope, Replaced it with a var

cjeanner:
/approve
/lgtm

thanks for the cleaning :)

PR #1016, opened 2024-01-10
Ensure we deploy the ssh configuration on controller-0

cjeanner:
Until now, the ssh configuration was deployed only for multi-hypervisor
use-case. It's wrong, since we still want to access remote nodes, being
OCP, CRC or Computes from the controller-0, wherever they are running.

This patch ensures we configure ssh the right way.

It also improves the user and used SSH key in order to ensure seamless
accesses.


raukadah (approved):
/approve

PR #1018, opened 2024-01-11
Generate kustomization snippet for deployment

cjeanner:
The CI must provide dynamic values for networking and instance related
configuration, consumed during the product deployment.
This first patch is an attempt at providing the "values.yaml" as
described in the architecture[1] repository.

[1] https://github.com/openstack-k8s-operators/architecture/blob/main/examples/va/hci/values-ci-framework.yaml

  - [ ] README in the role


roles/networking_mapper/templates/network-config-values.j2
```diff
@@ -0,0 +1,103 @@
...
+    mtu: {{ network.mtu | default(1500) }}
+{%  if network.vlan_id is defined  %}
+    vlan: {{ network.vlan_id }}
+    iface: {{ network.network_name }}
+    base_iface: {{ ns.interfaces[network.network_name] }}
+{%  else %}
+    iface: {{ ns.interfaces[network.network_name] }}
+{%  endif %}
+    lb_addresses:
+      - {{ network.network_v4 | ansible.utils.ipmath(80) }}-{{ network.network_v4 | ansible.utils.ipmath(90) }}
```
pablintino, line 46: Nope, for this purpose we have the tools section in the mapper, with a range for it. Same goes for the NNCPs, there is no need to hardcode this here.

roles/networking_mapper/molecule/default/converge.yml
```diff
@@ -214,3 +239,44 @@
...
+            gateway: 192.168.122.1
+            name: multus
+          - allocationRanges:
+              - end: 192.168.122.49
+                start: 192.168.122.40
+            cidr: 192.168.122.0/24
+            gateway: 192.168.122.1
+            name: netconfig
+      ansible.builtin.assert:
+        that:
```
pablintino, line 277: Isn´t this thing too loose? Should we use a golden file?

roles/ci_gen_kustomize/tasks/main.yml
```diff
@@ -0,0 +1,71 @@
...
+  register: _snippets
+  ansible.builtin.find:
+    paths: "{{ cifmw_ci_gen_kustomize_src_dir }}"
+    recurse: false
+
+- name: Load various snippet files
+  register: _snippets_content
+  ansible.builtin.slurp:
+    path: "{{ file.path }}"
+  loop: "{{ _snippets.files }}"
```
pablintino, line 36: Should we sort them?

roles/networking_mapper/templates/network-config-values.j2
cjeanner: hm care to point to it? Is it in the "tools" maybe?

roles/ci_gen_kustomize/README.md
```diff
@@ -0,0 +1,28 @@
+# ci_gen_kustomize
+Generate complete kustomization file based on multiple snippets
+
+## Privilege escalation
+None
+
+## Parameters
+* `cifmw_ci_gen_kustomize_basedir`: (String) Base directory. Defaults to `cifmw_basedir` which defaults to `~/ci-framework-data`.
+* `cifmw_ci_gen_kustomize_src_dir`: (String) Source directory where your snippets are located. Defaults to `{{ cifmw_ci_gen_kustomize_basedir }}/artifacts/ci_k8s_snippets`.
+* `cifmw_ci_gen_kustomize_dest_dir`: (String) Destination directory of the generated file. Defaults to `{{ cifmw_ci_gen_kustomize_basedir }}/artifacts/ci_kustomize/`.
```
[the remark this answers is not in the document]
cjeanner: any other fancy name I can use then? `generated_kustomize` or something like that?

roles/ci_gen_kustomize/tasks/main.yml
cjeanner: ah, yeah, might be good indeed. Allowing then to get some ordering and, eventually, overrides.

roles/ci_gen_kustomize/README.md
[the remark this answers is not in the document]
cjeanner: changed to `ci_gen_kustomize` - it matches the role name.

roles/networking_mapper/molecule/default/converge.yml
pablintino: Me neither, but, if we do not add tests here, in molecule, we can merge a broken jinja without notice. We will discover it's broken in the periodic e2e, far far later and probably with an error that doesn´t point directly to the real issue.

cjeanner:
/hold

there seems to be some issue with the way lb are configured in the template. Need some more details.

roles/networking_mapper/molecule/default/converge.yml
cjeanner: I went for a "partial, split golden file" - seems easier to maintain on the medium term. Also took the main "test config" out in its own file, to make the `converge.yml` a bit smaller and easier to read.

roles/ci_gen_kustomize/README.md
```diff
@@ -0,0 +1,28 @@
+# ci_gen_kustomize
+Generate complete kustomization file based on multiple snippets
```
fultonj, line 2: Can we get a better name?

The CR that is generated isn't a kustomization file. It's a `ConfigMap` which kustomize can read. I've been calling it a kustomize data file or a kustomize values file. Maybe one of these names would work?

`ck_gen_kust_cf`
`ck_gen_kustomize_cf`
`ci_gen_kustomize_configmap`
`ci_gen_kustomize_values`

fultonj (commented):
Could the reproducer playbook be updated so that when I run the following I can see the results generated in `{{ cifmw_reproducer_basedir  }}` my environment so I can test it?

```
time ansible-playbook \
     -i custom/inventory.yml \
     -e cifmw_target_host=hypervisor-1 \
     -e @scenarios/reproducers/validated-architecture-1.yml \
     -e cifmw_reproducer_basedir="~/ci-framework-data" \
     -e @custom/validated-architecture-1.yml reproducer.yml
```

cjeanner:
> Could the reproducer playbook be updated so that when I run the following I can see the results generated in `{{ cifmw_reproducer_basedir }}` my environment so I can test it?
> 
> ```
> time ansible-playbook \
>      -i custom/inventory.yml \
>      -e cifmw_target_host=hypervisor-1 \
>      -e @scenarios/reproducers/validated-architecture-1.yml \
>      -e cifmw_reproducer_basedir="~/ci-framework-data" \
>      -e @custom/validated-architecture-1.yml reproducer.yml
> ```

hm, that will require some more work - and I'd nudge it in a dedicated patch anyway. Adding the feature and using it isn't the same imho, deserves its own patch (in that PR). Though modifying the reproducer for that will take some time I think. Lemme check today.

Also, my first intend was to run that role + networking_mapper from within the "[02-infra.yml](https://github.com/openstack-k8s-operators/ci-framework/blob/main/playbooks/02-infra.yml)" subplaybook, not from the reproducer role. It's that kind of thing between two stages imho.

fultonj:
> > Could the reproducer playbook be updated so that when I run the following I can see the results generated in `{{ cifmw_reproducer_basedir }}` my environment so I can test it?
> > ```
> > time ansible-playbook \
> >      -i custom/inventory.yml \
> >      -e cifmw_target_host=hypervisor-1 \
> >      -e @scenarios/reproducers/validated-architecture-1.yml \
> >      -e cifmw_reproducer_basedir="~/ci-framework-data" \
> >      -e @custom/validated-architecture-1.yml reproducer.yml
> > ```
> 
> hm, that will require some more work - and I'd nudge it in a dedicated patch anyway. Adding the feature and using it isn't the same imho, deserves its own patch (in that PR). Though modifying the reproducer for that will take some time I think. Lemme check today.
> 
> Also, my first intend was to run that role + networking_mapper from within the "[02-infra.yml](https://github.com/openstack-k8s-operators/ci-framework/blob/main/playbooks/02-infra.yml)" subplaybook, not from the reproducer role. It's that kind of thing between two stages imho.

The reproducer playbook doesn't need to call the new role directly and the change does not need to be in this PR.

However, I would like to be able to run the reproducer.yml playbook and then run kustomize using the generated values as input. I should be able to to trigger the control plane and edpm deployment using CRs independent of the CI framework. At least that was my understanding from our previous conversations.

PR #1023, opened 2024-01-15
Add update role and update playbooks to ci-framework

ciecierski:
Add update playbooks and update role which is going to be used to orchestrate minor update procedure. This change adds base structure of role and playbooks. A content of role is going to be added in next pull request.

  - [x] README in the role


playbooks/075-update.yml
```diff
@@ -0,0 +1,21 @@
+- name: Run pre_update hooks
+  vars:
+    hooks: "{{ pre_update | default([]) }}"
+    step: pre_update
+  ansible.builtin.import_playbook: ./hooks.yml
+
+- name: Update playbook
```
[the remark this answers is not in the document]
ciecierski: I followed convention used in other playbooks. I don't think we need any rescue block, because there is no rollback operation yet. We might change it in future, but for now current structure is enough.

roles/update/README.md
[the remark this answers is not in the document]
ciecierski: This is template and not much to document yet.

roles/update/files/.gitkeep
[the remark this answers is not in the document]
ciecierski: done

ci/playbooks/edpm/update.yml
```diff
@@ -0,0 +1,35 @@
...
+          {%- if edpm_file.stat.exists %}
+          -e @{{ ansible_user_dir }}/ci-framework-data/artifacts/edpm-ansible.yml
+          {%- endif %}
+          {%- if cifmw_extras is defined %}
+          {%-   for extra_var in cifmw_extras %}
+          -e "{{   extra_var }}"
+          {%-   endfor %}
+          {%- endif %}
+          -e "@{{ ansible_user_dir }}/ci-framework-data/artifacts/parameters/zuul-params.yml"
+          -e "cifmw_repo_setup_promotion={{ cifmw_repo_setup_promotion_target }}"
```
raukadah, line 35: Can we move `zuul-params.yml` line in last so that vars passed from zuul jobs have higher precedence. Like this
```suggestion
          -e "cifmw_repo_setup_promotion={{ cifmw_repo_setup_promotion_target }}"
          -e "@{{ ansible_user_dir }}/ci-framework-data/artifacts/parameters/zuul-params.yml"
         
```

marios (approved):
as discussed yesterday this is scaffolding and will not be wired up anywhere yet so should not affect anything.

ci/playbooks/edpm/update.yml
ciecierski: No, this is done on purpose. We want to override cifmw_repo_setup_promotion with  cifmw_repo_setup_promotion_target on update run.

ciecierski: If we move zuul-params last  -e "cifmw_repo_setup_promotion={{ cifmw_repo_setup_promotion_target }}" has no effect, because last -e param always win.

raukadah: Thank you for explaining it on call. I can mark this resolved.

PR #1025, opened 2024-01-16
Switch reproducer to networking_mapper

cjeanner:
### Remove most of the IP management from ansible
All of the IP management will mostly rely on the networking_mapper role
and associated tasks.

In order to avoid duplications, we're now ensuring we're just keeping
the minimal parts to get the network interfaces attached to the VMs;
we're also removing the static_ip support and the support for the
`cifmw_libvirt_manager_networks` parameter.

This patch has been locally tested against:
- 3-nodes.yml reproducer scenario
- validated-architecture-1.yml reproducer scenario

### Refactor reproducer role to consume networking_mapper
This patch switches the reproducer role to networking_mapper when it
comes to networking configuration.

It means we have to make a deep refactoring, as well as changing some of
the logic underneath the various tasks.

Mostly, we have to ensure the controller-0 is accessed using the right
user (see previous patch in this same PR), but also ensure controller-0
is ready before we get to the CRC (and OCP) node(s). This is mandatory,
especially for CRC, because we need to get its fixed IP.
This IP is generated via the networking_mapper, so we have to call it
early enough.

Since we'll be consuming some crafted output generated from the
networking_mapper data during the product deployment, we also have to
ensure we have all the needed pieces on the controller-0 - meaning it's
easier to run everything from that node directly, so that the content
gets generated in the right location since the beginning.

There's also a deep cleaning about the delegate_to and related
remote_user, delegate_facts and others since we're consuming a known,
fixed user via the inventory we update dynamically from the
libvirt_manager.

Most of the features should be properly tested from within the molecule
tests running against the reproducer role - but also the ones running
against the libvirt_manager directly.

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/1036
Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/1037
Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/1032
Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/1043

  - [X] README in the role


cjeanner:
/hold

I want #1043 to get in first.

roles/libvirt_manager/tasks/start_manage_vms.yml
```diff
@@ -54,7 +54,7 @@
       Host {{ vm_ip.nic.host }} cifmw-{{ vm_ip.nic.host }} {{ extracted_ip }}
         ProxyJump {{ ansible_user | default(lookup('env', 'USER')) }}@{{ proxy_hostname }}
         Hostname {{ extracted_ip }}
-        User {{ 'core' if vm_ip.nic.host.startswith('crc') else 'zuul' }}
+        User {{ 'core' if vm_ip.nic.host is match('^(crc|ocp).*') else 'zuul' }}
```
SeanMooney, line 57: so this and the other code like it is making sure we use the core user for vms deploy by the ci framework if ht hostname starts with crc or ocp.

would it not be better to just have a way to map the user to hostname?

by the way zuul jobs are not meant to assume they are logging in as the zuul user either for portability reasons. but that fine for now i guess.

the main thing to ensure is that we do not just use root when sshing in.
so I'm happy that we are using core or zuul and not root.

i would have just tried to make this contolabel via the inventory file using host/group vars instead but this is functionl

roles/reproducer/tasks/default_repositories.yml
```diff
@@ -1,6 +1,7 @@
 ---
 - name: "Ensure repository is present: {{ repository.src | basename }}"
   delegate_to: controller-0
+  remote_user: zuul
```
SeanMooney, line 4: so this will requrie that we create a zuul user on the image we user for the hosts and vms

that probably should be configurable but that could come later and it can default to zuul.

roles/reproducer/tasks/ci_job.yml
```diff
@@ -27,34 +26,34 @@
             - bootstrap
           ansible.builtin.copy:
             src: "{{ _reproducer_basedir }}/parameters/"
-            dest: "./{{ job_id }}-params"
+            dest: "/home/zuul/{{ job_id }}-params"
```
SeanMooney, line 29: perhaps this instead
"{{ansible_env.HOME}}/{{ job_id }}-params"

roles/libvirt_manager/tasks/start_manage_vms.yml
cjeanner: so there's a way to pass the `admin_user` from the configuration dict - but there are some other issues with that. I had to hack my way in the past, maybe a follow-up would help converging again with that parameter (not really the scope of this PR)

roles/reproducer/tasks/default_repositories.yml
cjeanner: controller-0 VM is 100% managed by the libvirt_manager, and the user is created there. No need to do anything on your side.

roles/reproducer/tasks/ci_job.yml
cjeanner: since the user is fixed, we can hardcode it. Also, I got bitten far too many times with delegate_to, facts and ansible internals...

fultonj:
I tested this with the reproducer and then passed the genereated values.yaml to the architecture kustomize.

The output looks good to me so far though the NNCPs get FailedToConfigure because the `base_inteface` needs to not have the "dot vlan" name. In other words:
```
[zuul@controller-0 hci]$ grep base_iface values.yaml    #BAD 
        base_iface: enp6s0.20
        base_iface: enp6s0.21
        base_iface: enp6s0.22
[zuul@controller-0 hci]$ grep base_iface values-ci-framework.yaml    # GOOD
    base_iface: enp6s0
    base_iface: enp6s0
    base_iface: enp6s0
[zuul@controller-0 hci]$ 
```
Otherwise I see all my CRs generated by this:
```
cd ~/architecture/examples/va/hci
cp ~/ci-framework-data/artifacts/ci_gen_kustomize_values/values.yaml .
kustomize build .
```

roles/reproducer/tasks/default_repositories.yml
cjeanner: (in fact I think this one isn't required - probably a leftover from when I hit so many issues with delegate_to, inventory and ssh_config)

cjeanner:
> I tested this with the reproducer and then passed the genereated values.yaml to the architecture kustomize.
> 
> The output looks good to me so far though the NNCPs get FailedToConfigure because the `base_inteface` needs to not have the "dot vlan" name. In other words:
> 
> ```
> [zuul@controller-0 hci]$ grep base_iface values.yaml    #BAD 
>         base_iface: enp6s0.20
>         base_iface: enp6s0.21
>         base_iface: enp6s0.22
> [zuul@controller-0 hci]$ grep base_iface values-ci-framework.yaml    # GOOD
>     base_iface: enp6s0
>     base_iface: enp6s0
>     base_iface: enp6s0
> [zuul@controller-0 hci]$ 
> ```
> 
> Otherwise I see all my CRs generated by this:
> 
> ```
> cd ~/architecture/examples/va/hci
> cp ~/ci-framework-data/artifacts/ci_gen_kustomize_values/values.yaml .
> kustomize build .
> ```

Thanks for confirming, @fultonj ! For the network interface, I nudged a dedicated patch #1046 (should merge shortly). And, also, there's now a patch in this current PR to configure the ctlplane interface on the compute nodes :). Sooooo.... you should be able to actually deploy :smile_cat:

psathyan (commented):
/lgtm

cjeanner:
recheck

cjeanner:
The `install_yamls_makes : Run netconfig_deploy` task returned `The connection to the server api.crc.testing:6443 was refused - did you specify the right host or port?` in BM job - I don't think it's due to the changes, just keeping this here in case it fails again, to compare.

fultonj:
> > I tested this with the reproducer and then passed the genereated values.yaml to the architecture kustomize.
> > The output looks good to me so far though the NNCPs get FailedToConfigure because the `base_inteface` needs to not have the "dot vlan" name. In other words:
> > ```
> > [zuul@controller-0 hci]$ grep base_iface values.yaml    #BAD 
> >         base_iface: enp6s0.20
> >         base_iface: enp6s0.21
> >         base_iface: enp6s0.22
> > [zuul@controller-0 hci]$ grep base_iface values-ci-framework.yaml    # GOOD
> >     base_iface: enp6s0
> >     base_iface: enp6s0
> >     base_iface: enp6s0
> > [zuul@controller-0 hci]$ 
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Otherwise I see all my CRs generated by this:
> > ```
> > cd ~/architecture/examples/va/hci
> > cp ~/ci-framework-data/artifacts/ci_gen_kustomize_values/values.yaml .
> > kustomize build .
> > ```
> 
> Thanks for confirming, @fultonj ! For the network interface, I nudged a dedicated patch #1046 (should merge shortly). And, also, there's now a patch in this current PR to configure the ctlplane interface on the compute nodes :). Sooooo.... you should be able to actually deploy 😸

Yes, the latest update seems to have resolved my problems. Thanks!

roles/reproducer/tasks/configure_controller.yml
```diff
@@ -300,3 +292,67 @@
...
+
+    - name: Generate configMap snippet
+      vars:
+        cifmw_networking_mapper_config_generator_output: >-
+          /home/zuul/ci-framework-data/artifacts/ci_k8s_snippets
+      ansible.builtin.import_role:
+        name: networking_mapper
+        tasks_from: network_config_values.yml
+
+    - name: Generate values.yaml for EDPM deploy
```
cjeanner, line 334: follow-up:
- change title to something like "Generate controlplan data"
- set the target file to something like "controlplan-values.yaml"

pablintino:
I'll review this PR on Monday's early morning.

roles/libvirt_manager/README.md
```diff
@@ -61,14 +61,13 @@ Specific `type_name`: `^crc.*` and `^ocp.*` are enabling specific paths in the m
 #### Example
 
 ```YAML
-cifmw_libvirt_manager_networks:
+_networks:
```
pablintino, line 64: Isn't documenting and  exposing a "private" variable a code smell? I'd say so. If this variable is encouraged to not be used I wouldn't document it, neither pass it. This case would collide with that mechanism we planed to put in place to detect avoid passing private vars into our playbooks.

scenarios/reproducers/3-nodes.yml
```diff
@@ -14,14 +14,12 @@ cifmw_rhol_crc_config:
...
-cifmw_libvirt_manager_networks:
+_networks:
   public:
     range: "192.168.100.0/24"
   osp_trunk:
-    default: true
     range: "192.168.122.0/24"
     mtu: 1500
-    static_ip: true
   internal-api:
```
pablintino, line 23: Do we really need to preserve `storage`,  `tenant`,  `storage-mgmt` and `internal-api`? Where are those values consumed?

roles/reproducer/tasks/ci_job.yml
pablintino: This is not the first place I see this kind of change. Can we provide more context in a comment or add it into the commit message? I'd avoid hard-coding it if we are not sure it's really really needed.

scenarios/reproducers/validated-architecture-1.yml
pablintino, line 1: ditto. And, why is there still a reference to static_ip (https://github.com/openstack-k8s-operators/ci-framework/pull/1025/files#diff-a27d3292259ca39642bd28c26baea1f33d2cc740203d86ebb1318bb396d15ab2R24)?
There are two more, in the crc role and in devscriopts.

roles/reproducer/tasks/configure_controller.yml
```diff
@@ -300,3 +292,67 @@
...
+      vars:
+        cifmw_ci_gen_kustomize_values_basedir: "/home/zuul/ci-framework-data"
+      ansible.builtin.import_role:
+        name: ci_gen_kustomize_values
+
+    - name: Debug
```
pablintino, line 340: Is this one intended to be here when merging or it's just a debug during coding?

roles/reproducer/tasks/configure_controller.yml
```diff
@@ -300,3 +292,67 @@
...
+        sleep: 2
+        timeout: 300
+      loop: "{{ groups.ocps }}"
+
+    - name: Generate networking definition
+      vars:
+        cifmw_networking_mapper_ifaces_info: >-
+          {{ _nic_info.content | b64decode | from_yaml }}
+        cifmw_networking_mapper_network_name: >-
+          {{ _layout.vms.controller.nets.1 }}
```
pablintino, line 322: Assuming that `osp_trunk` is the second item in the nets list sounds fragile.

roles/libvirt_manager/README.md
cjeanner: temporary for now. It's used only in the `cifmw_libvirt_manager_config.networks`. I'm waiting to see all the pieces together (ci_nmstate and others) and will converge things properly. More cleaning is also needed anyway.

scenarios/reproducers/3-nodes.yml
cjeanner: not consumed anymore - more cleaning and converging will be needed as well. I'll probably update doc and related to instruct "include the networking-definition first, then the layout" so that we can inherit parameters the right way.

roles/reproducer/tasks/ci_job.yml
cjeanner: context: delegate_to, facts and everything. Kind of a "leftover" from the time I wasn't sure about the used `user` - and I thought it would be far, far better to get absolute path. Clearer for everybody.

roles/reproducer/tasks/configure_controller.yml
cjeanner: it can't be first, that's a sure thing. Can't be third - or a least I don't know what kind of layout that would be.

There's no simple way. We can't enforce `osp_trunk` name (especially since I think we may use `default` if available), so there's no way to get it without fixing one thing. Or getting yet a new parameter to namely write the network name - but it's not really good either imho :(. Already have one for the "public network", and I'm unhappy with it.

scenarios/reproducers/validated-architecture-1.yml
cjeanner: ah, right - need to clean deeper. I didn't edit the structure of the param in this file. As said: I want to converge consuming the networking_definition directly.

roles/reproducer/tasks/configure_controller.yml
pablintino: Agree that both solutions looks ugly :(  I don't have other options on top of head

roles/libvirt_manager/README.md
cjeanner: followup is on its way to remove that parameter and consume the networking definition instead.

scenarios/reproducers/validated-architecture-1.yml
cjeanner: followup is on its way to remove that parameter and consume the networking definition instead.
It will also chase down all of the remaining `static_ip`

roles/reproducer/tasks/configure_controller.yml
cjeanner: followup on its way.

roles/reproducer/tasks/configure_controller.yml
cjeanner: done

scenarios/reproducers/3-nodes.yml
cjeanner: cleaned.

PR #1030, opened 2024-01-16
Converge "ocps" and "crcs" group names

cjeanner:
Doing so, we make things slightly easier to consume from the
networking_mapper perspective: no matter if we're using CRC or the OCP
cluster, they will be known as "ocps" group.

It should allow easier management.


cjeanner:
@pablintino any thoughts? It would be "easy" to converge names I think.

pablintino:
> @pablintino any thoughts? It would be "easy" to converge names I think.

@cjeanner I think converging makes sense for the two envs we have on top of the table. I'd go that way.

cjeanner:
> > @pablintino any thoughts? It would be "easy" to converge names I think.
> 
> @cjeanner I think converging makes sense for the two envs we have on top of the table. I'd go that way.

cool - I'll check tomorrow. FYI, I just added a patch in #1025 to integrate the networking_mapper - and I'll need the networking definition, since it will be consumed by molecule as launched in the reproducer role (yessss - we do have tests!)

scenarios/reproducers/networking-definition.yml
cjeanner, line 1: @pablintino any reasons you didn't put the whole network definition under `cifmw_networking_definition` ? Since we're expecting ppl to pass that file using `-e @scenarios/reproducers/networking-definition.yml` I think it would make sense to define the parameter properly.. ?

pablintino: @cjeanner not really, but it makes sense. Feel free to change the content to nest everything under `cifmw_networking_definition`

cjeanner:
recheck

pablintino:
recheck

raukadah (commented):
/approve

PR #1034, opened 2024-01-17
[ci_nmstate] Adopt the Networking Env Definition as input

pablintino:
  - [x] README in the role


roles/ci_nmstate/molecule/default/converge.yml
```diff
@@ -41,6 +41,7 @@
           - "!all"
           - "!min"
           - network
+          - dns
```
cjeanner, line 44: isn't "dns" a subset of "network" ?

roles/networking_mapper/tasks/load_env_definition.yml
```diff
@@ -0,0 +1,25 @@
...
+- name: Load the Network Environment Definition from file if not given
+  when: "cifmw_networking_env_definition is not defined"
+  block:
+    - name: Check for Networking Environment Definition file existence
+      ansible.builtin.stat:
+        path: "{{ cifmw_networking_mapper_networking_env_def_path }}"
+      register: _net_env_def_stat
+
+    - name: Load the Networking Definition from file
+      when: "_net_env_def_stat.stat.exists"
```
cjeanner, line 11: shouldn't we `assert` the file exists instead? Else we may end with a failure later, and it may be hard to find the root cause.

pablintino: @cjeanner In this case, ci_nmstate can be used without the mapper. Anyone can directly pass NMstates though `cifmw_ci_nmstate_instance_config` .

roles/ci_nmstate/molecule/default/converge.yml
pablintino: Cannot make it to work without this addition. Agree that, in theory, it's part of the network facts.

pablintino:
recheck

roles/networking_mapper/tasks/load_env_definition.yml
cjeanner: so it means:
- add the `assert` in this file
- from the ci_nmstate, check if this `instance_config` is set, if not, call the load_env_defintiion.yml from networking_mapper role

Seems to be the best way to avoid weird, hard to debug issues?

roles/ci_nmstate/filter_plugins/ci_nmstate_map_instance_nets.py
```diff
@@ -0,0 +1,299 @@
...
+        default:
+            dns_v4:
+            - 192.168.122.1
+            dns_v6: []
+            gw_v4: 192.168.122.1
+            mtu: 1500
+            network_name: default
+            network_v4: 192.168.122.0/24
+            search_domain: default.example.com
+        internal-api:
```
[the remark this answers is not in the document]
pablintino: Correct. Updated.

pablintino:
recheck

PR #1035, opened 2024-01-17
Set use_libvirt false in adoption job

cescgina:
Since in adoption job we no longer need any nested vm, we can set
use_libvirt to false and not run the adoption job for libvirt_manager
role, since it should not use it anymore.


cjeanner:
/approve

adoption job was triggered by this change, and is green.

PR #1036, opened 2024-01-17
Follow-up and corrections to the networking_mapper

cjeanner:
There were some issues with the way network_config_values was working:
- include_vars doesn't work with remote content, switched to slurp
- template faced issues for some networks that were defined, but not
  associated to instances, leading to empty "interface"
- renamed "internal-api" to match actual architecture

### DIFF
--- a/roles/networking_mapper/molecule/default/vars/internalapi_output.yml
+++ b/roles/networking_mapper/molecule/default/vars/internalapi_output.yml
@@ -1,6 +1,6 @@
 ---
-internal-api:
-  dnsDomain: internal-api.example.com
+internalapi:
+  dnsDomain: internalapi.example.com
   subnets:
     - allocationRanges:
         - end: 172.17.0.49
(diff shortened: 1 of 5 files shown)


cjeanner:
recheck

roles/networking_mapper/molecule/default/vars/internalapi_output.yml
```diff
@@ -1,6 +1,6 @@
 ---
-internal-api:
-  dnsDomain: internal-api.example.com
+internalapi:
+  dnsDomain: internalapi.example.com
```
pablintino, line 3: We don't really need this line, as the mapper generated the search domain based on the network name. Just FYI, no need to remove it here, as it's a molecule.

cjeanner: that's the output, yes. so it's in, and has to be nudged in the generated snippet used for the product deploy :).

PR #1038, opened 2024-01-17
Remove ci_framework folder

cescgina:
When moving the ci-framework to a collection, the ci_framework folder
was deprecated and contained only symlinks. After checking that it's
safe to remove it, we can clean it up.


psathyan (approved):
/lgtm

PR #1040, opened 2024-01-18
Make /etc/yum.repos.d writeable for component repos

raukadah:
On using cifmw_repo_setup_output: '/etc/yum.repos.d'.
The component_repo tasks are failing with
```
The destination directory (/etc/yum.repos.d) is not writable by the current user
```

In order to fix it, we need to set become: true for these tasks.
We are already using the same conditional for repo-setup commands.


cjeanner:
/approve
/lgtm

(it's more about "ensure proper right accesses" but ok)

PR #1042, opened 2024-01-18
Ensure hosts are ready before collecting facts

cjeanner:
With automation, it may happen hosts are being booted, ansible goes on
other tasks, and when we hit this fact collection, hosts are in a middle
stage where they seem to have some networking issues.

This is especially visible with the OCP cluster - the libvirt_manager
role starts them, wait for them, gets access on them and everything's
good, moves on, but later the OCP cluster is unreachable for a limited
amount of time.

This new task ensure everything's as ready as possible before looping on
the fact collection, avoiding a crash.

(Mostly tested while debugging #1025 against OCP cluster)

### DIFF
--- a/roles/networking_mapper/tasks/main.yml
+++ b/roles/networking_mapper/tasks/main.yml
@@ -73,6 +73,13 @@
         _net_def_groups_instances + _net_def_instances_filtered
       }}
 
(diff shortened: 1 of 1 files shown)


cjeanner:
There may be a better way.

PR #1044, opened 2024-01-18
[Tobiko] Create tobiko.conf and ssh keys

eduolivares:
With this change, ci-framework will create tobiko.conf based on the input param cifmw_tobiko_tobikoconf, which should be defined as a dictionary, providing total flexibility to define tobiko.conf.

Besides, the framework will generate ssh key files that can be shared between different tobiko containers running tests, so that they can connect to the VMs created by a previous tobiko container (the Openstack workloads created by tobiko are persistent).

Related: https://github.com/openstack-k8s-operators/tcib/pull/120

  - [X] README in the role

### DIFF
--- a/roles/tobiko/tasks/main.yml
+++ b/roles/tobiko/tasks/main.yml
@@ -33,6 +33,21 @@
     tasks_from: create-clouds-file
   when: not cifmw_tobiko_dry_run | bool
 
+
+- name: Generate ssh keys used for the VMs that tobiko creates
+  block:
+    - name: Check if the ssh keys already exist
+      ansible.builtin.stat:
+        path: "{{ cifmw_tobiko_artifacts_basedir }}/id_ecdsa"
+      register: check_id_ecdsa
+
+    - name: Create the ssh keys
+      community.crypto.openssh_keypair:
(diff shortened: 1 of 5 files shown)


roles/tobiko/tasks/main.yml
```diff
@@ -33,6 +33,29 @@
...
+    - name: Create the ssh keys
+      community.crypto.openssh_keypair:
+        path: "{{ cifmw_tobiko_artifacts_basedir }}/id_ecdsa"
+        type: ecdsa
+        passphrase: ''
+      when: not check_id_ecdsa.stat.exists
+
+- name: Create tobiko.conf
+  vars:
+    tobikoconf_combined: "{{ cifmw_tobiko_tobikoconf_default | combine(cifmw_tobiko_tobikoconf ,recursive=True) }}"
```
cjeanner, line 53: So that we avoid a long line (also, note the space location in the `combine` call)
```suggestion
    tobikoconf_combined: >-
      {{
          cifmw_tobiko_tobikoconf_default |
          combine(cifmw_tobiko_tobikoconf, recursive=True)
      }}
```

psathyan (approved):
/lgtm

PR #1053, opened 2024-01-22
Fix refspec, url for cloning projects in reproducer

cescgina:
Zuul reproducer does only seem to support cloning projects from github,
while trying to clone projects from gerrit it uses a wrong url and
refspec. This change introduces two filters to get the right values for
url and refspec for the different sources (github and gerrit).

  - [x] README in the role

### DIFF
--- a/plugins/filter/reproducer_gerrit_infix.py
+++ b/plugins/filter/reproducer_gerrit_infix.py
@@ -1,0 +1,58 @@
+#!/usr/bin/env python3
+
+__metaclass__ = type
+
+DOCUMENTATION = """
+    name: reproducer_gerrit_infix
+    short_description: Maps a hostname to the infix needed for cloning a repo
+    for that hostname.
+    description:
+        - Maps a hostname to the infix needed for cloning a repo from that
+          hostname.Some gerrit instances add an infix like "gerrit", or "r"
+          to the repo clone url.
+    options:
+        _input:
+            description:
+                - repo hostname where we want to clone from.
+            type: str
+            required: true
+"""
+
+EXAMPLES = """
+    - name: Get infix to clone repo from rdo gerrit
+      ansible.builtin.set_fact:
+        gerrit_infix: >-
+            {{
+              "review.rdoproject.org" | reproducer_gerrit_infix
+            }}
+"""
+
+RETURN = """
+  _value:
+    description: The infix to add to the url to clone the repo.
+    type: str
+    sample:
+      "gerrit/"
+
+"""
+
+from ansible.errors import AnsibleFilterTypeError
+
+
+class FilterModule:
+    @classmethod
+    def __map_hostname_infix(cls, hostname):
+        if not isinstance(hostname, str):
+            raise AnsibleFilterTypeError(
+                f"reproducer_gerrit_infix requires a str, got {type(hostname)}"
+            )
+        if "rdoproject" in hostname:
+            return "r/"
+        elif "code.eng" in hostname:
+            return "gerrit/"
+        return ""
+
+    def filters(self):
+        return {
+            "reproducer_gerrit_infix": self.__map_hostname_infix,
+        }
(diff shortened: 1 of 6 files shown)


roles/reproducer/filter_plugins/reproducer_gerrit_infix.py
cjeanner, line 1: is there a way to get some unit tests for both filters?

cescgina: that's a good point, I'll look into it tomorrow

cescgina: added some unit tests, I could not find any convention around unit testing filters, so I had to improvise, let me know what you think

pablintino: I think that the ansible teams just ensure integration tests works. I have checked the `urlsplit` filter in the core repo and they only test it that way. https://github.com/ansible/ansible/blob/devel/test/integration/targets/filter_urlsplit/tasks/main.yml

cescgina: That was my thought as well if I couldn't get the python importing to work, it might make sense to use integration, that way we don't need the python `sys.path` hacks, wdyt?

cescgina: @pablintino @cjeanner updated the PR to have the tests in ansible

tests/integration/integration.cfg
pablintino, line 1: Uhmm. Is the ansible team doing something like this to be able to test? Sounds strange.... Have you tested to refer to your filters using the FQN?

cescgina: you're right, with FQN it seems to work if I move the filters to `plugins/filters`, I want to do some more testing locally and will update the PR tomorrow with the change

cescgina:
recheck

cjeanner:
/approve
there's a known(?) issue with the failing job.

plugins/filter/reproducer_gerrit_infix.py
```diff
@@ -0,0 +1,58 @@
...
+    options:
+        _input:
+            description:
+                - repo hostname where we want to clone from.
+            type: str
+            required: true
+"""
+
+EXAMPLES = """
+    - name: Get infix to clone repo from rdo gerrit
```
marios, line 22: i thought you were making up new words for a minute there :)
TIL "infix" ... I knew about prefix and suffix but had not heard of this word before

plugins/filter/reproducer_refspec.py
```diff
@@ -0,0 +1,75 @@
...
+            raise AnsibleFilterError(
+                "repo information does not contain 'project' field"
+            )
+        if "canonical_hostname" not in repo["project"]:
+            raise AnsibleFilterError(
+                "repo information does not contain 'canonical_hostname' field"
+            )
+        hostname = repo["project"]["canonical_hostname"]
+        if "rdoproject" in hostname or "code.eng" in hostname:
+            if "patchset" not in repo:
```
marios, line 61: trying to think how to tidy up a bit but it is hard as the fields are all different. I was thinking along the lines of moving the checks right at the start:
```
for check in ["change", "project", "patchset"]:
    if check not in repo.keys():
        error
```

marios (approved):
lgtm just added a note about a possible tidy up but nothing major. if you update i'll revisit

plugins/filter/reproducer_refspec.py
cescgina: I'm not sure we can easily do that, as there are many cases where some of those keys are not present and it should not raise an error, e.g, periodic jobs have no `check` field, and changes coming from github like the framework do not have a `patchset` field

PR #1058, opened 2024-01-24
Bring crio log gathering back to failling runs

pablintino:
  - [x] README in the role


roles/artifacts/defaults/main.yml
```diff
@@ -18,4 +18,7 @@
 # All variables intended for modification should be placed in this file.
 # All variables within this role should have a prefix of "cifmw_artifacts"
 cifmw_artifacts_basedir: "{{ cifmw_basedir | default(ansible_user_dir ~ '/ci-framework-data') }}"
+cifmw_artifacts_crc_host: "api.crc.testing"
+cifmw_artifacts_crc_user: "core"
+cifmw_artifacts_crc_sshkey: "~/.crc/machines/crc/id_ecdsa"
```
cjeanner, line 23: do we have this key on extracted layout? we may want to override it in such case to use zuul's key instead.

pablintino: it's in the job definition I think

rebtoor (approved):

roles/artifacts/defaults/main.yml
cjeanner: key exists, and is working, we're all good.

rebtoor (approved):

roles/artifacts/tasks/main.yml
```diff
@@ -65,6 +65,11 @@
   ansible.builtin.import_role:
     name: os_must_gather
 
+- name: Collect crc logs when tests failed
+  ignore_errors: true  # noqa: ignore-errors
+#  when: not cifmw_success_flag.stat.exists
```
lewisdenny, line 70: I think this was just used for testing, now we are collecting crc logs on every run, not only failed ones which doesn't match the commit message.

PR #1066, opened 2024-01-25
Get zuul reproducer back on track

cjeanner:
Since the introduction of networking_mapper in the reproducer role, the
zuul reproducer feature was broken.

This patch provides the needed update to re-enable the feature, by
remapping the networking_mapper environment output to the old
networking-info.yml format/file, allowing to re-run the job until we
switch the CI itself to the new networking_mapper usage.

This patch also provides the needed bits to properly support downstream
zuul and the RHEL ecosystem specificities. Most of the involved changes
are transparent to the user, since the tasks were already done on the
controller-0 - we had to extend them to the compute nodes.

### DIFF
--- a/roles/reproducer/tasks/configure_computes.yml
+++ b/roles/reproducer/tasks/configure_computes.yml
@@ -18,3 +18,28 @@
       ansible.builtin.command:
         cmd: |
           ping -c2 {{ _controller_ip4 }}
+
+    - name: RHEL related tasks on computes
+      become: true
+      when:
+        - cifmw_repo_setup_rhos_release_rpm is defined
+      block:
+        - name: Get rhos-release and CA
+          ansible.builtin.import_tasks: rhos_release.yml
+
+        - name: Configure rhos-release
+          ansible.builtin.command:
+            cmd: "rhos-release {{ cifmw_repo_setup_rhos_release_args }}"
+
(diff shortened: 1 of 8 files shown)


cjeanner:
Switching to draft, some more tests are needed.

cescgina (commented):
/lgtm

roles/reproducer/tasks/configure_computes.yml
```diff
@@ -18,3 +18,28 @@
...
+        - cifmw_repo_setup_rhos_release_rpm is defined
+      block:
+        - name: Get rhos-release and CA
+          ansible.builtin.import_tasks: rhos_release.yml
+
+        - name: Configure rhos-release
+          ansible.builtin.command:
+            cmd: "rhos-release {{ cifmw_repo_setup_rhos_release_args }}"
+
+    - name: Create repositories on computes
```
rebtoor, line 34: compute nodes?

cjeanner: `computes` matches the group name in the ansible inventory. Both are valid imho.

rebtoor: oh, that makes sense. :+1:

rebtoor (approved):

PR #1069, opened 2024-01-26
Add dmesg log in post-run

danpawlik:
Sometimes, the job might fail not because of the change but because of the hypervisor issue. This commit starts collecting dmesg logs into a separate file for later analysis.

### DIFF
--- a/ci/playbooks/collect-logs.yml
+++ b/ci/playbooks/collect-logs.yml
@@ -67,6 +67,7 @@
               pip3 --version >> ./python.log;
               command -v ansible && ansible --version >> ./python.log;
               pip3 freeze >> ./python.log;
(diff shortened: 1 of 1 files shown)


ci/playbooks/collect-logs.yml
```diff
@@ -67,6 +67,7 @@
               pip3 --version >> ./python.log;
               command -v ansible && ansible --version >> ./python.log;
               pip3 freeze >> ./python.log;
+              dmesg > ./dmesg.log;
```
cjeanner, line 70: `dmesg -T` to get proper, readable timestamp?

pablintino: @cjeanner fair point!

PR #1073, opened 2024-01-27
Extend ci_gen_kustomize_values test coverage

cjeanner:
Ensuring generation happens isn't enough, we also want to ensure the
generated values actually compile with kustomize.

Those new tests will ensure we can modify the role later without
breaking the feature.


marios (approved):
/lgtm

stared for a bit and it seems like a sane thing to do ;)

cjeanner:
/hold
I want to see if I can't, by chance, get over the ansible-lint issue

cjeanner:
/unhold

apparently we can't skip that one. Best I could do was to transform it as a "warning", but that's not really better (exit 2, so failing).

roles/ci_gen_kustomize_values/molecule/default/prepare.yml
```diff
@@ -20,7 +20,7 @@
   vars:
     cifmw_ci_gen_kustomize_values_src_dir: >-
       {{
-        (ansible_user_dir,
+        (lookup('env', 'HOME', '/home/zuul'),
```
cescgina, line 23: question, I assume that `/home/zuul/` is the default value, don't we need a `default` keyword? The examples in the docs all seem to use it https://docs.ansible.com/ansible/latest/collections/ansible/builtin/env_lookup.html

cjeanner: It seems to work without the keyword - at least molecule isn't failing with that, and seems to provide the correct value

cescgina (commented):
/lgtm

roles/ci_gen_kustomize_values/molecule/default/converge.yml
```diff
@@ -79,6 +79,52 @@
...
+        remote_src: true
+        src: "{{ artifacts }}/ci_gen_kustomize_values/{{ item.key }}/values.yaml"
+        dest: "{{ cifmw_ci_gen_kustomize_values_architecure_repo }}/{{ item.value }}"
+      loop:
+        - key: 'network-values'
+          value: 'examples/va/nfv/sriov/values.yaml'
+        - key: 'edpm-values'
+          value: 'examples/va/nfv/sriov/edpm/values.yaml'
+
+    - name: Ensure kustomize is able to build
```
fultonj, line 120: I like having this extra test so overall I like this patch.

One thought: if you are going to pull from the architecture repo and then modify that file but the source at the architecture repo broke for some reason, then this test might turn up false positives. One way to prevent that is to have a sanity test which pulls the CR from the architecture repos and runs kustomize on it before anything changes. That would tell you that if this new job fails that it wasn't due to a problem in a proposed patch to ci-framework but a problem in the architecture repo.

cjeanner: so running twice? Might be a thing indeed. BTW, maybe a job against the `architecture` repository should do the `run kustomize` ? To ensure no breaking change may slip in?

fultonj: Yeah, I'd like it if the architecture repository had a job but that repository has no automated jobs right now. Thus, adding an extra task or two to this PR to run kustomize first will prevent this problem for you now.

cjeanner: just added that as a look, and we'll generate "vanilla-{{ file }}" we may compare with the ones we generate later if we want.

cjeanner: Seems to work as expected:
https://logserver.rdoproject.org/73/1073/f3f345dd3fbc860319094cbd7c89c97955b3513a/github-check/cifmw-molecule-ci_gen_kustomize_values/0c2016f/ci-framework-data/artifacts/

We have the 4 wanted files.

fultonj: +2

roles/ci_gen_kustomize_values/molecule/default/converge.yml
```diff
@@ -25,8 +25,20 @@
...
       }}
+    artifacts: "{{ cifmw_ci_gen_kustomize_values_basedir }}/artifacts"
   tasks:
-    # We first create a dedicated keypair.
+    - name: Ensure architecture repository is building
+      ansible.builtin.shell:
+        chdir: "{{ cifmw_ci_gen_kustomize_values_architecure_repo }}/{{ item.path }}"
+        cmd: >-
+          {{ ansible_user_dir }}/bin/kustomize build >
+          {{ artifacts}}/vanilla-{{ item.name }}.yaml
```
rebtoor, line 35: a space missing after `artifacts`

roles/ci_gen_kustomize_values/molecule/default/converge.yml
```diff
@@ -79,6 +90,52 @@
...
+          value: 'examples/va/nfv/sriov/values.yaml'
+        - key: 'edpm-values'
+          value: 'examples/va/nfv/sriov/edpm/values.yaml'
+
+    - name: Ensure kustomize is able to build
+      ansible.builtin.shell:
+        chdir: "{{ cifmw_ci_gen_kustomize_values_architecure_repo }}/{{ item.path }}"
+        cmd: >-
+          {{ ansible_user_dir }}/bin/kustomize build >
+          {{ artifacts}}/{{ item.name }}.yaml
```
rebtoor, line 136: ditto

roles/ci_gen_kustomize_values/molecule/default/converge.yml
```diff
@@ -25,8 +25,20 @@
...
   tasks:
-    # We first create a dedicated keypair.
+    - name: Ensure architecture repository is building
+      ansible.builtin.shell:
+        chdir: "{{ cifmw_ci_gen_kustomize_values_architecure_repo }}/{{ item.path }}"
+        cmd: >-
+          {{ ansible_user_dir }}/bin/kustomize build >
+          {{ artifacts}}/vanilla-{{ item.name }}.yaml
+      loop:
+        - name: 'dataplane'
```
pablintino, line 37: Are we going to maintain this list?

roles/ci_gen_kustomize_values/molecule/default/converge.yml
cjeanner: more than that, I should use the `artifacts` parameter instead of re-writing the whole path in some other places. Updated.

rebtoor (approved):

roles/ci_gen_kustomize_values/molecule/default/prepare.yml
```diff
@@ -56,3 +56,8 @@
       ansible.builtin.git:  # noqa: latest[git]
         repo: "https://github.com/openstack-k8s-operators/architecture"
         dest: "{{ ansible_user_dir }}/src/github.com/openstack-k8s-operators/architecture"
+
+- name: Download tools for later testing and validations
+  ansible.builtin.import_playbook: >-
```
pablintino, line 61: Why not using this same notation? https://github.com/openstack-k8s-operators/ci-framework/pull/1073/files#diff-7f9fbb85b66eb3b26a6abc6a28034e080eb0c8e7629216f1659be3691c612913R58
Or the one that uses something like this? `zuul.projects['github.com/openstack-k8s-operators/install_yamls'].src_dir`

roles/ci_gen_kustomize_values/molecule/default/converge.yml
cjeanner: that one here? it's for the testing - unless there's a plan to remove that specific example, we shouldn't have any update... ?

roles/ci_gen_kustomize_values/molecule/default/prepare.yml
cjeanner: that may have been a rest of the tests (and failures) I tried to correct with the ansible-lint + "missing playbook".

PR #1075, opened 2024-01-29
[devscripts] Enable ip forwarding at global level

psathyan:
This patch enables IP forwarding when the network type is OVNKubernetes by default. This is required as the default number of NICs in the deployed cluster is 2.

It fixes the DNS query issue observed when configuring the OpenStack compute nodes.

*Network Operator details*
````
Name:         cluster
Namespace:    
Labels:       <none>
Annotations:  networkoperator.openshift.io/ovn-cluster-initiator: 192.168.111.22
API Version:  operator.openshift.io/v1
Kind:         Network
Metadata:
  Creation Timestamp:  2024-01-29T09:12:43Z
  Generation:          74
  Resource Version:    35543
  UID:                 e3805527-d2ed-486b-9b89-06de28e363d1
Spec:
  Cluster Network:
    Cidr:         192.168.16.0/20
    Host Prefix:  22
  Default Network:
    Ovn Kubernetes Config:
      Egress IP Config:
      Gateway Config:
        Ip Forwarding:     Global
        Routing Via Host:  false
      Geneve Port:         6081
```

### DIFF
--- a/roles/devscripts/molecule/default/converge.yml
+++ b/roles/devscripts/molecule/default/converge.yml
@@ -26,7 +26,7 @@
       domains:
         - "ocp.openstack.lab"
     cifmw_devscripts_dry_run: true
-    cifmw_devscripts_ocp_version: "4.13.12"
+    cifmw_devscripts_ocp_version: "4.14.9"
     cifmw_devscripts_user: "{{ ansible_user_id }}"
     cifmw_devscripts_repo_dir: "{{ (ansible_user_dir, 'src', 'dev-scripts') | path_join }}"
     cifmw_devscripts_data_dir: "{{ (ansible_user_dir, 'ci-framework-data') | path_join }}"
@@ -61,6 +61,7 @@
           - cifmw_devscripts_config.provisioning_network_profile == 'Managed'
           - cifmw_devscripts_config.num_masters == 3
           - cifmw_devscripts_config.num_workers == 0
+          - cifmw_devscripts_config.assets_extra_folder == '/home/dev-scripts/assets'
 
     - name: Collect stat information
       ansible.builtin.stat:
@@ -71,6 +72,7 @@
         - "{{ cifmw_devscripts_repo_dir }}"
         - "{{ cifmw_devscripts_logs_dir }}"
         - "{{ cifmw_devscripts_output_dir }}"
+        - "{{ cifmw_devscripts_config.assets_extra_folder }}"
       register: stat_results
 
     - name: Test directory exists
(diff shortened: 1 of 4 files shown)


roles/devscripts/tasks/sub_tasks/14_user.yml
```diff
@@ -44,3 +45,14 @@
     mode: "0640"
     content: |
       {{ cifmw_devscripts_user }}    ALL=(ALL)    NOPASSWD: ALL
+
+- name: Enable IP forwarding in the Network Operator.
+  when:
+    - cifmw_devscripts_config['network_type'] == 'OVNKubernetes'
+    - cifmw_devscripts_config['openshift_version'] is match("4\.14.*")
```
karelyatin, line 52: what about future versions i.e 4.15+ ?

psathyan: Since we are seeing the problem with `4.14`, I have added the specific check

psathyan:
recheck

roles/devscripts/tasks/sub_tasks/14_user.yml
cjeanner: I'd use the `is version()` test instead:
https://docs.ansible.com/ansible/latest/collections/ansible/builtin/version_test.html

karelyatin: With https://issues.redhat.com/browse/OCPBUGS-3176 it changed default behavior so should apply to future versions, so would be good to cover those cases.

psathyan: Thank you @cjeanner ... something new. :)
Thanks @karelyatin , I agree that the check should be 4.14 and +

Made the necessary modifications in the next commit.

karelyatin (commented):
/lgtm

roles/devscripts/molecule/default/converge.yml
```diff
@@ -26,7 +26,7 @@
       domains:
         - "ocp.openstack.lab"
     cifmw_devscripts_dry_run: true
-    cifmw_devscripts_ocp_version: "4.13.12"
+    cifmw_devscripts_ocp_version: "4.14.9"
```
cjeanner, line 29: Should we bump the version in the defaults?

roles/devscripts/vars/main.yml
```diff
@@ -33,6 +33,7 @@ cifmw_devscripts_rhsm_enabled_repos: false
 
 cifmw_devscripts_config_defaults:
   working_dir: "/home/dev-scripts"
+  assets_extra_folder: "/home/dev-scripts/assets"
```
cjeanner, line 36: a pity dev-scripts doesn't have converging name to either `*_dir` or `*_folder` - we'd be able to just extract those in a loop...

cjeanner:
/approve

Looks sane - just a couple of notes, maybe a follow-up would be a release bump in the defaults?

roles/devscripts/molecule/default/converge.yml
psathyan: Let's move to the latest version in a separate PR. I am running some tests with `4.14.9` based on the results will spawn a separate PR. (easy for revert also)

PR #1077, opened 2024-01-29
Ensure we have install_yamls in prow container

cjeanner:
This is needed when we want to run ansible-lint and other things on
ci-framwork code that may depend on install_yamls (mostly when we
`import_playbook`).

### DIFF
--- a/containerfiles/Containerfile.ci
+++ b/containerfiles/Containerfile.ci
@@ -5,3 +5,5 @@ USER root
 # workaround for RHBZ#2184640
 RUN rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-centosofficial
 RUN dnf update -y && dnf install -y git python3-pip python3-netaddr make gcc sudo rsync && yum clean all
+RUN mkdir -p /home/zuul/src/github.com/openstack-k8s-operators
+RUN git clone https://github.com/openstack-k8s-operators/install_yamls /home/zuul/src/github.com/openstack-k8s-operators/install_yamls


containerfiles/Containerfile.ci
```diff
@@ -5,3 +5,5 @@ USER root
 # workaround for RHBZ#2184640
 RUN rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-centosofficial
 RUN dnf update -y && dnf install -y git python3-pip python3-netaddr make gcc sudo rsync && yum clean all
+RUN mkdir -p /home/zuul/src/github.com/openstack-k8s-operators
+RUN git clone https://github.com/openstack-k8s-operators/install_yamls /home/zuul/src/github.com/openstack-k8s-operators/install_yamls
```
rebtoor, line 9: can we export `/home/zuul/src/github.com/openstack-k8s-operators` as a var so we can reuse it inside the Containerfile?

rebtoor (changes requested):

marios (approved):
lgtm

containerfiles/Containerfile.ci
cjeanner: Any actual usecase ?

marios: vars makes sense when it is used >1 and I only see it once in this file

rebtoor: in fact, we are using that dir twice in 2 lines. It's not so important BTW, it could be useful if we needed to change it but it's not the case for now I guess.

rebtoor (approved):

rebtoor (approved):

PR #1078, opened 2024-01-29
Correct ci_script issues

cjeanner:
Since we're consuming ci_script more widely in the project, we're seeing
some issues:
- it was checking for local directories to exist, while we may not need
  them in case of remote run.
- it wasn't properly cleaning parameters when passed down to the
  ansible.builtin.script, leading to issues

This patch corrects both issues, by not checking for directory
availability anymore, and properly cleaning parameters when needed.


marios (approved):
thanks

PR #1082, opened 2024-01-30
Don't remove bootstrap service from sevice list

rabi:
This removes the first service from the service list which happens to be `bootstrap` service and hence the required users and groups are not created as expected.

### DIFF
--- a/hooks/playbooks/fetch_compute_facts.yml
+++ b/hooks/playbooks/fetch_compute_facts.yml
@@ -233,12 +233,6 @@
                     value: ["{{ content_provider_registry_ip }}:5001"]
               {% endif %}
 
-              {% if zuul is defined and 'periodic' in zuul.job %}
-                  - op: remove
-                    path: /spec/services/0
-                    value: repo-setup
-              {% endif %}
-
                   - op: add
                     path: /spec/nodeTemplate/ansible/ansibleVars/edpm_sshd_allowed_ranges
                     value: ["0.0.0.0/0"]


hooks/playbooks/fetch_compute_facts.yml
```diff
@@ -233,12 +233,6 @@
                     value: ["{{ content_provider_registry_ip }}:5001"]
               {% endif %}
 
-              {% if zuul is defined and 'periodic' in zuul.job %}
-                  - op: remove
-                    path: /spec/services/0
-                    value: repo-setup
```
cjeanner, line 239: Shouldn't the "remove" match the "value" as a security?

rabi: "AFAIK Kustomize does not provide a built-in mechanism for removing a key based on its value directly using declarative configuration."  As you can see it's removing the `bootstrap` service [1]. 

Can you point me to docs where you have seen that feature documented?

[1] https://sf.hosted.upshift.rdu2.redhat.com/logs/01/01da771c5a9cdc5a6150d2d8a43de93162bb2f0a/openstack-periodic-integration-rhos-18-rhel9/periodic-podified-multinode-edpm-deployment-rhel9-osp18-crc/5f3970e/controller/ci-framework-data/artifacts/manifests/openstack/dataplane/cr/cifmw-kustomization-result.yaml

cjeanner: great... That's how we see good tools :(.

Since repo-setup has been removed, we can drop that removal indeed.

pablintino: Agree, is not possible, it's unfortunate but json-patching a dict/list by value is not possible.

PR #1045, opened 2024-01-18
[kustomize_deploy] Introduce `kustomize_deploy` role

rebtoor:
This PR introduce the `kustomize_deploy` role which is used to deploy VA scenarios using as input:

1. VAs taken from [architecture](https://github.com/openstack-k8s-operators/architecture) repo
2. configurations taken `values.yml` file(s)

Combining those two input sources, and following the manual steps defined [here](https://github.com/openstack-k8s-operators/architecture/tree/main/examples/va/hci), `kustomize` tool is used to generate CR files so they can be applied on an Openshift cluster.

A proper molecule scenario for the hci architecture is provided.

This PR will cover the deployment of the OLM dependencies and the deployment of the OSP control-plane.

As a pull request owner and reviewers, I checked that:
- [x] Appropriate testing is done and actually running
- [x] Appropriate documentation exists and/or is up-to-date:
  - [x] README in the role
  - [x] Content of the docs/source is reflecting the changes

PR #1084, opened 2024-01-30
Converge usage of the architecture repository

cjeanner:
With #1045, we're seeing a new top-level parameter introduced in order
to point to the architecture repository.
We're also seeing some differences with the way scenarios are defined,
showing some more "defensive" approach toward the users, ensuring typos
should be avoided.

This patch should hopefully help limiting the differences to the strict
minimum, while correcting a couple of typo introduced in the first
iteration.

It also includes a couple of examples for the role, and some updates to
the reproducer role to match the new conventions.

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/1045

  - [X] README in the role


cjeanner:
recheck

rebtoor (approved):

PR #1086, opened 2024-01-30
[devscripts] Enable routing via host

psathyan:
When we use NetworkManagers dns services, it is required to enable routing via the OCP nodes.

*Log snippet*
```
zuul@controller-0 ~]$ oc describe network.operator
Name:         cluster
Namespace:    
Labels:       <none>
Annotations:  networkoperator.openshift.io/ovn-cluster-initiator: 192.168.111.22
API Version:  operator.openshift.io/v1
Kind:         Network
Metadata:
  Creation Timestamp:  2024-01-31T02:25:52Z
  Generation:          76
  Resource Version:    32708
  UID:                 f9989752-d237-4c03-89c5-36716883f0be
Spec:
  Cluster Network:
    Cidr:         192.168.16.0/20
    Host Prefix:  22
  Default Network:
    Ovn Kubernetes Config:
      Egress IP Config:
      Gateway Config:
        Ip Forwarding:     Global
        Routing Via Host:  true
      Geneve Port:         6081
      Mtu:                 1400
      Policy Audit Config:
        Destination:            null
        Max File Size:          50
        Max Log Files:          5
        Rate Limit:             20
        Syslog Facility:        local0
    Type:                       OVNKubernetes
  Deploy Kube Proxy:            false
  Disable Multi Network:        false
  Disable Network Diagnostics:  false
  Log Level:                    Normal
  Management State:             Managed
  Observed Config:              <nil>
  Operator Log Level:           Normal
  Service Network:
    172.30.0.0/16
  Unsupported Config Overrides:  <nil>
  Use Multi Network Policy:      false
Status:
```
*Job progress*
```
[zuul@controller-0 custom]$ oc get jobs
NAME                                           COMPLETIONS   DURATION   AGE
bootstrap-dep-pre-ceph                         1/1           2m31s      7m51s
configure-network-dep-pre-ceph                 0/1           2m1s       2m1s
dataplane-deployment-repo-setup-dep-pre-ceph   1/1           29s        8m20s
download-cache-dep-pre-ceph                    1/1           3m19s      5m20s
keystone-cron-28444621                         1/1           5s         23m
```

### DIFF
--- a/roles/devscripts/tasks/sub_tasks/14_user.yml
+++ b/roles/devscripts/tasks/sub_tasks/14_user.yml
@@ -49,10 +49,20 @@
 - name: Enable IP forwarding in the Network Operator.
   when:
     - cifmw_devscripts_config['network_type'] == 'OVNKubernetes'
-    - "cifmw_devscripts_config.openshift_version is ansible.builtin.version('4.14.0', '>=')"
-  ansible.builtin.copy:
-    src: "files/forwarding.yaml"
-    dest: "{{ cifmw_devscripts_config['assets_extra_folder'] }}/forwarding.yaml"
+    - ("cifmw_devscripts_config.openshift_version is ansible.builtin.version('4.14.0', '>=')") or
+      (cifmw_devscripts_host_routing | bool)
+  vars:
+    ip_forward: "{{ cifmw_devscripts_config.openshift_version is ansible.builtin.version('4.14.0', '>=') }}"
+    host_routing: "{{ cifmw_devscripts_host_routing | bool }}"
+  ansible.builtin.template:
+    src: "templates/ovn_config.j2"
+    dest: >-
+      {{
+        [
+          cifmw_devscripts_config['assets_extra_folder'],
+          'ovn_k8s_config.yaml'
+        ] | ansible.builtin.path_join
+      }}
     owner: "{{ cifmw_devscripts_user }}"
     group: "{{ cifmw_devscripts_user }}"
     mode: "0644"
--- a/roles/devscripts/README.md
+++ b/roles/devscripts/README.md
@@ -32,6 +32,7 @@ managed services.
 * `cifmw_devscripts_use_layers` (bool) Toggle overlay support. Specifically, this boolean will instruct the role to
   shutdown the whole OCP cluster, dump the XML, undefine the nodes, and prevents running the "post" tasks. Defaults to `false`.
 * `cifmw_devscripts_remove_default_net` (bool) Remove the default virtual network. Defaults to `false`.
+* `cifmw_devscripts_host_routing` (bool) Enable routing via host for OCP nodes in case of OVNKubernetes. Defaults to `false`.
 
 ### Secrets management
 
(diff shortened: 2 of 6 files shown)


lewisdenny:
recheck

roles/devscripts/tasks/sub_tasks/14_user.yml
```diff
@@ -50,8 +50,10 @@
   when:
     - cifmw_devscripts_config['network_type'] == 'OVNKubernetes'
     - "cifmw_devscripts_config.openshift_version is ansible.builtin.version('4.14.0', '>=')"
-  ansible.builtin.copy:
-    src: "files/forwarding.yaml"
+  vars:
+    host_routing: "{{ cifmw_devscripts_host_routing }}"
+  ansible.builtin.template:
+    src: "templates/forwarding.yaml.j2"
```
karelyatin, line 56: would be good to use different name than forwarding.yaml as scope is now increased to also set routeviahost conditionally. Something like patch_networkoperator.yaml ?

roles/devscripts/templates/forwarding.yaml.j2
```diff
@@ -8,3 +8,6 @@ spec:
     ovnKubernetesConfig:
       gatewayConfig:
         ipForwarding: Global
+{% if host_routing %}
```
karelyatin, line 11: host_routing|bool else it will get true with cifmw_devscripts_host_routing: "false"

roles/devscripts/tasks/sub_tasks/14_user.yml
karelyatin: Routingviahost need to be set even for versions <4.14.0

karelyatin (changes requested):

roles/devscripts/tasks/sub_tasks/14_user.yml
psathyan: Thank you for the clarification.

Yes, the file can be renamed. Let me rename it to `net_op_defaults.yaml`. `patch` may be misleading as we are providing the custom configuration during installation.

roles/devscripts/templates/forwarding.yaml.j2
psathyan: Good point. In the next patch, let me ensure the value set for `host_routing` is a boolean one.

roles/devscripts/tasks/sub_tasks/14_user.yml
```diff
@@ -49,10 +49,18 @@
 - name: Enable IP forwarding in the Network Operator.
   when:
     - cifmw_devscripts_config['network_type'] == 'OVNKubernetes'
-    - "cifmw_devscripts_config.openshift_version is ansible.builtin.version('4.14.0', '>=')"
-  ansible.builtin.copy:
-    src: "files/forwarding.yaml"
-    dest: "{{ cifmw_devscripts_config['assets_extra_folder'] }}/forwarding.yaml"
+    - cifmw_devscripts_config.openshift_version is ansible.builtin.version('4.14.0', '>=') or
+      cifmw_devscripts_host_routing
```
karelyatin, line 53: cifmw_devscripts_host_routing|bool or may be can drop this condition completely as already checked in template

psathyan: I missed that one... thank you. Let's have the condition so that the config is not empty with
`OCP version is less than 4.14.0` and `routingViaHost: false`.

karelyatin (approved):
/lgtm

PR #1089, opened 2024-01-31
Increase default dataplane deployment timeout

rabi:
We've reached the threshold of 30 mins now and many jobs fail when ansibleee is still deploying services as a result of increased number of services and tasks.


cjeanner:
/lgtm

would a bump of resources in CRC or some resource consumption assessment help with that, or is it just "more services, it's taking more time, can't do anything about it" case?

PR #1092, opened 2024-01-31
Fix zuul HCI job that is missing the bootstrap service

pablintino:

### DIFF
--- a/roles/hci_prepare/tasks/phase1.yml
+++ b/roles/hci_prepare/tasks/phase1.yml
@@ -61,6 +61,7 @@
             path: /spec/services
             value:
               - repo-setup
+              - bootstrap
               - download-cache
               - configure-network
               - validate-network


roles/hci_prepare/tasks/phase1.yml
```diff
@@ -60,6 +60,7 @@
           - op: replace
             path: /spec/services
             value:
+              - bootstrap
```
viroel, line 63: phase-2 also chnage services:
https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/hci_prepare/tasks/phase2.yml#L95
But since bootstrap will run on phase-1, It won't be a problem. Just a matter of making things in sync.

pablintino: I have checked how it's done in the VA repo and that one is not getting called in the second deploy. So I think it's fine like this.

viroel:
/lgtm

zuul is going to validate in the hci job. lets see

PR #1093, opened 2024-01-31
Correct detection of the content-provider

cjeanner:
It seems DS isn't working like US, leading to some issues. We can't rely
on two parameters to check, only on registry_ip one - and its detection
doesn't need to rely on the zuul inventory.

### DIFF
--- a/roles/reproducer/tasks/ci_data.yml
+++ b/roles/reproducer/tasks/ci_data.yml
@@ -52,13 +52,17 @@
             zuul_inventory.all.hosts.keys() |
             select('match', '^compute.*') | length
            }}
+        zuul_params: >-
+          {{
+            lookup('file', _reproducer_basedir ~ '/parameters/zuul-params.yml') |
+            from_yaml
+          }}
       ansible.builtin.set_fact:
         cacheable: true
         ci_job_networking: "{{ zuul_inventory.all.hosts.crc.crc_ci_bootstrap_networking }}"
         use_content_provider: >-
           {{
-            zuul_inventory.all.hosts.crc.content_provider_registry_ip is defined and
-            zuul_inventory.all.hosts.crc.cifmw_operator_build_output is defined
+            zuul_params.content_provider_registry_ip is defined
           }}
         updated_layout:
           vms:


lewisdenny:
@cjeanner do you have a test project D/S testing this we can check?

cjeanner:
> @cjeanner do you have a test project D/S testing this we can check?

nope - tested locally for both upstream and downstream. I can check with @cescgina for a job to prove it

cjeanner:
> > @cjeanner do you have a test project D/S testing this we can check?
> 
> nope - tested locally for both upstream and downstream. I can check with @cescgina for a job to prove it

@lewisdenny so I got a proof of work in DS testproject - it's correctly setting up the content-provider now. I'll link the job privately (internal)

roles/reproducer/tasks/ci_data.yml
```diff
@@ -52,13 +52,17 @@
...
+            from_yaml
+          }}
       ansible.builtin.set_fact:
         cacheable: true
         ci_job_networking: "{{ zuul_inventory.all.hosts.crc.crc_ci_bootstrap_networking }}"
         use_content_provider: >-
           {{
-            zuul_inventory.all.hosts.crc.content_provider_registry_ip is defined and
-            zuul_inventory.all.hosts.crc.cifmw_operator_build_output is defined
+            zuul_params.content_provider_registry_ip is defined
```
marios, line 65: looks good. we only  set that when content provider runs  https://github.com/openstack-k8s-operators/ci-framework/blob/5bae6d325fee17999f93d4d4dfb0c4873febcd11/ci/playbooks/content_provider/run.yml#L50

only unknown to me is where/how you are grabbing the parameters/zuul-params.yml but I will assume that the content provider zuul vars are included in that

cescgina: @marios for reference the zuul params are fetched in the `ci_data.yml` task file: https://github.com/openstack-k8s-operators/ci-framework/blob/5bae6d325fee17999f93d4d4dfb0c4873febcd11/roles/reproducer/tasks/ci_data.yml#L23

cjeanner: indeed, we fetch earlier selected CI artifacts in order to know how many computes are needed, among things. We're then just parsing the zuul parameters and are able to find out if we have a content-provider or not.

PR #1094, opened 2024-01-31
Ensure we clear some directories

cjeanner:
There's a need to clear some directories, even without deepscrub tag.

Without this clean on the hypervisor, we may end with broken ansible
inventory, listing 2 "ocps" groups, one with CRC, the other for OCP
cluster.

Regarding the "local" directory (from the laptop/desktop/whatever),
it wasn't cleaned, meaning we may end with an ever growing directory, if
we're running many zuul job reproducers.

### DIFF
--- a/reproducer-clean.yml
+++ b/reproducer-clean.yml
@@ -25,6 +25,24 @@
         name: rhol_crc
         tasks_from: cleanup.yml
 
+    - name: Remove remote selected data directories
+      tags:
+        - deepscrub
+      ansible.builtin.file:
+        path: "{{ cifmw_reproducer_basedir }}/{{ item }}"
+        state: absent
+      loop:
+        - reproducer-inventory
+        - reproducer-network-env
+
+    - name: Remove local ci-reproducer data directory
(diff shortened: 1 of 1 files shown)


reproducer-clean.yml
```diff
@@ -25,6 +25,24 @@
...
+      tags:
+        - deepscrub
+      ansible.builtin.file:
+        path: "{{ cifmw_reproducer_basedir }}/{{ item }}"
+        state: absent
+      loop:
+        - reproducer-inventory
+        - reproducer-network-env
+
+    - name: Remove local selected data directories
```
marios, line 38: just a comment on the name, this one is not selecting directories more like "nuke entire ci-framework-data directory"

marios (approved):
lgtm

reproducer-clean.yml
cjeanner: yeah, I'll update the directory to only a selected one related to reproducer.

marios (approved):
thanks for update lgtm

PR #1065, opened 2024-01-24
[WIP] Install telemetry_tempest_plugin from master branch

elfiesmelfie:
The version of the telemetry-tempest-plugin in the tempest container is behind master.
The test in the tempest image require gnocchi, which is not included by the telemetry operator. In order to test the telemetry-operator tempest job, the newer version of the telemetry-tempest-plugin needs to be used.

In order to test this, the tempest role has been modified to update the version of telemetery-tempest-plugin to be included.

As a pull request owner and reviewers, I checked that:
- [ ] Appropriate testing is done and actually running
- [ ] Appropriate documentation exists and/or is up-to-date:
  - [ ] README in the role
  - [ ] Content of the docs/source is reflecting the changes

PR #1095, opened 2024-01-31
[test_operator] Update var name for tempest override_scenario bools

elfiesmelfie:
After passing the cifmw_test_operator_tempest_include_list var to the role, the following error occurred in roles/test_operator/tasks/tempest-tests.yml': line 22

'cifmw_test_operator_tempest_tests_include_override_scenario' is undefined

Previously, the first condition was false and caused the check to fail. Now that that var is defined, the second condition is being evaluated.

The cifmw_test_operator_tempest_tests_include_override_scenario var does not have a default set in default/main.yml

However, defaults/main.yml defines:
cifmw_test_operator_tests_include_override_scenario: false cifmw_test_operator_tests_exclude_override_scenario: false

The names of the vars were updated so that this bool is defined.

  - [x] README in the role

### DIFF
--- a/roles/test_operator/defaults/main.yml
+++ b/roles/test_operator/defaults/main.yml
@@ -53,5 +53,5 @@ cifmw_test_operator_default_groups:
   - default
 cifmw_test_operator_default_jobs:
   - default
-cifmw_test_operator_tests_include_override_scenario: false
-cifmw_test_operator_tests_exclude_override_scenario: false
+cifmw_test_operator_tempest_tests_include_override_scenario: false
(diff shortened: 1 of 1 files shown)


elfiesmelfie:
This is being tested in [1].

This commit is also part of  [2], which is a dependency of [1]

[1] https://github.com/openstack-k8s-operators/telemetry-operator/pull/284
[2] https://github.com/openstack-k8s-operators/ci-framework/pull/1065

lpiwowar (approved):
Thanks a lot! LGTM:)

kopecmartin (approved):
lgtm, thanks

PR #1097, opened 2024-02-01
Adds override checkout so that RPs can be overriden in child jobs

frenzyfriday:
  - [X] README in the role


pablintino:
recheck

PR #1100, opened 2024-02-01
Properly detect partition id for 3-nodes layout as well

cjeanner:
The validated-architecture-1.yml file has a way to detect what partition
has to be grown depending on the OS. This was missing from the
3-nodes.yml.

### DIFF
--- a/scenarios/reproducers/3-nodes.yml
+++ b/scenarios/reproducers/3-nodes.yml
@@ -16,6 +16,11 @@ cifmw_libvirt_manager_configuration:
   vms:
     compute:
       amount: 1
+      root_part_id: >-
+        {{
+          (cifmw_repo_setup_os_release is defined and cifmw_repo_setup_os_release == 'rhel') |
+          ternary(4, 1)
+        }}
       image_url: "{{ cifmw_discovered_image_url }}"
       sha256_image_name: "{{ cifmw_discovered_hash }}"
       image_local_dir: "{{ cifmw_basedir }}/images/"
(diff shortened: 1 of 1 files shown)


cescgina (commented):
/lgtm

scenarios/reproducers/3-nodes.yml
```diff
@@ -16,6 +16,11 @@ cifmw_libvirt_manager_configuration:
   vms:
     compute:
       amount: 1
+      root_part_id: >-
+        {{
+          (cifmw_repo_setup_os_release is defined and cifmw_repo_setup_os_release == 'rhel') |
+          ternary(4, 1)
```
marios, line 22: ack seems consistent with what we have in the validated-arch 1 https://github.com/marios/ci-framework/blob/37fd6398b31860e16cc30c82d9e9a2486b5c6e6d/scenarios/reproducers/validated-architecture-1.yml#L45 

is this something that will depend on the actual setup we run it on though? I am guessing the above works for the systems we are testing on but in theory the partition system used by the target machines may vary?

Just wondering if we need some note about this unless we already mention that in the docs?

marios (commented):
looks good just a thought about the chosen 4/1 partition ID

scenarios/reproducers/3-nodes.yml
cjeanner: it's not the target hypervisor, this is really the virtual machine drive (the .qcow2). I don't really expect it to change, sooo... That parameter is in the doc iirc.

PR #1101, opened 2024-02-01
Clone all projects in the zuul reproducer

cescgina:
Clone all projects listed in the zuul inventory when reproducing a zuul
job, not just those coming from github. This is needed when reproducing
jobs defined in downstream gitlab that consume playbooks from internal
gerrit repos. If the internal gerrit repos are not in a depends-on, they
are not pulled and the playbooks are not found.

### DIFF
--- a/roles/reproducer/tasks/push_code.yml
+++ b/roles/reproducer/tasks/push_code.yml
@@ -41,11 +41,10 @@
 
     - name: Fetch zuul.projects repositories for dependencies
       when:
-        - repo.key is match('^github.com')
         - "repo.key not in (zuul['items'] | map(attribute='project.canonical_name'))"
       ansible.builtin.git:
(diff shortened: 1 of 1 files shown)


cjeanner (commented):
/approve

It may slightly extend the time a reproducer gets ready - in such a case we may consider using some `async` in order to parallelize the clones.

cescgina:
> /approve
> 
> It may slightly extend the time a reproducer gets ready - in such a case we may consider using some `async` in order to parallelize the clones.

that's a fair point, as reference for a future discussion here is how long it took in the only run I have with this change:
```
2024-02-02 03:41:28,252 p=26483 u=zuul n=ansible | reproducer : Fetch zuul.projects repositories for dependencies --------- 22.44s
```
for now it's probably ok, but should definitely keep an eye on the time

PR #1102, opened 2024-02-01
Split out content-provider reproducer

cjeanner:
It may happen people want to run the content-provider automatically, but
the actual job manually.

It may also happen the content-provider just passes, but the subsequent
"job" run fails - and we of course don't want to re-run the
content-provider.

  - [X] README in the role


roles/reproducer/README.md
```diff
@@ -10,6 +10,7 @@ None
 * `cifmw_reproducer_kubecfg`: (String) Path to the CRC kubeconfig file. Defaults to the image_local_dir defined in the cifmw_libvirt_manager_configuration dict.
 * `cifmw_reproducer_repositories`: (List[mapping]) List of repositories you want to synchronize from your local machine to the ansible controller.
 * `cifmw_reproducer_run_job`: (Bool) Run actual CI job. Defaults to `true`.
+* `cifmw_reproducer_run_content_provider`: (Bool) Run content-provider job id needed. Defaults to `true`.
```
cescgina, line 13: typo? `id` should be `if`?

cjeanner: ah, right.

cescgina (commented):
/lgtm

pablintino:
/lgtm
Adding lgtm back

PR #1108, opened 2024-02-06
Ensure support for newer CRC (>=2.32)

cjeanner:
dnsmasq container was removed from newer CRC releases[1], leading to a
crash whenever we want to update its configuration.

This patch updates both reproducer and Zuul to ensure it's compatible
with older and newer releases.

[1] https://github.com/crc-org/snc/commit/35fce67a31e469f84e1047cc204771d12b5f6434

### DIFF
--- a/ci/playbooks/multinode-customizations.yml
+++ b/ci/playbooks/multinode-customizations.yml
@@ -9,85 +9,134 @@
       ansible.builtin.slurp:
         src: "/etc/ci/env/networking-info.yml"
 
-    - name: Harcode crc-pre script
-      become: true
+    - name: Check which dnsmasq config we must edit
+      register: _dnsmasq
+      ansible.builtin.stat:
+        path: '/srv/dnsmasq.conf'
+        get_attributes: false
+        get_checksum: false
+        get_mime: false
+
+    - name: Shared parameter across blocks
       vars:
-        _decoded_net_env: "{{ _cifmw_multinode_customizations_crc_net_env_slurp['content'] | b64decode | from_yaml }}"
+        _decoded_net_env: >-
+          {{
+            _cifmw_multinode_customizations_crc_net_env_slurp['content'] |
+            b64decode | from_yaml
+          }}
         _crc_default_net_ip: >-
           {{
-            _decoded_net_env.crc_ci_bootstrap_networks_out.crc.default.ip | ansible.utils.ipaddr('address')
+            _decoded_net_env.crc_ci_bootstrap_networks_out.crc.default.ip
           }}
-        _crc_default_net_dns: "{{ _decoded_net_env.crc_ci_bootstrap_provider_dns | default([]) }}"
-      ansible.builtin.copy:
-        dest: "/usr/local/bin/configure-pre-crc.sh"
-        mode: '0755'
-        owner: root
-        group: root
-        content: |-
-          #!/bin/bash
+      block:
+        - name: Manage name resolution and interfaces
+          become: true
+          vars:
+            _crc_default_iface: >-
+              {{
+                 _decoded_net_env.crc_ci_bootstrap_networks_out.crc.default.iface
+              }}
+            _crc_default_gw: >-
+              {{
+                 _decoded_net_env.crc_ci_bootstrap_networks_out.crc.default.gw
+              }}
+            _crc_private_connection_name: >-
+              {{
+                 _decoded_net_env.crc_ci_bootstrap_networks_out.crc.default.connection
+              }}
+          block:
+            - name: Ensure crc knows about its second NIC
+              community.general.nmcli:
+                autoconnect: true
+                conn_name: "{{ _crc_private_connection_name }}"
+                dns4: 127.0.0.1
+                ifname: "{{ _crc_default_iface }}"
+                type: ethernet
+                ip4: "{{ _crc_default_net_ip }}"
+                gw4: "{{ _crc_default_gw }}"
+                state: present
 
-          set -x
+            - name: Ensure crc does not get "public" DNS
+              become: true
+              community.general.nmcli:
+                autoconnect: true
+                conn_name: "Wired connection 1"
+                dns4_ignore_auto: true
+                state: present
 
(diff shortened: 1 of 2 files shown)


d34dh0r53 (commented):
/lgtm

ci/playbooks/multinode-customizations.yml
```diff
@@ -18,6 +26,11 @@
             _decoded_net_env.crc_ci_bootstrap_networks_out.crc.default.ip | ansible.utils.ipaddr('address')
           }}
         _crc_default_net_dns: "{{ _decoded_net_env.crc_ci_bootstrap_provider_dns | default([]) }}"
+        _dnsmasq_config: >-
+          {{
+            _dnsmasq.stat.exists |
+            ternary('/srv/dnsmasq.conf', '/etc/dnsmasq.d/crc-dnsmasq.conf')
```
marios, line 32: maybe can be
```
ternary(_dnsmasq.stat.path, '/etc/dnsmasq.d/crc-dnsmasq.conf')
```
to prevent duplication

pablintino (changes requested):
This requires deeper analysis. Why are we still calling the crc-pre? And even worst, why is the reproducer not failing if crc-pre is a service that exists only in the rhos-infra crc images?

cjeanner:
> This requires deeper analysis. Why are we still calling the crc-pre? And even worst, why is the reproducer not failing if crc-pre is a service that exists only in the rhos-infra crc images?

reproucer doesn't use crc-pre. I didn't want to dig too deep in the whole bootstrap playbook, to be fair.

ci/playbooks/multinode-customizations.yml
cjeanner: good point

ci/playbooks/multinode-customizations.yml
```diff
@@ -9,6 +9,14 @@
       ansible.builtin.slurp:
         src: "/etc/ci/env/networking-info.yml"
 
+    - name: Check which dnsmask config we must edit
```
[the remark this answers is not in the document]
cjeanner: corrected

cjeanner:
@pablintino I think we can re-open that PR for review and unhold it? Once we get the new nested image ready, molecule should work properly.
The `multinode-customize.yml` play is now converged with what the reproducer `configure_crc.yml` tasks file does, ensuring we're not relying on the custom `pre-crc` service anymore.
The nmcli changes in that play are actually testing in any CI run we have here - the only thing "not tested in CI" is the dnsmasq service reload - I don't expect it to fail at that point, to be fair... ?

cjeanner:
Molecule's failing due to nested virt (for a change), there's an on-going issue with that in RDO CI.

IMHO, the change is safe to merge, up to the team.

ci/playbooks/multinode-customizations.yml
```diff
@@ -9,85 +9,135 @@
...
-        state: absent
+            # TODO: ensure we know CRC IP on osp_trunk interface
+            # Needs the networking_mapper output.
+            - name: Configure local DNS for CRC pod
+              become: true
+              register: last_modification
+              ansible.builtin.replace:
+                path: "{{ _dnsmasq_config }}"
+                regexp: "192.168.130.11"
+                replace: "{{ _crc_default_net_ip }}"
```
marios, line 102: hard to follow the indentation here but... are we really still inside the outer block here so we have this available default_net_ip (line 27)

ci/playbooks/multinode-customizations.yml
```diff
@@ -9,85 +9,135 @@
...
+              ansible.builtin.service:
+                name: NetworkManager
+                state: reloaded
+
+        - name: Configure dnsmasq
+          vars:
+            _dnsmasq_config: >-
+              {{
+                _dnsmasq.stat.exists |
+                ternary('/srv/dnsmasq.conf', '/etc/dnsmasq.d/crc-dnsmasq.conf')
```
marios, line 77: added an earlier comment perhaps use _dnsmasq.stat.path as the true result instead of repeating the path.

do we *also* want to stat /etc/dnsmasq.d/crc-dnsmasq.conf as an extra sanity then just error if neither?

marios (commented):
stared again and it grew since the last visit :)

generally lgtm some inline thoughts.

holding approve for the review call

ci/playbooks/multinode-customizations.yml
cjeanner: yes, we are - there are 2 levels of blocks. I know, it may be slightly hard, but it allow a lot of code de-duplication. Also, it did properly pass CI before RDO got down - molecule tests are failing due to the nested virt issue currently roaming the jobs :(

ci/playbooks/multinode-customizations.yml
cjeanner: ah, right for the .path - missed that one while doing the deeper update.

As for checking the other file: it would fail if it doesn't exist anyway, because we don't instruct ansible to `create: true` the file.
AFAIK, the next move CRC *could* do is moving dnsmasq to the NetworkManager service... Not sure they want to go that way though.

ci/playbooks/multinode-customizations.yml
```diff
@@ -9,85 +9,133 @@
...
-      containers.podman.podman_container:
-        name: crc-dnsmasq
-        state: absent
+            - name: Configure local DNS for CRC pod
+              become: true
+              register: last_modification
+              ansible.builtin.replace:
+                path: "{{ _dnsmasq_config }}"
+                regexp: "192.168.130.11"
+                replace: "{{ _crc_default_net_ip }}"
```
pablintino, line 100: There is a small change in the behavior here. The old code set there the IP only, without the network prefix length, as the `_decoded_net_env.crc_ci_bootstrap_networks_out.crc.default.ip` value was filtered with `ansible.utils.ipaddr('address')`.

cjeanner: oh, good catch. It seems to be working though, but indeed, for the sake of consistency - lemme update it.

raukadah (approved):

marios (approved):
/lgtm

PR #1109, opened 2024-02-06
[test-operator] Introduce params for ssh-key and custom config files

lpiwowar:
In this patch [1] a new set of parameters were introduced to the test-operatr. The 'SSHKeySecretName' enables the user to mount custom private key via openshift secret and 'customOverwrite' gives the user the ability to mount an arbitrary file with a content into the tempest test pod.

This change exposes the newly introduced parameters via the test-operator role.

[1] https://github.com/openstack-k8s-operators/test-operator/pull/39

  - [x] README in the role

### DIFF
--- a/docs/dictionary/en-custom.txt
+++ b/docs/dictionary/en-custom.txt
@@ -339,6 +339,7 @@ polarion
 polkit
 pre
 prikey
+privatekey
 projectquay
 prometheus
 provisioner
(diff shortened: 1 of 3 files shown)


kopecmartin (approved):
lgtm

SeanMooney (commented):
this looks reasonable to me
i have not tested it yet but i belive it would enable our usecase for the whitebox-tempest-plugin.

arxcruz (approved):
/lgtm
lgtm

PR #1111, opened 2024-02-06
[test-operator] Collect logs when the pod times out

lpiwowar:
Currently, we do not collect the logs from the test pod when the pod times out. This patch makes sure that:
- we do not end the role when the test execution takes too long and
- that we collect the logs from the pod even when the test execution was not fully completed.

  - [x] README in the role

### DIFF
--- a/roles/test_operator/tasks/main.yml
+++ b/roles/test_operator/tasks/main.yml
@@ -148,10 +148,14 @@
     wait_condition:
       type: Complete
       status: true
+  ignore_errors: true
+  register: tempest
   when: not cifmw_test_operator_dry_run | bool
 
 - name: Start test-operator-logs-pod
-  when: not cifmw_test_operator_dry_run | bool
+  when:
+    - not cifmw_test_operator_dry_run | bool
+    - not tempest.failed
   kubernetes.core.k8s:
     kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
     api_key: "{{ cifmw_openshift_token | default(omit)}}"
@@ -190,7 +194,9 @@
   until: logs_pod.resources[0].status.phase == "Running"
   delay: 10
   retries: 20
-  when: not cifmw_test_operator_dry_run | bool
+  when:
+    - not cifmw_test_operator_dry_run | bool
+    - not tempest.failed
 
 - name: Get logs from test-operator-logs-pod
   environment:
@@ -199,7 +205,9 @@
   ansible.builtin.shell: >
     oc cp -n {{ cifmw_test_operator_namespace }} openstack/test-operator-logs-pod:mnt/
     {{ cifmw_test_operator_artifacts_basedir }}
-  when: not cifmw_test_operator_dry_run | bool
+  when:
+    - not cifmw_test_operator_dry_run | bool
+    - not tempest.failed
 
 - name: Get list of all pods
   kubernetes.core.k8s_info:


roles/test_operator/tasks/main.yml
```diff
@@ -148,10 +148,14 @@
...
       status: true
+  ignore_errors: true
+  register: tempest
   when: not cifmw_test_operator_dry_run | bool
 
 - name: Start test-operator-logs-pod
-  when: not cifmw_test_operator_dry_run | bool
+  when:
+    - not cifmw_test_operator_dry_run | bool
+    - not tempest.failed
```
SeanMooney, line 158: so we still want to get the logs if the wait for the results to complete failed.

so i don't think this will actually help as written

even if it failed i think we need to start the logs pod and extract the logs for the in progress job

otherwise isn't this the same as before?

lpiwowar: > so we still want to get the logs if the wait for the results to complete failed.

Yes, this is not a good approach. I updated the patch so that we collect the logs even when to pod fails. I merged this patch into this PR as the changes are connected [1]. 

> even if it failed i think we need to start the logs pod and extract the logs for the in progress job

All the logs are copied into the persistent volume at the end of the tests execution [2]. So I think it is safe to assume that there are no logs to be collected when the pod times out. Do you think I'm missing something? What we can do is collect the output of (oc logs).

[1] https://github.com/openstack-k8s-operators/ci-framework/pull/1112
[2] https://github.com/openstack-k8s-operators/tcib/blob/106264ab0a0b18be278d69b0664dad6a4f9be29a/container-images/tcib/base/os/tempest/run_tempest.sh#L265

SeanMooney: ywhen the job times out there are still files that can be collected so no its not a good assumption that if it times out there are no logs.

in addition to the pod logs for the pod standard out for the pod running tempest you can also extract the subunit file that tempest is creating and then use that to generate a report of the test that ran.

i have previously asked for the execution of tempest and report generation to be extracted into two separate scripts.
ideally from a zuul point of view the report generation and log collection could be run as a post action.

one of the requirement i express to arks last year was the ability to replicate how we run tempest with devstack like this

https://github.com/openstack/tempest/blob/master/tox.ini#L109C1-L120C123

or specifically like this

https://github.com/openstack/tempest/blob/master/tox.ini#L189-L201

we want wanted to be able to provide 2 sets of regexes of test to run

i.e. the test that are safe to run in parallel and the test that should run serial at the end

then we wanted dot have a single log collection phase to produce a single test report containing all results form both tempst invocations.

when jobs time out in our devstack jobs we get almost the same level of logs as when they don't. that is because we run tempest with timeout in the job so that after the timeout expires the tempest client exits and we can gather the logs normally.

i think we need to rework https://github.com/openstack-k8s-operators/tcib/blob/106264ab0a0b18be278d69b0664dad6a4f9be29a/container-images/tcib/base/os/tempest/run_tempest.sh#L215 to actually enforce the timeout. we may also be able to do this by setting some kind of job timeout on the kubernetes jobs if that is how the tests are being run today.

ideally we would extract the report generation into a separate script that would be then invoked by the ci framework after the pod terminates to generate the test report 

i.e. if we have n tempest CRs it would be nice to be able to aggregate them into one report when all of them have run but that a slightly different topic.

ill try and review #1112

PR #1112, opened 2024-02-06
[test-operator] Fix handling of failed job

lpiwowar:
[test-operator] Fix handling of failed job

There are two issues in the way of how the role does handle failed job

- Currently, the role waits for the Job that starts the execution of
  the test pod to reach Completed state but this state is only reached
  when the execution of the tests was successful. If there is an error
  the job finishes with Failed state.

- Logs are not collected when the the test pod times out

This patch updates the role so that it correctly waits for the job
completion (either with success or failure) and ensures that we collect
the logs (output of oc logs) even when the test pod times out.

  - [x] README in the role

### DIFF
--- a/roles/test_operator/tasks/main.yml
+++ b/roles/test_operator/tasks/main.yml
@@ -143,15 +143,25 @@
     namespace: "{{ cifmw_test_operator_namespace }}"
     kind: Job
     name: "{{ cifmw_test_operator_tempest_name }}"
-    wait: true
-    wait_timeout: "{{ cifmw_test_operator_timeout }}"
-    wait_condition:
-      type: Complete
-      status: true
+  retries: "{{ (cifmw_test_operator_timeout / 10) | round | int }}"
+  delay: 10
+  until: >
+    tempest.resources[0].status.succeeded | default(0) | int >= 1 or
+    tempest.resources[0].status.failed | default(0) | int >= 1
+  ignore_errors: true
+  register: tempest
+  when: not cifmw_test_operator_dry_run | bool
+
+- name: Check whether tempest timed out
+  ansible.builtin.set_fact:
+    tempest_timed_out: >-
+      {{ tempest.attempts == (cifmw_test_operator_timeout / 10) | round | int }}
   when: not cifmw_test_operator_dry_run | bool
 
 - name: Start test-operator-logs-pod
-  when: not cifmw_test_operator_dry_run | bool
+  when:
+    - not cifmw_test_operator_dry_run | bool
+    - not tempest_timed_out
   kubernetes.core.k8s:
     kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
     api_key: "{{ cifmw_openshift_token | default(omit)}}"
@@ -190,7 +200,9 @@
   until: logs_pod.resources[0].status.phase == "Running"
   delay: 10
   retries: 20
-  when: not cifmw_test_operator_dry_run | bool
+  when:
+    - not cifmw_test_operator_dry_run | bool
+    - not tempest_timed_out
 
 - name: Get logs from test-operator-logs-pod
   environment:
@@ -199,7 +211,9 @@
   ansible.builtin.shell: >
     oc cp -n {{ cifmw_test_operator_namespace }} openstack/test-operator-logs-pod:mnt/
     {{ cifmw_test_operator_artifacts_basedir }}
-  when: not cifmw_test_operator_dry_run | bool
+  when:
+    - not cifmw_test_operator_dry_run | bool
+    - not tempest_timed_out
 
 - name: Get list of all pods
   kubernetes.core.k8s_info:


kopecmartin (approved):
lgtm

roles/test_operator/tasks/main.yml
```diff
@@ -143,15 +143,25 @@
     namespace: "{{ cifmw_test_operator_namespace }}"
     kind: Job
     name: "{{ cifmw_test_operator_tempest_name }}"
-    wait: true
-    wait_timeout: "{{ cifmw_test_operator_timeout }}"
-    wait_condition:
-      type: Complete
-      status: true
+  retries: "{{ (cifmw_test_operator_timeout / 10) | round | int }}"
```
marios, line 146: ok default 3600
 https://github.com/openstack-k8s-operators/ci-framework/blob/fa2fcee6fec40f7ede3f2868a368149c583c51d5/roles/test_operator/defaults/main.yml#L23

roles/test_operator/tasks/main.yml
```diff
@@ -143,15 +143,25 @@
...
+    tempest.resources[0].status.succeeded | default(0) | int >= 1 or
+    tempest.resources[0].status.failed | default(0) | int >= 1
+  ignore_errors: true
+  register: tempest
   when: not cifmw_test_operator_dry_run | bool
 
-- name: Start test-operator-logs-pod
+- name: Check whether tempest timed out
+  ansible.builtin.set_fact:
+    tempest_timed_out: >-
```
marios, line 157: is there nothing returned for us here from the tempest.resources ? feels a bit fragile the way you are determining the timeout

lpiwowar: As we've already discussed there is unfortunately probably no other way how to determine the timeout.

roles/test_operator/tasks/main.yml
```diff
@@ -143,15 +143,25 @@
...
+      {{ tempest.attempts == (cifmw_test_operator_timeout / 10) | round | int }}
   when: not cifmw_test_operator_dry_run | bool
+
+- name: Start test-operator-logs-pod
+  when:
+    - not cifmw_test_operator_dry_run | bool
```
queria, line 163: would this mean there will be no logs collected in case it hanged/timed-out?

just wild guessing but if its maybe due to access to persistent storage or such, could there be some read-only parallel access, or maybe force terminate the execution on timeout to get access to logs?

lpiwowar: I wanted to skip these because there are no longs to be collected in case of time out [1]. The logs are moved to the directory where the PV is mounted at the end of the test execution.

However, it is possible that there might be an issue with parallel access to the logs. 

[1] https://github.com/openstack-k8s-operators/tcib/blob/106264ab0a0b18be278d69b0664dad6a4f9be29a/container-images/tcib/base/os/tempest/run_tempest.sh#L265

marios (approved):
as discussed yesterday this won't affect any currently running job until we wire it up so lets merge to unblock work here

SeanMooney:
i used depend on to include it in https://github.com/openstack-k8s-operators/nova-operator/pull/665
and it seamed to function fine. that last run did not time out but the logs are present
https://logserver.rdoproject.org/65/665/8a814593c6acbbf86b6ad3022bb565bb23d73247/github-check/nova-operator-tempest-multinode/ac65ef4/controller/ci-framework-data/tests/test_operator/stestr_results.html
so this is at least not breaking that existing behaivor

roles/test_operator/tasks/main.yml
SeanMooney: that is not correct

generate logs is post processing the tempest logs.

the tempest logs are stored in a binary format by stester when tempest is run in parrall with the execution fo the these

this https://github.com/openstack-k8s-operators/tcib/blob/106264ab0a0b18be278d69b0664dad6a4f9be29a/container-images/tcib/base/os/tempest/run_tempest.sh#L226C5-L226C66

stestr last --subunit > ${TEMPEST_PATH}testrepository.subunit

is just copying the last log to a new location.

however what we should do doing is making sure that the default .stestr folder
is mind mounted to a log volume so that the logs are directly written to a PVC and we can post process them even when a time out happens.


alternatively we shoudl be passing the time out to the tempest run invocation

timeout ${TEMPEST_TIMEOUT} tempest run ${TEMPEST_ARGS}

hhttps://github.com/openstack-k8s-operators/tcib/blob/106264ab0a0b18be278d69b0664dad6a4f9be29a/container-images/tcib/base/os/tempest/run_tempest.sh#L215


then the job will finish in a roghyly fixed time and it will generate the logs properly then we done need to wait for a fix period of time here in this script.

SeanMooney: making the run_tempest.sh script enforce the timeout when invoking tempest is obviously the simpler option

PR #1117, opened 2024-02-07
Mount oc and kubeconfig files on the tobiko container

eduolivares:
This is needed to run whitebox tests on a podified environment. Tobiko needs to access to those files, so proper read and exec permissions are necessary.


eduolivares:
This is being implemented on test-operator. We are deprecating the tobiko role and will use test-operator instead to run tobiko.

PR #1119, opened 2024-02-07
Configure Ceph Object Store to use external Ceph rgw

katarimanojk:
This patch configures ceph object store to use external ceph object gateway (rgw) by running set of openstack commands in [1].

[1]
https://github.com/openstack-k8s-operators/docs/blob/main/ceph.md#configure-swift-with-a-rgw-backend

  - [x] README in the role

### DIFF
--- a/roles/cifmw_cephadm/tasks/configure_object.yml
+++ b/roles/cifmw_cephadm/tasks/configure_object.yml
@@ -1,0 +1,61 @@
+---
+# Copyright 2024 Red Hat, Inc.
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+- name: Check for existing object store config
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+  delegate_to: localhost
+  when: cifmw_openshift_kubeconfig is defined
+  block:
+    - name: Check if swift is enabled in deployed controlplane
+      ansible.builtin.shell: "set -o pipefail && oc get $(oc get oscp -n openstack -o name) -o json| jq .spec.swift.enabled"
+      register: swift_in_ctlplane
+
+    # checking swift_endpoints_count will avoid unnecessary errors during ceph deployment re-run
+    - name: Check if swift endpoint is already created
+      ansible.builtin.shell: "set -o pipefail && oc rsh openstackclient openstack endpoint list | grep 'swift.*object-store' | wc -l"
+      register: swift_endpoints_count
+      ignore_errors: true
+
+- name: Display a note about swift deployment
+  ansible.builtin.debug:
+    msg: "WARNING: Swift is deployed and the endpoint exists already, ceph RGW cannot be configured as object store service"
+  when:
+    - cifmw_openshift_kubeconfig is defined
+    - cifmw_ceph_daemons_layout.rgw_enabled | default(true) | bool
+    - swift_in_ctlplane.stdout | bool
+
+- name: Configure object store to use rgw
+  cifmw.general.ci_script:
+    extra_args:
+      KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    output_dir: "/home/zuul/ci-framework-data/artifacts"
+    script: |-
+      oc rsh openstackclient openstack service create --name swift --description 'OpenStack Object Storage' object-store
+      oc rsh openstackclient openstack user create --project service --password {{ cifmw_ceph_rgw_keystone_psw }}  swift
+      oc rsh openstackclient openstack role create swiftoperator
+      oc rsh openstackclient openstack role create ResellerAdmin
+      oc rsh openstackclient openstack role add --user swift --project service member
+      oc rsh openstackclient openstack role add --user swift --project service admin
+      oc rsh openstackclient openstack endpoint create --region regionOne object-store public http://{{ cifmw_cephadm_vip }}:8080/swift/v1/AUTH_%\(tenant_id\)s
+      oc rsh openstackclient openstack endpoint create --region regionOne object-store internal http://{{ cifmw_cephadm_vip }}:8080/swift/v1/AUTH_%\(tenant_id\)s
+      oc rsh openstackclient openstack role add --project admin --user admin swiftoperator
+  delegate_to: localhost
+  when:
+    - cifmw_openshift_kubeconfig is defined
+    - cifmw_ceph_daemons_layout.rgw_enabled | default(true) | bool
+    - not swift_in_ctlplane.stdout | bool
+    - swift_endpoints_count.stdout == "0"
--- a/roles/cifmw_cephadm/tasks/post.yml
+++ b/roles/cifmw_cephadm/tasks/post.yml
@@ -17,6 +17,9 @@
 - name: Get ceph_cli
   ansible.builtin.include_tasks: ceph_cli.yml
 
+- name: Configure ceph object store to use external ceph object gateway
+  ansible.builtin.include_tasks: configure_object.yml
+
 - name: Get the ceph orchestrator status
   ansible.builtin.command: "{{ cifmw_cephadm_ceph_cli }} orch status --format json"
   register: ceph_orch_status


roles/cifmw_cephadm/tasks/post.yml
```diff
@@ -17,6 +17,27 @@
 - name: Get ceph_cli
   ansible.builtin.include_tasks: ceph_cli.yml
 
+- name: Check if swift endpoint is already created
+  shell: "oc rsh openstackclient openstack endpoint list | grep 'swift.*object-store' | wc -l"
+  register: swift_endpoint
+  delegate_to: localhost
+
+- name: Configure ceph rgw as backend to Swift
```
tosky, line 25: I may be wrong, but this doesn't configure Ceph RGW as backend for Swift, but it configures RGW to expose the same Object Storage API that Swift exposes. It is a compatible implementation of the same API, not a backend..
If tha's the case, please fix the task name and the commit message and the PR message.

katarimanojk: I think the intention here is to configure an external Ceph Object Gateway (RGW) as an Object Store service and Swift client tool can be used to interact with object store.

I just copied the commands and task name from [1] , @fmount can you confirm on this ?

[1] https://github.com/openstack-k8s-operators/docs/blob/main/ceph.md#configure-swift-with-a-rgw-backend

fmount: I think we can improve the wording to be something like: `Configuring Ceph Object Store to use external Ceph Object Gateway`, including the doc patch. We commonly refer to rgw as a backend, but technically we're not setting up any backend for the swift service, we're just replacing it w/ RGW.
@katarimanojk I think it's good we can improve the wording. I can do that in the docs you mentioned, but +1 to move to a better wording.

roles/cifmw_cephadm/tasks/post.yml
```diff
@@ -17,6 +17,27 @@
...
+- name: Check if swift endpoint is already created
+  shell: "oc rsh openstackclient openstack endpoint list | grep 'swift.*object-store' | wc -l"
+  register: swift_endpoint
+  delegate_to: localhost
+
+- name: Configure ceph object store to use external ceph object gateway
+  cifmw.general.ci_script:
+    output_dir: "/home/zuul/ci-framework-data/artifacts"
+    script: |-
+      oc rsh  openstackclient openstack service create --name swift --description 'OpenStack Object Storage' object-store
```
fmount, line 29: Nit, one more space after `oc rsh `

fmount, line 29: The last failure is weird:

```
2024-02-12 07:02:49.908197 | controller | TASK [cifmw_cephadm : Configure ceph object store to use external ceph object gateway output_dir=/home/zuul/ci-framework-data/artifacts, script=oc rsh  openstackclient openstack service create --name swift --description 'OpenStack Object Storage' object-store
2024-02-12 07:02:49.908218 | controller | oc rsh  openstackclient openstack user create --project service --password {{ cifmw_ceph_rgw_keystone_psw }}  swift
2024-02-12 07:02:49.908223 | controller | oc rsh  openstackclient openstack role create swiftoperator
2024-02-12 07:02:49.908238 | controller | oc rsh  openstackclient openstack role create ResellerAdmin
2024-02-12 07:02:49.908242 | controller | oc rsh  openstackclient openstack role add --user swift --project service member
2024-02-12 07:02:49.908247 | controller | oc rsh  openstackclient openstack role add --user swift --project service admin
2024-02-12 07:02:49.908252 | controller | oc rsh  openstackclient openstack endpoint create --region regionOne object-store public http://{{ cifmw_cephadm_vip }}:8080/swift/v1/AUTH_%\(tenant_id\)s
2024-02-12 07:02:49.908262 | controller | oc rsh  openstackclient openstack endpoint create --region regionOne object-store internal http://{{ cifmw_cephadm_vip }}:8080/swift/v1/AUTH_%\(tenant_id\)s
2024-02-12 07:02:49.908267 | controller | oc rsh  openstackclient openstack role add --project admin --user admin swiftoperator] ***
2024-02-12 07:02:49.908271 | controller | Monday 12 February 2024  07:02:49 -0500 (0:00:00.331)       0:33:53.693 *******
2024-02-12 07:02:50.123385 | controller | An exception occurred during task execution. To see the full traceback, use -vvv. The error was: NoneType: None
2024-02-12 07:02:50.123426 | controller | fatal: [compute-0 -> localhost]: FAILED! => {"changed": true, "msg": "non-zero return code", "rc": 1, "stderr": "", "stderr_lines": [], "stdout": "error: Missing or incomplete configuration info.  Please point to an existing, complete config file:\n\n\n  1. Via the command-line flag --kubeconfig\n  2. Via the KUBECONFIG environment variable\n  3. In your home directory as ~/.kube/config\n\nTo view or setup config directly use the 'config' command.\n", "stdout_lines": ["error: Missing or incomplete configuration info.  Please point to an existing, complete config file:", "", "", "  1. Via the command-line flag --kubeconfig", "  2. Via the KUBECONFIG environment variable", "  3. In your home directory as ~/.kube/config", "", "To view or setup config directly use the 'config' command."]}
```

so it's like it delegates the tasks to a node where KUBECONFIG is not set.

roles/cifmw_cephadm/tasks/post.yml
```diff
@@ -17,6 +17,27 @@
 - name: Get ceph_cli
   ansible.builtin.include_tasks: ceph_cli.yml
 
+- name: Check if swift endpoint is already created
```
fmount, line 20: Put a note about the fact we do not support updates on the existing endpoints

fultonj:
I think it would be good for you to run a few `oc` commands to confirm that swift is NOT running before doing the configuration. If swift IS running, then make these ansible tasks fail before they start.

I.e. swift must be disabled as a pre-requisite so let the user know early if they don't meet that pre-req.

You should see this:
```
apiVersion: core.openstack.org/v1beta1
kind: OpenStackControlPlane
metadata:
  name: controlplane
  namespace: openstack
spec:
  ...
  swift:
    ...
    enabled: false
```
After you identify the name of the `OpenStackControlPlane` object.
```
[zuul@controller-0 hci]$ oc get OpenStackControlPlane
NAME           STATUS   MESSAGE
controlplane   False    OpenStackControlPlane RabbitMQ in progress
[zuul@controller-0 hci]$ oc get OpenStackControlPlane controlplane -o yaml | grep -i swift -A 1
```

roles/cifmw_cephadm/tasks/post.yml
```diff
@@ -70,7 +78,9 @@
       oc rsh openstackclient openstack endpoint create --region regionOne object-store internal http://{{ cifmw_cephadm_vip }}:8080/swift/v1/AUTH_%\(tenant_id\)s
       oc rsh openstackclient openstack role add --project admin --user admin swiftoperator
   delegate_to: localhost
-  when: swift_endpoints_count.stdout == "0"
+  when:
+    - not swift_in_ctlplane.stdout | bool
+    - swift_endpoints_count.stdout == "0"
```
fultonj, line 83: I think you're on the right track. Maybe we want something like this:
```
fail:
  msg: "WARNING: Swift is already deployed so you cannot deploy RGW"
when:
- cifmw_ceph_daemons_layout.rgw_enabled | bool
- swift_in_ctlplane.stdout | bool
- swift_endpoints_count.stdout > 0
```
or maybe change from `failure` to `debug`

fultonj: and move that warning earlier in the ceph.yml playbook so we fail early before doing the deployment.

katarimanojk: In the latest patch i moved the warning to pre.yaml but i updated the msg as
 "WARNING: Swift is deployed and the endpoint exists already, ceph RGW cannot be configured as object store"

because we always deploy rgw here [1]
iiuc,  in this PR we are just configuring the deployed rgw as the object store at the end.

[1] https://github.com/openstack-k8s-operators/ci-framework/blob/main/playbooks/ceph.yml#L268

roles/cifmw_cephadm/tasks/post.yml
```diff
@@ -17,6 +17,27 @@
...
+      oc rsh openstackclient openstack role create swiftoperator
+      oc rsh openstackclient openstack role create ResellerAdmin
+      oc rsh openstackclient openstack role add --user swift --project service member
+      oc rsh openstackclient openstack role add --user swift --project service admin
+      oc rsh openstackclient openstack endpoint create --region regionOne object-store public http://{{ cifmw_cephadm_vip }}:8080/swift/v1/AUTH_%\(tenant_id\)s
+      oc rsh openstackclient openstack endpoint create --region regionOne object-store internal http://{{ cifmw_cephadm_vip }}:8080/swift/v1/AUTH_%\(tenant_id\)s
+      oc rsh openstackclient openstack role add --project admin --user admin swiftoperator
+  delegate_to: localhost
+  when:
+    - cifmw_ceph_daemons_layout.rgw_enabled | bool
```
katarimanojk, line 37: @fultonj the zuul job podified-multinode-hci-deployment-crc fails to identify sruct 'cifmw_ceph_daemons_layout' which is supposed to be passed via ceph_overrides.

fmount: and you probably need a `| default('true')` to the boolean in case no overrides are passed

katarimanojk: @fmount ack will do that.

roles/cifmw_cephadm/tasks/post.yml
viroel: hi @katarimanojk - Just wondering if we should add this openstack configuration inside cifmw_cephadm role. Is there anything that don't allow us to do this configuration outside of the role?

fultonj:
The molecule test seems to have failed because when moelecule is run `cifmw_openshift_kubeconfig` is not defined.

- https://review.rdoproject.org/zuul/build/62dfb533cf50444e994f8e8b8c1ce8d0

You could modify the code that uses that variable to only do so in a block when `cifmw_openshift_kubeconfig` is defined. The cifmw_cephadm role needs to be able to work when k8s is not available.

roles/cifmw_cephadm/tasks/post.yml
```diff
@@ -17,6 +17,43 @@
...
+      oc rsh openstackclient openstack role add --user swift --project service member
+      oc rsh openstackclient openstack role add --user swift --project service admin
+      oc rsh openstackclient openstack endpoint create --region regionOne object-store public http://{{ cifmw_cephadm_vip }}:8080/swift/v1/AUTH_%\(tenant_id\)s
+      oc rsh openstackclient openstack endpoint create --region regionOne object-store internal http://{{ cifmw_cephadm_vip }}:8080/swift/v1/AUTH_%\(tenant_id\)s
+      oc rsh openstackclient openstack role add --project admin --user admin swiftoperator
+  delegate_to: localhost
+  when:
+    - cifmw_ceph_daemons_layout.rgw_enabled | default(true) | bool
+    - not swift_in_ctlplane.stdout | bool
+    - swift_endpoints_count.stdout == "0"
```
fultonj, line 39: Add `cifmw_openshift_kubeconfig is defined` to the when (that should fix molecule too).

roles/cifmw_cephadm/tasks/pre.yml
```diff
@@ -158,3 +158,49 @@
     mode: "0644"
     config_overrides: "{{ cifmw_cephadm_conf_overrides }}"
     config_type: ini
+
+- name: Check the list of services
```
fultonj, line 162: Can we please put lines 162 to 206 in a separate file like `k8s_openstack_object_config.yml` and then in pre.yml include that tasks file? The inclusion of that tasks file should be wrapped in a `when: cifmw_openshift_kubeconfig is defined`.

roles/cifmw_cephadm/tasks/post.yml
```diff
@@ -17,6 +17,43 @@
...
+  shell: |
+    oc rsh openstackclient openstack service list
+    oc rsh openstackclient openstack user list
+    oc rsh openstackclient openstack role list
+    oc rsh openstackclient openstack endpoint list
+    oc get pods
+  register: service_list
+  delegate_to: localhost
+
+- name: print service list
```
fultonj, line 53: I assume this debug info will be removed and that it's only there as we try to get green CI.

roles/cifmw_cephadm/tasks/pre.yml
```diff
@@ -158,3 +158,49 @@
...
+  shell: |
+    oc rsh openstackclient openstack service list
+    oc rsh openstackclient openstack user list
+    oc rsh openstackclient openstack role list
+    oc rsh openstackclient openstack endpoint list
+    oc get pods
+  register: service_list
+  delegate_to: localhost
+
+- name: print service list
```
fultonj, line 174: I assume this debug info will be removed and that it's only there as we try to get green CI.

roles/cifmw_cephadm/tasks/pre.yml
```diff
@@ -158,3 +158,49 @@
...
+  debug:
+    msg: "{{ service_list }}"
+
+- name: Check if swift endpoint is already created
```
fultonj, line 178: We probably don't need this task since checking if swift is enabled or disabled should be sufficient.

roles/cifmw_cephadm/tasks/pre.yml
```diff
@@ -158,3 +158,49 @@
...
+  delegate_to: localhost
+
+- name: Check if swift is enabled in deployed controlplane
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+  shell: "oc get $(oc get oscp -n openstack -o name) -o json| jq .spec.swift.enabled"
+  register: swift_in_ctlplane
+  delegate_to: localhost
+
+- name: print swift_in_ctlplane
```
fultonj, line 192: I assume lines 192 to 195 will be removed after we get green CI.

roles/cifmw_cephadm/tasks/post.yml
fultonj: @viroel has a valid point. `cifmw_cephadm` is already a big role and it's about configuring ceph not openstack, but these changes configure openstack.

You could introduce a separate role like `cifmw_obj_store_keystone` (since this ultimately results in changes to keystone so that the object store is registered). You would then call this role from the ceph.yml playbook after `Render Ceph client configuration`. When the new role is called, you would have to pass variables which were defined by `cifmw_cephadm` like `{{ cifmw_ceph_rgw_keystone_psw }}` and `{{ cifmw_cephadm_vip }}`.

https://github.com/openstack-k8s-operators/ci-framework/blob/main/playbooks/ceph.yml#L316

Because we need `cifmw_cephadm` to not depend on k8s, it's important that we only call these tasks if a k8s env (`cifmw_openshift_kubeconfig`) is present. I would require that condition. I.e. the `cifmw_obj_store_keystone`

The role itself can be simple. main.yml can basically include the items being added to pre.yml and post.yml.

fmount:
@katarimanojk can we rebase? I suspect this patch is ready and it's required by VA1 to get RGW in place and configured as object-storage

roles/cifmw_cephadm/tasks/post.yml
fmount: I'm not sure it's good to have yet another role for a couple of tasks but we can definitely talk more about that if it's the desired direction.
In general I think that if `Swift` is not detected in the ctlplane (e.g. VA1, but in `architectures` is disabled by default), we might need to be able to configure a Ceph service (`RGW`) in keystone, but I think we should be responsible to do such configuration here, because we have all the variables in a single place, and we don't risk to complicate more the flow.
I would propose a follow up if we think about a refactoring of this patch, but personally I like to have a `post.yaml` when we perform some post-ceph operations.

fultonj: OK, we don't need to block on it being a separate role. Let's at least put it in a separate tasks files and call it from post.yml though.

katarimanojk: Sure thing, these tasks will not be called if rgw is disabled as per this condition
- cifmw_ceph_daemons_layout.rgw_enabled | default(true) | bool

roles/cifmw_cephadm/tasks/configure_object.yml
```diff
@@ -0,0 +1,60 @@
...
+- name: Check for exisiting object store config
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+  delegate_to: localhost
+  when: cifmw_openshift_kubeconfig is defined
+  block:
+    - name: Check if swift is enabled in deployed controlplane
+      ansible.builtin.shell: "set -o pipefail && oc get $(oc get oscp -n openstack -o name) -o json| jq .spec.swift.enabled"
+      register: swift_in_ctlplane
+    # checking swift_endpoints_count will avoid unnecessary errors during ceph deployment re-run
```
fmount, line 26: nit: comment indentation not aligned

fmount (approved):
overall I think it's good enough to approve and we can move forward. thanks @katarimanojk

roles/cifmw_cephadm/tasks/configure_object.yml
```diff
@@ -0,0 +1,60 @@
...
+    - name: Check if swift endpoint is already created
+      ansible.builtin.shell: "set -o pipefail && oc rsh openstackclient openstack endpoint list | grep 'swift.*object-store' | wc -l"
+      register: swift_endpoints_count
+      ignore_errors: true
```
fmount, line 30: This line was my only concern, but I can't think about a wrong case because I think, worst case scenario, the `Configure object store to use rgw` is skipped

katarimanojk: if the swift endpoints are not configured, exit status of the command is non-zero so it may be considered as a failure.

fmount (approved):
/lgtm

fmount:
@fultonj if you are ok with the current implementation I would /approve this patch (and discuss about any follow up activity)

fmount:
@cjeanner  || @pablintino wondering if you can approve this patch (that is ready)

fultonj (commented):
+2

fultonj:
@katarimanojk this cannot merge as is because it depends on a patch in draft https://github.com/openstack-k8s-operators/openstack-operator/pull/691 

Please update the depends on structure. Do we need a CI job update so that swift is not deployed where this is used?

PR #1126, opened 2024-02-09
Improve logging for nested ansible-playbook calls

cjeanner:
The main ansible.log is usually unreadable once we reach a nested
ansible-playbook call, such as hook, or in the reproducer.

This patch should improve the situation by:
- ensuring nested ansible-playbook calls don't push logs in the main
  ansible.log file
- ensuring nested ansible-playbook calls have their own log file by
  exporting an environment variable

This should greatly improve the situation regarding logging in the
project.


cjeanner:
recheck

cjeanner:
recheck

cescgina (commented):
/lgtm

PR #1139, opened 2024-02-13
Add wait for openstack operator

lewisdenny:
\- Add wait on openstack operator status being ready

Problem:
We're seeing issues applying the OpenStackControlPlane CR because the
openstack operator webhook is not up yet. Waiting for the ready condition
will ensure the operator is ready for requests.

[1]
```
2024-02-12 18:46:38.263494 TASK [edpm_prepare : Apply the OpenStackControlPlane CR
`Internal error occurred: failed calling webhook \"mopenstackcontrolplane.kb.io\": failed to call webhook`
```

[2]
```
2024-02-12T23:46:45.169Z	INFO	controller-runtime.webhook	Serving webhook server	{"host": "", "port": 9443}
```

[1] https://logserver.rdoproject.org/22/722/e2dd0f26cbf33fba8cf4aeff668ade3575a53de0/github-check/install-yamls-crc-podified-edpm-baremetal/1b424aa/job-output.txt
[2] https://logserver.rdoproject.org/22/722/e2dd0f26cbf33fba8cf4aeff668ade3575a53de0/github-check/install-yamls-crc-podified-edpm-baremetal/1b424aa/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack-operators/pods/openstack-operator-controller-manager-699f5c6f99-c9cgh/logs/manager.log

### DIFF
--- a/roles/edpm_prepare/tasks/main.yml
+++ b/roles/edpm_prepare/tasks/main.yml
@@ -193,6 +193,22 @@
       ansible.builtin.debug:
         var: cifmw_edpm_prepare_crs_kustomize_result
 
+    - name: Wait for openstack-operator-controller-manager pod to be ready
+      kubernetes.core.k8s_info:
+        kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
+        api_key: "{{ cifmw_openshift_token | default(omit)}}"
+        context: "{{ cifmw_openshift_context | default(omit) }}"
+        namespace: openstack-operators
+        kind: Pod
+        wait: true
+        wait_sleep: 10
+        wait_timeout: 600
+        wait_condition:
+          type: Ready
+          status: "True"
+        label_selectors:
+          - "openstack.org/operator-name = openstack"
+
     - name: Apply the OpenStackControlPlane CR
       environment:
         KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"


marios (commented):
seems reasonable 

waiting for green test run before approving

roles/edpm_prepare/tasks/main.yml
```diff
@@ -193,6 +193,22 @@
       ansible.builtin.debug:
         var: cifmw_edpm_prepare_crs_kustomize_result
 
+    - name: Wait for openstack-operator-controller-manager pod to be ready
```
karelyatin, line 196: can't we just used make_openstack_wait instead at line 77? or that check too have some issue?

cjeanner:
/lgtm

Good to rely a bit less on install_yamls - though we're far from being freed of it.

roles/edpm_prepare/tasks/main.yml
lewisdenny: The wait condition you mention is useless for telling us if the webhook is alive. the webhook comes up at `23:46:38`  and install plan phase condition = Complete at `23:44:57`:
```
  conditions:
  - lastTransitionTime: "2024-02-12T23:44:57Z"
    lastUpdateTime: "2024-02-12T23:44:57Z"
    status: "True"
    type: Installed
  phase: Complete
  ```
  
  Let me check the operator and see if I can use that rather than watching the pod :)

Comment (lewisdenny) [reply roles/edpm_prepare/tasks/main.yml:196]: I can't check this instance of the error because it seems we don't collect the `oc get operators openstack-operator.openstack-operators` output, the reason I checked the pod was because I felt it was the most safe and accurate.

`oc get operators openstack-operator.openstack-operators` only has one condition to check and it's:
```
      conditions:
      - lastTransitionTime: "2024-02-13T07:22:15Z"
        message: all available catalogsources are healthy
        reason: AllCatalogSourcesHealthy
        status: "False"
        type: CatalogSourcesUnhealthy
      kind: Subscription
      name: openstack-operator
      namespace: openstack-operators
```

Comment (abays) [reply roles/edpm_prepare/tasks/main.yml:196]: Waiting for the Openstack operator controller-manager pod to be "available" is the best way to know if the webhook is ready, since we changed the controller-manager pod's "available" state to be determined by the webhooks availability [1].

[1] https://github.com/openstack-k8s-operators/openstack-operator/pull/434

Comment (lewisdenny) [reply roles/edpm_prepare/tasks/main.yml:196]: Sorry @abays for the stupid question but I don't see an available type condision for the Openstack operator controller-manager pod? How do we check that? 

```
  status:
    conditions:
    - lastProbeTime: null
      lastTransitionTime: "2024-02-13T07:23:05Z"
      status: "True"
      type: Initialized
    - lastProbeTime: null
      lastTransitionTime: "2024-02-13T07:23:25Z"
      status: "True"
      type: Ready
    - lastProbeTime: null
      lastTransitionTime: "2024-02-13T07:23:25Z"
      status: "True"
      type: ContainersReady
    - lastProbeTime: null
      lastTransitionTime: "2024-02-13T07:23:05Z"
      status: "True"
      type: PodScheduled
```

Comment (abays) [reply roles/edpm_prepare/tasks/main.yml:196]: No worries.  I actually slightly misspoke.  I meant to say wait for the `Deployment` for the OpenStack operator controller-manager to be ready.  It could be done like so:

```
oc wait deployments/openstack-operator-controller-manager -n openstack-operators --for condition=Available
```

The `Deployment` shouldn't reach the `Available` state until the `Pod` probes respond affirmatively, and the probes shouldn't respond properly until the webhooks are available due to [1] that I mentioned in the last comment.

[1] https://github.com/openstack-k8s-operators/openstack-operator/pull/434

Comment (abays) [reply roles/edpm_prepare/tasks/main.yml:196]: You can also actually wait for the `Pod` itself as another option though, so I guess I should share that too:

```
oc wait pod -n openstack-operators -l openstack.org/operator-name=openstack --for condition=Ready
```

lewisdenny: Ah right, thank you for the clarification, the second option is what I'm doing in this patch. As both seem to be functionally the same could you folks please let me know if there is a preference in checking the deployment VS pod and I will update accordingly. :)

lewisdenny:
Test patch: https://github.com/openstack-k8s-operators/install_yamls/pull/722

PR #1145, opened 2024-02-14
[libvirt_manager] Fix undefined variable cifmw_reproducer_basedir

psathyan:
This change removes the reference of `cifmw_reproducer_basedir` in `libvirt_manager` role. Referencing the tasks from `start_manage_vms` throws `undefined variable name`.

### DIFF
--- a/roles/libvirt_manager/tasks/start_manage_vms.yml
+++ b/roles/libvirt_manager/tasks/start_manage_vms.yml
@@ -67,7 +67,7 @@
     extracted_ip: "{{ (vm_ip.stdout.split())[3] | ansible.utils.ipaddr('address') }}"
     identity_file: >-
       {{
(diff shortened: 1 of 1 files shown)


roles/libvirt_manager/tasks/start_manage_vms.yml
```diff
@@ -67,7 +67,7 @@
     extracted_ip: "{{ (vm_ip.stdout.split())[3] | ansible.utils.ipaddr('address') }}"
     identity_file: >-
       {{
-        cifmw_reproducer_basedir ~ '/artifacts/cifmw_ocp_access_key' if vm_type == 'ocp' else
+        cifmw_libvirt_manager_basedir ~ '/artifacts/cifmw_ocp_access_key' if vm_type == 'ocp' else
```
cjeanner, line 70: right - it was called from the reproducer until now, so it wasn't failing. Values should be the same at this point.

PR #1146, opened 2024-02-14
Update content provider job to use Vexxhost label

lewisdenny:
Currently the content provider node can get scheduled onto IBM cloud
and get a private IP, only jobs also scheduled on IBM cloud can access
and pull containers. You can see it here how only one job passed:
https://review.rdoproject.org/zuul/buildset/682107b0c1b7406383063a285e4a2cb1

The job that passed and the content-provider are on IBM cloud, the rest
are running in Vexxhost and failing to pull containers from the
content-provider.

- Add new nodeset `centos-stream-9-vexxhost` that only targets Vexxhost
cloud
- Update `openstack-k8s-operators-content-provider` to use new nodeset

### DIFF
--- a/zuul.d/nodeset.yaml
+++ b/zuul.d/nodeset.yaml
@@ -42,6 +42,18 @@
         nodes: []
 
 - nodeset:
+    name: centos-stream-9-vexxhost
+    nodes:
+      - name: controller
+        label: cloud-centos-9-stream-tripleo-vexxhost
+    groups:
+      - name: switch
+        nodes:
+          - controller
+      - name: peers
+        nodes: []
+
+- nodeset:
     name: multinode-centos-9-stream-crc
     nodes:
       - name: controller
--- a/zuul.d/base.yaml
+++ b/zuul.d/base.yaml
@@ -23,6 +23,7 @@
 - job:
     name: openstack-k8s-operators-content-provider
     parent: cifmw-base-minimal
+    nodeset: centos-stream-9-vexxhost
     irrelevant-files: &ir_files
       - .*/*.md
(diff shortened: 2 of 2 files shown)


zuul.d/nodeset.yaml
```diff
@@ -41,6 +41,18 @@
...
+    nodes:
+      - name: controller
+        label: cloud-centos-9-stream-tripleo-vexxhost
+    groups:
+      - name: switch
+        nodes:
+          - controller
+      - name: peers
+        nodes: []
+
```
marios, line 55: so the content provider will always run in vexx
but what if child jobs get scheduled on ibm cloud don't we have the same problem?

or can ibm child jobs reach the vexx content provider?

pablintino: vexx IPs are public

marios: yeah indeed... i guess I was wondering if there was administrative issue with reaching vexx from ibm

lewisdenny: There shouldn't be an issue reaching Vexxhost from IBM cloud, not sure how to find a job with that scenario but as Vexxhost is a public IP it works.

marios (approved):
as discussed yesterday, we would ideally schedule all jobs in a given buildset on the same provider. For now we definitely need this, so that content providers never run in ibm cloud. Any further tweaking to the scheduling will have to be a followup (and likely require help from the sf/rhos-ops folks with the nodesets and scheduling)

rlandy (commented):
Against my better judgement, agreeing to merge this.
As thing stand right now, we have no other option

PR #1148, opened 2024-02-14
Make dnsmasq listen to localhost in the reproducer

cescgina:
Change the listen-address parameter in the dnsmasq config in the crc
node so that it listens in localhost. This was causing reproducer
deployments with crc 2.32 to fail as the crc node had no working dns
server.


psathyan (approved):
/lgtm

PR #1149, opened 2024-02-15
Ensure second interface doesn't configure second default route.

lewisdenny:
Generally nodes should only have a single default route, when adding
the second interface to the nodes a second route is being added that
can cause routing issues.

This patch adds the `never_default4: true` parameter to ensure only
one default route is configured.

### DIFF
--- a/roles/reproducer/tasks/configure_computes.yml
+++ b/roles/reproducer/tasks/configure_computes.yml
@@ -11,7 +11,7 @@
         ifname: "{{ _iface }}"
         type: ethernet
         ip4: "{{ _ip4 }}/{{ _prefix }}"
-        gw4: "{{ _gw4 }}"
+        never_default4: true
         state: present
 
     - name: Ensure we can ping controller-0 from ctlplane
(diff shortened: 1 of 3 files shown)


lewisdenny (commented):
https://docs.ansible.com/ansible/latest/collections/community/general/nmcli_module.html#parameter-never_default4
`This parameter is mutually_exclusive with gw4 parameter.` :(

lewisdenny:
```
[zuul@controller-0 ~]$ cat reproducer-inventory/* > inventory.yaml

[zuul@controller-0 ~]$ ansible -i inventory.yaml all -m shell -a 'ip r |grep default && ip r |grep "192.168.122"'
controller-0 | CHANGED | rc=0 >>
default via 192.168.100.1 dev eth0 proto dhcp src 192.168.100.21 metric 100
192.168.122.0/24 dev eth1 proto kernel scope link src 192.168.122.9 metric 101

compute-0 | CHANGED | rc=0 >>
default via 192.168.100.1 dev eth0 proto dhcp src 192.168.100.20 metric 101
192.168.122.0/24 dev eth1 proto kernel scope link src 192.168.122.100 metric 102

crc-0 | CHANGED | rc=0 >>
default via 192.168.100.1 dev enp5s0 proto dhcp src 192.168.100.27 metric 100
192.168.122.0/24 dev enp6s0 proto kernel scope link src 192.168.122.10 metric 101
```

psathyan (approved):
/lgtm

PR #1150, opened 2024-02-15
Allow loading ceph playbook network ranges from Net Env Def

pablintino:
To align the ceph.yml playbook with the new way of configuring networking in the framework, but keeping its standalone functionality, we now try to load `storage_network_range` and `storage_mgmt_network_range` from the Networking Environment Definition file. If the file is not present, like using it without the rest of framework, we still rely on the existing defaults.

  - [x] README in the role


rebtoor (approved):

fmount (approved):
/lgtm

PR #1152, opened 2024-02-15
Allow zuul reproducer to pull multiple depends-on

cescgina:
Allow the zuul reproducer to work with a job that contains more than one
depends-on to a same repo. Until now, the reproducer will fail because
the second invocation of the git ansible module finds an already checked-out
branch when trying to pull the second depends-on.

### DIFF
--- a/roles/reproducer/tasks/push_code.yml
+++ b/roles/reproducer/tasks/push_code.yml
@@ -27,18 +27,35 @@
       loop: "{{ zuul['items'] | zip(repos_dir_stats.results) | list }}"
 
     - name: Fetch zuul.items repositories
-      vars:
-        _skip_refspec: "{{ job_id is not defined or repo.change is not defined }}"
-      ansible.builtin.git:
-        dest: "{{ repo.project.src_dir }}"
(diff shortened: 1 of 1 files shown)


roles/reproducer/tasks/push_code.yml
```diff
@@ -27,18 +27,35 @@
...
+        - name: Rebase on top of existing change if we tried to pull on the same repo twice
+          vars:
+            refspec: "{{ item['invocation']['module_args']['refspec'] }}"
+            dest_dir: "{{ item['invocation']['module_args']['dest'] }}"
+          when:
+            - item.failed
+            - '"Failed to download remote objects and refs:" in item.msg'
+            - '"fatal: refusing to fetch into branch" in item.msg'
+            - '"checked out at" in item.msg'
+          ansible.builtin.command: # noqa: command-instead-of-module
```
cjeanner, line 54: fun, usually ansible-lint is complaining when there's only one space before the `#` for comments o_O

marios (approved):
lgtm

PR #1156, opened 2024-02-15
Allow to install custom CA on reproducer env

cjeanner:
This may be handy when you're using some squid proxy configured to bump
TLS connection - or if you want to access some self-signed content
(repositories or anything).

### DIFF
--- a/roles/reproducer/tasks/configure_computes.yml
+++ b/roles/reproducer/tasks/configure_computes.yml
@@ -19,6 +19,24 @@
         cmd: |
           ping -c2 {{ _controller_ip4 }}
 
+    - name: Tweak dnf configuration
+      become: true
+      community.general.ini_file:
+        no_extra_spaces: true
+        option: "{{ config.option }}"
+        path: "/etc/dnf/dnf.conf"
+        section: "{{ config.section | default('main') }}"
+        state: "{{ config.state | default(omit) }}"
+        value: "{{ config.value | default(omit) }}"
+      loop: "{{ cifmw_reproducer_dnf_tweaks }}"
+      loop_control:
+        label: "{{ config.option }}"
+        loop_var: 'config'
+
+    - name: Install custom CA if needed
+      ansible.builtin.import_role:
+        name: install_ca
+
     - name: RHEL related tasks on computes
       become: true
       when:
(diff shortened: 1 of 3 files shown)


roles/reproducer/tasks/configure_computes.yml
```diff
@@ -19,6 +19,24 @@
...
+        state: "{{ config.state | default(omit) }}"
+        value: "{{ config.value | default(omit) }}"
+      loop: "{{ cifmw_reproducer_dnf_tweaks }}"
+      loop_control:
+        label: "{{ config.option }}"
+        loop_var: 'config'
+
+    - name: Install custom CA if needed
+      ansible.builtin.import_role:
+        name: install_ca
```
marios, line 38: so you'd have to set the relevant parameters for the install_ca role to take advantage of this
do we need to note this in the readme or we already have that

otherwise i think the role does nothing? https://github.com/marios/ci-framework/blob/6b71222f3a7384a41c7264606cca905ecd922541/roles/install_ca/defaults/main.yml#L21C1-L22C32 (default '')

cescgina (commented):
/lgtm

marios (commented):
sounds right

roles/reproducer/tasks/configure_computes.yml
cjeanner: exactly, yes. I configured a squid proxy locally to intercept TLS connections and cache things such as packages, soooo... I kind of needed a possibility to inject custom CA.

Note that this mechanism may, short-term, replace the way we install RH internal CA, avoiding the RPM and all of its heavy dependencies.

roles/reproducer/tasks/configure_computes.yml
```diff
@@ -19,6 +19,24 @@
         cmd: |
           ping -c2 {{ _controller_ip4 }}
 
+    - name: Tweak dnf configuration
```
pablintino, line 22: A couple of comments, nothing really blocker:

- The title of the block is not describing what's done with this change.
- The dnf tweak may not be "directly" related to the purpose of the change.

cjeanner: It's part of it in fact - all related to "consume proxy". We're really applying DNF configuration, i.e. tweaking it - the CA is injected after it. Not really sure it deserves its own patch for the tweaking though...

PR #1159, opened 2024-02-16
Set srbac by default

kopecmartin:
Let's set srbac (enforce_scope) by default. In jobs where it does
not make sense it should be disabled explicitly.

  - [x] README in the role

### DIFF
--- a/roles/test_operator/defaults/main.yml
+++ b/roles/test_operator/defaults/main.yml
@@ -45,6 +45,32 @@ cifmw_test_operator_tempest_network_attachments: []
 cifmw_test_operator_tempest_tests_include_override_scenario: false
 cifmw_test_operator_tempest_tests_exclude_override_scenario: false
 cifmw_test_operator_tempest_workflow: []
+
+# Enabling SRBAC by default, in jobs where this does not make sense should be turned off explicitly
+cifmw_tempest_tempestconf_config_defaults:
+  deployerInput: |
+    [enforce_scope]
+    nova = true
+    neutron = true
+    glance = true
+    cinder = true
+    keystone = true
+    placement = true
+    manila = true
+
+    [barbican_rbac_scope_verification]
+    enforce_scope = true
+
+    [identity-feature-enabled]
+    enforce_scope = true
+
+    [load_balancer]
+    member_role = load-balancer_member
+    admin_role = load-balancer_admin
+    RBAC_test_type = keystone_default_roles
+    enforce_new_defaults = true
+    enforce_scope = false
+
 # Please refer to https://openstack-k8s-operators.github.io/test-operator/guide.html#executing-tempest-tests
 cifmw_test_operator_tempest_config:
   apiVersion: test.openstack.org/v1beta1
@@ -68,7 +94,7 @@ cifmw_test_operator_tempest_config:
       concurrency: "{{ cifmw_test_operator_concurrency }}"
       externalPlugin: "{{ cifmw_test_operator_tempest_external_plugin | default([]) }}"
       extraImages: "{{ cifmw_test_operator_tempest_extra_images | default([]) }}"
-    tempestconfRun: "{{ cifmw_tempest_tempestconf_config | default(omit) }}"
+    tempestconfRun: "{{ cifmw_tempest_tempestconf_config_defaults | combine(cifmw_tempest_tempestconf_config | default({})) }}"
     workflow: "{{ cifmw_test_operator_tempest_workflow }}"
 
 # Section 3: tobiko parameters - used when run_test_fw is 'tobiko'


roles/test_operator/defaults/main.yml
```diff
@@ -31,6 +31,16 @@ cifmw_test_operator_tempest_container: openstack-tempest
 cifmw_test_operator_tempest_image: "{{ cifmw_test_operator_tempest_registry }}/{{ cifmw_test_operator_tempest_namespace }}/{{ cifmw_test_operator_tempest_container }}"
 cifmw_test_operator_tempest_image_tag: current-podified
 
+# Enabling SRBAC by default, in jobs where this does not make sense should be turned off explicitly
+cifmw_tempest_tempestconf_config_defaults:
```
tosky, line 35: maybe a new variable, set to true by default, which enables these settings? Or too much?

lpiwowar: It maybe depends on how "difficult" we want to make the process of turning this off to be and how much we want to force this? Setting one variable to `false` is lot easier then turning SRBAC for every service individually.

roles/test_operator/defaults/main.yml
```diff
@@ -49,7 +59,7 @@ cifmw_test_operator_tempest_config:
         {{ cifmw_test_operator_tempest_exclude_list | default('') }}
       concurrency: "{{ cifmw_test_operator_concurrency }}"
       externalPlugin: "{{ cifmw_test_operator_tempest_external_plugin | default([]) }}"
-    tempestconfRun: "{{ cifmw_tempest_tempestconf_config | default(omit) }}"
+    tempestconfRun: "{{ cifmw_tempest_tempestconf_config_defaults | combine(cifmw_tempest_tempestconf_config) }}"
```
lpiwowar, line 62: I've just tested this quickly locally. This will fail because of `cifmw_tempest_tempestconf_config` is not defined. 

```
...
'cifmw_tempest_tempestconf_config' is undefined\n\nThe error appears to be in '/home/cloud-user/ci-framework/roles/test_operator/tasks/main.yml': line 117, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Start Tempest tests\n 
```

roles/test_operator/defaults/main.yml
kopecmartin: @tosky i was thinking about that but what about a case when someone wants to enable srbac only for a specific service and not all of them?

roles/test_operator/defaults/main.yml
lpiwowar: The `cifmw_tempest_tempestconf_config_defaults` variable should probably look like this:
```
cifmw_tempest_tempestconf_config_defaults:
  overrides: |
    enforce_scope.nova true
    enforce_scope.neutron true
    enforce_scope.glance true
    enforce_scope.cinder true
    enforce_scope.keystone true
    enforce_scope.placement true
```

lpiwowar: @kopecmartin as you pointed out this seems to be working together with the change of `cifmw_tempest_tempestconf_config_defaults` (above ^^):

```
tempestconfRun: "{{ cifmw_tempest_tempestconf_config_defaults | combine(cifmw_tempest_tempestconf_config | default({})) }}"
```

lpiwowar: @kopecmartin as I'm thinking about this. The `combine` will does not work then, right? Because `override` section is a string.

kopecmartin: that's a problem, yes, does it have to be a string?

kopecmartin: .. as overrides are a string, we need to set these defaults in another way and leave 'overrides' for jobs to set. 
We have 2 options then .. try to use 'deployerInput' or set srbac by default in python-tempestconf. I'd like to avoid the latter.

kopecmartin:
the options were set in tempest.conf as expected - https://logserver.rdoproject.org/46/46/605e4aabbf174c084d65dbd8f24464167cd7df8a/github-check/podified-multinode-edpm-deployment-crc/82f7d45/controller/ci-framework-data/tests/test_operator/etc/tempest.conf via this testing PR - https://github.com/openstack-k8s-operators/test-operator/pull/46 .. now we need to verify if those are all the options we want to set by default

roles/test_operator/defaults/main.yml
SeanMooney: secure rbac will be on by default for all operators and im not sure if we will supprot turning it off.

my understanding was we wanted to only support srbac

you could use a CustomServiceConfig to diable it in nova or placement but we do not intend to test that.

form an upstrema point of view your are not ment to mix and match services where you use the new policy in some and not other so this is an all or nothing choice for your cloud.

we are not plannig to provide an easy switch to diable srbac in the contol plane so if you wanted to do that as a cutomer you would have to overried the config of every service.
its technically possible but its intetionally not trivial.
we really dont want cutomers turning this off.

roles/test_operator/defaults/main.yml
```diff
@@ -40,6 +40,32 @@ cifmw_test_operator_tempest_image: "{{ cifmw_test_operator_tempest_registry }}/{
 cifmw_test_operator_tempest_image_tag: current-podified
 cifmw_test_operator_tempest_tests_include_override_scenario: false
 cifmw_test_operator_tempest_tests_exclude_override_scenario: false
+
+# Enabling SRBAC by default, in jobs where this does not make sense should be turned off explicitly
+cifmw_tempest_tempestconf_config_defaults:
+  deployerInput: |
+    [enforce_scope]
+    nova = true
```
SeanMooney, line 48: scope enforcemt is slightly different as almost all project removed scopes other then project scope as the desgin fo system scope fundementally broke many workflows.

so nova for example only uses project scope

scope are similar to using the new defaults in that really you should not mix an match enforcement in differnt services.

its technially possible to do but i dont think it is correct to do.

roles/test_operator/defaults/main.yml
kopecmartin: exactly, as srbac is on by default for all operators, this is meant to enable srbac in tempest by default = anyone who consumes this role will have srbac set by default in their tempest.conf

the motivation for this comes from this ticket - https://issues.redhat.com/browse/OSPRH-1515

roles/test_operator/defaults/main.yml
kopecmartin: hmm, i thought enforce scope means srbac, if not, than this change isn't needed as there is no other setting in tempest for srbac functionality

tosky: Isn't it about setting the enforce_scope.<service>=True enabled? 
I see that SRBAC tests relies on that, so shouldn't we enable it?
For example https://opendev.org/openstack/glance-tempest-plugin/src/branch/master/glance_tempest_plugin/tests/rbac/v2/base.py#L32

SeanMooney: no unless tempest ahve really badly named there tests.

enforce_new_defaults = true enables the new rbac policies that provide the reader member, admin  and service user personas


enforce_scope = true enabled enforcement of system vs project scopted apis.

as afar as I'm aware scopes are only used by keystone and ironic.

nova and placement deleted all scopes other then project scope

we should have both 
enforce_new_defaults and enforce_scope set in all operators at this point
but enforce_new_defaults is the primary part of srbac.

if you not familiar with why scopes were removed from most of the projects following
the zed ptg

https://governance.openstack.org/tc/goals/selected/consistent-and-secure-rbac.html#direction-change

SeanMooney: jsut to be clear all apis have scopes_enforcement and the new defaults enabled
only ironic an keyston use a scope other the project scope in there policy.
and we do not plan to add any additional usage fo scopes like system scope or domain scope.

the primary features of SRBAC are related to the enforce_new_defaults which enable the use of the admin, member, reader and service roles in the policies.

tosky: I've followed the story, but I think it may be, as you suggested, a case of bad naming for some tests, and I'd advise people to check that before just dismissing that option as not needed (for sure we wouldn't run relevant glance tests).

dmendiza: I think we do need `[oslo_policy] enforce_scope = True` in every service for Secure RBAC to be fully enforced.  An important part of the direction change in the RBAC spec is this guidance:

> Services with project resources that have already implemented scope (or have yet to) should make all policy rules set `scope_types=[‘project’]`

Keystone will be issuing tokens with `project`, `domain`, and `system` scopes, and the `enforce_scope` option in oslo.policy ensures that tokens with scopes that are not part of the `scope_type` list for a specific policy will get rejected before the policy is enforced.  For projects that have not added `scope_types=[...]` to their policies, the enforce_scope option is basically a no-op, but we probably still need to test to make sure that requests with scopes other than `'project'` do get rejected by those APIs with a 403.

Some policies might reject a request with the wrong scope e.g. if the policy checks the token's project_id against a target.project_id. However, other policies might not be able to reject a request and return errors instead.  e.g. a request to `POST /v1/some_entity` that assigns ownership based on the context project_id might error out with a domain-scoped token.

kopecmartin: ack, thanks, so do we agree that defaults look ok as they are now so that we can proceed with this?

SeanMooney: the assertion i was orignaly making is not that we should not be enabling scope enforcement it was that 99% for the srbac feature is implemented by enforce_new_defaults

so if you only enabled enforce_scope then the reader role would not work for example.
it would have fully write access

almost none of the project work with domain or system scoped tokens even without 
enforce_scope (only keystone and ironic support scopes other then project) but you should enable both options in the service it you want SRBAC to be enabled.

in the tempestconf I'm not sure if there is 1 or 2 options but
we have been defaultign to srbac in the operator for some time now so we are cornfiug the service to use the srbac functionally today.

lewisdenny:
Hi @kopecmartin, is this something we want in? Was there a conclusion after all? If not can we mark this draft until you want it reviewed?

lewisdenny:
Converting to draft for now

kopecmartin:
> Hi @kopecmartin, is this something we want in? Was there a conclusion after all? If not can we mark this draft until you want it reviewed?

no, no conclusion, it seems we won't proceed with a global setting like this, which is ok, as the required opts can be set in specific jobs where we want to set them ...

roles/test_operator/defaults/main.yml
tosky: Some plugins did the wrong thing, so that option, even if the name is wrong, enables the correct SRBAC tests. I think the confusion comes from that. The name is wrong but we need to set these options.

tosky:
Please reopen! We  need this.

kopecmartin:
Tested here: https://github.com/openstack-k8s-operators/test-operator/pull/98
Job run: https://review.rdoproject.org/zuul/build/c47bc93c5a474c5ea2b9b286922463e2
tempest.conf where we can see that the values got set there:
https://logserver.rdoproject.org/98/98/8f0fdf18be2d24225b50b5e655d11805d9dafba2/github-check/podified-multinode-edpm-deployment-crc-test-operator/c47bc93/controller/ci-framework-data/tests/test_operator/tempest-tests-single-test-workflow-step-1/etc/tempest.conf

lewisdenny:
/lgtm

@kopecmartin Would be good to get some other DFG members on the review list for this one IMO

tosky:
For the storage components, and from a quick glance from whatever I know of the other components, this looks fine, thanks (additional virtual lgtm :)

roles/test_operator/defaults/main.yml
```diff
@@ -43,6 +43,34 @@ cifmw_test_operator_tempest_network_attachments: []
...
 cifmw_test_operator_tempest_tests_exclude_override_scenario: false
 cifmw_test_operator_tempest_workflow: []
+
+# Enabling SRBAC by default, in jobs where this does not make sense should be turned off explicitly
+cifmw_tempest_tempestconf_config_defaults:
+  deployerInput: |
+    [enforce_scope]
+    # this doesn't work for nova team, let's enable srbac everywhere else
+    # and revisit nova later
+    # nova = true
```
SeanMooney, line 53: if this is form my feedback you have not understood what i said at all.

i explicitly said that enforce_scope is just a minor part of SRBAC and that the primary change was the enforce_new_defaults option.

so that twe need to test both.

it was then pointed out by others that this si just unfortunate nameing in tempeast and that enforce_scope is being used to feature gate both features.

i repeatliy tried to make it clear that i was not objecting to enabling these test, and highlighting that test only scope enforcement would not be sufficient to validate srbac as it only a minor part of it. since this apparently it testing both the new default and scope enforcement unlike what the config option name implies i don't see why we would not have this enabled for nova as well.

kopecmartin:
this patch has been around for more than 3 months and we only exchange messages here, it doesn't lead anywhere, let's set some rules to move it forward because I don't want to invest a minute more on this:
1. I want to hear what other options (if there are any) should be set, with those I proposed, together with their specific value
2. if some of the options I set are not correct and doesn't work for someone, we have 2 options, either a new value of that config option is proposed or we remove that option (from this default setting)

SeanMooney:
> this patch has been around for more than 3 months and we only exchange messages here, it doesn't lead anywhere, let's set some rules to move it forward because I don't want to invest a minute more on this:
> 
>     1. I want to hear what other options (if there are any) should be set, with those I proposed, together with their specific value
> 
>     2. if some of the options I set are not correct and doesn't work for someone, we have 2 options, either a new value of that config option is proposed or we remove that option (from this default setting)

well i think that more the problem then anything else 

if its only discussed here and not in any of the meetings and especially if its not associated with a jira issue (if it is it should be in the pr message) then it not going to be prioritized.

as the author of the patch i would have expected @kopecmartin to bring this up in the ci framework slack channel and ensure this is properly tracked under an epic targeted to ga to ensure its reflect as a priority to stakeholders.

in this case the primary reviews of this repo.

@abays @arxcruz could you bring this up in the relevant ci framework call and see if this can be expedited.

SRBAC is a ga requirement so i expect correct testing of SRBAC is also a ga requirement to this should be reviewed as a priority

kopecmartin:
> > this patch has been around for more than 3 months and we only exchange messages here, it doesn't lead anywhere, let's set some rules to move it forward because I don't want to invest a minute more on this:
> > ```
> > 1. I want to hear what other options (if there are any) should be set, with those I proposed, together with their specific value
> > 
> > 2. if some of the options I set are not correct and doesn't work for someone, we have 2 options, either a new value of that config option is proposed or we remove that option (from this default setting)
> > ```
> 
> well i think that more the problem then anything else
> 
> if its only discussed here and not in any of the meetings and especially if its not associated with a jira issue (if it is it should be in the pr message) then it not going to be prioritized.
> 
> as the author of the patch i would have expected @kopecmartin to bring this up in the ci framework slack channel and ensure this is properly tracked under an epic targeted to ga to ensure its reflect as a priority to stakeholders.
> 
> in this case the primary reviews of this repo.
> 
> @abays @arxcruz could you bring this up in the relevant ci framework call and see if this can be expedited.
> 
> SRBAC is a ga requirement so i expect correct testing of SRBAC is also a ga requirement to this should be reviewed as a priority

I ask for feedback only on the tempest options and instead I get a feedback on jira triaging .. no wonder this took 3 months

tosky:
> > @abays @arxcruz could you bring this up in the relevant ci framework call and see if this can be expedited.
> > SRBAC is a ga requirement so i expect correct testing of SRBAC is also a ga requirement to this should be reviewed as a priority
> 
> I ask for feedback only on the tempest options and instead I get a feedback on jira triaging .. no wonder this took 3 months

The point here is: which options should be changed, if you @SeanMooney realize that some should be changed?

That's why it was slow: changes were asked but without pointing out what exactly.

Now, as this patch improves the coverage for basically everyone the way it is, I suggest (this is for the ci-framework core reviewers) to merge it as it is. Additional options can be tuned and amended later.

roles/test_operator/defaults/main.yml
SeanMooney: > Some plugins did the wrong thing, so that option, even if the name is wrong, enables the correct SRBAC tests. I think the confusion comes from that. The name is wrong but we need to set these options.

so this is the sailiant point 

https://github.com/openstack/tempest/blob/master/tempest/config.py#L1213-L1219
the config option states that option also enables testing of enforce_new_defaults

which is the primary feature in the SRBAC community goal

so this should be set to true for nova too.

and an upstream tempest bug should be filed to fix the name of the config option
given the current one is very misleading.

SeanMooney:
> > > @abays @arxcruz could you bring this up in the relevant ci framework call and see if this can be expedited.
> > > SRBAC is a ga requirement so i expect correct testing of SRBAC is also a ga requirement to this should be reviewed as a priority
> > 
> > 
> > I ask for feedback only on the tempest options and instead I get a feedback on jira triaging .. no wonder this took 3 months
> 
> The point here is: which options should be changed, if you @SeanMooney realize that some should be changed?
> 
> That's why it was slow: changes were asked but without pointing out what exactly.
> 
> Now, as this patch improves the coverage for basically everyone the way it is, I suggest (this is for the ci-framework core reviewers) to merge it as it is. Additional options can be tuned and amended later.

no as i said before i never ask for this patch to not be merged, ever.
i asked questions to ensure that this is testing what we expect its testing.

you confirmed that it is in your last week https://github.com/openstack-k8s-operators/ci-framework/pull/1159#discussion_r1616858358 and https://github.com/openstack-k8s-operators/ci-framework/pull/1159#discussion_r1534114566

so I'm really not sure why nova is not being ensled in this patch and why it was changed.

i intentionally did not use the requested change feature in the review tab as i was never requesting a change to the patch i was asking everyone to confirm this is actually test the new defaults.


this is review is very frustrating as it feel like we keep talking past each other without people actually reading the replies.

roles/test_operator/defaults/main.yml
```diff
@@ -45,6 +45,32 @@ cifmw_test_operator_tempest_network_attachments: []
...
+    neutron = true
+    glance = true
+    cinder = true
+    keystone = true
+    placement = true
+    manila = true
+
+    [barbican_rbac_scope_verification]
+    enforce_scope = true
+
```
vakwetu, line 63: I think the barbican one has changed.  @dmendiza   should verify this, but I think we now use the same mechanism [enforce_scope] barbican = true  as everyone else.

vakwetu: And what about ironic, ironic_inspector, octavia, designate ?  Are the load_balancer options still correct? @johnsom ?

vakwetu (commented):
Thanks for sticking with this @kopecmartin .  

We definitely need to do this, so +2 to getting this in.  But some of the options may have been updated since you started your patch.  I have left some comments to that effect.

roles/test_operator/defaults/main.yml
vakwetu: That said, also , I'm ok with merging and amending later as needed.

vakwetu (approved):

PR #1160, opened 2024-02-16
Ensure target repository directories exists

pablintino:
The current implementation of the repository copy from the reproducer Ansible controller to the controller of the fresh generated environment assumes that the directory where repos are copied already exists, which is sadly false for non GitHub based repositories.
This commit creates the parent directories before rsync or git creates the target destination.


cescgina (approved):

cjeanner:
/approve

never faced any issue with that, but it won't hurt ensuring parent tree exists.

PR #1161, opened 2024-02-16
Allow installing CA from a url in install_ca

cescgina:
Allow downloading a certificate file directly into
/etc/pki/ca-trust/source/anchors/ in install_ca role, whitout it
being part of an rpm. This change also introduces new tasks to add CAs
to an openshift image registry in the openshift_setup role.

  - [x] README in the role

### DIFF
--- a/roles/install_ca/tasks/main.yml
+++ b/roles/install_ca/tasks/main.yml
@@ -22,6 +22,13 @@
         path: "{{ cifmw_install_ca_trust_dir }}"
         state: directory
 
+    - name: Install internal CA from url
+      when: cifmw_install_ca_url is defined
+      ansible.builtin.get_url:
+        url: "{{ cifmw_install_ca_url }}"
+        dest: "{{ cifmw_install_ca_trust_dir }}"
+        validate_certs: "{{ cifmw_install_ca_url_validate_certs | default(omit) }}"
+
     - name: Install custom CA bundle from inline
       register: ca_inline
       when:
--- a/roles/install_ca/README.md
+++ b/roles/install_ca/README.md
@@ -13,6 +13,10 @@ our capabilities to access remote package repositories, among possible targets.
 * `cifmw_install_ca_bundle_src`: (String) Path to the CA bundle we want to inject.
 * `cifmw_install_ca_trust_dir`: (String) Directory where we put custom CA. Must be known by the update-ca-trust command. Defaults to `/etc/pki/ca-trust/source/anchors/`.
 * `cifmw_install_ca_update_cmd`: (String) Command to run in order to update the CA trust ring. Defaults to `update-ca-trust`.
+* `cifmw_install_ca_url`: (String) URL pointing to a CA bundle that will be stored in `cifmw_install_ca_trust_dir`.
+* `cifmw_install_ca_url_validate_certs`: (Bool) Whether to validate SSL
+certificates when pulling a CA bundle from a url, will have no effect if
+`cifmw_install_ca_url` is not set.
 
 ## Examples
 ```YAML
@@ -28,6 +32,12 @@ our capabilities to access remote package repositories, among possible targets.
         -----BEGIN CERTIFICATE-----
         ....
         -----END CERTIFICATE-----
+  ansible.builtin.include_role:
+    role: install_ca
+
+- name: Inject custom CA from url
+  vars:
+    cifmw_install_ca_url: https://dummyurl.com/ca_file.pem
   ansible.builtin.include_role:
     role: install_ca
 ```
--- a/roles/openshift_setup/README.md
+++ b/roles/openshift_setup/README.md
@@ -10,3 +10,8 @@ No privilege escalation needed.
 * `cifmw_openshift_setup_create_namespaces`: (Strings) Namespaces to create beforehand. Defaults to `[]`.
 * `cifmw_openshift_setup_skip_internal_registry`: (Boolean) Skips login and setup of the Internal Openshift registry. Defaults to `false`.
 * `cifmw_openshift_setup_skip_internal_registry_tls_verify`: (Boolean) Skip TLS verification for Podman login. Defaults to `false`.
+* `cifmw_openshift_setup_ca_registry_to_add`: (String) Registry to which the CA
+should be configured for in an OCP/CRC cluster.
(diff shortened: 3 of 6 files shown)


roles/reproducer/tasks/install_ca.yml
cjeanner, line 1: Did you consider extending the install_ca role to support downloading a remote CA?
If you download the file, you'll need to run the `sudo update-ca-trust` once it's in the `anchors` directory.

cescgina:
@cjeanner @viroel I've updated the PR. The change to install_ca to pull from ca was simple enough, but I also introduced a new role to do the same job that was done by the downstream hook. I'm not sure I covered all cases there, please take a look when free

cescgina:
recheck

roles/configure_ocp_registry_ca/tasks/main.yml
```diff
@@ -0,0 +1,39 @@
...
+    - cifmw_configure_ocp_registry_ca_registry_to_add is defined
+    - cifmw_path is defined
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  block:
+    - name: Create config map with registry CAs
+      ansible.builtin.command:
+        cmd: >-
+          "oc create configmap registry-cas -n openshift-config --from-file={{ cifmw_configure_ocp_registry_ca_registry_to_add }}={{ cifmw_configure_ocp_registry_ca_bundle_path }}"
```
cjeanner, line 34:
```suggestion
          oc create configmap registry-cas -n openshift-config
          --from-file={{ cifmw_configure_ocp_registry_ca_registry_to_add }}={{ cifmw_configure_ocp_registry_ca_bundle_path }}
```

would be slightly better - but the `--from-file` is long on its own ^^'

roles/configure_ocp_registry_ca/tasks/main.yml
```diff
@@ -0,0 +1,39 @@
...
+
+    - name: Install Red Hat CA for pulling images from internal registry
+      ansible.builtin.command:
+        cmd: >-
+          oc patch image.config.openshift.io/cluster --patch '{"spec":{"additionalTrustedCA":{"name":"registry-cas"}}}' --type=merge
```
cjeanner, line 39:
```suggestion
          oc patch image.config.openshift.io/cluster
          --patch '{"spec":{"additionalTrustedCA":{"name":"registry-cas"}}}'
          --type=merge
```

roles/install_ca/tasks/main.yml
```diff
@@ -22,6 +22,12 @@
         path: "{{ cifmw_install_ca_trust_dir }}"
         state: directory
 
+    - name: Install internal CA from url
+      when: cifmw_install_ca_url is defined
+      ansible.builtin.get_url:
```
cjeanner, line 27: we may want to allow to not check TLS certificate at this point?

roles/configure_ocp_registry_ca/tasks/main.yml
cescgina: I've split the command in two lines and moved the `--from-file` to a private var so it's a bit shorter

roles/configure_ocp_registry_ca/tasks/main.yml
cescgina: done, thanks

roles/install_ca/tasks/main.yml
cescgina: I'm not sure what you're asking, is there some parameter to `get_url` that it's missing?

cescgina: I missed that parameter, thanks @cjeanner updated the patch with a variable that controls whether we validate the certs

roles/configure_ocp_registry_ca/molecule/default/converge.yml
```diff
@@ -0,0 +1,60 @@
...
+        kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
+        api_key: "{{ cifmw_openshift_token | default(omit)}}"
+        context: "{{ cifmw_openshift_context | default(omit)}}"
+        api_version: config.openshift.io/v1
+        kind: Image
+        name: cluster
+      register: _image_patch
+
+    - name: Debug cluster was patched
+      ansible.builtin.deub:
```
cjeanner, line 54: *debug

cescgina: indeed, thanks!

cescgina:
recheck

cescgina:
recheck

cescgina:
recheck

roles/configure_ocp_registry_ca/tasks/main.yml
```diff
@@ -0,0 +1,44 @@
...
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  vars:
+    _config_map_data: "{{ cifmw_configure_ocp_registry_ca_registry_to_add }}={{ cifmw_configure_ocp_registry_ca_bundle_path }}"
+  block:
+    - name: Create config map with registry CAs
+      ansible.builtin.command:
+        cmd: >-
+          oc create configmap registry-cas -n openshift-config
```
pablintino, line 36: I'm ok creating a role for such of a simple task, but the main goal of extracting code in a dedicated role is to allow it to be reused. The way this is coded will prevent it's re-usability, cause 2 calls to` oc create` for the same type and name will fail.

roles/configure_ocp_registry_ca/tasks/main.yml
```diff
@@ -0,0 +1,44 @@
...
+          --from-file="{{ _config_map_data }}"
+
+    - name: Install Red Hat CA for pulling images from internal registry
```
pablintino, line 39: While the command used here is fine the task will be flagged as changed always. Can we make it idempotent or just skip it if the CM is already added?

pablintino (changes requested):
While I'm ok with the general approach of this and I see that it's needed, do we really need a dedicated role for this?

pablintino (approved):

PR #1162, opened 2024-02-17
Target only known pool for refresh

cjeanner:
It seems some infra may have inactive pool, leading to a crash when we
try to refresh it.

This patch corrects the situation by:
- targeting only the needed pool
- ensuring that pool exists before refreshing


lewisdenny (approved):

cjeanner:
(working on a better approach - targeting only running state. It should be better in many ways)

lewisdenny:
Nice!

/lgtm

roles/libvirt_manager/tasks/start_manage_vms.yml
```diff
@@ -1,9 +1,13 @@
 ---
-- name: Refresh pools before starting VMs
+- name: Refresh oooq_pool before starting VMs
+  when:
+    - item.value.state == 'running'
```
pablintino, line 4: Based on the checks in the beaker node, this should be `active` not `running`.

cjeanner: I seem to have mixed things up:
```YAML
        "default": {
            "autostart": "yes",
            "path": "/var/lib/libvirt/images",
            "persistent": "yes",
            "size_available": "497357316096",
            "size_total": "506051166208",
            "size_used": "8693850112",
            "state": "active",
            "status": "running",
            "type": "dir",
            "uuid": "c9eb02d6-c91b-4994-8761-669dd2c8a10d",
            "volume_count": 0,
            "volumes": []
        },
```

IMHO checking for `status: running` is better than `state: active`. Updated accordingly.

cjeanner:
recheck

cjeanner:
(note that the fact is collected here: https://github.com/openstack-k8s-operators/ci-framework/blob/fc711e17c75f45e1aa76ee01758b115868e128cc/roles/libvirt_manager/tasks/deploy_layout.yml#L114-L118)

PR #1167, opened 2024-02-19
[automation] Modify reproducer role to support VA deploy

cjeanner:
This patch will help supporting architecture automation testing on
premises.

It cleans some wrong/deprecated code, while ensuring we have a clearer
path to the VA automation.

It also ensures we can access molecule logs so that we can debug the
current state better.

It also exposes a new environment file on the controller-0 to make it
easier to run any kind of reproducer - not only VA deployment. Note
that, for now, only VA deployment has its own nice and friendly
"deploy-va.sh" script in the zuul's home. More scripts will come as
needed.

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/1166

  - [X] README in the role


roles/reproducer/tasks/configure_controller.yml
```diff
@@ -270,6 +270,41 @@
...
+               'architecture/automation/vars/default.yaml') |
+               path_join
+            }}
+          pre_infra:
+            - name: Download needed tools
+              inventory: 'localhost,'
+              connection: local
+              type: playbook
+              source: >-
+                {{ cifmw_installyamls_repos }}/devsetup/download_tools.yaml
```
pablintino, line 305: I wouldn't expect install_yamls, VA or tools (not related to openshift) to be part of a file that is called openshift-environment.yml.

roles/reproducer/molecule/crc_layout/converge.yml
```diff
@@ -91,6 +91,20 @@
             <ip family='ipv4' address='{{ _networks.osp_trunk.range | ansible.utils.nthhost(1) }}' prefix='24'>
             </ip>
           </network>
+    cifmw_deploy_va_steps:
```
pablintino, line 94: I this variable consumed as part of this PR? Cause I think it's part of the next ones and it may be part of those ones.

roles/reproducer/tasks/main.yml
```diff
@@ -128,9 +122,15 @@
...
+          playbooks/01-bootstrap.yml
 
     - name: Emulate CI job
       when:
         - cifmw_job_uri is defined
       ansible.builtin.import_tasks: ci_job.yml
+
+    - name: Prepare VA deployment
+      when:
+        - cifmw_architecture_va_scenario is defined
```
pablintino, line 135: For me this one seems confusing, cause when a user calls the reproducer to fire up a new environment, he/she needs to pass something like this at the command line `-e cifmw_architecture_va_scenario=hci` (or put it in an env file). That, explicitly tells the user that the deployment will be for that particular VA, what is completely wrong, as the env can be used for whatever VA/DT they want (if it's layout matches the requirements, of course).

cjeanner: It's not wrong - this is automation, for the records. If they don't want to get any VA in place, so be it, just don't pass anything and build the needed bits on their own.
That parameter should be passed via the scenarios/reproducers/VA_NAME.yml environment file, in order to allow a full automation.
Don't forget that we don't have only VA. We therefore have to get something to switch the path between "non-VA/DT" vs "VA/DT".

That said, as done in the previous patch, removing the `_va` from the param name may be a good move, since we may support things that aren't VA, if properly described in the architecture repository.

roles/reproducer/molecule/crc_layout/converge.yml
cjeanner: actually, that one isn't needed anymore. Have to double-check, but molecule would load the architecture content directly.

roles/reproducer/tasks/configure_controller.yml
cjeanner: (more as a TODO to address later, really - too many moving parts)

roles/reproducer/tasks/main.yml
cjeanner: (TODO: rename top-level parameter)

roles/reproducer/molecule/crc_layout/converge.yml
cjeanner: useless, removed.

roles/reproducer/tasks/generate_deploy_values.yml
```diff
@@ -1 +0,0 @@
----
```
cescgina, line 1: is this file empty now? does it have any use?

pablintino:
recheck

roles/reproducer/tasks/generate_deploy_values.yml
cjeanner: it's removed in fact :)

PR #1168, opened 2024-02-19
[automation] Introduce flexible deploy loop

cjeanner:
This new tasks file will allow a more flexible approach for VA
deployment.

It will rely on the automation file provided in the architecture
repository, and execute each stage.

It also allows to stop the loop in 3 "strategic" steps, for each
described stage.

This patch also remove the install_dataplane and install_controlplane
tasks files, since they won't be used anymore. It allows to simplify
molecule tests.

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/1167

  - [X] README in the role


cjeanner:
recheck

roles/kustomize_deploy/tasks/execute_step.yml
```diff
@@ -0,0 +1,189 @@
...
+
+- name: Group tasks under the same tag
+  tags:
+    - "deploy_va_stage_{{ stage_id }}"
+  block:
+    - name: Ensure source files exists
+      register: _src
+      when:
+        - item.src_file is not none
+        - item.src_file != 'none'
```
pablintino, line 21: Where's the "none" string comming from? Sonds like a weird value. Ok to `=! None` cause you check their against the python value, but I don't see where is this value coming from?

roles/kustomize_deploy/tasks/execute_step.yml
```diff
@@ -0,0 +1,189 @@
...
+        name: ci_gen_kustomize_values
+      loop: "{{ stage['values'] }}"
+      loop_control:
+        loop_var: _val
+        index_var: _val_id
+        label: "{{ _val['name'] }}"
+
+    - name: "Copy generated values for {{ stage.path }}"
+      when:
+        - _val.src_file is not none
```
pablintino, line 83: ditto.

roles/kustomize_deploy/tasks/execute_step.yml
```diff
@@ -0,0 +1,189 @@
...
+        label: "{{ _val['name'] }}"
+
+    - name: Stop before building kustomization if requested
+      when:
+        - cifmw_deploy_va_stopper is defined
+        - cifmw_deploy_va_stopper == _stage_stopper
+      vars:
+        _stage_stopper: "pre_kustomize_stage_{{ stage_id }}"
+      ansible.builtin.fail:
+        msg: "Failing on demand {{ cifmw_deploy_va_stopper }}"
```
pablintino, line 113: I know we already talked about this, and that this may be the best possible approach. I have been checking how the `end_play` and `end_host` meta works and based on the code I think it works at play level, meaning, in this case, that calling it inside a loop ruins the entire purpose. Have you tried to do the following?

1. Remove the fail call
2. Call the `ansible.builtin.meta` with the `end_play` argument when needed and set an internal fact that signal that we want to stop (non-cacheable).
3. Add it as a condition in every call. We already have a when clause in every task here so I don't think its a big deal.
4. In the upper caller loop, add that flag as a condition. As the loop is in a playbook I'm pretty sure the iterator that handles it will properly stop.

roles/kustomize_deploy/molecule/flexible_loop/converge.yml
pablintino, line 1: Given the fact that this test runs with the dry-run option set its value is rather limited. I'm ok with that, as to run a proper VA we, for now, cannot use crc. But, there are verifications we can do, as an example. If someone touches the role and deletes a key task like [this](https://github.com/openstack-k8s-operators/ci-framework/pull/1168/files#diff-888ede7c7622d73459cf89a99ce06525e383be568e73d2fed5453eff970810a8R115-R134) one the molecule will still pass. I think we need to have a thought if some basic verifications can be done.

roles/kustomize_deploy/molecule/flexible_loop/converge.yml
```diff
@@ -0,0 +1,67 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+- name: Download tools
+  ansible.builtin.import_playbook: >-
+    /home/zuul/src/github.com/openstack-k8s-operators/install_yamls/devsetup/download_tools.yaml
```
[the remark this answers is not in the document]
cjeanner: ah, right, I had "some" issues due to "missing playbook" - due to that war between `.yaml` and `.yml` thing....

roles/kustomize_deploy/molecule/flexible_loop/molecule.yml
```diff
@@ -15,5 +15,8 @@ scenario:
     - create
     - prepare
     - converge
+    - verify
     - cleanup
     - destroy
+verifier:
+  name: ansible
```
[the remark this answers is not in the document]
cjeanner: cleaned to match other molecule.yml

roles/kustomize_deploy/tasks/execute_step.yml
```diff
@@ -0,0 +1,189 @@
...
+- name: Assert mandatory bits are defined
+  ansible.builtin.assert:
+    that:
+      - stage is defined
+      - stage_id is defined
+      - stage['path'] is defined
+      - stage.validations is defined
+      - stage.validations | length > 0
+      - stage['values'] is defined
+      - stage['values'] | length > 0
```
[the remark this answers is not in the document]
cjeanner: well, my expectation is to get a linter in the architecture repository to ensure the automation file is correct - then, those checks would be meaningless. For now, there's no real way to validate YAML, is there?

Also, we may be able to re-start the task at a specific stage, by using the `--skip-tags stage_...` introduced in the role.

roles/kustomize_deploy/tasks/execute_step.yml
cjeanner: so I'm 100% sure `end_host` doesn't work. I'll do another test with `end_play` - but I have little to no confidence it will really stop the whole playbook.

Also, there is a need to have an actual "break point" in the loop itself - being before or after some specific parts. Removing that capability will just call for "hey we need this to debug that specific part" - so I really don't think removing it is a good idea.
Moving the end_play to the upper level (and having some more `when` in the tasks) would just make the whole code more complicated to maintain imho.

roles/kustomize_deploy/tasks/execute_step.yml
```diff
@@ -0,0 +1,189 @@
...
+        PATH: "{{ cifmw_path }}"
+      ansible.builtin.shell:
+        cmd: >-
+          kustomize build > {{ _output }}
+        chdir: >-
+          {{
+            (cifmw_kustomize_deploy_architecture_repo_dest_dir,
+             stage['path']
+             ) | path_join
+          }}
```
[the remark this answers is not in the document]
cjeanner: done

roles/kustomize_deploy/tasks/execute_step.yml
```diff
@@ -0,0 +1,189 @@
...
+        KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+        PATH: "{{ cifmw_path }}"
+      ansible.builtin.command:
+        cmd: oc apply -f {{ stage['build_output'] }}
+        chdir: >-
+          {{
+            (cifmw_kustomize_deploy_architecture_repo_dest_dir,
+             stage['path']
+             ) | path_join
+          }}
```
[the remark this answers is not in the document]
cjeanner: done

roles/kustomize_deploy/molecule/flexible_loop/molecule.yml
```diff
@@ -15,5 +15,8 @@ scenario:
     - create
     - prepare
     - converge
+    - verify
     - cleanup
```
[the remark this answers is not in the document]
cjeanner: done (removed the scenario, it's defined in the main config)

roles/kustomize_deploy/molecule/flexible_loop/converge.yml
cjeanner: indeed. Maybe as a follow-up though? Unless you already have any ideas we can implement "now"?

roles/kustomize_deploy/tasks/execute_step.yml
cjeanner: So, I did some more tests here:
- define a fact before the loop
- start of the loop is checking if the defined fact exists, and if its value (true|false)
- during the loop, move tasks in blocks with `when` conditions based on an iterative approach (easier to manage)
- last task in the loop is to update the fact defined before the loop, in case we want to stop
- run, run, run
Result:
- the 06-deploy-va.yml "main" playbook is indeed stopped (both with `end_play` and `end_host`)
- but then it goes straight to the next playbook we import (hooks, then the rest of the list as set in the deploy-edpm.yml)
- eventually fails because of missing data, since we didn't finish to deploy OpenStack itself

sooo.... apparently, failing is the "best" way to actually stop the whole process....

roles/kustomize_deploy/tasks/execute_step.yml
cjeanner: Right thing to do:
- automation file gets rid of the "src_file" if not needed
- we check "is defined"

roles/kustomize_deploy/tasks/execute_step.yml
cjeanner: Nothing to do there then: we can't do it better without hacking really deep in some custom plugin.

roles/kustomize_deploy/molecule/flexible_loop/converge.yml
cjeanner: TODO: find the best way to test that role and especially the tasks.

PR #1169, opened 2024-02-19
[automation] Introduce flexible loop in deploy-edpm.yml

cjeanner:
Consuming architecture automation, this patch, with the rest of the
"[automation]" related patches, will allow to deploy VAs directly by
running the deploy-edpm.yml playbook.

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/1168

### DIFF
--- a/deploy-edpm.yml
+++ b/deploy-edpm.yml
@@ -38,7 +38,16 @@
     - build-operators
 
 - name: Import deploy edpm playbook
+  when:
+    - cifmw_architecture_va_scenario is undefined
   ansible.builtin.import_playbook: playbooks/06-deploy-edpm.yml
+  tags:
+    - edpm
+
+- name: Import VA deployment playbook
+  when:
+    - cifmw_architecture_va_scenario is defined
+  ansible.builtin.import_playbook: playbooks/06-deploy-architecture.yml
   tags:
     - edpm
 
--- a/playbooks/06-deploy-architecture.yml
+++ b/playbooks/06-deploy-architecture.yml
@@ -1,0 +1,124 @@
+---
+- name: Run pre_deploy hooks
+  vars:
+    hooks: "{{ pre_deploy | default([]) }}"
+    step: pre_deploy
+  ansible.builtin.import_playbook: ./hooks.yml
+
+- name: Deploy VA
+  hosts: "{{ cifmw_target_host | default('localhost') }}"
+  tasks:
+    - name: Load Networking Environment Definition
+      ansible.builtin.include_role:
+        name: networking_mapper
+        tasks_from: load_env_definition.yml
+
+    - name: Fetch controller-0 network facts
+      when: "'ansible_interfaces' not in hostvars['controller-0']"
+      ansible.builtin.setup:
+        gather_subset: network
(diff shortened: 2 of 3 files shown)


marios (commented):
where/what is the flexible loop though i was looking for it here :)

playbooks/06-deploy-va.yml
```diff
@@ -0,0 +1,124 @@
...
+
+    - name: Check requirements
+      ansible.builtin.import_role:
+        name: kustomize_deploy
+        tasks_from: check_requirements.yml
+
+    - name: Deploy OSP operators
+      ansible.builtin.import_role:
+        name: kustomize_deploy
+        tasks_from: install_operators.yml
```
pablintino, line 90: The "kustomization" capabilities we have here are really poor, basically none. How is a user going to kustomize this? https://github.com/openstack-k8s-operators/architecture/blob/main/lib/olm-openstack/kustomization.yaml (needs work in the architecture repo too, so the scope of this comment is not only this PR)

playbooks/06-deploy-va.yml
```diff
@@ -0,0 +1,124 @@
...
+      loop_control:
+        label: "{{ stage.path }}"
+        loop_var: stage
+        index_var: stage_id
+
+    - name: Extract and install OpenStackControlplane CA
+      ansible.builtin.include_role:
+        role: install_openstack_ca
+
+    - name: Run nova host discover process
```
pablintino, line 110: I'm sure all the VAs/DTs will deploy a dataplane, but, if at somepoint we want to use this for CI and deploy a cropped OSP with only control-plane, we won't run this.

playbooks/06-deploy-va.yml
rebtoor: +1

cjeanner: I already talked about this with @rebtoor - there's a need to extend a bit the `install_operators.yml` to allow kustomization. This won't be part of the current patch though, the feature needs to be implemented first.

playbooks/06-deploy-va.yml
cjeanner: maybe this should be part of a post_run hook in the latest stage as described in the architecture automation? @fultonj @abays any thoughts?

fultonj: So add a `post_stage_run` to [1] like this?
```
        post_stage_run:
          - name: Run nova host discover process
            type: command?
            ?: oc rsh -n openstack nova-cell0-conductor-0 nova-manage cell_v2 discover_hosts --verbose
```

[1] https://github.com/openstack-k8s-operators/architecture/blob/main/automation/vars/default.yaml#L43

cjeanner: or maybe a validation? Not really happy to inject a command support in hooks, to be honest. The "playbook" approach sounds overkill - sooo... bleh. TBD.

cjeanner: TODO: find a better way to trigger that command - for now we keep it as is, and will revisit.

cjeanner:
recheck

PR #1170, opened 2024-02-19
[os_net_setup] Update public_subnet default allocation pool

karelyatin:
192.168.122.200 also configured for netconfig end_range, so for public subnet use 192.168.122.201 as starting range. Also extend end range till 192.168.122.250 to allow more FIPs by default.

  - [x] README in the role

### DIFF
--- a/roles/os_net_setup/defaults/main.yml
+++ b/roles/os_net_setup/defaults/main.yml
@@ -11,8 +11,8 @@ cifmw_os_net_setup_config:
     subnets:
       - name: public_subnet
         cidr: 192.168.122.0/24
-        allocation_pool_start: 192.168.122.200
-        allocation_pool_end: 192.168.122.240
+        allocation_pool_start: 192.168.122.201
+        allocation_pool_end: 192.168.122.250
         gateway_ip: 192.168.122.1
         enable_dhcp: false
 cifmw_os_net_setup_verify_tls: true


roles/os_net_setup/defaults/main.yml
```diff
@@ -11,8 +11,8 @@ cifmw_os_net_setup_config:
     subnets:
       - name: public_subnet
         cidr: 192.168.122.0/24
```
cjeanner, line 13: so that subnet is the same as the osp_trunk we have, which collide with the `default` network in libvirt.

Would it be possible to get a parameter for that subnet (and use ansible.builtin.ipmath filter to calculate IPs based on that parameter subnet) ? That way, we may be able to consume what networking-environment-definition provides instead of that hard coded subnet, without having to override the whole dict. WDYT?

@pablintino any thoughts? Am I completely out of topic?

pablintino: @cjeanner It has nothing to do with that. Here we create a new neutron network leveraging the unused  very end of the controlplane network, that lasts till the .200 IP, included. As @karelyatin says, that IP is used by the control-plane, so it shouldn't be used by the new OSP network.

cjeanner:
/lgtm

probably too early to get rid of hardcoded subnets :/.

roles/os_net_setup/defaults/main.yml
karelyatin: Yes right if the underlying network is change it would have an impact here. So would need a way to sync these.

PR #1171, opened 2024-02-19
Tobiko can be executed with test-operator

eduolivares:
Before this patch, tobiko tests could only be executed using the ci-fmw's tobiko role and the ci-fmw's test_operator tole only supported tempest.
With this patch, tobiko tests can be run using the test_operator role as well.

TODO:
- support tobiko workflows - requires changes in tcib
- pass relevant env variables to the tobiko container - requires changes
   in test-operator

[OSPRH-4831](https://issues.redhat.com//browse/OSPRH-4831)

  - [X] README in the role


playbooks/08-run-tests.yml
```diff
@@ -7,14 +7,34 @@
 - name: "Test playbook for {{ run_test_fw }}"
   vars:
     run_test_fw: "{{ test_fw | default('tempest') }}"
+    run_test_role: "{{ cifmw_run_test_role | default(run_test_fw) }}"
   hosts: "{{ cifmw_target_host | default('localhost') }}"
   gather_facts: false
   tasks:
-    - name: "Run {{ run_test_fw }}"
+    - name: "Ensure tests can be run using the selected role"
```
lpiwowar, line 14: I'm not sure whether it really belongs here to decide which framework will be executed as part of the test role. In my opinion it belongs to the test-operator role but I'm not sure. I think that most likely the test-operator role will be the only role capable of executing different testing frameworks (if there every will be another test role aside from tempest,tobiko or test_operator).

I would keep the `08-run-tests.yml` simple (just deciding which role will be executed) and then I would leave any further logic for the specific roles. But this is just an idea.

Also right now the `test_fw` is used in different projects to turn on execution of tests via the test_operator. This patch would change the meaning of the test_fw variable and therefore break already running test operator based jobs.

roles/test_operator/README.md
```diff
@@ -46,3 +46,32 @@ Execute tests via the [test-operator](https://openstack-k8s-operators.github.io/
       externalPlugin: "{{ cifmw_test_operator_tempest_external_plugin | default([]) }}"
     tempestconfRun: "{{ cifmw_tempest_tempestconf_config | default(omit) }}"
 ```
+
+## Tempest specific parameters
```
lpiwowar, line 50: nit: Tobiko specific parameters

playbooks/08-run-tests.yml
eduolivares: I am going to try to apply your proposal. For this, I will have to remove one of these two calls to the `08-run-tests.yml` playbook:
https://github.com/openstack-k8s-operators/ci-framework/blob/main/deploy-edpm.yml#L50-L72

Hence, the mutually exclusive options will be:
- to execute tempest once using the tempest role
- to execute tobiiko once using the tobiko role (which is not used by any CI job and this role is being deprecated)
- to execute tempest, tobiko or both using the test_operator role.

> Also right now the test_fw is used in different projects
I would insist in renaming it to either cifmw_test_fw or cifmw_test_role, in order to align it to other cifmw params.
Changing the existing jobs is not a big deal.
Right now, only upstream jobs are using it:
https://github.com/search?q=org%3Aopenstack-k8s-operators+%22test_fw%3A+test_operator%22&type=code
And a few downstream jobs that can be simply identified and fixed too.

lpiwowar (approved):
LGTM! It seems to be working nicely! Thank you!:) Also the change was tested in a job here [1].

My only concern is that we should handle the renaming of the test_fw variable properly. Ideally we should rename the `test_fw` variable in the few jobs that are already using the test-operator.

Idea: It would be nice if we in the future modified the role so that the logs are in separate folder `ci-framework-data/logs/test_operator/[tobiko|tempest]/*`

[1] https://github.com/openstack-k8s-operators/test-operator/pull/47

playbooks/08-run-tests.yml
cjeanner: (as a matter of fact, prefixing parameters with `cifmw_` is a good practice, since it will allow to expose them later in the parameters that would then be consumable from the controller-0, or any hooks)

cjeanner:
/approve

first step of some refactoring - let's move on. Seems to have successful runs already, so we should be able to merge as-is.

PR #1172, opened 2024-02-19
[FIPS] Adding small fixes/alternatives to avoid issues in FIPS mode

viroel:
* ssh-keyscan is failing even with two FIPS-enabled nodes. Needs further investigation. For now, try to replace with a simple ssh operation.

  - [x] README in the role

### DIFF
--- a/ci/playbooks/multinode-customizations.yml
+++ b/ci/playbooks/multinode-customizations.yml
@@ -171,9 +171,11 @@
         state: present
 
     - name: Ensure we know compute host keys
-      ansible.builtin.shell:
-        cmd: |
-          ssh-keyscan {{ hostvars[item].ansible_host }} >> ~/.ssh/known_hosts
+      # Using a dumb ssh command to get host keys
+      # ssh-keyscan is failing when running under FIPS mode
+      ansible.builtin.command: >-
+          ssh -o StrictHostKeyChecking=no
+          {{ hostvars[item].ansible_user }}@{{ hostvars[item].ansible_host }} uptime
       loop: >-
         {{
           hostvars | dict2items |
--- a/hooks/playbooks/fetch_compute_facts.yml
+++ b/hooks/playbooks/fetch_compute_facts.yml
@@ -238,9 +238,9 @@
                     value: ["0.0.0.0/0"]
 
         - name: Ensure we know about the private host keys
-          ansible.builtin.shell:
-            cmd: |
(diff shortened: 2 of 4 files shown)


viroel:
recheck

ci/playbooks/multinode-customizations.yml
```diff
@@ -173,7 +173,7 @@
     - name: Ensure we know compute host keys
       ansible.builtin.shell:
```
pablintino, line 174: With this change in place we can replace `shell` by `command`.

ci/playbooks/multinode-customizations.yml
```diff
@@ -173,7 +173,7 @@
...
         cmd: |
-          ssh-keyscan {{ hostvars[item].ansible_host }} >> ~/.ssh/known_hosts
+          ssh -o StrictHostKeyChecking=no {{ hostvars[item].ansible_user }}@{{ hostvars[item].ansible_host }} uptime
```
pablintino, line 176: I'd add a comment here saying why we run such of a dumb command here using SSH, clearly explaining that is to replace `ssh-keyscan` because of blabla.

viroel:
/test pre-commit

`[2mfatal: unable to access 'https://github.com/ansible-collections/ansible.netcommon/': error:0A000126:SSL routines::unexpected eof while reading[0m`

ci/playbooks/multinode-customizations.yml
```diff
@@ -171,9 +171,11 @@
         state: present
 
     - name: Ensure we know compute host keys
-      ansible.builtin.shell:
-        cmd: |
-          ssh-keyscan {{ hostvars[item].ansible_host }} >> ~/.ssh/known_hosts
```
cjeanner, line 176: I think we "just" need to set the `-t <type>` for ssh-keyscan to NOT try to get RSA keys - or any non-FIPS compatible type?

viroel: Already tried, without success. Might be some other detail that is hard to debug since sshd fails without too much details (not even with debug level set).

cjeanner: sshd itself? "nice". Is this something we may debug later using the reproducer? is anything needed on that field actually, other than this fix/workaround? Since we have console access...

viroel:
closing this PR since small changes in sshd config will fix this issue.

PR #1163, opened 2024-02-19
[WIP] Added Podified Multinode Ironic Job

raukadah:
The goal of this pull request is to
- Add a base job to deploy podified control plane. It will be useful for testing non-EDPM services like ironic operator.
- It drops old e2e notagg build ironic job (It is not used anywhere)
- It adds an Ironic job template which will be consumed against ironic operator.

As a pull request owner and reviewers, I checked that:
- [ ] Appropriate testing is done and actually running
- [ ] Appropriate documentation exists and/or is up-to-date:
  - [ ] README in the role
  - [ ] Content of the docs/source is reflecting the changes

PR #1173, opened 2024-02-20
Ironic jobs

raukadah:
  - [ ] README in the role

### DIFF
--- a/scenarios/centos-9/podified_common.yml
+++ b/scenarios/centos-9/podified_common.yml
@@ -1,0 +1,41 @@
+---
+# It is the common scenario file for EDPM multinode podified deployment
+
+ansible_user_dir: "{{ lookup('env', 'HOME') }}"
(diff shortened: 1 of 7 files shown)


scenarios/centos-9/podified_common.yml
```diff
@@ -0,0 +1,41 @@
+---
```
tosky, line 1: The comment talks about EDPM multinode, but then cifmw_deploy_edpm: false. If this is a common file for EDPM and non-EDPM deployment maybe the comment should reflect it

tosky:
Why a new pull request? Didn't we have https://github.com/openstack-k8s-operators/ci-framework/pull/1163 already?

scenarios/centos-9/podified_common.yml
raukadah: Thank you @tosky .Good point, I will update that in the main pr.

PR #1174, opened 2024-02-20
[reproducer] Fix wrongly coded when condition that was always true

pablintino:


cescgina (commented):
/lgtm

PR #1178, opened 2024-02-21
Ensure libvirt_manager cleanup is safe when libvirt is not deployed

pablintino:

### DIFF
--- a/roles/libvirt_manager/tasks/clean_layout.yml
+++ b/roles/libvirt_manager/tasks/clean_layout.yml
@@ -1,147 +1,156 @@
 ---
-- name: List all of the existing virtual machines
-  register: vms_list
-  community.libvirt.virt:
-    command: list_vms
-    uri: "qemu:///system"
+- name: Get installed packages list
+  ansible.builtin.package_facts: {}
 
-- name: Get pool configuration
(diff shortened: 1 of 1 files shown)


roles/libvirt_manager/tasks/clean_layout.yml
```diff
@@ -1,147 +1,155 @@
...
-    path: "{{ ansible_user_dir }}/.ssh/cifmw_reproducer_key.pub"
-
-- name: Remove temporary ssh key from authorized_keys
+- name: Get installed packages list
+  ansible.builtin.package_facts: {}
+
+- name: Perform the libvirt cleanup
   when:
-    - _tmp_key.stat.exists
+    - "'libvirt' in ansible_facts.packages"
```
cjeanner, line 7: should be enough to ensure we're facing a deployed env - even if we do have those packages but nothing is deployed, it will just loop on empty lists and remove absent directories/files.

cjeanner (commented):
/approve

cescgina (commented):
/lgtm

cjeanner:
/lgtm
re-adding the missing tag - the current solution (checking for all packages instead of a small subset) is better and more future-proof :+1:

PR #1181, opened 2024-02-21
Allow to properly scale nodes

cjeanner:
Until now, the network-info.yml wasn't updated in case of re-deploy.
While it was OK with the first iteration of its generation, it doesn't
make sense anymore: we're gathering all the interfaces, always, for old
and new VM.

Removing the "force: false" will ensure we're properly populating the VM
interfaces, even in case of a scale-out, meaning a second run.

### DIFF
--- a/roles/libvirt_manager/tasks/deploy_layout.yml
+++ b/roles/libvirt_manager/tasks/deploy_layout.yml
@@ -162,7 +162,6 @@
   ansible.builtin.copy:
     dest: "{{ cifmw_libvirt_manager_basedir }}/artifacts/interfaces-info.yml"
     content: "{{ cifmw_libvirt_manager_mac_map | to_nice_yaml }}"
-    force: false
 
 - name: Ensure we get proper access to CRC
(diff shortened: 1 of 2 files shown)


psathyan (approved):
/lgtm

roles/libvirt_manager/tasks/deploy_layout.yml
```diff
@@ -162,7 +162,6 @@
   ansible.builtin.copy:
     dest: "{{ cifmw_libvirt_manager_basedir }}/artifacts/interfaces-info.yml"
     content: "{{ cifmw_libvirt_manager_mac_map | to_nice_yaml }}"
-    force: false
 
 - name: Ensure we get proper access to CRC
   when:
```
marios, line 167: ack (defaults true had to check ;)) https://docs.ansible.com/ansible/latest/collections/ansible/builtin/copy_module.html#parameter-force

marios (approved):
sounds sane

PR #1182, opened 2024-02-21
Allow to bump the amount of compute with a parameter

cjeanner:
This patch, with #1181, should allow to easily scale-out the amount of
computes, without having to edit the validated-architecture-1.yml nor
the 3-nodes.yml.

It's taking the same parameter as the libvirt_manager role,
"cifmw_libvirt_manager_compute_amount".

The VA-1 scenario will pick the maximum amount, meaning we'll get at
least 3 computes; the 3-nodes will do the same, meaning we'll get at
least 1 node.


psathyan (approved):
/lgtm

PR #1184, opened 2024-02-22
Add extra skip tests to nova-operator

lewisdenny:
Add additional tempest tests to exclude from the nova-operator that are
currently not supported with default deployment. [1]

These tests fail because the default deployment doesn't have a compute node.

[1] https://sf.hosted.upshift.rdu2.redhat.com/logs/55/55260f435033f8469699ecba87475748a8bc0790/openstack-operators-periodic-integration-osp18-rhel9/podified-multinode-tempest-internal-integration-rhel9-osp18/942bb62/controller/ci-framework-data/tests/tempest/stestr_results.html

### DIFF
--- a/roles/tempest/files/list_skipped.yml
+++ b/roles/tempest/files/list_skipped.yml
@@ -1127,6 +1127,42 @@ known_failures:
         lp: http://no.bug
     jobs:
       - nova-operator
+  - test: tempest.api.compute.servers.test_server_actions.ServerActionsTestOtherA
+    deployment:
+      - overcloud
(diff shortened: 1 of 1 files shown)


marios (approved):

SeanMooney:
so in general we dont expect to maintain or use this file for compute maintened jobs.

for example all jobs in the nova-operator repo wont use this file and it wont be used downstream

i strongly dislike centralising a skip list like this so please ensure this file is only pulled in for jobs that ask for it.

PR #1186, opened 2024-02-22
Remove cifmw_extras from the reproducer vars

pablintino:
The cifmw_extras is a variable we use to pass extra variables to the nested ansible-playbooks we usually use in CI. That variable, that has nothing to do with the framework itself, may contain jinja templates, to shortcut the access to zuul project dirs for example. That jinja, that the reproducer takes as a string, it's problematic when playbooks inside the reproducer, like the autogenerated cifmw-bootstrap.yml, tries to resolve it, cause variables, like zuul there, are not available.
To avoid these issues, and based on the fact that cifmw_extras is not a variable that the framework should care about out of our zuul CI, we now filter it out.

  - [x] README in the role

### DIFF
--- a/roles/reproducer/tasks/configure_controller.yml
+++ b/roles/reproducer/tasks/configure_controller.yml
@@ -258,6 +258,7 @@
             rejectattr('key', 'equalto', 'cifmw_target_host') |
             rejectattr('key', 'equalto', 'cifmw_basedir') |
             rejectattr('key', 'equalto', 'cifmw_path') |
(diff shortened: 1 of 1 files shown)


cjeanner:
/approve

makes perfectly sense. If really the reproducer must parse it, it should happen beforehand, in the ci_data.yml, where we have access to almost everything from the CI, zuul vars included.

rebtoor (approved):

PR #1187, opened 2024-02-22
fix: remove ironic allocation test from run

juliakreger:
Ironic, internally, has a lightweight allocation engine, originally added for metalsmith usage ages ago. Turns out its unit tests can fail when running automated cleaning combined with default deployment interface which is not the fake driver. This was not detected upstream because only the fake drivers are executed together.

Needed for: https://github.com/openstack-k8s-operators/ironic-operator/pull/401

  - [x]  Appropriate testing is done and actually running

### DIFF
--- a/roles/tempest/files/list_skipped.yml
+++ b/roles/tempest/files/list_skipped.yml
@@ -1310,3 +1310,12 @@ known_failures:
         lp: http://no.bug
     jobs:
       - ironic-operator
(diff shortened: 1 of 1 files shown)


hjensas (approved):

raukadah:
@juliakreger thank you for adding the pull request. I am going to update the PR description for making check-main-comment job passing.

PR #1190, opened 2024-02-24
Ensure libvirt manager layout is consistent

psathyan:
This change enforces `cifmw_libvirt_manager_configuration` as a single source and is consistent. It removes `devscripts_layout` that tends to change the precedence.

  - [x] README in the role


raukadah (approved):
It is not used in downstream reproducer devscript BM job. Safe to merge.

PR #1193, opened 2024-02-26
Ensure we keep short file names

cjeanner:
While system may support long file name (over 250 chars), it seems
ansible doesn't really like that.

Also, it may look really weird having a hook name so long it takes
multiple lines to display.

In order to make it more readable, and avoid any file name length issue,
we're now truncating the hook name to 20 chars. This will lead to better
UX, and may avoid some weird issues.


queria (commented):
/lgtm

PR #1194, opened 2024-02-26
Dump dlrn md5 hash in cifmw_repo_setup_output path

raukadah:
Currently dlrn md5 hash is written in a hardcoded directory. If cifmw_repo_setup_output is set to something else then dlrn md5 file will not be generated in cifmw_repo_setup_output directory and any role depending on the cifmw_repo_setup_output dlrn md5 file breaks.

Generating it in cifmw_repo_setup_output directory fixes the issue.

### DIFF
--- a/roles/repo_setup/tasks/sync_repos.yml
+++ b/roles/repo_setup/tasks/sync_repos.yml
@@ -1,6 +1,6 @@
 ---
 - name: Copy generated repos to /etc/yum.repos.d directory
-  when: not cifmw_repo_setup_output.startswith(ansible_user_dir)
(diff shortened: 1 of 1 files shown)


ci/playbooks/tcib/tcib.yml
```diff
@@ -17,7 +17,7 @@
     - name: Copy repo-setup generated repos to yum.repos.d directory
       tags:
         - container_img_build
-      become: true
+      become: "{{ not cifmw_build_containers_repo_dir.startswith(ansible_user_dir) }}"
```
cjeanner, line 20: oh, that one must stay "true", missed it!

raukadah:
No longer needed, closing it.

PR #1196, opened 2024-02-26
Minor refactor in edpm_prepare and new hook for ceph scenarios

viroel:
When deploying an HCI environment (together with edpm deploy), the
Control Plane resources are already ready and we don't have a chance of
kustomizing them with ceph backends. To enable some storage backend
with Ceph, we need to provide the Ceph FSID, which is available only
after edpm_deploy.
This change refactor edpm_prepare to split a set of tasks used to
kustomize and deploy the control plane again, if needed, after edpm
deploy. These tasks can be called in a hook (added in this PR as example)
to update the Control Plane resources.

  - [X] README in the role

### DIFF
--- a/roles/edpm_prepare/tasks/kustomize_and_deploy.yml
+++ b/roles/edpm_prepare/tasks/kustomize_and_deploy.yml
@@ -1,0 +1,84 @@
+---
+# Copyright Red Hat, Inc.
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+- name: Perform kustomizations to the OpenStackControlPlane CR
+  vars:
+    cifmw_edpm_prepare_openstack_crs_path: >-
+      {{
+        [
+          cifmw_edpm_prepare_manifests_dir,
+          cifmw_install_yamls_defaults['NAMESPACE'],
+          'openstack',
+          'cr'
+        ] | ansible.builtin.path_join
+      }}
+  ci_kustomize:
+    target_path: "{{ cifmw_edpm_prepare_openstack_crs_path }}"
+    sort_ascending: false
+    kustomizations: "{{ cifmw_edpm_prepare_kustomizations }}"
+    kustomizations_paths: >-
+      {{
+        [
+          (
+            [
+              cifmw_edpm_prepare_manifests_dir,
+              'kustomizations',
+              'controlplane'
+            ] | ansible.builtin.path_join
+          )
+        ]
+      }}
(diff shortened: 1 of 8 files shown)


viroel:
/hold

under tests

viroel:
Some testproject results from https://review.rdoproject.org/r/c/testproject/+/51865:
* Running simple scenario test: podified-multinode-hci-deployment-crc-1comp-testproject-default https://review.rdoproject.org/zuul/build/dbd028a21cb9410b89ae3344589d665c : SUCCESS in 1h 19m 36s
* Running with cinder/manila tests: podified-multinode-hci-deployment-crc-1comp-testproject https://review.rdoproject.org/zuul/build/1face28934a443adbac88110c1932752 : FAILURE in 2h 16m 53s
 * Manila tests are failing because Ceph MDS daemon is failing to start, needs further investigation and testing in single node ceph deploy. 
 
 Besides the failure, we were able to configure both cinder/manila with the Ceph backend

marios (commented):
looks ok from visual inspection but its a big change  
noting that this  does more than just "add job definition"

viroel:
> looks ok from visual inspection but its a big change noting that this does more than just "add job definition"

You are correct. This started as adding a simple single HCI job and is now a complete different thing. I already fixed the description, and I will also update the commit message.
I decide to remove the job definition since it is not going to be used here. Can be added afterwards, if needed.

viroel:
testproject with for with cinder tests: podified-multinode-hci-deployment-crc-1comp-testproject-cinder https://review.rdoproject.org/zuul/build/d9b93d832c034b71b44c43036935632d : SUCCESS in 1h 12m 27s
Manila still needs some fixes in ceph deploy, but will not be covered by this PR.

roles/edpm_prepare/tasks/kustomize_and_deploy.yml
```diff
@@ -0,0 +1,84 @@
...
+
+- name: Log the CR that is about to be applied
+  ansible.builtin.debug:
+    var: cifmw_edpm_prepare_crs_kustomize_result
+
+- name: Apply the OpenStackControlPlane CR
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  cifmw.general.ci_script:
```
marios, line 54: i know this is a copy of what was there but just FYI we added 'extra_args' when removing ci_make so you can use that instead of environment

```
  extra_args:
    PATH: "{{ cifmw_path }}"
```
environment: works fine, but we don't get this in the resulting logs or output from ci_script (see https://github.com/openstack-k8s-operators/ci-framework/blob/f957797867789b42ffdf6031945a844853d7ab4b/plugins/action/ci_script.py#L118)

I am in 2 minds about updating this here but up to you.

marios (approved):
looks good as a tidy up thanks for the update to the commit message and job removal

agree adding the job can be a followup after this merges

pablintino:
Hi @viroel, I'm not sure about the agreement you have made with @marios, but I think we need a job definition to be sure anyone that wants to use this scenario file know which other files he/she needs to pass. I'd add it even if the job is not used. No problem to do so in a follow up if you want
.

roles/edpm_prepare/tasks/kustomize_and_deploy.yml
viroel: yeah, it was just a code move from a file to another, so I didn't pay attention to improvements. I would say that we can make a hard this as hard requirement for new code. For this role, would be good to update in all places (at least for the role)

PR #1200, opened 2024-02-27
Support non-default OCP cluster name

psathyan:
Add support for gathering the cluster access information for non-default OCP Cluster name

The cluster name is configured in `cifmw_devscripts_configs.cluster_name` and the default value is `ocp`. `dev-scripts` stores the authentication information at `<dev-scripts-repo-dir>/ocp/<cluster-name>/auth/` directory.

_Testing scenarios_
[x] reproducer play with defaults
[x] reproducer play with non-default cluster name.


son-vyas:
are we testing this anywhere or we are good without testing here?

psathyan:
Testing is mostly manual as we need the cluster deployed and it is conditionally off during molecule testing. We don't have any specific jobs testing this feature at the moment.

roles/reproducer/tasks/devscripts_post.yml
```diff
@@ -17,18 +17,24 @@
       ansible.builtin.slurp:
         path: >-
           {{
-            (ansible_user_dir,
```
pablintino, line 20: The fact that we need to edit this in a couple of tasks is just suggesting we may want to move the path (till the /auth) to a variable in the block.

psathyan: Thank you. It is a good suggestion. Let me incorporate the change.

psathyan:
`cifmw_devscripts_repo_dir` and `cifmw_devscripts_config.cluster_name` variables are defined by default. The former is set in the defaults while the later comes from `cifmw_devscripts_config_defaults` in vars.

PR #1202, opened 2024-02-27
Change top-level parameters

cjeanner:
The CI Framework shouldn't care about scenario type - VAs and DTs are
mostly the same. We want to care about defined automation only.

This patch removes most of the "va" from parameter names to be agnostic.

We also correct a bug introduced in the latest patches related to
automation: the missing shebang in the `deploy-architecture.sh` script
prevented using `ansible.builtin.command`.

  - [X] README in the role

### DIFF
--- a/roles/reproducer/tasks/configure_architecture.yml
+++ b/roles/reproducer/tasks/configure_architecture.yml
@@ -4,13 +4,14 @@
   block:
     - name: Push script
       ansible.builtin.copy:
-        dest: "/home/zuul/deploy-va.sh"
+        dest: "/home/zuul/deploy-architecture.sh"
         mode: "0755"
         owner: "zuul"
         group: "zuul"
         content: |-
+          #!/bin/bash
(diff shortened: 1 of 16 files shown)


cjeanner:
/hold

we need to sync with DS.

roles/reproducer/tasks/configure_architecture.yml
```diff
@@ -4,13 +4,13 @@
   block:
     - name: Push script
       ansible.builtin.copy:
-        dest: "/home/zuul/deploy-va.sh"
+        dest: "/home/zuul/deploy-architecture.sh"
         mode: "0755"
         owner: "zuul"
         group: "zuul"
         content: |-
           pushd src/github.com/openstack-k8s-operators/ci-framework
```
cjeanner, line 12: there's an issue with the script, apparently calling it via `command` isn't working, I suspect the missing shebang to be the issue - currently testing.

cjeanner: corrected.

PR #1203, opened 2024-02-27
Simplify checks on the src_file

cjeanner:
With [1] we're now able to remove most of the checks, ensuring we only
rely on defined/undefined instead of lame string comparison.

[1] https://github.com/openstack-k8s-operators/architecture/pull/110

Depends-On: https://github.com/openstack-k8s-operators/architecture/pull/110


lewisdenny (approved):
/lgtm

"instead of lame string comparison." 😆

marios (approved):
looks right and grep didn't give me any more examples  "src_file.*none"

PR #1205, opened 2024-02-29
Fix iSCSI variables

maximsava12:
This fix naming and format for variables


psathyan (approved):
/lgtm

PR #1206, opened 2024-02-29
Precisely target values.yaml file for user kustomization

cjeanner:
Until now, the user provided kustomization would be applied to all of
the values.yaml files in a specific stage. This may lead to errors, and
isn't the right way.

This patch now adds a level of precision, so that we can target the
right file without leaking any data outside of it.


psathyan (approved):
/lgtm

PR #1083, opened 2024-01-30
Converge edpm group to the computes one

pablintino:
As a pull request owner and reviewers, I checked that:
- [x] Appropriate testing is done and actually running
- [x] Appropriate documentation exists and/or is up-to-date:
  - [x] README in the role
  - [x] Content of the docs/source is reflecting the changes

PR #1207, opened 2024-03-01
Add support for deploying external ceph cluster

rachael-george:
[OSPRH-2737](https://issues.redhat.com//browse/OSPRH-2737) - Support external Ceph deployment

Adds support for deploying external ceph cluster via the architecture repo

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/1224


fultonj:
This commit message needs to be fixed.

`Deploys ceph on non-edpm nodes by using two nodesets, one defined for EDPM and the other for Ceph nodes`

I think you meant  this.

`Deploys ceph on EDPM nodes which are not Compute nodes.`

Not all EDPM nodes are compute nodes. For example, the EDPM nodes in this patch are not compute nodes.

EDPM nodes are dataplane nodes. They are usually compute nodes, but not always.

playbooks/deploy-external-ceph.yml
```diff
@@ -0,0 +1,253 @@
...
+    - name: Create a nodeset for ceph
+      block:
+        - name: Ensure that directory exists
+          ansible.builtin.file:
+            path: "{{ cifmw_ceph_dir }}"
+            state: directory
+
+        - name: Get sample nodeset file from URL
+          get_url:
+            url: "https://raw.githubusercontent.com/openstack-k8s-operators/dataplane-operator/main/config/samples/dataplane_v1beta1_openstackdataplanenodeset.yaml"
```
fultonj, line 66: This will break after this merges.

  https://github.com/openstack-k8s-operators/dataplane-operator/pull/698

playbooks/deploy-external-ceph.yml
```diff
@@ -0,0 +1,253 @@
...
+# License for the specific language governing permissions and limitations
+# under the License.
+
+- name: Pre-requisites for deploying external ceph
+  hosts: localhost
+  vars:
+    cifmw_ceph_dir: "{{ ansible_user_dir }}/ci-framework-data/manifests/ceph"
+    cifmw_ceph_node_num: "{{ cifmw_ceph_node_num | default(3) }}"
+  tasks:
+    - name: Prepare inventory for ceph nodes
```
fultonj, line 23: If I have `N` compute nodes how is the subset of `cifmw_ceph_node_num` ceph nodes identified?

The first `cifmw_ceph_node_num` of them or are they expected to have a certain name?

I think you might be relying on some convention, and that convention should at least be documented in  comment.

fultonj (commented):
I like that this playbook works and that it uses node sets so that changes to the operators or the ci-framework are unnecessary. I understand that it will also solve a problem we want to solve today.

However, we're trying to move to a pattern where an external structure [1] defines the CRs and what order they are applied in. Following that pattern would make this contriubtion to the ci-framework make more sense from a design perspective.

Thus, you could instead create a DT in the architecture repos which has two nodesets and use the posthooks to call ceph. This is really a variation of the HCI deployment and it fitst that pattern.

It might also involve undoing part of this change [2] though it would need to be done without breaking the VA1 automation. Not all EDPM nodes are compute nodes. They are for HCI but not for this case. The name could be hacked but I think it's better to support any group name as a variable.

I encourage you to try the DT and structure approach instead of adding this playbook as it will fit a pattern which is more easy to automate. You could use this playbook (outside of ci-framework) in the meantime to make a quick job if you're balancing priority.

[1] https://github.com/openstack-k8s-operators/architecture/blob/main/automation/vars/default.yaml
[2] https://github.com/openstack-k8s-operators/ci-framework/pull/1083

fultonj:
@rachael-george thanks but could you try this as a DT instead?

Ultimately the same executions will happen but the implantation will be described in the architecture repos as CRs  and a structure [1] of the order that they will be run

[1] https://github.com/openstack-k8s-operators/architecture/blob/main/automation/vars/default.yaml

rachael-george:
I have updated the PR going in the direction of supporting external Ceph deployment through the architecture repo. However, there might be some changes required in the ci-framework repo as well, which this PR will track now

scenarios/reproducers/external-ceph.yml
```diff
@@ -0,0 +1,165 @@
...
+# the OCP nodes. This one is a "private network", and will hold
+# the VLANs used for network isolation.
+
+# Please create a custom env file to provide:
+# cifmw_devscripts_ci_token:
+# cifmw_devscripts_pull_secret:
+
+# type and size of ssh keys injected into the OCP workers and compute nodes
+cifmw_ssh_keytype: ecdsa
+cifmw_ssh_keysize: 521
```
fultonj, line 165: Please add the following:

```
cifmw_ceph_daemons_layout:
  rgw_enabled: true
  dashboard_enabled: false
  cephfs_enabled: true
```
Because of https://github.com/openstack-k8s-operators/ci-framework/pull/1208 

You need these if you're going to be testing Manila and Object Storage. If you don't intend to test them, they will be disabled by default.

scenarios/reproducers/networking-definition-ceph.yml
fultonj, line 1: Why do we need this? Can't we use the exact same values that va1 is using? I would think a lot of this is duplicate code.

scenarios/reproducers/external-ceph.yml
fultonj, line 1: Why do we need this? Can't we use the exact same values that va1 is using? I would think a lot of this is duplicate code.

I understand that we want to reference a different scenario `external-ceph.yml` when we run ci-framework to indicate that we're deploying external ceph and not va1 (or hci) but should we find a way to not have all of the same variables duplicated?

roles/reproducer/tasks/libvirt_layout.yml
```diff
@@ -91,7 +91,7 @@
 # match the multi-hypervisor schema.
 - name: Configure compute network
   when:
-    - compute.key is match('^compute.*')
+    - compute.key is match('^(compute|ceph).*')
```
fultonj, line 94: I understand why this is done. Because compute and ceph refer to different nodes in the inventory. So I'm good with this chagne.

roles/ci_gen_kustomize_values/templates/hci/ceph-values/values.yaml.j2
pablintino, line 1: I'd prefer to avoid adding a new type of config map just because the type of node is call ceph instead of compute. Let me propose a PR to allow us to reuse the existing one.

scenarios/reproducers/external-ceph.yml
fultonj: FYI https://github.com/openstack-k8s-operators/ci-framework/pull/1216

roles/reproducer/tasks/libvirt_layout.yml
pablintino: Yeah, anyway, one thing that grind my gears is that we are basing these kind of decisions based on the name of the instance, instead of relaying on the groups we already have.

playbooks/deploy-external-ceph.yml
```diff
@@ -0,0 +1,253 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+- name: Pre-requisites for deploying external ceph
+  hosts: localhost
+  vars:
+    cifmw_ceph_dir: "{{ ansible_user_dir }}/ci-framework-data/manifests/ceph"
+    cifmw_ceph_node_num: "{{ cifmw_ceph_node_num | default(3) }}"
```
pablintino, line 21: Why not base the amount of ceph nodes in the size of a group called, ie, ceph, that it's part of the inventory used to deploy the external ceph?
Anyway, if this use-case is done with a DT I'm not sure why we need such of a playbook like this one.

roles/ci_gen_kustomize_values/templates/hci/ceph-values/values.yaml.j2
```diff
@@ -0,0 +1,43 @@
+data:
+  ssh_keys:
```
fultonj, line 2: Do we need this in order to ensure the EDPM nodes have SSH?

roles/ci_gen_kustomize_values/templates/hci/ceph-values/values.yaml.j2
```diff
@@ -0,0 +1,43 @@
...
+    authorized: {{ cifmw_ci_gen_kustomize_values_ssh_authorizedkeys | b64encode }}
+    private: {{ cifmw_ci_gen_kustomize_values_ssh_private_key | b64encode }}
+    public: {{ cifmw_ci_gen_kustomize_values_ssh_public_key | b64encode }}
+  nova:
```
fultonj, line 6: If this is just for Ceph nodes, then we don't need a live migration SSH key

roles/reproducer/tasks/libvirt_layout.yml
fultonj: OK, can we update this patch to do that. E.g. do we have a group name to refer to those nodes?

rachael-george:
Depends on #1224

scenarios/reproducers/networking-definition-ceph.yml
```diff
@@ -0,0 +1,132 @@
...
+        range:
+          start: 100
+          length: 21
+      networks:
+        ctlplane: {}
+        internalapi: {}
+        tenant: {}
+        storage: {}
+        storagemgmt: {}
+    cephs:
```
fultonj, line 117: Instead of introducing `networking-definition-ceph.yml`, what would happen if you just added the `cephs` definition to the existing `networking-definition.yml` (since that is the only way they differ [1])? 

If the presence of the `cephs` structure doesn't break VA1, I think it would be better to let it have a single networking definition with an item that is not used so there is less reproduced content.

Also, if `cephs` is just for nodes which host ceph then, please you could remove `internalapi` and `tentant` network.

[1] 
```
$ diff -u ../ci-framework/scenarios/reproducers/networking-definition.yml networking-definition-ceph.yml
--- ../ci-framework/scenarios/reproducers/networking-definition.yml	2024-03-04 11:41:06.757530796 -0500
+++ networking-definition-ceph.yml	2024-03-08 07:55:24.749427209 -0500
@@ -114,6 +114,17 @@
         tenant: {}
         storage: {}
         storagemgmt: {}
+    cephs:
+      network-template:
+        range:
+          start: 150
+          length: 21
+      networks:
+        ctlplane: {}
+        internalapi: {}
+        tenant: {}
+        storage: {}
+        storagemgmt: {}
   instances:
     controller-0:
       networks:
```

scenarios/reproducers/external-ceph.yml
fultonj: `validated-architecture-1.yml` and `external-ceph.yml` differ only in the `cifmw_libvirt_manager_configuration` variable in that the one in `external-ceph.yml` contains  `vms.ceph` [1].

AFAICT these are just variable files which are passed to ansible with `-e @_file_name`. In theory you could introduce a common file which they both use and only make them differ by having each `cifmw_libvirt_manager_configuration` in a separate file. That would allow us to inherit bug fixes, keep consistent, and avoid duplication as per the DRY principle.

I'm not saying it has to be this way, but I think that would be cleaner so I want to leave this comment. 

I'll let other ci-framework maintainers (@pablintino @cjeanner) comment on if they want it though.

[1]
```
$ diff ../ci-framework/scenarios/reproducers/validated-architecture-1.yml external-ceph.yml 
--- ../ci-framework/scenarios/reproducers/validated-architecture-1.yml	2024-03-04 11:45:02.003026965 -0500
+++ external-ceph.yml	2024-03-08 07:55:56.077627251 -0500
@@ -94,6 +94,23 @@
       nets:
         - ocpbm
         - osp_trunk
+    ceph:
+      root_part_id: >-
+        {{
+          (cifmw_repo_setup_os_release is defined and cifmw_repo_setup_os_release == 'rhel') |
+          ternary(4, 1)
+        }}
+      amount: 3
+      disk_file_name: "base-os.qcow2"
+      disksize: 50
+      memory: 8
+      cpus: 4
+      nets:
+        - ocpbm
+        - osp_trunk
     controller:
       root_part_id: >-
         {{
@@ -147,8 +164,7 @@
 cifmw_ssh_keytype: ecdsa
 cifmw_ssh_keysize: 521
 
-# Test Ceph file and object storage (block is enabled by default)
 cifmw_ceph_daemons_layout:
   rgw_enabled: true
   dashboard_enabled: false
-  cephfs_enabled: true
+  cephfs_enabled: true
\ No newline at end of file
$
```

fultonj (commented):
Thanks @rachael-george I think this is getting there. I have two ideas for improvement that I'll comment on below.

Let me know what you think in this review and we can let a core from the ci-framework team decide if more changes are needed.

scenarios/reproducers/networking-definition-ceph.yml
pablintino: +1, that will work. If the Networking Definition is used in a deployment without one of the groups been a group in the Ansible inventory nothing will break.

scenarios/reproducers/external-ceph.yml
pablintino: +1 to that that suggestion.

roles/reproducer/tasks/libvirt_layout.yml
pablintino, line 1: Wasn't easier to update the regex?

scenarios/reproducers/external-ceph.yml
cjeanner: Limiting the copy/pasting is a must, else we'll lose time finding why things are working. Adding a new layout file with only the needed overrides is better, really.

roles/reproducer/tasks/libvirt_layout.yml
rachael-george: Something like this?

scenarios/reproducers/external-ceph.yml
rachael-george: Removed the duplicated content from the `external-ceph.yml` file. I haven't made changes to the original validated-architecture file, just in case I break something there

scenarios/reproducers/external-ceph.yml
```diff
@@ -0,0 +1,79 @@
+---
+cifmw_use_libvirt: true
```
cjeanner, line 2: I think we can remove that `cifmw_use_libvirt: true` - it's already present in the "parent" parameter file.

roles/reproducer/tasks/libvirt_layout.yml
```diff
@@ -91,7 +91,7 @@
 # match the multi-hypervisor schema.
 - name: Configure compute network
   when:
-    - compute.key is match('^compute.*')
+    - "compute.key in groups['computes'] or compute.key in (groups['cephs'] | default([]))"
```
cjeanner, line 94: my only issue here is about the way to match the hosts - there's an assumption of names being, really "compute" and "ceph" in the libvirt_manager_configuration.vms
For now, it's OK, because we kind of know and agree on that. But maybe we should get a better way - later, once we hit actual issue.

cjeanner:
/approve

we can refine if needed.

rebtoor (approved):

PR #1208, opened 2024-03-01
Reduce ceph resources to support single node jobs

katarimanojk:
This patch limits the default set of ceph services to be run to avoid issues in single node setup.

Changes made in this patch are
- Dashbaord service will be disabled by default and can be enabled by updating the parameter 'cifmw_cephadm_dashboard_enabled'

- ingress service will be deployed if the edpm nodes > 1

  - [x] README in the role

### DIFF
--- a/playbooks/ceph.yml
+++ b/playbooks/ceph.yml
@@ -266,6 +266,7 @@
         tasks_from: pools
 
     - name: Deploy RGW
+      when: cifmw_ceph_daemons_layout.rgw_enabled | default(true) | bool
       ansible.builtin.import_role:
         name: cifmw_cephadm
         tasks_from: rgw
@@ -274,6 +275,7 @@
         cifmw_cephadm_rgw_vip: "{{ cifmw_cephadm_vip }}/24"
 
     - name: Configure Monitoring Stack
+      when: cifmw_ceph_daemons_layout.dashboard_enabled  | default(false) | bool
       ansible.builtin.import_role:
         name: cifmw_cephadm
         tasks_from: monitoring
@@ -282,8 +284,8 @@
         cifmw_cephadm_dashboard_crt: "{{ cifmw_cephadm_certificate }}"
         cifmw_cephadm_dashboard_key: "{{ cifmw_cephadm_key }}"
 
-
     - name: Create cephfs volume
+      when: cifmw_ceph_daemons_layout.cephfs_enabled  | default(true) | bool
       ansible.builtin.import_role:
         name: cifmw_cephadm
         tasks_from: cephfs
--- a/roles/cifmw_cephadm/templates/ceph_rgw.yml.j2
+++ b/roles/cifmw_cephadm/templates/ceph_rgw.yml.j2
@@ -19,19 +19,21 @@ spec:
     {{ rgw_frontend_cert | indent( width=4 ) }}
 {% endif %}
 ---
-service_type: ingress
-service_id: rgw.default
-service_name: ingress.rgw.default
-placement:
-  count: 1
-spec:
-  backend_service: rgw.rgw
-  frontend_port: 8080
-  monitor_port: 8999
-  virtual_ip: {{ cifmw_cephadm_rgw_vip }}
-  virtual_interface_networks:
-  - {{ cifmw_cephadm_rgw_network }}
-{% if rgw_frontend_cert is defined %}
-  ssl_cert: |
-    {{ rgw_frontend_cert | indent( width=4 ) }}
+{% if _hosts|length > 1 %}
+  service_type: ingress
+  service_id: rgw.default
+  service_name: ingress.rgw.default
+  placement:
+    count: 1
+  spec:
+    backend_service: rgw.rgw
+    frontend_port: 8080
+    monitor_port: 8999
+    virtual_ip: {{ cifmw_cephadm_rgw_vip }}
+    virtual_interface_networks:
+    - {{ cifmw_cephadm_rgw_network }}
+  {% if rgw_frontend_cert is defined %}
+    ssl_cert: |
+      {{ rgw_frontend_cert | indent( width=4 ) }}
+  {% endif %}
 {% endif %}


katarimanojk:
@fmount FYI

roles/cifmw_cephadm/tasks/monitoring.yml
```diff
@@ -14,40 +14,44 @@
...
-    mode: '0644'
-    force: true
-
-- name: Config ssl cert(s) and key(s) for the exposed components
-  become: true
-  when: cifmw_cephadm_dashboard_crt | length > 0 and
-        cifmw_cephadm_dashboard_key | length > 0
+- name: Build and apply the monitoring stack daemons
+  when:
+    - cifmw_cephadm_dashboard_enabled | bool
```
fmount, line 19: I'm wondering if, instead of changing the code in the role itself, it's better to have a variable in the ceph playbook [1] and skip the dashboard entirely.

[1] https://github.com/openstack-k8s-operators/ci-framework/blob/main/playbooks/ceph.yml#L276

fmount (changes requested):

fmount:
Can you point out in the description that this is related to Jira: https://issues.redhat.com/browse/OSPRH-5300

roles/cifmw_cephadm/tasks/monitoring.yml
cjeanner: :+1: would be better indeed

cjeanner:
/ok-to-test

@katarimanojk will be part of the group, probably the last manual ok-to-test to date :).

roles/cifmw_cephadm/tasks/monitoring.yml
katarimanojk: yes we can do that but we should also update the ceph_overrides in the documeantion: https://ci-framework.pages.redhat.com/docs/ci-framework/05_deploy_va.html#ceph

playbooks/ceph.yml
```diff
@@ -266,6 +266,7 @@
         tasks_from: pools
 
     - name: Deploy RGW
+      when: cifmw_ceph_daemons_layout.rgw_enabled | bool
```
fultonj, line 269: `cifmw_ceph_daemons_layout.rgw_enabled | default(false) | bool`

and same pattern for others

katarimanojk: Thank john, updated it in the latest patch.

playbooks/ceph.yml
```diff
@@ -274,6 +275,7 @@
         cifmw_cephadm_rgw_vip: "{{ cifmw_cephadm_vip }}/24"
 
     - name: Configure Monitoring Stack
+      when: cifmw_ceph_daemons_layout.dashboard_enabled  | default(false) | bool
```
fmount, line 278: +1

playbooks/ceph.yml
```diff
@@ -282,8 +284,8 @@
         cifmw_cephadm_dashboard_crt: "{{ cifmw_cephadm_certificate }}"
         cifmw_cephadm_dashboard_key: "{{ cifmw_cephadm_key }}"
 
-
     - name: Create cephfs volume
+      when: cifmw_ceph_daemons_layout.cephfs_enabled  | default(false) | bool
```
fmount, line 288: deploy by default: in case of issues we can change this behavior or override for single node

roles/cifmw_cephadm/templates/ceph_rgw.yml.j2
```diff
@@ -19,19 +19,21 @@ spec:
...
-  backend_service: rgw.rgw
-  frontend_port: 8080
-  monitor_port: 8999
-  virtual_ip: {{ cifmw_cephadm_rgw_vip }}
-  virtual_interface_networks:
-  - {{ cifmw_cephadm_rgw_network }}
-{% if rgw_frontend_cert is defined %}
-  ssl_cert: |
-    {{ rgw_frontend_cert | indent( width=4 ) }}
+{% if _hosts|length > 1 %}
```
fmount, line 22: +1

playbooks/ceph.yml
katarimanojk: i will make it as defualt(true) for cephfs also

katarimanojk:
updated documentation here: https://gitlab.cee.redhat.com/ci-framework/docs/-/merge_requests/105

fmount:
This should be good to go. @katarimanojk did you test the override using a struct like:

```yaml
cifmw_ceph_daemons_layout:
  rgw_enabled: true
  dashboard_enabled: false
  cephfs_enabled: false
```
to see how the behavior changes in a single node deployment?
Also, @fultonj do you think we should change the "default" to `false` and always use this data structure to deploy more than `rbd`?  I'm tempted to move everything to false and drive everything via overrides.

playbooks/ceph.yml
```diff
@@ -282,8 +284,8 @@
         cifmw_cephadm_dashboard_crt: "{{ cifmw_cephadm_certificate }}"
         cifmw_cephadm_dashboard_key: "{{ cifmw_cephadm_key }}"
 
-
     - name: Create cephfs volume
+      when: cifmw_ceph_daemons_layout.cephfs_enabled  | default(true) | bool
```
fmount, line 288: I'm sorry to go back here, but I think we should move all these defaults to `false` (as John suggested at the beginning). We deploy rbd only unless something else is specified via overrides.
@fultonj @katarimanojk is that a sane approach?

katarimanojk:
> This should be good to go. @katarimanojk did you test the override using a struct like:
> 
> ```yaml
> cifmw_ceph_daemons_layout:
>   rgw_enabled: true
>   dashboard_enabled: false
>   cephfs_enabled: false
> ```
> 
> to see how the behavior changes in a single node deployment? Also, @fultonj do you think we should change the "default" to `false` and always use this data structure to deploy more than `rbd`? I'm tempted to move everything to false and drive everything via overrides.

@fmount i tested the override struct in my 3-node VA1 deployment and saw the tasks were skipped accordingly.

fmount:
> > This should be good to go. @katarimanojk did you test the override using a struct like:
> > ```yaml
> > cifmw_ceph_daemons_layout:
> >   rgw_enabled: true
> >   dashboard_enabled: false
> >   cephfs_enabled: false
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > to see how the behavior changes in a single node deployment? Also, @fultonj do you think we should change the "default" to `false` and always use this data structure to deploy more than `rbd`? I'm tempted to move everything to false and drive everything via overrides.
> 
> @fmount i tested the override struct in my 3-node VA1 deployment and saw the tasks were skipped accordingly.

ack as per our conversation this patch is good to go.

fmount (approved):
+2 from me

fultonj (commented):
+2

fultonj (approved):
/lgtm

PR #1215, opened 2024-03-04
Copy yum.repos.d directory to logs

raukadah:
In EDPM image build and container build jobs, we are generating repos in /etc/yum.repos.d directory. If these jobs breaks, Devs looks for repos in the logs which does not get collected. It makes debugging hard.

Copying the repos in the logs will make the life easier.

### DIFF
--- a/roles/artifacts/tasks/environment.yml
+++ b/roles/artifacts/tasks/environment.yml
@@ -19,6 +19,7 @@
     cmd: |-
       cp /etc/resolv.conf /etc/hosts {{ cifmw_artifacts_basedir }}/artifacts/
       cp -r /etc/NetworkManager/system-connections {{ cifmw_artifacts_basedir }}/artifacts/NetworkManager
+      cp -r /etc/yum.repos.d {{ cifmw_artifacts_basedir }}/artifacts/yum_repos
       test -d /etc/ci/env && cp -r /etc/ci/env {{ cifmw_artifacts_basedir }}/artifacts/ci-env
       test -d /var/log/bmaas_console_logs && cp -r /var/log/bmaas_console_logs {{ cifmw_artifacts_basedir }}/logs
       ip ro ls > {{ cifmw_artifacts_basedir }}/artifacts/ip-network.txt


roles/artifacts/tasks/environment.yml
```diff
@@ -24,3 +24,29 @@
       ip ro ls > {{ cifmw_artifacts_basedir }}/artifacts/ip-network.txt
       ip rule ls >> {{ cifmw_artifacts_basedir }}/artifacts/ip-network.txt
       ip -j -p link ls >> {{ cifmw_artifacts_basedir }}/artifacts/ip-network.txt
+
+- name: Find yum repos files
+  become: true
```
pablintino, line 29: Do we need sudo to read and search? I don't think so.

roles/artifacts/tasks/environment.yml
```diff
@@ -24,3 +24,29 @@
...
+    paths: "/etc/yum.repos.d/"
+    patterns: "*.repo"
+    recurse: false
+  register: _repo_files
+
+- name: Create repos directory
+  ansible.builtin.file:
+    path: "{{ cifmw_artifacts_basedir }}/logs/yum_repos"
+    state: directory
+    mode: "0777"
```
pablintino, line 40: I don't we need to set 777. It will work even without this, mainly because the user and group are set just fine.

roles/artifacts/tasks/environment.yml
```diff
@@ -24,3 +24,29 @@
...
+    owner: "{{ ansible_user_id }}"
+    group: "{{ ansible_user_id }}"
+
+- name: Copy repos to /tmp
```
pablintino, line 44: ditto. I don't think we need to become or set 777 as mode.

roles/artifacts/tasks/environment.yml
```diff
@@ -24,3 +24,29 @@
       ip ro ls > {{ cifmw_artifacts_basedir }}/artifacts/ip-network.txt
       ip rule ls >> {{ cifmw_artifacts_basedir }}/artifacts/ip-network.txt
       ip -j -p link ls >> {{ cifmw_artifacts_basedir }}/artifacts/ip-network.txt
+
+- name: Find yum repos files
```
cjeanner, line 28: Wondering why you want to find and loop on the files instead of copying the directory directly? Care to explain a bit? Why won't a simple `cp -r /etc/yum.repos.d {{ cifmw_architect_basedir }}/artifacts/yum_repos` in the above `ansible.builtin.shell` be sufficient?

raukadah: Sorry, I thought too much on this task not sure why. Yes, simple cp can do the job. Thank you @pablintino @cjeanner :-)

roles/artifacts/tasks/environment.yml
```diff
@@ -19,6 +19,7 @@
     cmd: |-
       cp /etc/resolv.conf /etc/hosts {{ cifmw_artifacts_basedir }}/artifacts/
       cp -r /etc/NetworkManager/system-connections {{ cifmw_artifacts_basedir }}/artifacts/NetworkManager
+      cp -r /etc/yum.repos.d {{ cifmw_architect_basedir }}/artifacts/yum_repos
```
cjeanner, line 22: As a side note, by default we generate the repositories in cifmw_basedir/artifacts/repositories:
https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/repo_setup/defaults/main.yml#L31

Having a distinct name is good imho.

PR #1218, opened 2024-03-04
net-mapper: is_trunk_parent, trunk_parent, routers

hjensas:
Adds the new parameters used by ci-bootstrap to control instance trunk port creation in:
  https://github.com/openstack-k8s-operators/ci-bootstrap/pull/33

It also adds the routers definition which enable creating one or more routers, for example to simulate DCN/Spine-and-Leaf architecture.

Related: https://issues.redhat.com/browse/OSPRH-5089
Partial: https://issues.redhat.com/browse/OSPRH-3376
Partial: https://issues.redhat.com/browse/OSPRH-2994

  - [x] README in the role

### DIFF
--- a/plugins/module_utils/net_map/networking_definition.py
+++ b/plugins/module_utils/net_map/networking_definition.py
@@ -129,6 +129,39 @@ def _validate_parse_field_type(
     return raw_value
 
 
+def _validate_parse_trunk_parent_field(
+    field_name: str,
+    raw_definition: typing.Dict[str, typing.Any],
+    parent_name: str = None,
+    parent_type: str = None,
+    trunk_parents: set = None,
+) -> typing.Any:
+    if trunk_parents is None:
+        trunk_parents = set()
+
+    trunk_parent = _validate_parse_field_type(
+        field_name,
+        raw_definition,
+        expected_type=str,
+        parent_name=parent_name,
+        parent_type=parent_type,
+        mandatory=False,
+    )
+    if trunk_parent and trunk_parent not in trunk_parents:
+        raise exceptions.NetworkMappingTrunkParentValidationError(
+            message=(
+                f"trunk parent '{trunk_parent}' does not exist in "
+                f"'{parent_type}' '{parent_name}'"
+            ),
+            field=field_name,
+            invalid_value=trunk_parent,
+            parent_name=parent_name,
+            parent_type=parent_type,
+        )
+
+    return trunk_parent
+
+
 def _validate_fields_one_of(
     fields_list: typing.List[str],
     raw_definition: typing.Dict[str, typing.Any],
@@ -1008,6 +1041,106 @@ class NetconfigNetworkDefinition(SubnetBasedNetworkToolDefinition):
         super().__init__(network, raw_config, self.__OBJECT_NAME)
 
 
+class RouterDefinition:
+    """Parser and validator of a signle router config element
+
+    Handles the parsing and validation of a router item in the
+    dict of routers of the Networking Definition.
+
+    The network definition that this class parses should follow
+    this format:
+
+    .. code-block:: YAML
+
+       <router-name>:
+         external_network: <external gateway network: optional>
+         networks:  <List of networks attached to the router>
+           - <Network 1>
+
+    Raises:
+        ValueError: If name it's not provided.
+        exceptions.NetworkMappingValidationError: When the format
+            of the provided instance is not correct.
+    """
+
+    __OBJECT_NAME = "router"
+
+    __FIELD_EXTERNAL_NETWORK = "external_network"
+    __FIELD_NETWORKS = "networks"
+
+    def __init__(self, name: str, raw_definition: typing.Dict[str, typing.Any]) -> None:
+        """Initializes a RouterDefinition from dict with its parameters
+
+        Args:
+            name: The name of router.
+            raw_definitions: The dictionary that contains the configuration
+                of the router
+
+        Raises:
+            ValueError: If name is not provided.
+        """
+        if not name:
+            raise ValueError("name is a mandatory argument")
+
+        self.__name = name
+        self.__external_network = None
+        self.__networks = []
+        self.__parse_raw(raw_definition)
+
+    @property
+    def name(self) -> str:
+        """Name of the network."""
+        return self.__name
+
+    @property
+    def external_network(self) -> typing.Optional[str]:
+        """External gateway network"""
+        return self.__external_network
+
+    @property
+    def networks(self) -> typing.List[str]:
+        """List of networks to attach to the router"""
+        return self.__networks
+
+    def __parse_raw(self, raw_definition: typing.Dict[str, typing.Any]):
+        self.__external_network = _validate_parse_field_type(
+            self.__FIELD_EXTERNAL_NETWORK,
+            raw_definition,
+            str,
+            parent_name=self.__name,
+            parent_type=self.__OBJECT_NAME,
+            mandatory=False,
+        )
+
+        self.__networks = _validate_parse_field_type(
+            self.__FIELD_NETWORKS,
+            raw_definition,
+            list,
+            parent_name=self.name,
+            parent_type=self.__OBJECT_NAME,
+            mandatory=True,
+        )
+
+    def __hash__(self) -> int:
+        return hash(
+            (
+                self.__name,
+                self.__external_network,
+                frozenset(self.__networks),
+            )
+        )
+
+    def __eq__(self, other) -> bool:
+        if not isinstance(other, RouterDefinition):
+            return False
+
+        return (
+            self.__external_network == other.__external_network
+            and self.__networks == other.__networks
+            and self.__name == other.__name
+        )
+
+
 class NetworkDefinition:
     """Parser and validator of a single network config element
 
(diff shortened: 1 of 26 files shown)


plugins/module_utils/net_map/networking_env_definitions.py
```diff
@@ -186,6 +188,20 @@ class MappedNetwork:
     mtu: typing.Optional[int] = None
 
 
+@dataclasses.dataclass(frozen=True)
+class MappedRouter:
+    """Defines all the settings for a router
+    Attributes:
+        router_name: The name of the router
+        networks: Networks attached to the
```
pablintino, line 196: Typo.

scripts/run_ansible_test
pablintino, line 1: This change should be local only :)

plugins/module_utils/net_map/networking_mapper.py
```diff
@@ -570,6 +619,33 @@ def __build_network_tool_common(
         return tool_type(*args_list)
 
 
+class NetworkingRoutersMapper:
+    """
+    Handles the mapping of Networking Definition routers section
+    """
+
+    def __init__(self):
+        pass
```
pablintino, line 628: Not needed.

plugins/module_utils/net_map/networking_mapper.py
```diff
@@ -570,6 +619,33 @@ def __build_network_tool_common(
...
+    def map_routers(
+        self,
+        router_definitions: typing.Dict[str, networking_definition.RouterDefinition],
+    ) -> typing.Dict[str, networking_env_definitions.MappedRouter]:
+        return {
+            router_name: self.__map_single_router(router_name, router_def)
+            for router_name, router_def in router_definitions.items()
+        }
+
+    def __map_single_router(
```
pablintino, line 639: If this one is not accessing anything from the instance it can be a static or class method.

plugins/module_utils/net_map/networking_definition.py
```diff
@@ -1008,6 +1008,86 @@ def __init__(
...
+            self.__external_network = raw_definition[self.__FIELD_EXTERNAL_NETWORK]
+
+        self.__networks = raw_definition[self.__FIELD_NETWORKS]
+
+    def __hash__(self) -> int:
+        return hash(
+            (
+                self.__name,
+                self.__external_network,
+                self.__networks,
```
pablintino, line 1076: I'm sure this is not tested, cause a list cannot be hashed in python.

plugins/module_utils/net_map/networking_definition.py
```diff
@@ -1008,6 +1008,86 @@ def __init__(
...
+    @property
+    def networks(self) -> typing.List[str]:
+        """List of networks to attach to the router"""
+        return self.__networks
+
+    def __parse_raw(self, raw_definition: typing.Dict[str, typing.Any]):
+        if self.__FIELD_EXTERNAL_NETWORK in raw_definition:
+            self.__external_network = raw_definition[self.__FIELD_EXTERNAL_NETWORK]
+
+        self.__networks = raw_definition[self.__FIELD_NETWORKS]
```
pablintino, line 1069: I'd try to validate the type to ensure it's a list, as done everywhere. https://github.com/openstack-k8s-operators/ci-framework/blob/main/plugins/module_utils/net_map/networking_definition.py#L261-L267.
A UT, like this one will be needed: https://github.com/openstack-k8s-operators/ci-framework/blob/main/tests/unit/module_utils/net_map/test_networking_definitions_network.py#L601-L615
That's one of the strengths of the mapper, the input is validated and mapped to give a consistent output, in terms of fields and their types.
If you address this you need to update the docstring like this: https://github.com/openstack-k8s-operators/ci-framework/blob/main/plugins/module_utils/net_map/networking_definition.py#L1797-L1798

plugins/module_utils/net_map/networking_definition.py
```diff
@@ -1008,6 +1008,86 @@ def __init__(
...
+        return self.__external_network
+
+    @property
+    def networks(self) -> typing.List[str]:
+        """List of networks to attach to the router"""
+        return self.__networks
+
+    def __parse_raw(self, raw_definition: typing.Dict[str, typing.Any]):
+        if self.__FIELD_EXTERNAL_NETWORK in raw_definition:
+            self.__external_network = raw_definition[self.__FIELD_EXTERNAL_NETWORK]
```
pablintino, line 1067: ditto. Same as the other field.

plugins/module_utils/net_map/networking_definition.py
pablintino, line 1069: If `__FIELD_NETWORKS` is not present we emit a nice error that is properly passed all along Ansible. I suggest you to have a look at the previous comment, as the function suggested there already takes care of this if you tell it that the field is mandatory.

hjensas: Done

plugins/module_utils/net_map/networking_definition.py
hjensas: Done

plugins/module_utils/net_map/networking_definition.py
hjensas: Done.

plugins/module_utils/net_map/networking_definition.py
hjensas: Done, `frozenset(self.__networks)`

plugins/module_utils/net_map/networking_definition.py
```diff
@@ -1008,6 +1008,107 @@ def __init__(
...
+        """External gateway network"""
+        return self.__external_network
+
+    @property
+    def networks(self) -> typing.List[str]:
+        """List of networks to attach to the router"""
+        return self.__networks
+
+    def __parse_raw(self, raw_definition: typing.Dict[str, typing.Any]):
+        if self.__FIELD_EXTERNAL_NETWORK in raw_definition:
```
pablintino, line 1073: This check is not needed as the `_validate_parse_field_type` function, when `mandatory=False` and the field is not present returns just `None`.

plugins/module_utils/net_map/networking_definition.py
```diff
@@ -1654,13 +1765,27 @@ def __parse_raw_net(
...
+            if self.__FIELD_IS_TRUNK_PARENT in templated_net_data
+            else None
+        )
+
+        trunk_parent = (
+            str(templated_net_data[self.__FIELD_TRUNK_PARENT])
+            if self.__FIELD_TRUNK_PARENT in templated_net_data
+            else None
+        )
+
```
pablintino, line 1779: I'd suggest doing something similar to what's done here: https://github.com/openstack-k8s-operators/ci-framework/pull/1218/files#diff-35c36b377323c97ce9082f04b63f3dd52f3f32c572cc593a384a393659f1f680R1083
Anyway, may make sense to check if the `trunk_parent` exists, isn't it?

hjensas: I've added validation to check that `trunk_parent` exists. The validation also adds a boundary, the trunk parent port and sub port must be in the same group template.

plugins/module_utils/net_map/networking_definition.py
```diff
@@ -129,6 +129,36 @@ def _validate_parse_field_type(
     return raw_value
 
 
+def _validate_parse_trunk_parent_field(
+    field_name: str,
+    raw_definition: typing.Dict[str, typing.Any],
+    parent_name: str = None,
+    parent_type: str = None,
+    trunk_parents: set = set(),
```
pablintino, line 137: This is a Python trap. `set`, a mutable object, when used as default will make that every next calls to the function with no `trunk_parent` passed will default to the previous set instance. With other words, default mutable values acts as "globals". https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

plugins/module_utils/net_map/networking_definition.py
```diff
@@ -129,6 +129,36 @@ def _validate_parse_field_type(
...
+    trunk_parent = _validate_parse_field_type(
+        field_name,
+        raw_definition,
+        expected_type=str,
+        parent_name=parent_name,
+        parent_type=parent_type,
+        mandatory=False,
+    )
+    if trunk_parent and trunk_parent not in trunk_parents:
+        raise exceptions.NetworkMappingTrunkParentValidationError(
```
pablintino, line 148: Do we have a test that checks this condition is working?

plugins/module_utils/net_map/networking_definition.py
pablintino: Sounds good.

plugins/module_utils/net_map/networking_definition.py
hjensas: Good catch.
I changed it to default None, and then initialize a set if it is None.

plugins/module_utils/net_map/networking_definition.py
hjensas: In almost all the tests based on the stubs we would evaluate `trunk_parent` to `None`.
I added two more tests in `test_networking_definitions_group_template.py` and `test_networking_definitions_instance.py`. 

I think that covers it?

pablintino: That's more than enough, thanks!

cjeanner:
/lgtm

I base my vote on the availability of new unit tests matching the new features, with failure and success scenarios. We should be covered for the current behavior as well as future changes/updates to the code.

My only concern is: does current scenarios/reproducers/networking-definition.yml file need any update to match the new feature? I think it doesn't, since the new parameters seem to default to `None` if not set.

raukadah (approved):
Tests are green, if something breaks we can fix it.

PR #1219, opened 2024-03-04
Create empty file for dnsmasq.conf file

maximsava12:
  - [ ] README in the role

### DIFF
--- a/roles/reproducer/tasks/configure_crc.yml
+++ b/roles/reproducer/tasks/configure_crc.yml
@@ -32,6 +32,19 @@
         dns4_ignore_auto: true
         state: present
 
+    - name: Create dir for dnsmasq.conf
+      ansible.builtin.file:
+        path: "/var/srv"
+        state: directory
+        mode: '0755'
+
+    - name: Create dnsmasq.conf file
+      become: true
+      ansible.builtin.file:
+        path: "/var/srv/dnsmasq.conf"
+        state: touch
+        mode: '0644'
+
     - name: Check which dnsmasq config we must edit
       register: _dnsmasq
       ansible.builtin.stat:


lewisdenny:
Hi @maximsava12 

I'm not sure you want `c3e9bf6b20ca5bc57edcf4f9104e30770508e11d` in this PR, please squash the other two commits. 

Also please explain in the commit message/description why this change is required, what issue are you seeing?

pablintino:
> Hi @maximsava12
> 
> I'm not sure you want `c3e9bf6b20ca5bc57edcf4f9104e30770508e11d` in this PR, please squash the other two commits.
> 
> Also please explain in the commit message/description why this change is required, what issue are you seeing?

Hi @maximsava12, as @lewisdenny said, we need to understand why is this needed and in which circumstances this missing file is an issue.
Anyway, I'm not sure where that path comes from, cause the two (yes, crc uses two paths, depending on its version, as not that old versions used a dnsmasq container with a config file in a non-default path) we know about are: https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci/playbooks/multinode-customizations.yml#L74

roles/reproducer/tasks/configure_crc.yml
```diff
@@ -32,6 +32,19 @@
         dns4_ignore_auto: true
         state: present
 
+    - name: Create dir for dnsmasq.conf
+      ansible.builtin.file:
```
cjeanner, line 36: it's supposed to be a symlink pointing to /srv - do we really need this one? Care to provide some more context about the issue you're trying to solve?

maximsava12:
Hi,
The context is 3-node Deployment failed with missing dnsmasq.conf file on stable branch.

cjeanner:
@maximsava12 ok, so if consuming `stable` branch, please ensure you're using "older" CRC release (2.30 IIRC) so that it will follow the right path.

Or consume `main` branch instead - there's a partch in order to ensure dnsmasq is properly configured. The one you're proposing here will just blow the whole thing in `main`.

PR #1223, opened 2024-03-05
Set cifmw_networking_definition for ci-bootstap staging

hjensas:
Set cifmw_networking_definition for ci-bootstap staging

Use the new ìs_trunk_parent and trunk_parent hints for ci-bootstrap
to understand how to attach networks.

Ref:
 https://github.com/openstack-k8s-operators/ci-framework/pull/1218
 https://github.com/openstack-k8s-operators/ci-bootstrap/pull/33

### DIFF
--- a/zuul.d/edpm_multinode.yaml
+++ b/zuul.d/edpm_multinode.yaml
@@ -283,6 +283,68 @@
     name: podified-multinode-edpm-deployment-crc-bootstrap-staging
     parent: cifmw-podified-multinode-edpm-ci-bootstrap-staging
     extra-vars: *edpm_bootstrap_extra_vars
-    vars: *edpm_bootstrap_vars
+    vars:
+      cifmw_networking_definition:
+        networks:
+          default:
+            network: "192.168.122.0/24"
+            gateway: "192.168.122.1"
+            mtu: 1500
+          internal-api:
+            network: "172.17.0.0/24"
+            gateway: "172.17.0.1"
+            vlan: 20
+          storage:
+            network: "172.18.0.0/24"
+            gateway: "172.18.0.1"
+            vlan: 21
+          tenant:
+            network: "172.19.0.0/24"
+            gateway: "172.19.0.1"
+            vlan: 22
+        routers:
+          ci-router:
+            external_network: public
+            networks:
+              - default
+        group-templates:
+          computes:
+            network-template:
+              range:
+                start: 100
+                length: 21
+            networks:
+              default:
+                is_trunk_parent: true
+              internal-api:
+                trunk_parent: default
+                skip-nm-configuration: true
+              tenant:
+                trunk_parent: default
+                skip-nm-configuration: true
+              storage:
+                trunk_parent: default
+                skip-nm-configuration: true
+        instances:
+          controller:
+            networks:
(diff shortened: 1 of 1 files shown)


zuul.d/edpm_multinode.yaml
```diff
@@ -283,6 +283,67 @@
...
+          ci-router:
+            external_network: public
+            networks:
+              - default
+        instances:
+          controller:
+            networks:
+              default:
+                ip: "192.168.122.11"
+          crc:
```
cjeanner, line 315: I think we can omit the `crc` and `compute-*` network, provided we use a group/template (@pablintino care to confirm?)

hjensas: Where is the group template defined? If that is defined elsewhere we need to update that with the `is_trunk_parent` and `trunk_parent` definitions?

cjeanner: For instance:
https://github.com/openstack-k8s-operators/ci-framework/blob/main/scenarios/reproducers/networking-definition.yml#L95-L121

hjensas: If that is used in CI it should be updated, for the reproducer in libvirt I guess we really only support the 1 NIC + VLANs for now. We may still want to update that one in the reproducer to indicate that one NIC with VLANs is whats deployed.

In ci-bootstrap staging branch we now have support to set up multiple NICs and to use/not-to use VLANs on each NIC - for example 3 NIC's where the two first (`ctlplane` and `tenant`) have VLAN subport - and the last one (`baremetal`) is just a NIC. 

```
crc:
    ocps:
      network-template:
        range:
          start: 10
          length: 5
      networks:
        ctlplane:
          is_trunk_parent: true
        internalapi:
          trunk_parent: ctlplane
        tenant:
          is_trunk_parent: true
        storage:
          trunk_parent: tenant
        baremetal: {}
```  

It is also be possible to connect `storage` and `internal_api` as non-VLAN NICs on a different node type.
For example - compute-a-0 uses VLANs - compute-b-0 has just 4 NICs - so separate `OpenStackDataPlaneNodeSet`'s with differnt `network_config_template`'s can be tested in a single job i.e multi-nic, single-nic-vlans, two-nic-vlans etc.
```
          compute-a-0:
            networks:
              default:
                ip: "192.168.122.100"
                is_trunk_parent: true
              internal-api:
                ip: "172.17.0.100"
                trunk_parent: default
                skip-nm-configuration: true
              storage:
                ip: "172.18.0.100"
                trunk_parent: default
                skip-nm-configuration: true
              tenant:
                ip: "172.19.0.100"
                trunk_parent: default
                skip-nm-configuration: true
          compute-b-1:
            networks:
              default:
                ip: "192.168.122.100"
              internal-api:
                ip: "172.17.0.100"
                skip-nm-configuration: true
              storage:
                ip: "172.18.0.100"
                skip-nm-configuration: true
              tenant:
                ip: "172.19.0.100"
                skip-nm-configuration: true
```

hjensas: Marking resolved, moved compute to group-definition.

rebtoor (approved):

PR #1224, opened 2024-03-05
Generate values based on the original content

pablintino:
  - [x] README in the role

### DIFF
--- a/roles/ci_gen_kustomize_values/templates/hci/edpm-values/values.yaml.j2
+++ b/roles/ci_gen_kustomize_values/templates/hci/edpm-values/values.yaml.j2
@@ -1,6 +1,6 @@
 data:
   ssh_keys:
-    authorized: {{ cifmw_ci_gen_kustomize_values_ssh_authorizedkeys | b64encode }}
+    authorized: {{ cifmw_ci_gen_kustomize_values_ssh_authorizedkeys | b64encode }}
     private: {{ cifmw_ci_gen_kustomize_values_ssh_private_key | b64encode }}
     public: {{ cifmw_ci_gen_kustomize_values_ssh_public_key | b64encode }}
   nova:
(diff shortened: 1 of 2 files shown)


cjeanner (commented):
With that change, we MAY end with an edpm-values.yaml being generic enough to be common with, at least, 2 VA?

roles/ci_gen_kustomize_values/templates/hci/edpm-values/values.yaml.j2
```diff
@@ -24,20 +24,28 @@ data:
...
 {% endif %}
     nodes:
-{% for compute in cifmw_networking_env_definition.instances.keys() %}
-{%   if compute.startswith('compute') %}
-      edpm-{{ compute }}:
+{% set instances_names = []                                                                             %}
+{% set _original_nodeset = (original_content.data | default({})).nodeset | default({})                  %}
+{% set _original_nodes = _original_nodeset.nodes | default({})                                          %}
+{% set _has_computes = _original_nodes.keys() | select('contains', 'compute') | length > 0                %}
+{% set _has_cephs = _original_nodes.keys() | select('contains', 'ceph') | length > 0                      %}
```
psathyan, line 31: IMHO, we could do 
```
{% set _vm_type = _original_nodes.keys() | first | split('-')[1] %}
```

Alternatively, would it be possible to add `networker` type also.

roles/ci_gen_kustomize_values/templates/hci/edpm-values/values.yaml.j2
```diff
@@ -24,20 +24,28 @@ data:
...
+{% for _inst in cifmw_networking_env_definition.instances.keys()                                        %}
+{%   if _inst.startswith('compute') and _has_computes or  _inst.startswith('ceph') and _has_cephs       %}
```
psathyan, line 33: if we go with `_vm_type` then we can do 
```
{% if _inst.startswith(_vm_type) %}
```

rebtoor (approved):

PR #1225, opened 2024-03-06
Add flag to skip fetching code in the reproducer

cescgina:
Add a flag to skip fetching code from the zuul var, and instead simply
push whatever is in ~/src in the ansible controller. This allows to rely
on zuul cloning required projects and depends-on and not repeat the same
process.

### DIFF
--- a/roles/reproducer/README.md
+++ b/roles/reproducer/README.md
@@ -15,6 +15,9 @@ None
 * `cifmw_reproducer_dns_servers`: List of dns servers which should be used by the CRC VM as upstream dns servers. Defaults to 1.1.1.1, 8.8.8.8.
 * `cifmw_reproducer_hp_rhos_release`: (Bool) Allows to consume rhos-release on the hypervisor. Defaults to `false`.
 * `cifmw_reproducer_dnf_tweaks`: (List) Options you want to inject in dnf.conf, both on controller-0 and hypervisor. Defaults to `[]`.
+* `cifmw_reproducer_skip_fetch_repositories`: (Bool) Skip fetching repositories from zuul var and simply copy the code from the ansible controller. Defaults to `false`.
+* `cifmw_reproducer_repositories_path`: (String) Path where the repositories are found in the ansible controller, relative to $HOME. Only has any effect if
+`cifmw_reproducer_skip_fetch_repositories` is set to `true`. Defaults to `src`.
 
 ### run_job and run_content_provider booleans and risks.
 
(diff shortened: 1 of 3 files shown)


roles/reproducer/README.md
```diff
@@ -15,6 +15,9 @@ None
 * `cifmw_reproducer_dns_servers`: List of dns servers which should be used by the CRC VM as upstream dns servers. Defaults to 1.1.1.1, 8.8.8.8.
 * `cifmw_reproducer_hp_rhos_release`: (Bool) Allows to consume rhos-release on the hypervisor. Defaults to `false`.
 * `cifmw_reproducer_dnf_tweaks`: (List) Options you want to inject in dnf.conf, both on controller-0 and hypervisor. Defaults to `[]`.
+* `cifmw_reproducer_skip_fetch_repositories`: (Bool) Skip fetching repositories from zuul var and simply copy the code from the ansible controller.
```
pablintino, line 18: The default value is missing here.

roles/reproducer/tasks/push_code.yml
```diff
@@ -123,6 +126,26 @@
         loop_var: repository
         label: "{{ repository.src | basename }}"
 
+- name: Push code in ~/src to the controller-0
```
pablintino, line 129: If someone uses `cifmw_reproducer_skip_fetch_repositories` this `~/src` may not be true anymore.

roles/reproducer/tasks/push_code.yml
```diff
@@ -123,6 +126,26 @@
...
+    - name: Ensure repositories path exists in the controller-0
+      delegate_to: controller-0
+      ansible.builtin.file:
+        path: "/home/zuul/{{ cifmw_reproducer_repositories_path }}"
+        state: directory
+
+    - name: Copy code from ansible controller to controller-0
+      delegate_to: localhost
+      vars:
+        _source_path: "{{ cifmw_reproducer_repositories_path | regex_replace('/$', '') }}/"
```
pablintino, line 142: Shouldn't this var be used, with the / sanitized, be used in the `ansible.builtin.file` call?

roles/reproducer/tasks/push_code.yml
cescgina: good catch, I missed that

roles/reproducer/README.md
cescgina: thanks, added it

roles/reproducer/tasks/push_code.yml
cescgina: indeed, changed the block to use the var

cescgina:
@pab

cescgina:
@pablintino thanks for the comments, I've made the changes but I'll wait until the current run finishes to upload them to not stop the job run

pablintino:
/approve
/lgtm
This one will have a followup to reuse the previous existing rsync.

PR #1227, opened 2024-03-07
Add additional extra var to update plyabook run

ciecierski:
Add additional extra var to update-edpm.yml playbook run
named update_playbook_run to make it possible to skip some
ansible tasks on deployment and run them on update only.

Signed-off-by: Mikołaj Ciecierski <mikolaj.ciecierski@redhat.com>

As a pull request owner and reviewers, I checked that:
- [x] Appropriate testing is done and actually running
- [x] Appropriate documentation exists and/or is up-to-date:
  - [x] README in the role
  - [x] Content of the docs/source is reflecting the changes

PR #1229, opened 2024-03-07
Make sure component repos in update jobs are setup during update phase only

ciecierski:
Add conditionals to skip setting component repos during deployment phase
in update jobs and only update these repos during update playbook run.
This is required to be able to test update from previously deployed repos
to updated component repos.

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/1227

Signed-off-by: Mikołaj Ciecierski <mikolaj.ciecierski@redhat.com>

  - [x] README in the role

### DIFF
--- a/roles/repo_setup/tasks/main.yml
+++ b/roles/repo_setup/tasks/main.yml
@@ -20,7 +20,8 @@
   ansible.builtin.import_tasks: configure.yml
 - name: Include component repo if required
   ansible.builtin.import_tasks: component_repo.yml
-  when: cifmw_repo_setup_component_name|length > 0
+  when: ( cifmw_repo_setup_component_name|length > 0 and update_playbook_run is undefined and not cifmw_run_update|default(false)) or
+        ( cifmw_repo_setup_component_name|length > 0 and update_playbook_run is defined and cifmw_run_update|default(false))
 - name: Generate additional artifacts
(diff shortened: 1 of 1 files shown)


roles/repo_setup/tasks/main.yml
```diff
@@ -20,7 +20,8 @@
   ansible.builtin.import_tasks: configure.yml
 - name: Include component repo if required
   ansible.builtin.import_tasks: component_repo.yml
-  when: cifmw_repo_setup_component_name|length > 0
+  when: ( cifmw_repo_setup_component_name|length > 0 and update_playbook_run is undefined and not cifmw_run_update|default(false)) or
+        ( cifmw_repo_setup_component_name|length > 0 and update_playbook_run is defined and cifmw_run_update|default(false))
```
marios, line 24: i find it hard to follow here.
it seems the conditionals are negating each other

if update is defined and run update

or 

if update is undefined and not run update ?

ciecierski: Two conditions are there, because in update CI jobs we have both deployment phase and update phase. When deployment is run in update job cifmw_run_update variable set in zuul globally for deployment and update playbook makes sure this tasks are skipped.
On the other hand, when job with just deployment is started  cifmw_run_update is not set and defaults to false and tasked are imported.

PR #1231, opened 2024-03-07
Try all image slot

afazekas:
The image key/values are moved and also not all key defined
in all branches, the keys should be globally unique regardless
where it is in the tree, so trying all.

  - [x] README in the role


viroel:
recheck

PR #1233, opened 2024-03-07
Deploy BGP Spine/Leaf infrastructure on a HV

eduolivares:
This patch includes changes in the libvirt_manager role to deploy VM
that simulate a spine/leaf infrastructure using the reproducer.yaml
playbook.
The main change is that the VMs created on the hypervisor from a certain
type (computes, ocp, etc) can have interfaces connected to different
networks. Example:
- compute-0 is connected to leaf-0
- compute-1 is connected to leaf-1
- ocp-0 is connected to leaf-2
- ocp-1 is connected to leaf-3
etc

This patch also includes values.yaml.j2 templates included in the
ci_gen_kustomize_values role that will be used to generate BGP CRs using
kustomize.

Finally, a new scenario file, bgp-4-racks-crc.yml, is included with the
parameters necessary to use the previously described changes.

  - [X] README in the role

### DIFF
--- a/roles/ci_gen_kustomize_values/templates/bgp/network-values/values.yaml.j2
+++ b/roles/ci_gen_kustomize_values/templates/bgp/network-values/values.yaml.j2
@@ -1,0 +1,145 @@
+---
+# source: bgp/network-values/values.yaml.j2
+{% set ns = namespace(interfaces={},
+                      ocp_index=0,
+                      lb_tools={}) %}
+data:
+{% for host in cifmw_networking_env_definition.instances.keys() -%}
+{%   for network in cifmw_networking_env_definition.instances[host]['networks'].values() -%}
+{%     set ns.interfaces = ns.interfaces |
+                           combine({network.network_name: (network.parent_interface |
+                                                           default(network.interface_name)
+                                                          )
+                                   },
+                                   recursive=true) -%}
+{%   endfor -%}
+{%   if host is match('^(ocp|crc).*') %}
+  node_{{ ns.ocp_index }}:
+    name: {{ cifmw_networking_env_definition.instances[host]['hostname'] }}
+{%     for network in cifmw_networking_env_definition.instances[host]['networks'].values() %}
+    {{ network.network_name }}_ip: {{ network.ip_v4 }}
+{%     endfor %}
+{%     set node_net_orig_content = original_content.data.bgp['net-attach-def']['node' ~ ns.ocp_index] %}
+{%     set node_bgp_net0 = node_net_orig_content.bgpnet0 | from_json %}
+{%     set node_bgp_net1 = node_net_orig_content.bgpnet1 | from_json %}
+    bgp_peers:
+      - {{ node_bgp_net0.ipam.range_start }}
+      - {{ node_bgp_net1.ipam.range_start }}
+    bgp_ip:
+      - {{ node_bgp_net0.ipam.range_start | ansible.utils.ipmath(1) }}
+      - {{ node_bgp_net1.ipam.range_start | ansible.utils.ipmath(1) }}
+{#     loopback_ip and loopback_ipv6 correspond to subnet3 because all ocp nodes are deployed on rack3 #}
+{%     set loopback_ip = original_content.data.bgp.subnets.bgpmainnet[3].allocationRanges[0].start |
+                         ansible.utils.ipmath(ns.ocp_index) %}
+{%     set loopback_ipv6 = original_content.data.bgp.subnets.bgpmainnetv6[3].allocationRanges[0].start |
+                           ansible.utils.ipmath(ns.ocp_index) %}
+    loopback_ip: {{ loopback_ip }}
+    loopback_ipv6: {{ loopback_ipv6 }}
+{%     set ns.ocp_index = ns.ocp_index+1 %}
+{%   endif %}
+{% endfor %}
+
+{% for network in cifmw_networking_env_definition.networks.values() %}
+{% set ns.lb_tools = {} %}
+  {{ network.network_name }}:
+    dnsDomain: {{ network.search_domain }}
+{%  if network.tools is defined and network.tools.keys() | length > 0 %}
+    subnets:
+{%    for tool in network.tools.keys() %}
+{%      if tool is match('.*lb$') %}
+{%        set _ = ns.lb_tools.update({tool: []}) %}
+{%      endif %}
+{%    endfor %}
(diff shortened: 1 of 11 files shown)


mhuin:
recheck

scenarios/reproducers/bgp-4-racks-crc.yml
pablintino, line 1: Is the name correct? As the file is targeted for dev-scripts.

cjeanner:
@eduolivares heya! would it be possible to get a "small" molecule scenario to ensure the boolean, and related tasks, are working correctly? Not that I don't trust the code, but that would also ensure no regression arise in the future.

To do that, you can:
- copy the current directory in molecule to, for instance, `spin_leaf`
- modifiy the "prepare.yml" and "converge.yml" accordingly

That new scenario will be triggered directly from molecule, in the same run. Maybe ensuring things are properly cleaned before would be good (calling the cleanup_layout.yml from a `cleanup.yml` play in the molecule scenario directory).

No need to deploy 10th of thousands of NIC, just a couple, to ensure things are properly attached.

eduolivares:
> @eduolivares heya! would it be possible to get a "small" molecule scenario to ensure the boolean, and related tasks, are working correctly? Not that I don't trust the code, but that would also ensure no regression arise in the future.
> 
> To do that, you can:
> 
> * copy the current directory in molecule to, for instance, `spin_leaf`
> * modifiy the "prepare.yml" and "converge.yml" accordingly
> 
> That new scenario will be triggered directly from molecule, in the same run. Maybe ensuring things are properly cleaned before would be good (calling the cleanup_layout.yml from a `cleanup.yml` play in the molecule scenario directory).
> 
> No need to deploy 10th of thousands of NIC, just a couple, to ensure things are properly attached.

ok, I have just added something for this. I haven't been able to test it, so most probably it won't work yet.

cjeanner:
/approve

@eduolivares we'll have to figure out the impact of this patch on IPv6 support addition, especially #1649 (and subsequent tasks not yet started)

roles/libvirt_manager/tasks/attach_interface.yml
```diff
@@ -73,10 +73,12 @@
         (_net_name not in _net_bridge_map.keys()) |
         ternary(_net_name, _net_bridge_map[_net_name])
       }}
+    _type: "{{ cifmw_libvirt_manager_network_interface_types[network.name] | default('bridge') }}"
     _attached_bridges: >-
       {{
         _extracted_xml.matches | default([]) |
-        selectattr('source.bridge', 'equalto', _local_bridge_name) | default([])
+        selectattr('source.' + _type, 'defined') | default([]) |
+        selectattr('source.' + _type, 'equalto', _local_bridge_name) | default([])
```
pablintino, line 81: Those `| default([])` are not needed. It's safe to iterate the list/iterator `selectattr` returns in case it's empty.

roles/ci_gen_kustomize_values/templates/bgp/network-values/values.yaml.j2
```diff
@@ -0,0 +1,22 @@
+---
+# source: bgp/network-values/values.yaml.j2
+data:
+  # nodes
+{% for node_id in range(3) %}
```
pablintino, line 5: Why not base this range in the `original_content` variable? You can check the amount of nodes to patch there, allowing easy scaling if needed at some point.

roles/ci_gen_kustomize_values/templates/bgp/network-values/values.yaml.j2
pablintino, line 1: Are we sure we want to diverge that far from the common network-values? I know that the networking for this DT is totally different, but there are "common" points:
- storageClass
- annotations (rabbitmq)
- multus network attachments
- metalb pools

eduolivares: I have modified this file. Now it's similar to [1], just adding the BGP stuff to it.
I'm adding a DNM tag to this PR while I test this change. 

[1] https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/ci_gen_kustomize_values/templates/common/network-values/values.yaml.j2

scenarios/reproducers/bgp-4-racks-3-ocps.yml
```diff
@@ -0,0 +1,402 @@
...
+          ternary(4, 1)
+        }}
+      image_url: "{{ cifmw_discovered_image_url }}"
+      sha256_image_name: "{{ cifmw_discovered_hash }}"
+      image_local_dir: "{{ cifmw_basedir }}/images/"
+      disk_file_name: "centos-stream-9.qcow2"
+      disksize: 50
+      memory: 8
+      cpus: 4
+      spineleafnets:
```
hjensas, line 244: Why are we not using different "roles"
i.e there is a vm of type ocp-1 with amount = 1, a vm of type ocp-2 also amoutn = 1

```
vms:
  ocp-1:
    amount: 1
     xml_paths:
        - /home/dev-scripts/ocp_master_0.xml
    networks:
       - ocpbm
       - l10-ocp0
       - l20-ocp0
       - l30-ocp0
  ocp-1:
    networks:
       - ocpbm
       - l10-ocp1
       - l20-ocp1
       - l30-ocp1
    amount: 1
     xml_paths:
        - /home/dev-scripts/ocp_master_1.xml
  ocp-2:
    amount: 1
     xml_paths:
        - /home/dev-scripts/ocp_master_2.xml
    networks:
       - ocpbm
       - l10-ocp3
       - l20-ocp3
       - l30-ocp3
```

hjensas: If we need a group "ocp" and a group "compute" in one of these ansible inventories - we could add something to tell which group vms should be in.

```
vms:
  ocp-1:
     amount: 1
     ansible_group: ocp
  ocp-2:
     amount: 1
     ansible_group: ocp
  ocp-3:
     amount: 1
     ansible_group: ocp
```

eduolivares: Thanks for the proposal. If I have time, I will try to apply this change on a later PR.

roles/ci_gen_kustomize_values/templates/bgp/network-values/values.yaml.j2
eduolivares: Removing the DNM tag after having tested this change.

PR #1237, opened 2024-03-08
Update recent changes in jump

jirimacku:
  - [ ] README in the role


tosky:
I would change the commit message, because this is just fixing an error in the name of the variable. The reference has always been against a test run (you collect the results and report them to a single or to multiple test runs).

jirimacku:
> I would change the commit message, because this is just fixing an error in the name of the variable. The reference has always been against a test run (you collect the results and report them to a single or to multiple test runs).

Sure, this is just a draft I want to test. Once working I'll update the commit message.

PR #1238, opened 2024-03-08
Refactor reproducer push code to reuse existing code

cescgina:
In a recent change[1], we introduced a new block to copy repositories
from the ansible controller to controller-0 in the push_code file of the
reproducer role. The new code is duplicating existing tasks.
This change refactors the push_code file to reuse existing tasks while
respecting user input.

  - [x] README in the role

### DIFF
--- a/roles/reproducer/tasks/push_code.yml
+++ b/roles/reproducer/tasks/push_code.yml
@@ -74,15 +74,35 @@
 - name: Push random code into the proper location
   vars:
     repo_base_dir: '/home/zuul/src'
-  when:
-    - not cifmw_reproducer_skip_fetch_repositories | default(false)
+    _cifmw_reproducer_all_repositories: "{{ cifmw_reproducer_repositories | default([]) }}"
   block:
+    - name: Expand cifmw_reproducer_repositories to pull code from ansible controller to controller-0
+      when:
+        - cifmw_reproducer_skip_fetch_repositories | default(false)
+        - zuul is defined
+        - zuul.items is defined
+        - zuul.projects is defined
+        # check that the project is not alredy in the user defined
+        # cifmw_reproducer_repositories value
+        - repo.value.name not in _user_sources
+      vars:
+        _user_sources: "{{ cifmw_reproducer_repositories | default([]) | map(attribute='src') }}"
+        _repo_entry:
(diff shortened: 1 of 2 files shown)


roles/reproducer/tasks/push_code.yml
```diff
@@ -74,9 +74,28 @@
...
+        # check that the project is not alredy in the user defined
+        # cifmw_reproducer_repositories value
+        - repo.value.name not in _user_sources
+      vars:
+        _user_sources: "{{ cifmw_reproducer_repositories | default([]) | map(attribute='src') }}"
+        _repo_entry:
+          src: "{{ ansible_user_dir }}/{{ repo.value.src_dir | regex_replace('/$', '') }}/"
+          dest: "/home/zuul/{{ repo.value.src_dir }}"
+      ansible.builtin.set_fact:
+        cifmw_reproducer_repositories: "{{ cifmw_reproducer_repositories | default([]) + [_repo_entry] }}"
```
cjeanner, line 93: afaik we can't override a provided parameter with `set_fact`. You'll need to get some "internal" fact for that one.

cescgina: ah right, it has lower precedence. I tested it with a job without any value for `cifmw_reprodcuer_repositories`, so I did not see any issue, good catch!

roles/reproducer/tasks/push_code.yml
```diff
@@ -74,15 +74,35 @@
...
+        # check that the project is not alredy in the user defined
+        # cifmw_reproducer_repositories value
+        - repo.value.name not in _user_sources
+      vars:
+        _user_sources: "{{ cifmw_reproducer_repositories | default([]) | map(attribute='src') }}"
+        _repo_entry:
+          src: "{{ ansible_user_dir }}/{{ repo.value.src_dir | regex_replace('/$', '') }}/"
+          dest: "/home/zuul/{{ repo.value.src_dir }}"
+      ansible.builtin.set_fact:
+        _cifmw_reproducer_all_repositories: "{{ _cifmw_reproducer_all_repositories  + [_repo_entry] }}"
```
pablintino, line 94: Is this properly overriding the variable declared here? Cause set_fact cannot change the value of a variable.

PR #1239, opened 2024-03-08
Adding the capability of customize values.yaml for OLM resources

rebtoor:
Since we added the capability of customize OLM resources in the architecture repo, now we can use `ci_gen_kustomize_values` role to customize it for its usage in the `kustomize_deploy` role.

  - [x] README in the role

### DIFF
--- a/roles/repo_setup/tasks/ci_mirror.yml
+++ b/roles/repo_setup/tasks/ci_mirror.yml
@@ -1,0 +1,17 @@
+---
+- name: Check for /etc/ci/mirror_info.sh
+  ansible.builtin.stat:
+    path: "/etc/ci/mirror_info.sh"
+  register: mirror_path
+
+- name: Use proxy mirrors
+  become: "{{ not cifmw_repo_setup_output.startswith(ansible_user_dir) }}"
+  when:
+    - mirror_path.stat.exists
+  ansible.builtin.shell: |
+    set -o pipefail
+    source /etc/ci/mirror_info.sh
+    sed -i -e "s|https://trunk.rdoproject.org|$NODEPOOL_RDO_PROXY|g" *.repo
+    sed -i -e "s|http://mirror.stream.centos.org|$NODEPOOL_CENTOS_MIRROR|g" *.repo
+  args:
+    chdir: "{{ cifmw_repo_setup_output }}"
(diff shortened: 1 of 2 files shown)


roles/ci_gen_kustomize_values/tasks/olm_values.yml
```diff
@@ -0,0 +1,5 @@
+---
+- name: Assert we have needed parameters
+  ansible.builtin.assert:
+    that:
+      - cifmw_ci_gen_kustomize_values_ooi_image is defined
```
pablintino, line 5: Does this mean, that if the new config-map grows we will blow up just because this one is missing? :(

roles/ci_gen_kustomize_values/templates/common/olm/values.yaml.j2
```diff
@@ -0,0 +1,2 @@
+data:
+    openstack-operator-image: {{ cifmw_ci_gen_kustomize_values_ooi_image | default('quay.io/openstack-k8s-operators/openstack-operator-index:latest', true) }}
```
pablintino, line 2: It's never going to use the default value as we are enforcing the user to provide one here: https://github.com/openstack-k8s-operators/ci-framework/pull/1239/files#diff-335b6a30d3df971b4acd157f143baba7b7659ce77fdf4ac3d1c1a14dce6271e5

roles/ci_gen_kustomize_values/tasks/olm_values.yml
rebtoor: yeah, i'll probably remove this.

roles/ci_gen_kustomize_values/templates/common/olm/values.yaml.j2
rebtoor: fixed :D

rebtoor:
recheck

roles/kustomize_deploy/tasks/install_operators.yml
```diff
@@ -14,6 +14,35 @@
...
+- name: Generate values.yaml for OLM resources
+  vars:
+    cifmw_architecture_scenario: 'common/olm'
+    cifmw_ci_gen_kustomize_values_src_file: >-
+      {{
+        (
+          cifmw_kustomize_deploy_olm_source_files,
+          'values.yaml') | path_join
+      }}
+  ansible.builtin.include_role:
```
pablintino, line 26: We may like to expose a variable to allow the user to pass an entire override (`cifmw_ci_gen_kustomize_values_userdata`)

cjeanner: +1

rebtoor: ack!

PR #1245, opened 2024-03-11
[repo_setup] Update generated repos with nodepool mirror repos

raukadah:
The nodepool mirror repos variables are stored in `/etc/ci/mirror_info.sh` file. Let's source this file and replace the centos and rdo trunk urls with nodepool provided mirror urls.

It will help to avoid mirror timeout issue in CI.

Closes: https://issues.redhat.com/browse/OSPCIX-212

### DIFF
--- a/roles/repo_setup/tasks/ci_mirror.yml
+++ b/roles/repo_setup/tasks/ci_mirror.yml
@@ -1,0 +1,17 @@
+---
+- name: Check for /etc/ci/mirror_info.sh
+  ansible.builtin.stat:
+    path: "/etc/ci/mirror_info.sh"
+  register: mirror_path
+
+- name: Use proxy mirrors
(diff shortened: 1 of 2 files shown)


roles/repo_setup/tasks/ci_mirror.yml
```diff
@@ -0,0 +1,17 @@
+---
+- name: Check for /etc/ci/mirror_info.sh
+  ansible.builtin.stat:
+    path: "/etc/ci/mirror_info.sh"
+  register: mirror_path
+
+- name: Use proxy mirrors
+  become: "{{ not cifmw_repo_setup_output.startswith(ansible_user_dir) }}"
+  when:
+    - mirror_path.stat.exists
```
rlandy, line 10: ok - will only execute if mirror_path is there

rlandy (approved):

PR #1246, opened 2024-03-11
[repo_setup] Sync generated repos to yum.repos.d directory

raukadah:
We should always sync generated repos to yum.repos.d directory so that the deployment should install all content from same generated repos.

It also make sure that earlier yum contents are removed from /etc/yum.repos.d directory and metadata are cleaned.

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/1245

### DIFF
--- a/roles/repo_setup/tasks/sync_repos.yml
+++ b/roles/repo_setup/tasks/sync_repos.yml
@@ -1,0 +1,28 @@
+---
+- name: Copy generated repos to /etc/yum.repos.d directory
+  when: not cifmw_repo_setup_output.startswith(ansible_user_dir)
+  block:
+    - name: Find existing repos from /etc/yum.repos.d directory
+      ansible.builtin.find:
+        paths: "/etc/yum.repos.d/"
+        patterns: "*.repo"
+        recurse: false
+      register: _yum_repos
+
+    - name: Remove existing repos from /etc/yum.repos.d directory
+      become: true
+      ansible.builtin.file:
+        path: "{{ item }}"
+        state: absent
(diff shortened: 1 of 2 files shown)


cjeanner:
Needs a rebase against `main` once #1245 merges.

roles/repo_setup/tasks/sync_repos.yml
```diff
@@ -0,0 +1,27 @@
...
+  block:
+    - name: Find existing repos from /etc/yum.repos.d directory
+      ansible.builtin.find:
+        paths: "/etc/yum.repos.d/"
+        patterns: "*.repo"
+        recurse: false
+      register: _yum_repos
+
+    - name: Remove existing repos from /etc/yum.repos.d directory
+      become: "{{ not cifmw_repo_setup_output.startswith(ansible_user_dir) }}"
```
cjeanner, line 13: hmmm since you're calling `sudo` you want to remove it and put `become: true` I think?

Same goes for the 2 other occurrences bellow.

roles/repo_setup/tasks/sync_repos.yml
```diff
@@ -0,0 +1,27 @@
...
+      ansible.builtin.file:
+        path: "{{ item }}"
+        state: absent
+      with_items: "{{ _yum_repos.files|map(attribute='path') }}"
+
+    - name: Cleanup existing metadata
+      become: "{{ not cifmw_repo_setup_output.startswith(ansible_user_dir) }}"
+      ansible.builtin.command: "sudo dnf clean metadata"
```
cjeanner, line 21: and with `become: true` no need for `sudo` in the command itself.

PR #1247, opened 2024-03-11
[test-operator] Configure when test-operator asserts results

eduolivares:
With this patch, when the test-operator runs tests from different test frameworks, by default the results are asserted when all the execution from all the tests frameworks is completed.
By configuring cifmw_test_operator_fail_fast to true, the role will raise an error as soon as tests from a test framework fail.

  - [X] README in the role


lpiwowar (approved):
Looks good to me! This will fix the issue of tobiko not being executed when tempest fails. [1]

[1] https://softwarefactory-project.io/zuul/t/rdoproject.org/build/6c466bda4c3f46668c4d32543508b0bf

PR #1249, opened 2024-03-11
Sync repos only non yum.repos.d cifmw_repo_setup_output directory

raukadah:
We need to sync repos in yum.repos.d directory only when cifmw_repo_setup_output is different from yum.repos.d directory.

Otherwise the job will fail with missing packages.


cjeanner:
/approve
/lgtm

aw dang... "tiny" nit blowing the whole thing.

rlandy (commented):
/lgtm

PR #1251, opened 2024-03-11
Rework a bit conditions on playbook imports

cjeanner:
With import_*, the catalog is built with all the content flattened, and
any conditions or vars/environment are applied on any of the flattened
content.

This means ansible will go through all of the tasks in the included
playbooks and "skip" them - instead of skipping the playbook import.

This patch should make things slightly more readable in the logs, with
an early end_play whenever we're able to apply it.

Note about the (pre|post)_deploy hooks:
we can't move them to the main `deploy-edpm.yml` playbook, because the
`06-deploy-edpm.yml` sub-play is directly imported in the adoption jobs.

### DIFF
--- a/deploy-edpm.yml
+++ b/deploy-edpm.yml
@@ -38,15 +38,11 @@
     - build-operators
 
 - name: Import deploy edpm playbook
-  when:
-    - cifmw_architecture_scenario is undefined
   ansible.builtin.import_playbook: playbooks/06-deploy-edpm.yml
   tags:
     - edpm
 
 - name: Import VA deployment playbook
-  when:
-    - cifmw_architecture_scenario is defined
   ansible.builtin.import_playbook: playbooks/06-deploy-architecture.yml
   tags:
     - edpm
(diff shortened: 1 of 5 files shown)


cjeanner:
recheck

raukadah:
recheck

deploy-edpm.yml
```diff
@@ -37,20 +37,26 @@
   tags:
     - build-operators
 
+- name: Run pre_deploy hooks
```
marios, line 40: generally makes sense (and this is interesting https://docs.ansible.com/ansible/latest/collections/ansible/builtin/meta_module.html#ansible-builtin-meta-module-execute-ansible-actions)

not clear on the relevance of the pre_deploy hooks being added here though with the rest of this i.e. removing the whens and using end_play?

cjeanner: pre_deploy hooks were duplicated in the sub-playbook, leading to duplicated imported playbook, with the same conditions being applied in order to avoid running them twice - but due to `import_*` pattern, it was running, with the `when` condition added to the content - more "skipping" than needed.

cjeanner: So, we can't extract those hook runs: since adoption job is calling the `06-deploy-edpm.yml` directly, it has to run hooks from there:

https://github.com/rdo-infra/rdo-jobs/blob/13c683ea25eb0f2fd71412c5c8d305a674fd0540/playbooks/data_plane_adoption/setup_crc_use_install_yamls_extracted_crc.yaml#L48-L51

This explains why the adoption job was failing on this patch :).

marios (approved):
lgtm + the adoption job seems happy after the last update

PR #1252, opened 2024-03-11
Generate a script for deploy-edpm.yml run

cjeanner:
This deploy-edpm.yml run is the easiest one: it will be a "close to CI"
deployment, without much specificities.

It will NOT deploy any VA/DT - just plain edpm, like we'd do in
install_yamls project.

### DIFF
--- a/roles/reproducer/tasks/main.yml
+++ b/roles/reproducer/tasks/main.yml
@@ -134,3 +134,19 @@
       when:
         - cifmw_architecture_scenario is defined
       ansible.builtin.include_tasks: configure_architecture.yml
+
+    - name: Prepare ci-like EDPM deploy
+      ansible.builtin.copy:
+        dest: "/home/zuul/deploy-edpm.sh"
+        mode: "0755"
+        owner: "zuul"
+        group: "zuul"
+        content: |-
+          #!/bin/bash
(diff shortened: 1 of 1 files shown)


raukadah (approved):
/approve

roles/reproducer/tasks/main.yml
```diff
@@ -134,3 +134,19 @@
...
       ansible.builtin.include_tasks: configure_architecture.yml
+
+    - name: Prepare ci-like EDPM deploy
+      ansible.builtin.copy:
+        dest: "/home/zuul/deploy-edpm.sh"
+        mode: "0755"
+        owner: "zuul"
+        group: "zuul"
+        content: |-
+          #!/bin/bash
```
psathyan, line 145: Shouldn't this be `/usr/bin/bash`?

cjeanner: a quick `git grep /bin/bash` in the repository shows a vast majority of `/bin/bash` with maybe 3 `/usr/bin/bash` - both are valid.

psathyan: I was also under the impression that `/bin/bash` is valid until I hit an issue with coreOS. Since this is on the controller, I don't see a reason why we cannot use `/bin/bash`

PR #1254, opened 2024-03-12
Keep epel repo dependency in rescue mode for openstack_buid_packages role

raukadah:
Mock package is needed by dlrn to build openstack package and is shipped by delorean-antelope-build-deps repo.
    
Rest of the deps is already provided by repo-setup repos. So we donot need to enable repos here.

This pr move the epel package installation in rescue mode.
  
https://github.com/openstack-k8s-operators/ci-framework/pull/1246 syncs the repo-setup repos to /etc/yum.repos.d directory and drops existing repos.
    
epel-release package comes from extra-commons centos repo defined in centos-addons.repo. Since repo-setup deletes the existing repos. So the epel-release package installation fails which is expected.
    
In order to fix, we are going to install epel-release from urlinstead of packages.

### DIFF
--- a/roles/build_openstack_packages/tasks/install_dlrn.yml
+++ b/roles/build_openstack_packages/tasks/install_dlrn.yml
@@ -14,33 +14,25 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
-- name: Install epel-release repo
-  become: true
-  tags:
-    - bootstrap
-    - packages
-  ansible.builtin.package:
-    state: present
-    name: epel-release
-  when: ansible_distribution | lower == 'centos'
-
 - name: Install mock
-  become: true
   tags:
     - bootstrap
     - packages
-  ansible.builtin.package:
-    state: present
-    name: mock
+  block:
+    - name: Install Mock from DLRN repos
+      become: true
+      ansible.builtin.package:
+        name: mock
+        state: present
 
-- name: Remove epel-repos
-  become: true
-  tags:
-    - bootstrap
-    - packages
-  ansible.builtin.shell: |
-    sudo rm -rf /etc/yum.repos.d/epel*;
-    sudo rpm -e epel-release || true;
+  rescue:
+    - name: Install mock from epel repos
+      vars:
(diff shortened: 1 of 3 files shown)


roles/build_openstack_packages/tasks/install_dlrn.yml
```diff
@@ -15,32 +15,15 @@
 # under the License.
 
 - name: Install epel-release repo
```
cjeanner, line 17: You want to rename that task, maybe to "Install mock from EPEL" ?

roles/build_openstack_packages/tasks/install_dlrn.yml
```diff
@@ -14,33 +14,14 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
-- name: Install epel-release repo
+- name: Install Mock
```
cjeanner, line 17: we may want to install from EPEL if we don't have dlrn? Maybe a `block/except` to try installing from existing repositories, and fall back on EPEL in case it can't find it?

roles/ci_setup/tasks/epel.yml
```diff
@@ -6,12 +6,9 @@
...
 - name: Install EPEL
   become: true
-  vars:
-    rhel_uri: "https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.noarch.rpm"
-    pkg: "{{ (ansible_facts['distribution'] == 'RedHat') | ternary(rhel_uri, 'epel-release')}}"
   ansible.builtin.dnf:
-    name: "{{ pkg }}"
-    disable_gpg_check: "{{ (pkg != 'epel-release') }}"
+    name: "https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.noarch.rpm"
+    disable_gpg_check: false
```
cjeanner, line 11: note that FIPS may dislike that?

cjeanner:
/approve

We can work on the FIPS/gpg check in a follow-up. We may "just" install the EPEL gpg key beforehand.

PR #1255, opened 2024-03-12
Rework include/import usage in reproducer role

cjeanner:
In order to make the whole reproducer slightly faster, a small rework in
the way we call roles has to be done.

`import_role` and `import_tasks` are embedding the content while building
the catalog, at the very beginning, and any `when` conditions are then
added to the subsequent tasks, leading to a lot of tasks being evaluated
and "skipped", clogging the logs and taking more time than necessary.

Switching "conditioned" imports to `include` are ensuring we don't face
that same behavior.

With this, we need to set the `cifmw_libvirt_manager_compute_amount`
parameter in the scenario files, since the libvirt_manager role isn't
*imported* anymore, but dynamically included.

### DIFF
--- a/roles/reproducer/tasks/configure_controller.yml
+++ b/roles/reproducer/tasks/configure_controller.yml
@@ -158,7 +158,7 @@
         dest: "/home/zuul/.kube/config"
         content: >-
           {{
-            (cifmw_use_devscripts | default(false) | bool) |
+            (cifmw_libvirt_manager_configuration.vms.ocp is defined) |
             ternary(_devscripts_kubeconfig.content, _crc_kubeconfig.content) |
             b64decode
           }}
--- a/roles/reproducer/tasks/main.yml
+++ b/roles/reproducer/tasks/main.yml
@@ -25,9 +25,9 @@
   ansible.builtin.assert:
     that:
       - (cifmw_libvirt_manager_configuration.vms.crc is defined) or
-        (cifmw_use_devscripts | default(false) | bool)
+        (cifmw_libvirt_manager_configuration.vms.ocp is defined)
       - not ((cifmw_libvirt_manager_configuration.vms.crc is defined) and
-             (cifmw_use_devscripts | default(false) | bool))
+             (cifmw_libvirt_manager_configuration.vms.ocp is defined))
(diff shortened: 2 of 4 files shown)


psathyan (commented):
I believe this line also needs to be changed https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/reproducer/tasks/configure_controller.yml#L161

cjeanner:
> I believe this line also needs to be changed https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/reproducer/tasks/configure_controller.yml#L161

oh, right, good catch. In fact...... that parameter may be obsolete. Though it may be used in hybrid, meaning the libvirt_manager_configuration wouldn't be used?

psathyan:
> > I believe this line also needs to be changed https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/reproducer/tasks/configure_controller.yml#L161
> 
> oh, right, good catch. In fact...... that parameter may be obsolete. Though it may be used in hybrid, meaning the libvirt_manager_configuration wouldn't be used?

`libvirt_manager_configuration` is the backbone for the virtual networking setup :) also I am trying to converge to using `reproducer`. Most of the recent PRs help in converging both the approaches.

psathyan (approved):
nit: I not sure if we would need to check for max here

https://github.com/openstack-k8s-operators/ci-framework/pull/1255/files#diff-a27d3292259ca39642bd28c26baea1f33d2cc740203d86ebb1318bb396d15ab2R87

rebtoor (approved):

cjeanner:
> nit: I not sure if we would need to check for max here
> 
> https://github.com/openstack-k8s-operators/ci-framework/pull/1255/files#diff-a27d3292259ca39642bd28c26baea1f33d2cc740203d86ebb1318bb396d15ab2R87

sooo for now yes, because that VA-1 is set for a minimal of 3 nodes - with this, we ensure that no one may try a 1-compute deployment (at least using that specific file). We can refine it if needed, but imho we should stick to the "proper HCI needs", unless there are specific, documented use-cases (CI, maybe? Not even sure at this point).

PR #1257, opened 2024-03-12
[ci_script] Print log location to allow easy following of it

pablintino:

### DIFF
--- a/plugins/action/ci_script.py
+++ b/plugins/action/ci_script.py
@@ -101,10 +101,11 @@ import pathlib
 import re
 import uuid
 
-
+from ansible import constants as C
 from ansible.plugins.action import ActionBase
 from ansible.errors import AnsibleActionFail
 from ansible.module_utils import basic
+from ansible.utils.display import Display
 
 from ansible_collections.cifmw.general.plugins.module_utils.encoding import (
     ansible_encoding,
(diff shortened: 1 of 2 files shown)


plugins/action/ci_script.py
```diff
@@ -186,16 +187,17 @@ def run(self, tmp=None, task_vars=None):
             file_task.args.pop(arg_name, None)
 
         fnum = len(glob.glob(f"{output_dir}/ci_script_*"))
-        t_name = re.sub(r"([^\x00-\x7F]|\s)+", "_", self._task.name).lower()
+        t_name = re.sub(r"([^\x00-\x7F]|\s|-)+", "_", self._task.name).lower()
```
cjeanner, line 190: since you're reworking that regexp, I think we should go with the "readable" `[^A-Za-z0-9_\]+` regexp - that would replace any non-alphanumeric (plus - and _) by "_", like I did in the run_hook.

plugins/action/ci_script.py
```diff
@@ -186,16 +187,17 @@ def run(self, tmp=None, task_vars=None):
...
         chdir_path = task_args.get("chdir", None)
+        log_path = output_dir.parent.joinpath(
+            "logs", f"ci_script_{fnum:03}_{t_name}.log"
```
cjeanner, line 193: for the sake of consistency, maybe we should ensure names aren't too long? A truncate() (or equivalent) to 20 chars or so (like the run_hook role) ?

rebtoor (approved):

PR #1259, opened 2024-03-12
Adapt VA-HCI for shiftstack

rlobillo:
Adapt reproducer to include the needed changes to support openshift as workload on RHOSP18.

  - [x] README in the role

### DIFF
--- a/reproducer.yml
+++ b/reproducer.yml
@@ -55,3 +55,10 @@
   roles:
     - role: ci_setup
     - role: reproducer
+
+  post_tasks:
+    - name: Allow traffic from OSP VMs to OSP API (needed for shiftstack)
+      become: true
+      when: cifmw_allow_vms_to_reach_osp_api | default ('false') | bool
+      ansible.builtin.command: # noqa: command-instead-of-module
+        cmd: iptables -I LIBVIRT_FWI 1 -o ocpbm -j ACCEPT
--- a/scenarios/reproducers/shift-on-stack-overrides.yml
+++ b/scenarios/reproducers/shift-on-stack-overrides.yml
@@ -1,0 +1,6 @@
+cifmw_libvirt_manager_compute_disksize: 200
+cifmw_libvirt_manager_compute_memory: 50
+cifmw_libvirt_manager_compute_cpus: 8
+cifmw_allow_vms_to_reach_osp_api: true
+# HCI requires bigger size to hold OCP on OSP disks
+cifmw_block_device_size: 100G
--- a/scenarios/reproducers/va-hci.yml
+++ b/scenarios/reproducers/va-hci.yml
@@ -89,9 +89,9 @@ cifmw_libvirt_manager_configuration:
       sha256_image_name: "{{ cifmw_discovered_hash }}"
       image_local_dir: "{{ cifmw_basedir }}/images/"
       disk_file_name: "base-os.qcow2"
-      disksize: 50
-      memory: 8
-      cpus: 4
+      disksize: "{{ [cifmw_libvirt_manager_compute_disksize|int, 50] | max }}"
+      memory: "{{ [cifmw_libvirt_manager_compute_memory|int, 8] | max }}"
+      cpus: "{{ [cifmw_libvirt_manager_compute_cpus|int, 4] | max }}"
       nets:
         - ocpbm
         - osp_trunk


scenarios/reproducers/validated-architecture-shiftstack.yml
```diff
@@ -0,0 +1,159 @@
...
+          ternary(4, 1)
+        }}
+      amount: "{{ [cifmw_libvirt_manager_compute_amount|int, 3] | max }}"
+      image_url: "{{ cifmw_discovered_image_url }}"
+      sha256_image_name: "{{ cifmw_discovered_hash }}"
+      image_local_dir: "{{ cifmw_basedir }}/images/"
+      disk_file_name: "base-os.qcow2"
+      disksize: 200
+      memory: 50
+      cpus: 8
```
rlobillo, line 93: different from default validated-architecture-1.yml

scenarios/reproducers/validated-architecture-shiftstack.yml
```diff
@@ -0,0 +1,159 @@
...
+# Test Ceph file and object storage (block is enabled by default)
+cifmw_ceph_daemons_layout:
+  rgw_enabled: true
+  dashboard_enabled: false
+  cephfs_enabled: true
+
+# OCP on OSP needs to access OSP API from the VMs that behaves as OCP nodes
+cifmw_allow_vms_to_reach_osp_api: true
+# HCI requires bigger size to hold OCP on OSP disks
+cifmw_block_device_size: 100G
```
rlobillo, line 159: different from default validated-architecture-1.yml

reproducer.yml
```diff
@@ -55,3 +55,9 @@
   roles:
     - role: ci_setup
     - role: reproducer
+
+  post_tasks:
+    - name: Allow traffic from OSP VMs to OSP API (needed for shiftstack)
+      ansible.builtin.shell: iptables -I LIBVIRT_FWI 1 -o ocpbm -j ACCEPT
+      become: true
+      when: cifmw_allow_vms_to_reach_osp_api
```
rlobillo, line 63: This is needed for allowing the VMs to contact the OSP API, which is a requirement for the shiftstack OCP nodes. This workload needs to manage OSP resources during both the installation process and the shiftstack operators activity.

scenarios/reproducers/validated-architecture-shiftstack.yml
cjeanner: this may be passed as a parameter, no need to copy the whole file just for that diff.

scenarios/reproducers/validated-architecture-shiftstack.yml
cjeanner: if CPU is the only difference, we should consider doing the same thing done regarding the compute amount and allow passing it as an inline parameter

reproducer.yml
cjeanner: I'd rather not use `shell` here:
- there's a [module for iptables](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/iptables_module.html) ( ansible-lint will crash for this)
- if really, `ansible.builtin.command` is enough (ansible-lint will crash for this)

Also, the `when` condition will fail if the parameter isn't set, so you want `| default(false) | bool`.
Maybe it should be done in any cases? Not really sure what this would allow, to be fair.

scenarios/reproducers/validated-architecture-shiftstack.yml
```diff
@@ -0,0 +1,159 @@
...
+
+## devscript support for OCP deploy
+cifmw_use_devscripts: true
+cifmw_devscripts_config_overrides:
+  worker_memory: 16384
+  worker_disk: 100
+  worker_vcpu: 10
+  num_extra_workers: 0
+# Required for egress traffic from pods to the osp_trunk network
+cifmw_devscripts_enable_ocp_nodes_host_routing: true
```
cjeanner, line 135: ok, this is a the new param to enable the firewall - no need for a full copy-paste...

cjeanner (changes requested):
While I understand the need, I think this PR is wrong, especially on the scenario file.
You should add the parameter to raise the compute CPU as described in the comment, then create one single file, like "shift-on-stack-overrides.yml" and just add the specific overrides there. It would make the whole thing far, far easier to maintain...

reproducer.yml
rlobillo: iptables module does not look to support the out_interface param for LIBVIRT_FWI (https://docs.ansible.com/ansible/latest/collections/ansible/builtin/iptables_module.html#parameter-out_interface). Moving to command module.

scenarios/reproducers/validated-architecture-shiftstack.yml
rlobillo: Moved the var to shift-on-stack-overrides.yml as suggested

scenarios/reproducers/validated-architecture-shiftstack.yml
rlobillo: Moved the var to shift-on-stack-overrides.yml as suggested

scenarios/reproducers/validated-architecture-shiftstack.yml
rlobillo: it's CPU, memory and disk. I'm creating the variables for these 3. Moved these vars to shift-on-stack-overrides.yml as suggested

rlobillo:
> While I understand the need, I think this PR is wrong, especially on the scenario file. You should add the parameter to raise the compute CPU as described in the comment, then create one single file, like "shift-on-stack-overrides.yml" and just add the specific overrides there. It would make the whole thing far, far easier to maintain...

Thanks for your comments, @cjeanner 

I addressed your comments and we are working on validate the changes. Please kindly take another look to the latest.

scenarios/reproducers/shift-on-stack-overrides.yml
```diff
@@ -0,0 +1,8 @@
+cifmw_libvirt_manager_compute_amount: 3
+cifmw_libvirt_manager_compute_disksize: 200
+cifmw_libvirt_manager_compute_memory: 50
+cifmw_libvirt_manager_compute_cpus: 8
+cifmw_allow_vms_to_reach_osp_api: true
+# HCI requires bigger size to hold OCP on OSP disks
+cifmw_block_device_size: 100G
+
```
cjeanner, line 8: that empty line will crash ansible-lint

scenarios/reproducers/shift-on-stack-overrides.yml
```diff
@@ -0,0 +1,8 @@
+cifmw_libvirt_manager_compute_amount: 3
```
cjeanner, line 1: this is the default value for the amount - no need to repeat it in this case

reproducer.yml
```diff
@@ -55,3 +55,10 @@
   roles:
     - role: ci_setup
     - role: reproducer
+
+  post_tasks:
+    - name: Allow traffic from OSP VMs to OSP API (needed for shiftstack)
+      ansible.builtin.command:
```
cjeanner, line 61: you'll need a `  # noqa: command-instead-of-module` in order to pass ansible-lint.

reproducer.yml
```diff
@@ -55,3 +55,10 @@
...
+        cmd: iptables -I LIBVIRT_FWI 1 -o ocpbm -j ACCEPT
+      become: true
```
cjeanner, line 63: care to add the `become: true` and `when: ...` just under the `name:` key? It makes things slightly more readable when we scroll over a receipt :).

reproducer.yml
cjeanner: dang... do they really filter the chains?

reproducer.yml
rlobillo: Done

scenarios/reproducers/shift-on-stack-overrides.yml
```diff
@@ -0,0 +1,6 @@
+cifmw_libvirt_manager_compute_disksize: 200
+cifmw_libvirt_manager_compute_memory: 50
+cifmw_libvirt_manager_compute_cpus: 8
+cifmw_allow_vms_to_reach_osp_api: true
+# HCI requires bigger size to hold OCP on OSP disks
+cifmw_block_device_size: 100G
```
cjeanner, line 6: I think this isn't the right parameter name? Shouldn't it be something like:
```YAML
cifmw_libvirt_manager_compute_disksize: 100
```

rlobillo: Yes, that variable is present in line 1. This is the size that the root disk will have: 200G.

Then, I need to set the variable **cifmw_block_device_size** which is the size for the virtual block device that will used by ceph on the computes: https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/cifmw_block_device/README.md#parameters

So, the outcome will be computes with 200G where 100G will be used by Ceph and the rest will be used by the compute itself.

PR #1261, opened 2024-03-13
[ci_setup] Fix error pertaining to GPG signature keys for EPEL repositories.

psathyan:
This change should resolve the below error

```
TASK [ci_setup : Install EPEL] *************************************************
fatal: [instance]: FAILED! => changed=false
  msg: 'Failed to validate GPG signature for epel-release-9-7.el9.noarch: Public key for epel-release-latest-9.noarch10woiruw.rpm is not installed'
```


psathyan:
recheck

PR #1204, opened 2024-02-28
Upgrade how hooks are defined

cjeanner:
There's a need to add single hooks to a list of existing ones, depending
on the scenario we want to run.

Until now, this wasn't possible, since parameters override themselves
and only the last occurrence wins.

This patch introduces a notion of "single hook definition" using crafted
parameters names, consuming specific prefixes.

As a pull request owner and reviewers, I checked that:
- [X] Appropriate testing is done and actually running
- [X] Appropriate documentation exists and/or is up-to-date:
  - [X] README in the role
  - [X] Content of the docs/source is reflecting the changes

PR #1262, opened 2024-03-13
Allow users to inject their own custom hooks

cjeanner:
This change empowers users with custom hooks injections during specific
stages in the Architecture deployment loop.

For instance, if a user wants to add their own hook as a "pre_stage",
for stage_id 2, they can create a custom parameter named:
`pre_stage_2_run_my_super_hook`. It will then be injected directly in
the existing pre_hook list of the stage.

If the user uses an index (0-9+ or a-z+) in the hook name, they can even ensure
the place their hook kicks in in the hook list:
`pre_stage_2_run_00_my_super_hook` has all the chances to be the very
first one, while `pre_stage_2_run_ZZ_my_super_hook` would be the last
one to run.

### DIFF
--- a/roles/kustomize_deploy/tasks/execute_step.yml
+++ b/roles/kustomize_deploy/tasks/execute_step.yml
@@ -38,13 +38,13 @@
       loop_control:
         label: "{{ item.stat.path | default('none') | basename }}"
 
-    - name: "Executing pre_deploy step for {{ stage.path }}"
+    - name: "Executing pre_stage hooks for {{ stage.path }}"
       when:
         - not cifmw_kustomize_deploy_generate_crs_only | bool
         - stage.pre_stage_run is defined
       vars:
         hooks: "{{ stage.pre_stage_run }}"
-        step: "pre_stage_run"
+        step: "pre_stage_{{ stage_id }}_run"
       ansible.builtin.include_role:
         name: run_hook
 
@@ -182,12 +182,12 @@
       ansible.builtin.fail:
         msg: "Failing on demand {{ cifmw_deploy_architecture_stopper }}"
 
-    - name: "Executing post_deploy step for {{ stage.path }}"
+    - name: "Executing post_stage hooks for {{ stage.path }}"
       when:
         - not cifmw_kustomize_deploy_generate_crs_only | bool
         - stage.post_stage_run is defined
       vars:
         hooks: "{{ stage.post_stage_run }}"
-        step: "post_stage_run"
+        step: "post_stage_{{ stage_id }}_run"
       ansible.builtin.include_role:
         name: run_hook


roles/kustomize_deploy/tasks/execute_step.yml
```diff
@@ -38,13 +38,13 @@
...
 
-    - name: "Executing pre_deploy step for {{ stage.path }}"
+    - name: "Executing pre_stage hooks for {{ stage.path }}"
       when:
         - not cifmw_kustomize_deploy_generate_crs_only | bool
         - stage.pre_stage_run is defined
       vars:
         hooks: "{{ stage.pre_stage_run }}"
-        step: "pre_stage_run"
+        step: "pre_stage_{{ stage_id }}_run"
```
SeanMooney, line 47: how does adding `{{ stage_id }}`
acomplish that.

was this functionaly effectivly implemented before and jsut no used?

cjeanner: it's a new feature of the `run_hook` role, added a couple of days ago:
https://github.com/openstack-k8s-operators/ci-framework/pull/1204

With that feature, you can now create a tailored parameter, for instance `pre_stage_0_my_hook` with the same `hook` structure, and it will be caught by the `step` value here, and injected in the existing list of hooks (may be empty, of course) for the current stage :).

Doc is on its way, but we need to split stable vs main first, and this will break doc bookmarks ppl may have, soooo... it will happen next week.

SeanMooney: i thnk that new feature broke the ordering of the old hooks

https://github.com/openstack-k8s-operators/nova-operator/pull/685/files#diff-da9898abacc1825df3ed68299b22727ff815db83146ae6c7bed904456ba18b54R12

i defined 
post_deploy:
  - name: Kustomize OpenStack CR with Ceph
    type: playbook
    source: control_plane_ceph_backends.yml
  - name: Kustomize and update Control Plane
    type: playbook
    source: control_plane_kustomize_deploy.yml

but if we look at the logs form the job 
we see

[TXT]	[ci_script_019_run_kustomize_and_update_control_plane.sh](https://logserver.rdoproject.org/85/685/a9a233eff0ff15427f34d0663ec7da45d60882f3/github-check/nova-operator-tempest-multinode-ceph/91040cf/controller/ci-framework-data/artifacts/ci_script_019_run_kustomize_and_update_control_plane.sh)	2024-03-14 01:46 	1.1K
[TXT]	[ci_script_020_apply_the_openstackcontrolplane_cr.sh](https://logserver.rdoproject.org/85/685/a9a233eff0ff15427f34d0663ec7da45d60882f3/github-check/nova-operator-tempest-multinode-ceph/91040cf/controller/ci-framework-data/artifacts/ci_script_020_apply_the_openstackcontrolplane_cr.sh)	2024-03-14 01:46 	259
[TXT]	[ci_script_021_run_kustomize_openstack_cr_with_ceph.sh](https://logserver.rdoproject.org/85/685/a9a233eff0ff15427f34d0663ec7da45d60882f3/github-check/nova-operator-tempest-multinode-ceph/91040cf/controller/ci-framework-data/artifacts/ci_script_021_run_kustomize_openstack_cr_with_ceph.sh)

so the ordering has been broken when you are using the hooks directly breaking the previous API contract, so any fo the sernairo files or downstream files that depended on the ordering of files in the hook being the order that was specified are likely not broken. 

https://github.com/openstack-k8s-operators/ci-framework/blob/main/scenarios/centos-9/hci_ceph_backends.yml#L8-L14hci_ceph_backends.yml#L8-L14

cjeanner: if order is important, you should use an index prefix, such as "01, 02, 03,..." in the hook name. Until now, we didn't face "ordered hooks", and it was mostly working "by chance" until we had to find a way to inject hooks to an existing list :/. Ansible doesn't support `|combine` in the parameter files, it ends with an infinite loop...

SeanMooney: as far as im aware all downstrema hook usage was depending on it being ordered and ordering was a part of the API contract so that's a disconnect between your users and your expectation.

we can enforce lexagraphic ordering but i guess this is one of the reasons that trying to use hooks ahs been problematic and why i have been avoidinging them.

cjeanner: Well, in any cases, this discussion has nothing to do with this current PR.... As for "contract", I don't remember describing anything related to the order hooks were running.

Also, with that other PR in, the order will be stable and fixed, while allowing to inject other hook in case some are already defined for the targeted step, while ensuring we can properly manage the order, whenever the hook is coming from.

Sometimes things must change, and sometimes we can't keep the old support. Especially with something moving so fast as the framework (not to mention the whole EDPM world) :/.

I don't understand why pushing lexicographic ordering is a problem, btw - you want order, show you want order, else let the thing `|sort` the content as-is. It's even better, since it provides a clear view about the needed order.

Anyway - as said: this isn't a discussion affecting this specific PR.

SeanMooney: oh i have no issue with doing lexagrapical ordering just calling out that the merged examples we have in the senairo files are not correct so now that we are being explicit with the new feature that you have to order them its easy for use to just update our josb to express that ordering.

i came across this PR when i was trying to debug why the ordering was not being preserved.

i have not see anyone use "01, 02, 03,..."  to date so i think that might explain some of the stability issues peopepl were having with trying to write jobs in some cases

your right this is not related to this pr so ill resolve this conversation.,

cescgina (approved):

PR #1263, opened 2024-03-13
Copy ansible logs from user's home dir to artifacts log dir

raukadah:
We run artifacts role on controller-0 in the BM reproducer job. We generate different ansible log files for nested ansible call done via reproducer.

This pr will help to collect these logs in the log collection and make debugging easier.

### DIFF
--- a/roles/artifacts/tasks/ansible_logs.yml
+++ b/roles/artifacts/tasks/ansible_logs.yml
@@ -1,0 +1,13 @@
+---
+- name: Generate list of ansible logs to collect in home directory
+  ansible.builtin.find:
+    paths: "{{ ansible_user_dir }}"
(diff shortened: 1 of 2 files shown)


roles/artifacts/tasks/ansible_logs.yml
```diff
@@ -0,0 +1,13 @@
+---
+- name: Generate list of ansible logs to collect in home directory
```
cjeanner, line 2: isn't this a repetition of https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci/playbooks/collect-logs.yml#L72-L84?

Note that I'm all for having that in the artifacts instead of zuul - but we may want to clean that other play?

PR #1266, opened 2024-03-13
standalone logging

SeanMooney:
- extend edpm log collection to standalone nodes
- update ansible lint to run locally

  - [x] README in the role

### DIFF
--- a/roles/artifacts/tasks/edpm.yml
+++ b/roles/artifacts/tasks/edpm.yml
@@ -42,11 +42,21 @@
             - cifmw_edpm_deploy_extra_vars is defined
           ansible.builtin.set_fact:
             ssh_key_file: "{{ cifmw_edpm_deploy_extra_vars.SSH_KEY_FILE }}"
-            ssh_user: "{{ hostvars['compute-0'].ansible_user | default('zuul') }}"
+            # ssh_user should be the same as the user used to deploy the edpm nodes
+            # in adoption jobs the edpm node is called standalone and in greenfield jobs it is called compute-0, compute-1, etc
+            # the default user is zuul, but it can be overridden by the user in the zuul inventory so we need to check both
+            # via hostvars and default user if not found
+            ssh_user: >-
+              {{
+                hostvars | dict2items |
+                selectattr('key', 'match', '^compute.*$|^standalone.*$') |
+                map(attribute='value.ansible_user') | first | default('zuul')
+              }}
+
             edpm_vms: >-
               {{
                 crc_ci_bootstrap_networks_out | dict2items |
-                selectattr('key', 'match', '^compute.*$') |
+                selectattr('key', 'match', '^compute.*$|^standalone.*$') |
                 map(attribute='value.default.ip') | default([])
               }}
 
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -3,6 +3,8 @@ profile: production
 exclude_paths:
   - ../
   - .cache/
+  - .vscode/  # and the vscode files are not part of the codebase so we should not lint them
+  - .venv/ # we shoudl not try to lint any yaml files found in the venv
   - build/
   - containerfiles/
   - docs/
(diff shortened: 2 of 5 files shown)


SeanMooney:
so we do not run the adoption jobs in this repo so ill add a DNM patch to the adoption repo with a depends on an we can use that to confirm if this does indeed fix the log collection for the standalone node as i expect. i could also drop the ansible-lint changes or move them to a different pr that fixes some but not all of the warnigs/slow execution issues i was having locally

roles/artifacts/tasks/edpm.yml
```diff
@@ -42,11 +42,21 @@
             - cifmw_edpm_deploy_extra_vars is defined
           ansible.builtin.set_fact:
             ssh_key_file: "{{ cifmw_edpm_deploy_extra_vars.SSH_KEY_FILE }}"
```
SeanMooney, line 44: ok so i am seeing this error

fatal: [localhost]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'dict object' has no attribute 'SSH_KEY_FILE'. 'dict object' has no attribute 'SSH_KEY_FILE'\n\nThe error appears to be in '/home/zuul/src/github.com/openstack-k8s-operators/ci-framework/roles/artifacts/tasks/edpm.yml': line 39, column 11, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n        - name: Extract Compute from zuul mapping if any\n          ^ here\n"}

so that means cifmw_edpm_deploy_extra_vars does not have SSH_KEY_FILE defined for standalone host.
ssh_key_file: "{{ cifmw_edpm_deploy_extra_vars.SSH_KEY_FILE }}"

SeanMooney: i wonder if we can just use ssh_key_file: "{{ cifmw_artifacts_basedir }}/artifacts/edpm/ansibleee-ssh-key-id_rsa"

or 
SSH_KEY_FILE: "{{ ansible_user_dir }}/.ssh/id_cifw"

https://github.com/openstack-k8s-operators/ci-framework/blob/64d1de7421b9e1d0211862bb6d9f2789ef288e0b/hooks/playbooks/fetch_compute_facts.yml#L105

looking at https://review.rdoproject.org/zuul/job/data-plane-adoption-osp-17-to-extracted-crc

we could use standalone_private_key in the adoptin case sepcificly but it woudl probaly be better to define cifmw_edpm_deploy_extra_vars.SSH_KEY_FILE as {{standalone_private_key}} instead to make it work as expected.

SeanMooney: ok this is normlaly set here

https://github.com/openstack-k8s-operators/ci-framework/blob/64d1de7421b9e1d0211862bb6d9f2789ef288e0b/hooks/playbooks/fetch_compute_facts.yml#L105

in edpm_install_yamls_vars and then that get added cifmw_edpm_deploy_extra_vars
here
https://github.com/openstack-k8s-operators/ci-framework/blob/64d1de7421b9e1d0211862bb6d9f2789ef288e0b/hooks/playbooks/fetch_compute_facts.yml#L116-L117

so if that is not being set computes_len is not > 0 

and since that defined here

https://github.com/openstack-k8s-operators/ci-framework/blob/64d1de7421b9e1d0211862bb6d9f2789ef288e0b/hooks/playbooks/fetch_compute_facts.yml#L33-L35

it means the standalone node is not part of the compute group in the inventory

that is being exectued as a hook so its not the zuul inventory its the inventory used by the ansible-playbook command in that is runing hte hook.

SeanMooney: the groups by defualt come from the nodeset and we can see in this case the nodeset does not define the standalone node to be part of the compute group 

https://github.com/openstack-k8s-operators/ci-framework/blob/64d1de7421b9e1d0211862bb6d9f2789ef288e0b/zuul.d/nodeset.yaml#L46-L63


so if i fix that it should resolve that issue and the key shoudl be set properly.

bogdando (commented):
lgtm, thanks!

cjeanner:
@SeanMooney heya! is this still relevant? If yes, care to rebase and advertise it on the support channel, or join the review call (Mon, Wed, Fri) to get some reviews? Thanks!

SeanMooney:
The logging situation for adoption has improved but it's still not the same as greenfield. With that said I'm not planning to work on this anytime soon so I'll closer this for now

PR #1269, opened 2024-03-14
Allow single hooks to be a list

pablintino:
  - [x] README in the role

### DIFF
--- a/roles/run_hook/tasks/main.yml
+++ b/roles/run_hook/tasks/main.yml
@@ -29,19 +29,20 @@
         selectattr('key', 'match', _matcher)
       }}
     _single_hooks: >-
-      {% for _hook in _filtered_hooks -%}
-      - name: {{ _hook.key | regex_replace(_matcher, '\1') |
-                 replace('_', ' ') | capitalize }}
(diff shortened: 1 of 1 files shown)


roles/run_hook/tasks/main.yml
```diff
@@ -29,32 +29,44 @@
...
+    - name: Assert single hooks are all mappings
+      vars:
+        _not_mapping_hooks: >-
+          {{
+            _single_hooks | reject('mapping') | list
+          }}
+      ansible.builtin.assert:
+        quiet: true
+        that:
+          - _not_mapping_hooks | length == 0
```
cjeanner, line 62: maybe a clear message here? "At least one single hook doesn't seem to be a mapping" or something like that?

rebtoor (approved):

roles/run_hook/tasks/main.yml
```diff
@@ -29,32 +29,46 @@
...
+        msg: >-
+          All single hooks must be contains a list of mappings or a mapping.
```
rebtoor, line 64: `must contains` or `must be`?

rebtoor (approved):

PR #1270, opened 2024-03-14
Rename "validated-architecture-1" to "va-hci"

cjeanner:
This will help understanding the purpose of that file -
"validated-architecture-1" was the only name back then, until we got the
"architecture" repository and the "hci" name.

New VA and DT should also get the same kind of nommenclature (va-FOO, or
dt-BAR).

The `validated-architecture-1.yml` symlink will be removed later, once
everyone get used to the new name.


psathyan (approved):

cescgina (approved):
lgtm

PR #1274, opened 2024-03-15
Deprecate cifmw_reproducer_internal_ca parameter

cescgina:
Deprecate cifmw_reproducer_internal_ca parameter in the reproducer to
use the install_ca role. We can use cifmw_install_ca_url to download
a CA bundle from a url, or cifmw_install_ca_bundle_src to use a local CA
bundle file.

  - [x] README in the role

### DIFF
--- a/roles/install_ca/tasks/main.yml
+++ b/roles/install_ca/tasks/main.yml
@@ -50,7 +50,11 @@
         mode: "0644"
 
     - name: Update ca bundle
+      vars:
+        _ca_bundle_changed: ca_bundle is defined and ca_bundle is changed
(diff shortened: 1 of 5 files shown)


cescgina:
> /approve
> 
> it looks sane - not sure if you want more testing and what kind of doc update you want to do, @cescgina ? Note that github wants _all_ of the available checkboxes to be checked - you can remove the non-relevant ones if needed.

yes @cjeanner I left the checkboxes unmarked as a sort of hold while I was testing this downstream. It failed a couple of times for unrelated reasons, but I was waiting for a clean run before merging this. I'll update the downstream docs after this merges as well

PR #1275, opened 2024-03-15
Increasing the default number attempts to have a successful login and API connectivity check.

psathyan:
We are now waiting for 8 minutes and 20 seconds. This should be sufficient for the OpenShift cluster to settle down.


cjeanner:
/approve

OCP cluster takes time to get up'n'running, and there's the certificate renewal to take into account.

PR #1276, opened 2024-03-15
Add new top-level parameter to skip stage

cjeanner:
Until now, we could skip a stage only by leveraging `--skip-tags`
ansible parameter.

With this patch, we create a new parameter, allowing to list the stages
we want to skip. If passed inline
(-e cifmw_deploy_architecture_skip_stages=deploy_architecture_stage_0)
it will be `|split(',')` in order to support multiple stages to skip.

The tag is still supported.


rlandy (commented):
/lgtm

raukadah (approved):
/approve

PR #1277, opened 2024-03-15
Correct generated scripts from reproducer role

cjeanner:
`deploy-architecture.sh` was loading a generated parameter file, leading
to potential issues when a user is iterating deploys directly on
controller-0 and is passing parameters to the deployment: those
parameters would end in that generated file, and therefore would be
loaded on each run.
This patch corrects this, and ensure we're loading a "vanilla" generated
file, from the reproducer run itself.

`deploy-edpm.sh` script wasn't properly installed on the controller-0
due to a missing `delegate_to: controller-0` on the task. This wasn't
caught in CI, since we're using `zuul` user for both "hypervisor" and
"controller-0".
This patch adds the missing instruction, and the script lands in the
right place now.

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/1281

### DIFF
--- a/roles/reproducer/tasks/devscripts_post.yml
+++ b/roles/reproducer/tasks/devscripts_post.yml
@@ -35,6 +35,11 @@
             (_auth_path, 'kubeadmin-password') | ansible.builtin.path_join
           }}
 
+    - name: Ensure we expose openshift_login related facts
+      ansible.builtin.import_role:
+        name: devscripts
+        tasks_from: set_cluster_fact.yml
+
     - name: Ensure the OpenShift cluster is stable.
       vars:
         cifmw_openshift_adm_op: "stable"
--- a/roles/reproducer/tasks/main.yml
+++ b/roles/reproducer/tasks/main.yml
@@ -136,6 +136,7 @@
       ansible.builtin.include_tasks: configure_architecture.yml
 
     - name: Prepare ci-like EDPM deploy
+      delegate_to: controller-0
       ansible.builtin.copy:
         dest: "/home/zuul/deploy-edpm.sh"
         mode: "0755"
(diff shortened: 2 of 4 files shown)


roles/reproducer/tasks/main.yml
```diff
@@ -148,5 +149,6 @@
           ansible-playbook deploy-edpm.yml \
             -e @scenarios/centos-9/base.yml \
             -e @scenarios/centos-9/edpm_ci.yml \
-            -e cifmw_openshift_password=12345678 $@
+            -e cifmw_openshift_password="{{ cifmw_openshift_password }}" \
```
psathyan, line 152: molecule is complaining about `undefined variable`

cjeanner: need some more testing, I'll work on that on Monday

cjeanner:
recheck

roles/reproducer/tasks/devscripts_post.yml
```diff
@@ -35,6 +35,11 @@
             (_auth_path, 'kubeadmin-password') | ansible.builtin.path_join
           }}
 
+    - name: Ensure we expose openshift_login related facts
```
cjeanner, line 38: So it may not be needed - but to be fair, I didn't have time to test without it... It doesn't really hurt at that point imho, so we should be OK, unless anyone has time to make a quick test before?

@psathyan any thoughts?

raukadah:
/approve
thank you cedric for fixing it :-)

roles/reproducer/tasks/devscripts_post.yml
psathyan: This shouldn't block us.

psathyan (approved):

PR #1281, opened 2024-03-18
Ensure we're able to get OpenShift cluster facts

cjeanner:
devscripts role had a nice way to re-inject the OpenShift cluster
credentials in the facts. While role_crc also did it, it was only
"runtime".

With this patch, we add a new tasks file to openshift_login, allowing to
load/read the various files, and inject the needed values as facts.

This ensures we can export those needed facts even in the case of a
re-deployement (this is for a follow-up).


cjeanner:
recheck

psathyan (approved):

psathyan (approved):

cjeanner:
recheck

cjeanner:

all jobs passed at least once.

PR #1285, opened 2024-03-18
Ensure we install the dependencies soon enough

cjeanner:
The `pip install` happened too late, we need dependencies in order to
get ansible related libraries in


cescgina (approved):
seems reasonable :)

PR #1289, opened 2024-03-19
 Discover controller IP address and priv key

lpiwowar:
**Discover controller IP address and private key**

Whitebox-neutron-tempest-plugin requires controller node IP address and
private key that can be used to connect to it. This patch discovers
these values and passes them to the Tempest CR so that they can be
consumed by the plugin.

**Make cifmw_test_operator_*_config modifiable**

When a consumer of the role runs the test_operator role with:

-e cifmw_test_operator_tempest_config=value

then the role can not update the value of the variable as extra vars
have higher precedence in ansible. This patch changes the structure
of the test_operator role so that we always work with a copy of this
variable.

  - [x] README in the role

### DIFF
--- a/roles/test_operator/tasks/tempest-tests.yml
+++ b/roles/test_operator/tasks/tempest-tests.yml
@@ -56,3 +56,52 @@
     - name: Set variable
       ansible.builtin.set_fact:
         cifmw_test_operator_tempest_exclude_list: "{{ list_skipped.skipped_tests | join('\n') }}"
+
+- name: Ensure a secret for the cifmw private key file exists
+  when:
+    - not cifmw_test_operator_dry_run | bool
+    - cifmw_test_operator_tempest_ssh_key_secret_name is not defined
+  kubernetes.core.k8s:
+    kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
+    api_key: "{{ cifmw_openshift_token | default(omit) }}"
+    context: "{{ cifmw_openshift_context | default(omit) }}"
+    state: present
+    wait: true
+    definition:
+      apiVersion: v1
+      kind: Secret
+      type: Opaque
+      metadata:
+        name: "{{ cifmw_test_operator_controller_priv_key_secret_name }}"
+        namespace: "{{ cifmw_test_operator_namespace }}"
+      data:
+        ssh-privatekey: "{{ lookup('file', '~/.ssh/id_cifw') | b64encode }}"
+
+- name: Add SSHKeySecretName section to tobiko CR
+  when:
+    - not cifmw_test_operator_dry_run | bool
+    - cifmw_test_operator_tempest_ssh_key_secret_name is not defined
+  ansible.builtin.set_fact:
+    cifmw_test_operator_tempest_config: >-
+      {{
+          cifmw_test_operator_tempest_config |
+          combine({'spec': {'SSHKeySecretName':
+                    cifmw_test_operator_controller_priv_key_secret_name
+                  }}, recursive=true)
+      }}
+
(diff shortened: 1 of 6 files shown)


mhuin:
recheck

morucci:
recheck

roles/test_operator/tasks/tempest-tests.yml
```diff
@@ -56,3 +56,56 @@
...
+      {{
+          cifmw_test_operator_tempest_config |
+          combine({'spec': {'tempestconfRun': {'overrides':
+                    'whitebox_neutron_plugin_options.proxy_host_address ' +
+                    cifmw_test_operator_controller_ip
+                  }}}, recursive=true)
+      }}
+  when:
+    - not cifmw_test_operator_dry_run | bool
+    - cifmw_test_operator_controller_ip != ""
```
pablintino, line 111: No need to create to set_fact tasks for this. `cifmw_test_operator_controller_ip` can be variable in the second task and it will continue to work the same way. Less tasks -> less logs and less run time.

roles/test_operator/tasks/tempest-tests.yml
```diff
@@ -56,3 +56,56 @@
...
+        namespace: "{{ cifmw_test_operator_namespace }}"
+      data:
+        ssh-privatekey: "{{ lookup('file', '~/.ssh/id_cifw') | b64encode }}"
+  when:
+    - not cifmw_test_operator_dry_run | bool
+    - cifmw_test_operator_tempest_ssh_key_secret_name is not defined
+
+- name: Add SSHKeySecretName section to tobiko CR
+  ansible.builtin.set_fact:
+    cifmw_test_operator_tempest_config: >-
```
pablintino, line 82: Keep in mind that is someone passes the `cifmw_test_operator_tempest_config` as variable in the ansible-playbook call, using `-e @<path-that-contains-that>` that value won't be updated and this set_fact and its combine wont' work, as the following tasks will still see the "old" value comming from the file passed in the cli.

lpiwowar: Thank you for pointing this out. Hopefully addressed in the second patch in this PR. If you think that there is nicer solution than let me know please:).

lpiwowar:
recheck

roles/test_operator/tasks/tempest-tests.yml
```diff
@@ -56,3 +56,53 @@
...
+                    (test_operator_cr.spec.tempestconfRun.overrides | default('')) + ' ' +
+                    'whitebox_neutron_plugin_options.proxy_host_address ' + controller_ip
+                  }}}, recursive=true)
+      }}
+  when:
+    - not cifmw_test_operator_dry_run | bool
+    - controller_ip != ""
+  vars:
+    controller_ip: >-
+      {{ cifmw_test_operator_controller_ip | default(ansible_default_ipv4.address) | default('') }}
```
pablintino, line 108: Cosmetic/consistency comment: We usually put then clauses and vars at the very first of the task definition. Easier to read as you can skip reading the task as soon as you know the when condition is false.

pablintino:
recheck

roles/test_operator/tasks/tempest-tests.yml
lpiwowar: Thank you, fixed! I'm going to stick with this rule next time I'll be proposing a PR.

kopecmartin (approved):
lgtm

m-blue-m:
Hi @lpiwowar , thank you for all the changes.
It seems the private ECDSA key file doesn't have a new line at the end, and this is enough for ssh and libcrypto to throw error message, and test pod is read-only file system.

PR #1291, opened 2024-03-19
Add default values for ceph playbook in adoption

cescgina:
The adoption nodeset does not have any compute nodes, and the empty
computes groups since to crash the ceph playbook that is imported but
not run there.


cjeanner:
(I'm trying a different approach with #1294 - if it works, I'd propose it for real)

fmount (approved):
/lgtm

PR #1296, opened 2024-03-19
Add application_credential module

steveb:
Create or delete a Keystone application credential.  When the secret parameter is not set a secret will be generated and returned in the response. Existing credentials cannot be modified so running this module against an existing credential will result in it being deleted and recreated. This needs to be taken into account when the secret is generated, as the secret will change on each run of the module.

This will be removed when [1] lands. ci-bootstrap will use this module to create application credentials for the duration of the job. This will allow a sushy-tools container to have host cloud credentials which are not the primary cloud credentials.

[1] https://review.opendev.org/c/openstack/ansible-collections-openstack/+/910463

Change-Id: I0ed86dc8785b0e9d10cc89cd9137a11d02d03945

  - [x] README in the role

### DIFF
--- a/plugins/modules/application_credential.py
+++ b/plugins/modules/application_credential.py
@@ -1,0 +1,332 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright (c) 2024 Red Hat, Inc.
+# GNU General Public License v3.0+
+# (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+DOCUMENTATION = r"""
+---
+module: application_credential
+short_description: Manage OpenStack Identity (Keystone) application credentials
+author: OpenStack Ansible SIG
+description:
+  - Create or delete an OpenStack Identity (Keystone) application credential.
+  - When the secret parameter is not set a secret will be generated and returned
+  - in the response. Existing credentials cannot be modified so running this module
+  - against an existing credential will result in it being deleted and recreated.
+  - This needs to be taken into account when the secret is generated, as the secret
+  - will change on each run of the module.
+options:
+  name:
+    description:
+      - Name of the application credential.
+    required: true
+    type: str
+  description:
+    description:
+      - Application credential description.
+    type: str
+  secret:
+    description:
+      - Secret to use for authentication
+      - (if not provided, one will be generated).
+    type: str
+  roles:
+    description:
+      - Roles to authorize (name or ID).
+    type: list
+    elements: dict
+    suboptions:
+      name:
+        description: Name of role
+        type: str
+      id:
+        description: ID of role
+        type: str
+      domain_id:
+        description: Domain ID
+        type: str
+  expires_at:
+    description:
+      - Sets an expiration date for the application credential,
+      - format of YYYY-mm-ddTHH:MM:SS
+      - (if not provided, the application credential will not expire).
+    type: str
+  unrestricted:
+    description:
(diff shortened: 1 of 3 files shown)


steveb:
recheck
/retest
github connection failures

steveb:
Here is the ci-bootstrap PR which uses this module https://github.com/openstack-k8s-operators/ci-bootstrap/pull/52

plugins/modules/application_credential.py
```diff
@@ -0,0 +1,332 @@
...
+
+        roles = self.params["roles"]
+        if roles:
+            kwroles = []
+            for role in roles:
+                kwroles.append(
+                    dict(
+                        (k, role[k])
+                        for k in ["name", "id", "domain_id"]
+                        if role[k] is not None
```
pablintino, line 291: Shouldn't this be `if k in role`?

plugins/modules/application_credential.py
```diff
@@ -0,0 +1,332 @@
...
+                    )
+                )
+            kwargs["roles"] = kwroles
+
+        kwargs["user"] = self._get_user_id()
+        creds = self.conn.identity.create_application_credential(**kwargs)
+        return creds
+
+    def _get_cloud_config(self, creds):
+        cloud_region = openstack.config.OpenStackConfig().get_one(self.cloud)
```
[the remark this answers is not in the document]
steveb: Graceful openstack import is handled by the parent class https://opendev.org/openstack/ansible-collections-openstack/src/branch/master/plugins/module_utils/openstack.py#L161-L166

plugins/modules/application_credential.py
steveb: The aim is to skip attributes which exist in `role` but have a `None` value. It would be safer to do `if role.get(k) is not None`, and add some coverage to the [upstream tests](https://review.opendev.org/c/openstack/ansible-collections-openstack/+/910463/12/ci/roles/application_credential/tasks/main.yml).

However I have no plan to use the role field in ci-bootstrap, so impact is minimal. This module will be deleted from ci-framework once the upstream one is consumable, but that could be some time.

lewisdenny (approved):
lgtm, thanks Steve

lewisdenny:
One comment you can take or leave is adding a named todo into the module stating it can be removed after https://review.opendev.org/c/openstack/ansible-collections-openstack/+/910463 lands in case we forget :)

plugins/modules/application_credential.py
steveb: Actually `if role[k] is not None` is fine. The ansible argument schema inserts None values for missing attributes, and the keystone API errors on None values, so this exists to clean that up

steveb:
> One comment you can take or leave is adding a named todo into the module stating it can be removed after https://review.opendev.org/c/openstack/ansible-collections-openstack/+/910463 lands in case we forget :)

I will add a TODO to [the ci-bootstrap change](https://github.com/openstack-k8s-operators/ci-bootstrap/pull/52) which includes a specific ci-framework commit to revert after this lands

plugins/modules/application_credential.py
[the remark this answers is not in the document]
pablintino: Nice, thanks for the poiner.

plugins/modules/application_credential.py
pablintino: `The ansible argument schema inserts None values for missing attributes` That was the point I was missing, thanks.

PR #1298, opened 2024-03-20
Use test operator for running tempest

raukadah:
tempest role does not give much info about what tests are running before exit.
Tempest container gets stuck for longer period of time.

test_operator is going to be the way to run tempest. Let's use that one. It has better error message and reporting via k8s jobs.

Let's use that one for running.


raukadah:
recheck

PR #1300, opened 2024-03-20
Support deploying an NFS server in a EDPM node

cescgina:
Support deploying an NFS server from an EDPM node, by default  the first
compute node. The nfs.yml playbook setups the server, and is called from
the 06-deploy-edpm.yml plabyook bedore the control plane is deployed.


playbooks/06-deploy-edpm.yml
```diff
@@ -6,6 +6,12 @@
     step: pre_deploy
   ansible.builtin.import_playbook: ./hooks.yml
 
+- name: Deploy NFS server on target nodes
+  when:
+    - cifmw_edpm_deploy_nfs | default('false') | bool
+    - cifmw_architecture_scenario is undefined
```
tosky, line 12: I think this may be useful also when an architecture is used

playbooks/06-deploy-edpm.yml
```diff
@@ -6,6 +6,12 @@
...
+  ansible.builtin.import_playbook: "ceph.yml"
```
tosky, line 13: nfs.yml?

cescgina: indeed, good catch, thanks!

playbooks/06-deploy-edpm.yml
cescgina: maybe, I copied the conditions from the ceph one, removed it now

cescgina:
Tested it with an ad-hoc job in https://review.rdoproject.org/r/c/testproject/+/52395. Added a post-run playbook to verify that the nfs server can be mounted from the controller and can create files properly https://review.rdoproject.org/zuul/build/fed291350c654578b3c19688a9e2d68f/log/job-output.txt#7270

playbooks/06-deploy-edpm.yml
```diff
@@ -6,6 +6,11 @@
     step: pre_deploy
   ansible.builtin.import_playbook: ./hooks.yml
 
+- name: Deploy NFS server on target nodes
+  when:
```
cjeanner, line 10: you want to move that `when` in a pre_tasks of the imported playbook, else we'll end with a list of "skipping" tasks.

playbooks/nfs.yml
```diff
@@ -0,0 +1,100 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+- name: Deploy an NFS server
+  become: true
+  hosts: "{{ groups[cifmw_nfs_target | default('computes')][0] | default([]) }}"
```
cjeanner, line 19: here, inject this:
```YAML
   pre_tasks:
     - name: End play early if no NFS is needed
       when: 
         - not cifmw_edpm_deploy_nfs | default('false') | bool
       ansible.builtin.meta: end_play
```

playbooks/nfs.yml
```diff
@@ -0,0 +1,100 @@
...
+        state: restarted
+        enabled: true
+
+    - name: Add shares to /etc/exports
+      ansible.builtin.lineinfile:
+        path: /etc/exports
+        line: "/data/{{ item }} {{ _nfs_network_range }}(rw,sync,no_root_squash)"
+      loop: "{{ cifmw_nfs_shares }}"
+
+    - name: Export the shares
```
cjeanner, line 98: maybe reload it if the `/etc/exports` is changed only?

playbooks/06-deploy-edpm.yml
cescgina: I was waiting for your PR to use the same pattern here :)

playbooks/nfs.yml
cescgina: that's a good idea, thanks!

playbooks/nfs.yml
```diff
@@ -0,0 +1,107 @@
...
+# License for the specific language governing permissions and limitations
+# under the License.
+
+- name: Deploy an NFS server
+  become: true
+  hosts: "{{ groups[cifmw_nfs_target | default('computes')][0] | default([]) }}"
+  pre_tasks:
+    - name: End play early if no NFS is needed
+      when:
+        - not cifmw_edpm_deploy_nfs | default('false') | bool
```
cjeanner, line 23: hmm and we should document that new top-level parameter I think (sorry for not mentioning it earlier - too many things on my plate).
Somewhere in https://github.com/openstack-k8s-operators/ci-framework/blob/main/docs/source/usage/01_usage.md

cescgina: True, I introduce a few parameters in this change, I'll add them all to the docs and update once the current CI run finishes, thanks for the link!

docs/source/usage/01_usage.md
```diff
@@ -38,12 +38,16 @@ are shared among multiple roles:
 - `cifmw_architecture_repo`: (String) Path of the architecture repository on the controller node.
   Defaults to `~/src/github.com/openstack-k8s-operators/architecture`
 - `cifmw_architecture_scenario`: (String) The selected VA scenario to deploy.
-- `cifmw_ceph_target`: (String) the Ansible inventory group where ceph is deployed. Defaults to `computes`.
+- `cifmw_ceph_target`: (String) The Ansible inventory group where ceph is deployed. Defaults to `computes`.
```
tosky, line 41: uh, those are unrelated changes, shouldn't they go in their own change? And why are they showing up just now? A change in the linter?

cescgina: I just fixed a few typos. The parameter description is the same, I just changed them so the first letter is uppercase to keep the docs page consistent

cescgina:
recheck

cescgina:
recheck

raukadah:
recheck

playbooks/nfs.yml
```diff
@@ -0,0 +1,107 @@
...
+    - name: Install nfs-utils package
+      ansible.builtin.package:
+        name: "nfs-utils"
+
+    - name: Configure nfs to use v4 only
+      community.general.ini_file:
+        path: /etc/nfs.conf
+        section: nfsd
+        option: vers3
+        value: n
```
marios, line 35: i was mainly curious about the value of 'n' for this but it doesn't seem like a typo

So I guess, vers3 includes ipv6 so disabling it means we only get v4? At first I thought this was about IPv4 but probably it is about the version of NFS protocol? 

(reference for others https://www.man7.org/linux/man-pages/man8/rpc.nfsd.8.html)

marios (commented):
looks right but holding approve until today's reviews call we can merge then?

cjeanner:
recheck

playbooks/nfs.yml
cescgina: that's right, we only want v4, so this option disables v3. Openstack components only support v4, that's the underlying reason for this config

cescgina:
recheck

cescgina:
recheck

cescgina:
recheck

danpawlik:
@cescgina we need to wait for the Pablo PS results: https://github.com/openstack-k8s-operators/ci-framework/pull/1339 . If that will work, its needs to be merged first

cjeanner:
/lgtm

re-adding dropped label after rebase.

PR #1301, opened 2024-03-20
Ensure we keep track of generated content

cjeanner:
It may happen we want to generate twice the same values.yaml, but with
maybe different content - this is especially true in the VA automation,
for instance two stages share the same `src_file` and the same
`value.name`, but the output is different.

In such as case, we want to consume the exact same `values.yaml`
(untouched), the same CI generaged content, and maybe a different user
provided kustomization.

While it was 100% supported, we may lose track of the history of the
changes. This patch therefore ensures we:
- backup the files before override
- exclude those backup files from the final `values.yaml` generation

### DIFF
--- a/roles/ci_gen_kustomize_values/tasks/generate_snippets.yml
+++ b/roles/ci_gen_kustomize_values/tasks/generate_snippets.yml
@@ -88,6 +88,7 @@
         [_tmpl_path, _common_tmpl_path] | select('exists')
       }}
   ansible.builtin.template:
+    backup: true
     dest: >-
       {{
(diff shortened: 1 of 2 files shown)


fultonj (commented):
+2

cjeanner:
/hold

I tested the `backup: true` but I want to test the "filter" on the `find` - this wasn't in my local test..

cjeanner:
/hold cancel

seems to work properly. Molecule is running against that code, ensuring the filter is find - which is the heart of that role.

rebtoor (approved):

PR #1307, opened 2024-03-21
Exclude bots PR from checklist workflow

cjeanner:
With this PR, dependabot[bot] and renovate[bot] PR shouldn't require the
checklist workflow anymore.

- [X] according to the doc, it should work.


cjeanner:
Merging. Prow is dead again, and this change doesn't affect any CI but internal github workflows - and it's passing.

PR #1311, opened 2024-03-21
Donot generate nova migration key if exists

raukadah:
If we re-run the deploy-architecture.sh script. Then Create nova migration keypair task fails with error:
```
Unable to read the key. The key is protected with a passphrase or broken.
```

In order to fix this issue, we donot need to regenerate the key but to reuse it.

This pr checks for nova migration key if exists. Reuse the same.

### DIFF
--- a/playbooks/06-deploy-architecture.yml
+++ b/playbooks/06-deploy-architecture.yml
@@ -27,14 +27,24 @@
       delegate_to: controller-0
       delegate_facts: true
 
-    - name: Create nova migration keypair
+    - name: Look for nova migration keypair file
+      register: _nova_key_file
+      ansible.builtin.stat:
+        path: "{{ cifmw_basedir }}/artifacts/nova_migration_key"
+
+    - name: Set _nova_key var when nova migration keypair exists
+      when: _nova_key_file.stat.exists
+      ansible.builtin.set_fact:
+        _nova_key:
+          filename: "{{ cifmw_basedir }}/artifacts/nova_migration_key"
+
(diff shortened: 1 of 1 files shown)


playbooks/06-deploy-architecture.yml
```diff
@@ -27,7 +27,14 @@
       delegate_to: controller-0
       delegate_facts: true
 
-    - name: Create nova migration keypair
+    - name: Look for nova migration keypair file
+      register: _nova_key_file
+      ansible.builtin.stat:
+        path: "{{ cifmw_basedir }}/artifacts/nova_migration_key"
```
lewisdenny, line 33: I would either update line 41 to match line 33 or vice-versa

lewisdenny (changes requested):

playbooks/06-deploy-architecture.yml
```diff
@@ -27,7 +27,20 @@
...
+
+    - name: Set _nova_key var when nova migration keypair exists
+      when: _nova_key_file.stat.exists
+      ansible.builtin.set_fact:
+        _nova_key:
+          filename: "{{ _nova_key_file.stat.path }}"
```
cjeanner, line 39: you want to use the same "{{ _nova_key_file.stat.path }}" on line 47 so that we avoid repetition (I think this is what @lewisdenny said)

lewisdenny (approved):
/lgtm

Thanks Chandan!

cjeanner:
/lgtm

I'm too fast...

raukadah:
recheck

PR #1312, opened 2024-03-21
Use a bigger flavor for standalone node in adoption

cescgina:
The standalone deployment seems to fail with what seems to be
resource-related conditions, so trying to use a biggere flavor there
with more RAM.

### DIFF
--- a/zuul.d/nodeset.yaml
+++ b/zuul.d/nodeset.yaml
@@ -51,7 +51,7 @@
       - name: crc
         label: coreos-crc-extracted-2-30-0-3xl
       - name: standalone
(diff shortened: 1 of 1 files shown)


lewisdenny:
recheck failed on: `wait for dataplane node set to be ready`

marios (approved):
yes lets

currently 8gb vm is not enough, especially if we are expecting to launch nested vms on that

marios:
recheck

marios:
as discussed in call now with Ronelle and Joan we are going to force merge this one.

it just touches the label for the adoption job so we feel it is safe to merge and should help with our ongoing cix

PR #1315, opened 2024-03-21
Rebased: [must-gather]Collect SOS reports from EDPM nodes

bogdando:
- [x] Appropriate documentation exists and/or is up-to-date https://github.com/openstack-k8s-operators/openstack-must-gather/pull/18 :
  - [x] README in the role

JIRA: [OSPRH-2685](https://issues.redhat.com/browse/OSPRH-2685)

### DIFF
--- a/roles/os_must_gather/tasks/main.yml
+++ b/roles/os_must_gather/tasks/main.yml
@@ -55,6 +55,7 @@
       environment:
         KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
         PATH: "{{ cifmw_path }}"
+        SOS_EDPM: "all"
       cifmw.general.ci_script:
         output_dir: "{{ cifmw_os_must_gather_output_dir }}/artifacts"
         script: >-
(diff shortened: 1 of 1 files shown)


bogdando:
tested ini CI logs https://logserver.rdoproject.org/15/1315/ccd2b7951fe588bfbee85e6e1dd62f1db5033efe/github-check/podified-multinode-edpm-deployment-crc/38deb1f/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/sos-reports/_all_nodes/sosreport-compute-0.ci-rdo.local/

fmount (approved):
/lgtm

fmount:
I just approved [PR#33](https://github.com/openstack-k8s-operators/openstack-must-gather/pull/33) in `openstack-must-gather`. This will start using the `openstack_edpm` profile when `sosreports` are gathered from an `edpm` node.
Of course for upstream this requires `sos` version `4.6.1`, otherwise it will fail. 
I didn't check what we have here, but at this point, if `ci-framework` needs to enable gathering from `edpm` nodes and `sos-4.6.1` is not there, we have two choices:

1. install `epel-release` and update `sos` (`dnf install -y epel-release && dnf install -y sos`)
2. use a `copr` repository like [1] to update the package during the `edpm` bootstrap (`dnf install -y dnf-plugins-core && dnf -y copr enable fmount/sos && dnf install -y sos`)

This patch should be updated (but it's ok to do in a follow up) to check if we have that version available and take care about temporarily update it until it's present in the base OS repos. 
AFAIK we don't have this problem downstream because the right version has already been shipped for `RHEL`. 

/cc @rebtoor @pablintino who might have opinions on this.

[1] https://copr.fedorainfracloud.org/coprs/fmount/sos/

rebtoor (approved):

bogdando:
> Of course for upstream this requires `sos` version `4.6.1`, otherwise it will fail.
> I didn't check what we have here

I checked the sos version in cs9 stream container image, and it proposes 
sos.noarch 4.6.1-1.el9 from baseos repo, so no follow ups needed

fmount:
@bogdando thank you for checking!

PR #1325, opened 2024-03-22
increase cifw_edpm_deploy_timeout to 60mins

SeanMooney:
in ci we see that the current time to deploy the dataplane
can be very close to the current time out of 40m. we are currently
trying to enable tls on the edpm nodes by default which slightly increase
the time to deploy the dataplane. this change increase the default
timeout to 60mins to accomidate that new default.

as this will affect all jobs the change is done in the role
defaults.


raukadah:
recheck

raukadah:
recheck

PR #1294, opened 2024-03-19
Stop ceph playbooks as early as possible

cjeanner:
With `import_playbook` ansible stills goes through all of the tasks
listed in the plays, but will `skip` them, leading to some time loss,
and potential issues with the expected inventory we have to consume.

We also ensure the HCI path of 06-deploy-edpm.yml isn't `imported` in
case we don't want to run it, switching to `include_*` instead. This
will ensure we don't have a long list of skipped tasks in the logs.

The default bahavior of the `ceph.yml` playbook isn't affected, meaning
we don't need to pass any parameter to NOT deploy it.

As a pull request owner and reviewers, I checked that:
- [x] Appropriate testing is done and actually running

PR #1326, opened 2024-03-24
Establish framework for QE Validation jobs

bshephar:
We have differing requirements for QE validation jobs depending on the team. This change aims to establish a framework by which each team will be able to define their QE validation job. In doing so, the ci-framework will be the extensible "tool to rule them all" from upstream CI to downstream QE.

Accordingly, we establish a new role called `qe_validations`. Under which we have tasks/<team> directories where validation jobs can be defined. We can call these are part of a post-deployment execution in the ci-framework. Essentially this would look like:

1. Normal deployment job is executed - as it would be today in ci
2. Post deployment will call a task from the qe_validations role to validate that certain conditions have been satisfied.
3. These qe_validation tasks should `assert` a state is satisfied. Assuming all assertions are satisfied, the job will be considered passed.

  - [x] README in the role

### DIFF
--- a/playbooks/06-deploy-edpm.yml
+++ b/playbooks/06-deploy-edpm.yml
@@ -151,3 +151,7 @@
   vars:
     step: post_deploy
   ansible.builtin.import_playbook: ./hooks.yml
+
+- name: Validations workflow
+  ansible.builtin.import_playbook: validations.yml
+  when: cifmw_execute_validations | default('false') | bool
--- a/playbooks/validations.yml
+++ b/playbooks/validations.yml
@@ -1,0 +1,8 @@
+- name: Execute the validations role
+  hosts: "{{ cifmw_target_host | default('localhost') }}"
+  gather_facts: false
+  tasks:
+
+    - name: Run validations
+      ansible.builtin.include_role:
+        name: validations
--- a/roles/validations/README.md
+++ b/roles/validations/README.md
@@ -1,0 +1,29 @@
+# validations
+The Validations role provides a framework for performing state verification.
+This is useful for QE purposes and will be leveraged post-deployment to assert
+that required conditions have been satisfied. For example, if I update a certain
+value in my `OpenStackDataPlaneNodeSet`. I want to validate that the change has
+been applied as per my expectations on the relevant OpenStack Compute nodes.
+
+## Privilege escalation
+Privilege escalation will be dependent on the specific job being executed. Some jobs
+may require privilege escalation. Check with the individual jobs to determine
+requirements.
+
+## Parameters
+The parameters will ultimately depend on what is implemented in each validation task.
+But in order to launch this role, at a minimum the following should be defined:
+
+- `cifmw_validations_list`: (List) List of validation jobs to run. Found under `roles/validations/tasks/*`
+- `cifmw_validations_run_all`: (Bool) Defaults to false
+- `cifmw_validations_default_path`: (String) Defaults to `"{{ role_path }}/tasks"`
+
+## Examples
+The following variables would be used to trigger the EDPM job for Huge Pages:
+```yaml
+  vars:
+    cifmw_execute_validations: true
+    cifmw_validations_list:
+      - edpm/hugepages_and_reboot.yml
+    cifmw_validations_edpm_check_node: 192.168.122.100
+```
--- a/roles/validations/tasks/main.yml
+++ b/roles/validations/tasks/main.yml
@@ -1,0 +1,47 @@
+---
+# Copyright Red Hat, Inc.
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+# We can execute all defined validations when cifmw_validation_run_all is defined.
+# Else, we will skip this and run only the explicitly defined validations from
+# cifmw_validations_list
+- name: Run all validations if required
+  when: cifmw_validations_run_all | bool
+  block:
+    - name: Find all validations
+      register: found_validations
+      ansible.builtin.find:
+        paths: "{{ cifmw_validations_default_path }}"
+        patterns: '*.yml'
+        recurse: true
+
+    - name: Run all found validations
+      ansible.builtin.include_tasks:
+        file: "{{ item.path }}"
+      loop: "{{ found_validations.files }}"
+
+- name: Run selected validations
+  when: not cifmw_validations_run_all | bool
+  block:
+    - name: Assert all listed validations exist
+      ansible.builtin.stat:
+        path: "{{ cifmw_validations_default_path }}{{ item }}"
+      loop: "{{ cifmw_validations_list }}"
+      register: validation_exists
+      failed_when: not validation_exists.stat.exists
+
+    - name: Run validations
+      ansible.builtin.include_tasks: "{{ item }}"
+      loop: "{{ cifmw_validations_list }}"
--- a/roles/validations/tasks/networking/main.yml
+++ b/roles/validations/tasks/networking/main.yml
@@ -1,0 +1,19 @@
+---
+# Copyright Red Hat, Inc.
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+- name: Hello World
+  ansible.builtin.debug:
+    msg: "Hello world from {{ role_name }}"
--- a/playbooks/06-deploy-architecture.yml
+++ b/playbooks/06-deploy-architecture.yml
@@ -153,3 +153,7 @@
   vars:
     step: post_deploy
   ansible.builtin.import_playbook: ./hooks.yml
+
+- name: Validations workflow
+  ansible.builtin.import_playbook: validations.yml
+  when: cifmw_execute_validations | default('false') | bool
--- a/roles/validations/defaults/main.yml
+++ b/roles/validations/defaults/main.yml
@@ -1,0 +1,34 @@
+---
+# Copyright Red Hat, Inc.
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+
+# All variables intended for modification should be placed in this file.
+# All variables within this role should have a prefix of "cifmw_validations"
+
+
+# cifmw_validations_list can be used to provide a list of validations that should
+# be executed For example.
+# cifmw_validations_list:
+#   - edpm/hugepages_and_reboot.yml
+#   - nfv/dpdk_and_sriov.yml
+#   - storage/hci_and_res_mem.yml
+cifmw_validations_list: []
+
+# cifmw_validations_run_all will find all validations under the tasks directory, save
+# them to a variable called found_validations and the sequentially execute each of them.
+cifmw_validations_run_all: false
+
+cifmw_validations_default_path: "{{ role_path }}/tasks"
--- a/roles/validations/meta/main.yml
+++ b/roles/validations/meta/main.yml
@@ -1,0 +1,41 @@
+---
+# Copyright Red Hat, Inc.
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+
+galaxy_info:
+  author: CI Framework
+  description: CI Framework Role -- validations
+  company: Red Hat
+  license: Apache-2.0
+  min_ansible_version: 2.14
+  namespace: cifmw
+  #
+  # Provide a list of supported platforms, and for each platform a list of versions.
+  # If you don't wish to enumerate all versions for a particular platform, use 'all'.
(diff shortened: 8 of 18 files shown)


bshephar:
> IF those are generic validaitons, why should this role be called "qe_validations" ? Let's call it "deployment_validation" or "validations". Quality is not restricted to quality engineers, it could be always possible for anyone to write specific validation assertions.

Agreed, sounds good to me. I have renamed it to `validations`.

bshephar:
recheck

playbooks/06-deploy-edpm.yml
```diff
@@ -146,3 +146,13 @@
   vars:
     step: post_deploy
   ansible.builtin.import_playbook: ./hooks.yml
+
```
cjeanner, line 149: why make a new play just for that?

roles/validations/handlers/main.yml
cjeanner, line 1: if not used, we can remove this

roles/validations/vars/main.yml
cjeanner, line 1: if not used, we can remove

roles/validations/tasks/main.yml
```diff
@@ -0,0 +1,46 @@
...
+
+    - name: Print found validations
+      ansible.builtin.debug: "Found validations: {{ found_validations }}"
+
+    - name: Run all found validations
+      ansible.builtin.include_tasks: "{{ item }}"
+      loop: "{{ found_validations }}"
+      when: found_validations is exist
+
+- name: Assert all listed validations exist
```
cjeanner, line 38: I'd make a block with this assert and the include_tasks, in order to apply a `when: not cifmw_validation_run_all | bool` to both. It should be exclusive in a `list XOR all` way IMHO

roles/validations/tasks/cleanup.yml
cjeanner, line 1: we can drop that file if no cleanup is needed

roles/validations/tasks/main.yml
```diff
@@ -0,0 +1,46 @@
...
+# Else, we will skip this and run only the explicitly defined validations from
+# cifmw_validations_list
+- name: Run all validations if required
+  when: cifmw_validations_run_all | bool
+  block:
+    - name: Find all validations
+      register: found_validations
+      ansible.builtin.find:
+        path: "{{ role_path }}/tasks"
+        pattern: .yml
```
cjeanner, line 27: I think the `pattern` should be `*.yml` according to the example in the doc:
```YAML
- name: Find /var/log files equal or greater than 10 megabytes ending with .old or .log.gz
  ansible.builtin.find:
    paths: /var/log
    patterns: '*.old,*.log.gz'
    size: 10m
```
https://docs.ansible.com/ansible/latest/collections/ansible/builtin/find_module.html

roles/validations/tasks/main.yml
```diff
@@ -0,0 +1,46 @@
...
+        recursive: true
+
+    - name: Print found validations
+      ansible.builtin.debug: "Found validations: {{ found_validations }}"
+
+    - name: Run all found validations
+      ansible.builtin.include_tasks: "{{ item }}"
+      loop: "{{ found_validations }}"
+      when: found_validations is exist
```
cjeanner, line 36: since we're `find`ing them, they should exist anyway...?

rabi:
recheck

roles/validations/tasks/main.yml
```diff
@@ -0,0 +1,46 @@
...
+# We can execute all defined validations when cifmw_validation_run_all is defined.
+# Else, we will skip this and run only the explicitly defined validations from
+# cifmw_validations_list
+- name: Run all validations if required
+  when: cifmw_validations_run_all | bool
+  block:
+    - name: Find all validations
+      register: found_validations
+      ansible.builtin.find:
+        path: "{{ role_path }}/tasks"
```
cjeanner, line 26: we may play a bit here. Not sure if it would work, but:
`path: "{{ cifmw_validation_default_path }}"`

with, in the `defaults/main.yml`, `cifmw_validation_default_path: "{{ role_path }}/tasks"`

This MAY allow to set an external path, for instance if someone has another (downstream-only?) set of validations to run. Might be good to test if the `include_tasks` would work with that, but since it accepts a path, I don't see why it wouldn't.

Thoughts?

rlandy:
recheck

roles/validations/tasks/networking/main.yml
cjeanner, line 1: maybe adding a couple of comment lines to explain the expectations here? (same for all of the `main.yml`)

playbooks/06-deploy-edpm.yml
bshephar: Seems to be the established pattern? Everything else appears to initialise a new play to execute their tasks, like HCI above:
https://github.com/openstack-k8s-operators/ci-framework/blob/main/playbooks/06-deploy-edpm.yml#L119

Everything that is starting a new stream of work here appears to be starting a new play.

Happy to change it, but the main reason was just matching the pattern of the other jobs here.

roles/validations/tasks/main.yml
bshephar: Yeah, I removed it. I thought it might have been because there were no files in the tasks/<dfg> directories to find initially. But it was throwing an error because of this variable in molecule.

I added some empty files to the tasks/<dfg> directories. I'll just put a hello world debug message task in each of them for now while testing it.

roles/validations/tasks/main.yml
bshephar: Sounds good. I implemented this in the latest iteration.

roles/validations/tasks/main.yml
bshephar: You are correct sir.

roles/validations/tasks/networking/main.yml
bshephar: Ultimately, I would like to not have these main.yml files in each of the folders I think. I was thinking it would be better to just have files for specific tasks.

The `main.yml` files atm just exist while I figure out if that find is going to work properly.

I guess if they need to stay prior to teams creating jobs, then I'll need to put something in them to explain the process a bit. Or, maybe I can just document it in the README.md for the role?

bshephar:
recheck

bshephar:
While the job itself is failing in https://github.com/openstack-k8s-operators/ci-framework/pull/1356, the change demonstrates the capability of this framework to execute arbitrary tests from the validations role.

The only requirements to execute a specific job are defined by the example here:
https://github.com/openstack-k8s-operators/ci-framework/pull/1356/commits/0cfc9347dc128b8621c2fa6810d50a5a05ac2b1b

playbooks/06-deploy-edpm.yml
cjeanner: hm, in that case... well, there are pros and cons.

Since we probably want this new path to be called from the 06-deploy-architecture.yml, we may indeed get it in a dedicated playbook in a dedicated file (maybe named `validations.yml`?)  to avoid code duplication.

That's mostly why we'd split things in a dedicated play.

Also, if a dedicated play has to be set, please consider the pattern I'm trying to push for the ceph part as shown in #1294 PR (now merged). Namely:
in the play, add a `pre_tasks` block and check if we want or not to run the validations.

Sorry for that back'n'forth.

roles/validations/README.md
```diff
@@ -0,0 +1,14 @@
+# validations
+The QE Validations role provides a framework for performing state verification.
```
cjeanner, line 2: you may want to remove the mention of "QE"

roles/validations/README.md
```diff
@@ -0,0 +1,14 @@
...
+This is useful for QE purposes and will be leveraged post-deployment to assert
+that required conditions have been satisfied. For example, if I update a certain
+value in my `OpenStackDataPlaneNodeSet`. I want to validate that the change has
+been applied as per my expectations on the relevant OpenStack Compute nodes.
+
+## Privilege escalation
```
cjeanner, line 8: I guess this will depend on the validations. Putting a generic 
`Check the validations set you want to run for privilege escalations` may be something good in here.
Thoughts?

roles/validations/README.md
```diff
@@ -0,0 +1,14 @@
...
+If apply, please explain the privilege escalation done in this role.
+
+## Parameters
+* `param_1`: this is an example
```
cjeanner, line 12: this will also probably depend on the validations to run.

You can add the top-level ones though:
```
cifmw_validations_list: []
cifmw_validations_run_all: false
cifmw_validation_default_path: "{{ role_path }}/tasks"
```

and add a mention about "Check validations set you want to run for more parameters".

Unless we enforce README update when adding new validation - but it may end with an endless file. Not really sure about the best way to handle this. Maybe add a comment section at the top of each validation file?

In any cases, once the rule is defined, it must be shown in this README here.

roles/validations/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+## Examples
```
cjeanner, line 14: Do you want to show some usage examples? Might be good, once there's a working flow.

roles/validations/tasks/main.yml
```diff
@@ -0,0 +1,46 @@
...
+        file: "{{ item.path }}"
+      loop: "{{ found_validations.files }}"
+
+- name: Run selected validations
+  when: not cifmw_validations_run_all | bool
+  block:
+    - name: Assert all listed validations exist
+      ansible.builtin.assert:
+        that:
+          - item is exists
```
cjeanner, line 41: NOTE for other reviewers:
- `item` will be in the form of "compute/foo.yml` - it should allow to point to external resources if needed/wanted.
- `is exists` test is local on the ansible-controller.

cjeanner (commented):
Couple of things, but otherwise, looks promising!

roles/validations/tasks/main.yml
bshephar: I'm not really sure this assertion is necessary tbh. The next task will just fail saying the file doesn't exist anyway. We could probably remove this check.

roles/validations/README.md
bshephar: Yeah, I completely forgot this file existed. Easter was too long apparently. :) 

I'll fix this file up

playbooks/06-deploy-edpm.yml
bshephar: Ok, so I created a validations.yml and just include_tasks from this 06-deploy-edpm.yml file. Let me know if that captures your intentions there.

cjeanner: perfect!

roles/validations/README.md
```diff
@@ -0,0 +1,34 @@
...
+## Privilege escalation
+Privilege escalation will be dependent on the specific job being executed. Some jobs
+may require privilege escalation. Check with the individual jobs to determine
+requirements.
+
+## Parameters
+The parameters will ultimately depend on what is implemented in each validation task.
+But in order to launch this role, at a minimum the following should be defined:
+
+# If you want to specify a specific list of jobs to execute, populate The
```
cjeanner, line 17: so here you want to use 3 `#` - it should be nested under the Parameters. Though you may consider using the usual way to display parameters? such as https://github.com/openstack-k8s-operators/ci-framework/blob/e64341bb5e44e42d418fa98640c61b6c654d751b/roles/artifacts/README.md?plain=1#L7-L15

roles/validations/README.md
cjeanner: thanks!

roles/validations/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+# To run all of the defined validations for the validations role, set `true`
+cifmw_validations_run_all: false
+
+# The path in which the validations can be found.
+cifmw_validation_default_path: "{{ role_path }}/tasks"
+
+## Examples
+The following variables would be used to trigger the EDPM job for Huge Pages:
+~~~
```
cjeanner, line 29: Care to use the "standard"(?)
```
```YAML
```
and close the code block? :)

roles/validations/tasks/main.yml
cjeanner: I'd rather keep it, and maybe add `msg: "Selected task {{ item }} doesn't exist"`

This would ensure users are fully aware of the issue...

roles/validations/tasks/networking/main.yml
cjeanner: document in README may be nice indeed.

playbooks/validations.yml
cjeanner, line 1: I think this one would also be needed in the `06-deploy-architecture.yml`.

Another solution is to integrate that playbook in the `08-run-tests.yml` play?

roles/validations/README.md
```diff
@@ -0,0 +1,29 @@
...
+- `cifmw_validations_run_all`: (Bool) Defaults to false
+- `cifmw_validation_default_path`: (String) Defaults to `"{{ role_path }}/tasks"`
+
+## Examples
+The following variables would be used to trigger the EDPM job for Huge Pages:
+```yaml
+  vars:
+    cifmw_execute_validations: true
+    cifmw_validations_list:
+    - edpm/hugepages_and_reboot.yml
```
cjeanner, line 27: yaml-lint: care to nudge by 2 spaces?

playbooks/validations.yml
bshephar: Added to 06-deploy-architecture.yml

roles/validations/README.md
bshephar: Bumped

bshephar:
recheck

roles/validations/tasks/main.yml
bshephar: So, this `item is exists` didn't actually work. Because I need to prepend the `"{{ cifmw_validations_default_path }}"`,  and I couldn't figure out the combination of `"{{ cifmw_validations_default_path }} {{ tiem }}" to work with `assert`. 

It's easier to just use:
```
    - name: Assert all listed validations exist
      ansible.builtin.stat:
        path: "{{ cifmw_validations_default_path }}{{ item }}"
      loop: "{{ cifmw_validations_list }}"
      register: validation_exists
      failed_when: not validation_exists.stat.exists

```

cjeanner (commented):
/approve

I think we're in a state where we can consider merging seriously.

If needed, we can refine, as usual, but I think we're good on many levels.

The two small comments are mostly informative, I won't block on either of them (especially the first one, it's a note for other reviewers).

roles/validations/tasks/edpm/huge-pages-and-reboot.yml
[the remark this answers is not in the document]
bshephar: Oh damn, I didn't mean to commit this file...

cjeanner:
Ah, care to run `make role_molecule` and commit the diff?

bshephar:
> Ah, care to run `make role_molecule` and commit the diff?

Hmm, this didn't change anything:
```
❯ make role_molecule
scripts/create_role_molecule.py
WARNING:root:Skipping openshift_adm. No molecule directory
WARNING:root:Skipping os_net_setup. No molecule directory
WARNING:root:Skipping ovirt. No molecule directory
WARNING:root:Skipping polarion. No molecule directory
WARNING:root:Skipping switch_config. No molecule directory
```

No new changes to commit after that

bshephar:
> > Ah, care to run `make role_molecule` and commit the diff?
> 
> Hmm, this didn't change anything:
> 
> ```
> ❯ make role_molecule
> scripts/create_role_molecule.py
> WARNING:root:Skipping openshift_adm. No molecule directory
> WARNING:root:Skipping os_net_setup. No molecule directory
> WARNING:root:Skipping ovirt. No molecule directory
> WARNING:root:Skipping polarion. No molecule directory
> WARNING:root:Skipping switch_config. No molecule directory
> ```
> 
> No new changes to commit after that

Disregard. I needed to `git pull` first. Changes are commited now.

bshephar:
> > > Ah, care to run `make role_molecule` and commit the diff?
> > 
> > 
> > Hmm, this didn't change anything:
> > ```
> > ❯ make role_molecule
> > scripts/create_role_molecule.py
> > WARNING:root:Skipping openshift_adm. No molecule directory
> > WARNING:root:Skipping os_net_setup. No molecule directory
> > WARNING:root:Skipping ovirt. No molecule directory
> > WARNING:root:Skipping polarion. No molecule directory
> > WARNING:root:Skipping switch_config. No molecule directory
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > No new changes to commit after that
> 
> Disregard. I needed to `git pull` first. Changes are commited now.

Ok, cool, passing again now.

lewisdenny (approved):
/lgtm

PR #1265, opened 2024-03-13
[TEST_ONLY]Shiftstack on va1

rlobillo:
This PR is created to be able to test below two PRs at the same:

- https://github.com/openstack-k8s-operators/ci-framework/pull/1259
- https://github.com/openstack-k8s-operators/ci-framework/pull/1210

As they both are needed to validate the e2e test for OSASINFRA DFG.

PR #1327, opened 2024-03-25
Patch ManilaAPI when Ceph is applied as a backend

fmount:
- [x] README in the role

### DIFF
--- a/hooks/playbooks/control_plane_ceph_backends.yml
+++ b/hooks/playbooks/control_plane_ceph_backends.yml
@@ -109,6 +109,12 @@
                 value: {{ cifmw_services_manila_enabled | default('false') }}
 
               - op: add
+                path: /spec/manila/template/manilaAPI/customServiceConfig
+                value: |
+                  [DEFAULT]
+                  enabled_share_protocols=nfs,cephfs
+
+              - op: add
                 path: /spec/manila/template/customServiceConfig
                 value: |
                   [DEFAULT]


pablintino:
@fmount If you want me to test the PR with the HCI job, let me know

fmount:
> @fmount If you want me to test the PR with the HCI job, let me know

Yes, that would be great actually!

hooks/playbooks/control_plane_ceph_backends.yml
```diff
@@ -108,6 +108,12 @@
                 path: /spec/manila/enabled
                 value: {{ cifmw_services_manila_enabled | default('false') }}
 
+              - op: add
```
tosky, line 111: shouldn't this be applied only when cifmw_services_manila_enabled is true? Or given the fact that the definition of manilaAPI is always there it doesn't matter?

fmount: exactly, it doesn't matter, and it can simplify any potential customization in a zuul job (where we have the distinction between `native_cephfs` and `cephnfs`) because the protocols are already configured in the API in case Ceph is selected as a backend.

hooks/playbooks/control_plane_ceph_backends.yml
```diff
@@ -108,6 +108,12 @@
...
+                path: /spec/manila/template/manilaAPI/customServiceConfig
+                value: |
+                  [DEFAULT]
+                  enabled_share_protocols=nfs,cephfs
```
viroel, line 115: Is this really needed since the patch below (line 117) should enable share protocols for all manila services? I am ok with defining only the config needed for each service too, but that means also refactoring the patch below I think.

fmount: ok so I might be overriding the wrong template at this point. The environment I'm debugging didn't have a global `customServiceConfig` passed down to all services.
Let me investigate further to see where that particular config is missing.
Putting this change on hold for now.

fmount: Looks like the deployment comes from [1], but I'm not sure how it deploys VA1 with manila. Is this hook being executed in that case?

[1] https://github.com/openstack-k8s-operators/ci-framework/pull/1265

viroel: You are correct that VA1's has different services configuration but this file is, at this moment, only being used in ceph_toy jobs, and we are now planning to add it to HCI job (which uses install_yamls to deploy). This file is not used in VA jobs, which uses the reproducer.

fmount: yeah, it turned out that the issue I saw didn't come from this file, but from VA1 (in architecture repo). It is possible to deploy VA1 via zuul using the architecture repo, and apply some `kustomizations` to `Manila`. 

This https://github.com/openstack-k8s-operators/architecture/pull/147 basically is supposed to fix the issue I saw.
Abandoning this PR as it is not necessary, but thank you @viroel for the help and your comments!

PR #1330, opened 2024-03-25
Add adoption variant to run minimal without ceph

cescgina:
Add a variant of the cifmw-adoption job to run with a swift backend and
run the test-minimal test-case.

### DIFF
--- a/zuul.d/adoption.yaml
+++ b/zuul.d/adoption.yaml
@@ -126,6 +126,13 @@
       - .*/*.md
 
 - job:
+    name: cifmw-data-plane-adoption-osp-17-to-extracted-crc-minimal-no-ceph
+    parent: cifmw-data-plane-adoption-osp-17-to-extracted-crc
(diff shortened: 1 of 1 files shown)


zuul.d/adoption.yaml
```diff
@@ -125,6 +125,13 @@
       - OWNERS
       - .*/*.md
 
+- job:
+    name: cifmw-data-plane-adoption-osp-17-to-extracted-crc-minimal-no-ceph
+    parent: cifmw-data-plane-adoption-osp-17-to-extracted-crc
+    vars:
+      use_ceph: "false"
```
pablintino, line 132: Why using a string for a boolean?

cescgina: ultimately this var is passed to a bash script, so it ends up being consumed as a string anyway, but I can change it to a bool, it should not matter for the use case

PR #1331, opened 2024-03-25
Revert "Try ssh command and freshly booting vms"

afazekas:
The change does not helps and the root cause was not related,
The provisioning system had an ansible role which run after
provisioning late, copied the authorized_keys from ~root/.ssh to
~zuul/.ssh nuking the key generated on the nodepool node.
The flow kept working because the ControlPerist
until a new session had to created for the delegte_to,
however the authorized_keys was based long long before.

This reverts commit 29799ab0f8f1388707e713c7791a170e2dcf5181.

- [X] No checklist is needed.


cjeanner:
/approve

nice RCA, @afazekas !

PR #1335, opened 2024-03-25
Add UEFI support for the virtual machines

cjeanner:
Until now, only BIOS mode was supported for the VM managed by the
libvirt_manager role.

From now on, we add support for UEFI, provided the base image supports
it of course (note: this is NOT the case for the official
CentOS-stream-9 images!).

A quick check/automation allows to detect if the image "is supposed to
support" UEFI based on cifmw_repo_setup_os_release parameter value. RHEL
images do provide UEFI, so let's just use it.

### DIFF
--- a/scenarios/reproducers/external-ceph.yml
+++ b/scenarios/reproducers/external-ceph.yml
@@ -15,6 +15,7 @@ cifmw_libvirt_manager_configuration:
       </network>
   vms:
     compute:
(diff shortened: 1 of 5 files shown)


scenarios/reproducers/external-ceph.yml
cjeanner, line 1: In fact, no need to set the cifmw_use_uefi here: this file is used to overload the va-hci.yml which provides more parameters. So I won't touch it.

PR #1339, opened 2024-03-26
Workaround to install OKD opearators

pablintino:
  - [x] README in the role

### DIFF
--- a/scenarios/centos-9/ceph_backends.yml
+++ b/scenarios/centos-9/ceph_backends.yml
@@ -3,7 +3,7 @@ cifmw_control_plane_ceph_backend_include_vars: "{{ cifmw_basedir }}/artifacts/pr
 
 cifmw_install_yamls_vars:
   BMO_SETUP: false
-  INSTALL_CERT_MANAGER: false
+  OKD: "true"
 
 cifmw_edpm_prepare_skip_crc_storage_creation: true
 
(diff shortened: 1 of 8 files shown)


roles/ci_local_storage/tasks/cleanup.yml
```diff
@@ -44,29 +44,41 @@
...
+      {{
+          hostvars | dict2items |
+          selectattr("value.ansible_hostname", "defined") |
+          selectattr("value.ansible_hostname", "in", k8s_nodes_hostnames) |
+          map(attribute="key") |
+          list
+      }}
 
-- name: Delete directories on worker nodes
+- name: Create directories on worker node
```
lewisdenny, line 75: Small nit, this should be `Delete directories on worker node`

pablintino: Good one.

scenarios/reproducers/networking-env-definition.yml
cjeanner, line 1: you want to remove that one at some point I think ;).

pablintino: I won't commit both commits all-together, so I'll remove it at that point.

cescgina:
I was looking a the failures and the kuttl job runs the `crc_storage` target (and the cleanup one) internally https://github.com/openstack-k8s-operators/ci-framework/blob/370b0ff4d62e99e5794a6438950c7f5249a0a127/ci/playbooks/kuttl/run-kuttl-tests.yml#L16-L28, so it's bound to run into problems. It would probably be ok to merge this breaking it and then moving those calls to the role in a followup

cjeanner:
@cescgina good catch yes!

Hopefully #1344 should solve the issue.

pablintino:
Not needed anymore.

PR #1340, opened 2024-03-26
Ensure script properly reports failure in reproducer

cjeanner:
Until now, if the deploy-architecture.sh was failing, it was silent:
ansible didn't know about that failure, and just reported green.

Adding the `set -e` seems to ensure we're reporting the state correctly.

This flag is therefore added to the deploy-architecture.sh and
deploy-edpm.sh scripts (even though the second isn't called by the
reproducer - let's just be consistent).

- [X] Tested in my env, we don't have CI for architecture deploy


raukadah (approved):
/approve

PR #1342, opened 2024-03-26
update cert_manager role with configureable timout

SeanMooney:
This change add a new cifmw_cert_manager_verify_timeout
variable to the cert_manager role and set its default to 10m

This is larger then the previous hardcoded 2m value because
we were seeing that timeout in ci.

  - [ ] README in the role

### DIFF
--- a/roles/cert_manager/defaults/main.yml
+++ b/roles/cert_manager/defaults/main.yml
@@ -64,3 +64,5 @@ cifmw_cert_manager_crds:
   - Challenge
   - ClusterIssuer
   - Issuer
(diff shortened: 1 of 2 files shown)


cjeanner:
/invalid

This won't help: the timeout occurs because it can't fetch the new image, because it's from the redhat registry, and the pull-secret is removed during the job bootstrap... So you're basically adding 8 more minutes of "image failing to pull".

SeanMooney:
oh ok i tought htis was timing out cause it was failing interminently

PR #1345, opened 2024-03-26
Add deploy_bmh role

rebtoor:
This role can be used to create BareMetalHost CRs and, eventually, deploying them, starting from a parameters file.

Closes: https://issues.redhat.com/browse/OSPRH-5630

  - [x] README in the role


rebtoor:
/hold

until dependencies are merged.

roles/deploy_bmh/template/bmh.yml.j2
```diff
@@ -0,0 +1,41 @@
+{%- for node in baremetal_nodes if node.split('-')[0] not in [ 'crc', 'controller']  %}
```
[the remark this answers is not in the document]
rebtoor: ah, nice catch!

roles/deploy_bmh/template/bmh.yml.j2
```diff
@@ -0,0 +1,41 @@
...
+labels:
+  app: openstack
+  workload: {{ node.split('-')[0] }}
+spec:
+  bmc:
+{# TODO
+   we'll have to build the URI address from scratch when
+   we have a proper redshift support in ci-framework
+#}
+    address: ipmi://{{ baremetal_nodes[node]['address'] }}:{{ baremetal_nodes[node]['port'] }}
```
pablintino, line 28: Are we assuming ipmi always? We are going to need idrac shortly.
Anyway, the ammount of things this field takes is "huge", including ssl, and other stuff. I'd treat it as anyone treats DB connection string, as a string that the user provides.

roles/deploy_bmh/template/bmh.yml.j2
pablintino, line 1: While using a ninja it's not maybe a big deal, the fact that we continue using it to create CRs/yaml content is starting to be annoying. Moreover, it enforces us to not allow a user to "easily" override values. Can we follow an approach like the one we used in the networking roles that used ansible variables (defaults) as "base" content, set the needed pieces on a copy and write it to disk with to_nice_yaml?

roles/deploy_bmh/template/bmh.yml.j2
```diff
@@ -0,0 +1,41 @@
...
+stringData:
+  username: {{ baremetal_nodes[node]['username'] }}
+  password: {{ baremetal_nodes[node]['password'] }}
+
+---
+apiVersion: metal3.io/v1alpha1
+kind: BareMetalHost
+metadata:
+  name: {{ node }}
+  namespace: openstack
```
pablintino, line 18: this namespace is not watched by metal3. Any BMH created outside it seems to be ignored.
We should allow to parameterize this one (and the associated secret namespace)

pablintino: FTR: BMO watches the `ns` set here https://github.com/metal3-io/baremetal-operator/blob/main/main.go#L135, that, in dev-scripts deployments at least, is just this:
```yaml
- name: WATCH_NAMESPACE
  valueFrom:
    fieldRef:
      apiVersion: v1
      fieldPath: metadata.namespace
```

Meaning: BMO watches for resources only inside it's own namespace. All the BMH we will create for VAs need to be applied in the `openshift-machine-api` ns.

abays: That is what we encountered when using Metal3 with OSP Director operator as well.  It seemed that we had to put our `Secret`s and `BaremetalHost`s in the `openshift-machine-api` namespace.

@rabi Did you find a way to get the BMO to watch other namespaces?  I thought I recalled you figuring something out that allowed `openstack` namespace to be used.

rabi: Yes,oc patch provisioning provisioning-configuration --type merge -p '{"spec":{"watchAllNamespaces": true }}'`

pablintino: Thank you both for that feedback and pointer.

roles/deploy_bmh/template/bmh.yml.j2
```diff
@@ -0,0 +1,41 @@
...
+labels:
+  app: openstack
+  workload: {{ node.split('-')[0] }}
+spec:
+  bmc:
```
pablintino, line 23: Another thing we will need: `disableCertificateVerification`
There is no warranty about the compute's BMCs to have a non-selfsigned certificate provisioned.

PR #1346, opened 2024-03-26
Ensure we know all the hosts beforehand

cjeanner:
Since #1343, ci_local_storage is using plain ansible to create the
persistent storage on the OpenShift node (in this case, CRC).

This means it's using SSH - and since the job never connects to the
nested CRC instance, it's failing due to the unknown remote host key.

Having ssh-keyscan populating the known_hosts should prevent this issue.

### DIFF
--- a/ci/playbooks/edpm_baremetal_deployment/run.yml
+++ b/ci/playbooks/edpm_baremetal_deployment/run.yml
@@ -15,6 +15,40 @@
         path: "{{ ansible_user_dir }}/ci-framework-data/artifacts/edpm-ansible.yml"
       register: edpm_file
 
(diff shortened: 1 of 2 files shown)


cjeanner:
@pablintino so this one makes the ci_local_storage pass - and we're crashing in the next stage like the rest - but at least we're at the same level with the rest of the jobs.
I propose, if Prow is happy, to override the zuul result and merge it - it would clear at least one issue from our path.

lewisdenny (approved):
/lgtm :)

jpodivin (approved):

jpodivin (approved):

PR #1349, opened 2024-03-27
Disable disable_os_marketplace hook

raukadah:
It is no longer needed in our CI, so removing it.


cjeanner:

It's a partial fix to a major issue preventing Zuul to be green. Let's get it in if Prow is happy (lint and other things).

PR #1351, opened 2024-03-27
Ensure MAC addresses are quoted in the generated CI snippet

cjeanner:
Apparently, something in the journey of the MAC is casting it into some
integer, leading to a nice crash.

Let's just ensure this is a string in the generated content we handle,
and see how it goes further down the path.

### DIFF
--- a/roles/ci_gen_kustomize_values/molecule/default/converge.yml
+++ b/roles/ci_gen_kustomize_values/molecule/default/converge.yml
@@ -18,6 +18,7 @@
 - name: Converge
   hosts: all
   vars:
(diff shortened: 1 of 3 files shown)


cjeanner:
We just found out where that weird casting happens:
in the snippet combination. For some reasons, a full-digit MAC (such as `52:54:00:17:05:43`) gets "cast" as an INT (in this case, `41135101543`) and, of course, it crashes.

Quoting the MAC in the generated snippet should prevent this issue.

fultonj (commented):
+2

quotes to the rescue

PR #1354, opened 2024-03-27
Avoid repetition, remove useless task in nova_migration key

cjeanner:
The code had an issue, the _nova_key was overridden even if the key did
exist.
In addition, there were too many repetition of the same path, leading to
potential issues if we want to update it later.

This patch corrects both issues, removes a task, and re-use the
`stat.path` data since it's available even if the file doesn't exists.


viroel:
/approve

lets get this in

PR #1355, opened 2024-03-27
[FIPS] Allow to deploy VA-HCI with FIPS enabled

viroel:
This patch adds parameters to enabled FIPS in VA-HCI scenario. A new top level parameter was added to allow users to enable FIPS in scenarios that supports it.

  - [x] README in the role


viroel:
recheck

PR #1356, opened 2024-03-28
Hugepages and reboot

bshephar:
  - [x] README in the role

### DIFF
--- a/roles/validations/tasks/edpm/hugepages_and_reboot.yml
+++ b/roles/validations/tasks/edpm/hugepages_and_reboot.yml
@@ -1,0 +1,128 @@
+# This job tests the functionality of the dataplane-operator to update values for a given
+# openstackdataplanenodeset and have them applied to nodes within the cluster. Specifically,
+# we will update the value for hugepages, ensure it is applied to the node, and that the node
+# has been rebooted.
+#
+# This job was created to satisfy:
+# https://issues.redhat.com/browse/osprh-5324
+
+# Collect facts from the node in question. this will give us an initial state prior to our changes
+# that can be compared with the end state.
+- name: Collect initial state of edpm node
+  ansible.builtin.gather_facts:
+  delegate_to: "{{ cifmw_validations_edpm_check_node }}"
+  register: initial_node_state
+
+# Since jobs use different names for their OpenStackDataPlaneNodeSet's, we need to dynamically
+# determine the name of the currently deployed one
+- name: Determine name of deployed NodeSet
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  cifmw.general.ci_script:
+    output_dir: "{{ cifmw_validations_basedir }}/artifacts"
+    script: >-
+      oc get osdpns --no-headers -o custom-columns=":metadata.name"
+  register: deployed_nodeset_name
+
+# collect initial confighash from the nodeset
+- name: Collecting initial confighash from the nodeset
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  cifmw.general.ci_script:
+    output_dir: "{{ cifmw_validations_basedir }}/artifacts"
+    script: >-
+      set -o pipefail && oc get -n {{ cifmw_validations_namespace }} osdpns "{{ deployed_nodeset_name.stdout }}" -o yaml | yq .status.configHash
+  register: initial_config_hash
+
+- name: Update hugepages value
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  cifmw.general.ci_script:
+    output_dir: "{{ cifmw_validations_basedir }}/artifacts"
+    script: >-
+      oc patch -n {{ cifmw_validations_namespace }} osdpns/"{{ deployed_nodeset_name.stdout }}" --type=merge -p '{"spec": {"nodeTemplate": {"ansible": {"ansibleVars": {"edpm_kernel_args": "default_hugepagesz=1gb hugepagesz=1g hugepages=64 iommu=pt intel_iommu=on tsx=off isolcpus=2-11,14-23"}}}}}'
+
+# The reboot-os service doesn't exist by default. So we need to create this service in
+# order to reboot the nodes.
+- name: Create reboot service
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  cifmw.general.ci_script:
+    output_dir: "{{ cifmw_validations_basedir }}/artifacts"
+    script: |
+      oc apply -f - <<EOF
+      apiVersion: dataplane.openstack.org/v1beta1
+      kind: OpenStackDataPlaneService
+      metadata:
+        name: reboot-os
+      spec:
+        playbook: osp.edpm.reboot
+      EOF
+
+# Create a new OpenStackDataPlaneDeployment to set the reboot variable. We also only need to execute
+# the jobs required to update Hugepages and reboot the nodes.
+- name: Create openstackdataplanedeployment to rollout changes
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  cifmw.general.ci_script:
+    output_dir: "{{ cifmw_validations_basedir }}/artifacts"
+    script: |
+      oc apply -f - <<EOF
+      apiVersion: dataplane.openstack.org/v1beta1
+      kind: OpenStackDataPlaneDeployment
+      metadata:
+        name: edpm-hugepages-update
+        namespace: {{ cifmw_validations_namespace }}
+      spec:
+        nodeSets:
+          - "{{ deployed_nodeset_name.stdout }}"
+        ansibleExtraVars:
+          edpm_reboot_force_reboot: "true"
+        servicesOverride:
+          - bootstrap
+          - reboot-os
+      EOF
+
+# loop check the status of the openstackdataplanedeployment until it is either completed,
+# or reaches a defined timeout value.
+- name: Wait for deployment to be complete
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  cifmw.general.ci_script:
+    output_dir: "{{ cifmw_validations_basedir }}/artifacts"
+    script: >-
+      oc wait openstackdataplanedeployment edpm-hugepages-update
+      --namespace={{ cifmw_validations_namespace }}
+      --for=condition=ready
+      --timeout={{ cifmw_validations_timeout | default(100) }}m
+
+# Collect final confighash from the nodeset
+- name: Collecting final confighash from the nodeset
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  cifmw.general.ci_script:
+    output_dir: "{{ cifmw_validations_basedir }}/artifacts"
+    script: >-
+      set -o pipefail && oc get -n {{ cifmw_validations_namespace }} osdpns "{{ deployed_nodeset_name.stdout }}" -o yaml | yq .status.configHash
+  register: post_change_config_hash
+
+# gather facts again to compare against the initial state.
+- name: Collect post change state of edpm node
+  ansible.builtin.gather_facts:
+  delegate_to: "{{ cifmw_validations_edpm_check_node }}"
+  register: post_change_node_state
+
+# these assertions will determine whether this job has ultimately been successful or not.
+# here, we should list all of the assertions required to satisfy our feature requirements.
+- name: Assert state change reflects our requirements
+  ansible.builtin.assert:
+    that:
+      - initial_config_hash.stdout != post_change_config_hash.stdout
+      - initial_node_state.ansible_facts.ansible_cmdline != post_change_node_state.ansible_facts.ansible_cmdline
--- a/roles/validations/tasks/edpm/hugepages_and_reboot.yml
+++ b/roles/validations/tasks/edpm/hugepages_and_reboot.yml
@@ -26,7 +26,6 @@
   register: deployed_nodeset_name
 
 # collect initial confighash from the nodeset
-
 - name: Collecting initial confighash from the nodeset
   environment:
     KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
--- a/roles/validations/tasks/edpm/hugepages_and_reboot.yml
+++ b/roles/validations/tasks/edpm/hugepages_and_reboot.yml
@@ -26,6 +26,7 @@
   register: deployed_nodeset_name
 
 # collect initial confighash from the nodeset
+
 - name: Collecting initial confighash from the nodeset
   environment:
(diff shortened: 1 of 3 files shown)


lewisdenny:
One thing we may want to consider now before there are too many tests is should these tests live in each operators repo? 

We have the all cloned as part of the framework so we technically could search and load them.

To me that would make sense, this is not framework code and it's not something ci-framework team should review/approve but more on the DFG/QE as they have context.

For example, Kuttl and Tempest tests are handled by ci-framework but the tests live in other repos managed by the developers. We handle starting the test and reporting the failures but leave the rest.

WDYT @cjeanner ? It would require some changed to https://github.com/openstack-k8s-operators/ci-framework/pull/1326 which I think you we're generally happy with so far so I might not be on the right track.

bshephar:
> One thing we may want to consider now before there are too many tests is should these tests live in each operators repo?
> 
> We have the all cloned as part of the framework so we technically could search and load them.
> 
> To me that would make sense, this is not framework code and it's not something ci-framework team should review/approve but more on the DFG/QE as they have context.
> 
> For example, Kuttl and Tempest tests are handled by ci-framework but the tests live in other repos managed by the developers. We handle starting the test and reporting the failures but leave the rest.
> 
> WDYT @cjeanner ? It would require some changed to #1326 which I think you we're generally happy with so far so I might not be on the right track.

I would be happy for them to have their own place to live. But I wouldn't want these playbooks to live in the individual operator repos. Kuttl tests are more self confined and testing specific things about the operator itself. These QE validations aren't really testing any one operator. This one for example tests that the dataplane-operator can update the inventory secret, the ansibleee operator can create the jobs and mount the secret, edpm-ansible is able to execute against the node and apply the change correctly and then reboot the node.

So this test spans at least three different operators and validates that all of them do their job cohesively.

I can understand the concern with keeping them in ci-framework; it's not a great fit here either. We could create a separate repo for QE jobs if we wanted to keep them all separate from ci-framework. Then we just need to update the other PR to `find` from the qe repo, rather than finding from the tasks/ directory of the validations role here.

I don't know how many of these jobs we will ever run upstream. We could probably just have a downstream internal repo for the jobs? The path we find from is already a variable, so we can just play with that variable to tweak as required.

bshephar:
Ok, so it looks like the secret is updated fine:
```
❯ curl https://logserver.rdoproject.org/56/1356/57a605e8b424e65ffb03569c7243322969770f70/github-check/podified-multinode-edpm-deployment-crc/7327b59/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/secrets/dataplane/dataplanenodeset-openstack-edpm-ipam.yaml -s | yq .data.inventory | base64 -d | yq .openstack-edpm-ipam.vars.edpm_kernel_args

default_hugepagesz=1gb hugepagesz=1g hugepages=64 iommu=pt intel_iommu=on tsx=off isolcpus=2-11,14-23
```

KernelArgs are not updated on the node:
```
❯ curl -s https://logserver.rdoproject.org/56/1356/57a605e8b424e65ffb03569c7243322969770f70/github-check/podified-multinode-edpm-deployment-crc/7327b59/controller/ci-framework-data/logs/38.129.56.234/etc/default/grub | grep CMDLINE
GRUB_CMDLINE_LINUX="console=ttyS0,115200n8 no_timer_check net.ifnames=0 crashkernel=1G-4G:192M,4G-64G:256M,64G-:512M"
```

NodeSet shows that the node successfully passed `configure-os` and `reboot-os`:
```
❯ curl -s https://logserver.rdoproject.org/56/1356/57a605e8b424e65ffb03569c7243322969770f70/github-check/podified-multinode-edpm-deployment-crc/7327b59/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/crs/openstackdataplanedeployments.dataplane.openstack.org/edpm-hugepages-update.yaml | yq .status.nodeSetConditions
openstack-edpm-ipam:
  - lastTransitionTime: "2024-04-02T03:58:29Z"
    message: Deployment completed
    reason: Ready
    status: "True"
    type: NodeSetDeploymentReady
  - lastTransitionTime: "2024-04-02T03:58:15Z"
    message: configure-os Deployment ready
    reason: Ready
    status: "True"
    type: ServiceConfigureOsDeploymentReady
  - lastTransitionTime: "2024-04-02T03:58:29Z"
    message: reboot-os Deployment ready
    reason: Ready
    status: "True"
    type: ServiceRebootOsDeploymentReady

```

bshephar:
The success of this job demonstrates the capability of the validation in combination with the proposed framework:

https://review.rdoproject.org/zuul/build/a20be4b1d1614b7a98a8b6d462a56a50

I will remove the DNM commit now that this functionality has been demonstrated.

roles/validations/tasks/edpm/hugepages_and_reboot.yml
cjeanner, line 1: Linter isn't happy:
```
 roles/validations/tasks/edpm/hugepages_and_reboot.yml:11: name[casing]: All names should start with an uppercase letter.
roles/validations/tasks/edpm/hugepages_and_reboot.yml:28: name[casing]: All names should start with an uppercase letter.
roles/validations/tasks/edpm/hugepages_and_reboot.yml:28: risky-shell-pipe: Shells that use pipes should set the pipefail option.
roles/validations/tasks/edpm/hugepages_and_reboot.yml:37: name[casing]: All names should start with an uppercase letter.
roles/validations/tasks/edpm/hugepages_and_reboot.yml:64: name[casing]: All names should start with an uppercase letter.
roles/validations/tasks/edpm/hugepages_and_reboot.yml:88: name[casing]: All names should start with an uppercase letter.
roles/validations/tasks/edpm/hugepages_and_reboot.yml:100: name[casing]: All names should start with an uppercase letter.
roles/validations/tasks/edpm/hugepages_and_reboot.yml:100: risky-shell-pipe: Shells that use pipes should set the pipefail option.
roles/validations/tasks/edpm/hugepages_and_reboot.yml:110: name[casing]: All names should start with an uppercase letter.
roles/validations/tasks/edpm/hugepages_and_reboot.yml:117: name[casing]: All names should start with an uppercase letter.
roles/validations/tasks/edpm/hugepages_and_reboot.yml:120: yaml[indentation]: Wrong indentation: expected 6 but found 5 
```

Also... I'm wondering if we shouldn't use the ansible module for kubernetes. Or, at least, generate files in the artifacts, so that we can ensure the CRs are generated as expected, and do some debugging.

The idea here of the framework is to keep track of as many actions as possible - having the generated CR somewhere, and the logs of the commands, would be nice I think. Especially for QEs. You may want to replace the calls to `command` or `shell` by our internal `ci_script`?

We can of course discuss that, but... I think it would be really neat.

bshephar: Fixed the linter complaints here.

I'm not familiar with this internal `ci_script`. I'll have to find some examples and see if it can be used as a replacement

cjeanner: It's a wrapper around `script` usable as a `shell`. For instance:
https://github.com/openstack-k8s-operators/ci-framework/blob/51a0b9f6a505b9bd65b33bf3ec6fc1609047cc27/roles/openshift_login/tasks/try_login.yml#L29-L57

There are many examples in the framework. In this use-case, I think it would be good to use it, since it would allow to get logs and reproducers for further debugging, info and so on :).

bshephar: Just holding off making this change until I've discussed it the implementation with the QE folks. Want to make sure these playbooks are easy to understand, and implement with minimal abstraction. But once I've consulted with the teams, we can make this change before we merge jobs.

roles/validations/tasks/edpm/hugepages_and_reboot.yml
```diff
@@ -0,0 +1,121 @@
...
+- name: Wait for deployment to be complete
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  ansible.builtin.command:
+    cmd: >-
+      oc wait openstackdataplanedeployment edpm-hugepages-update
+      --namespace={{ cifmw_install_yamls_defaults['NAMESPACE'] }}
+      --for=condition=ready
+      --timeout={{ cifmw_validations_timeout | default(100) }}m
```
SeanMooney, line 97: enabling hugepages ideally woudl not require new task in roles

we should be able to do this in the ci framework by providing a customisation to the dataplen in the job

this can work but we probably should avoid special caseing thing like this with new tasks and try to be more genic since customer will not want to write and miantin playbooks to do multiple deploy like this.

lewisdenny: Hey Sean, I think you might be missing the context on this patch because the Github description isn't up to date with the commit message but this isn't for customer use.

bshephar: I'm not sure I'm picking up what you're putting down here. This is simply a QE validation job to test functionality of the dataplane-operator to take arbitrary kernel args as input, apply them to the edpm nodes and ensure the nodes are rebooted. I'm basically just paraphrasing my commit message though:
https://github.com/openstack-k8s-operators/ci-framework/pull/1356/commits/0876265aec15531002d18ab1e0459d355bcf0e41

In summary though, this change isn't about having the CI tests always test hugepages, it's just about testing functionality from a QE perspective.

So the idea being that any QE validation for any feature would exist in a single playbook, that can be executed using the framework defined in:
https://github.com/openstack-k8s-operators/ci-framework/pull/1326
https://issues.redhat.com/browse/OSPRH-5778

SeanMooney: what im saying is that oure QE jobs and the ci framework in general should try to emulate how a customer would do this.

the customer would do this as part of the initial greenfiled deployment ideally prior to the installation of ovs/nova

SeanMooney: i guess im pusshing back on the central premms of why this role exits

testing indigugal fution slike configuring hugepages should be tested in the edpm-ansible molecule test for the relevant role in this case this should be tested by the edpm_kernel molecule tests https://github.com/openstack-k8s-operators/edpm-ansible/tree/main/roles/edpm_kernel/molecule

bshephar: I guess the intention behind this QE test is the end-to-end capability of the product to take arbitrary user input via the `OpenStackDataPlaneNodeSet`, reconfigure the Ansible inventory, execute the required tasks to apply the arbitrary input, ensure the node is rebooted and the end result matches user expectations.

Testing in edpm_ansible alone doesn't guarantee the end-to-end functionality. It's good to have it there as well, but the whole solution needs to be validated. Consider the workflow required here, users first need to:

1. Update the `OpenStackDataPlaneNodeSet`;
2. The reboot service doesn't exist by default, so that needs to be created;
3. We need a new `OpenStackDataPlaneDeployment` to rollout the changes. Changes like this can be done in a limited subset of tasks, in this case just `bootstrap` and `reboot`. So we are also covering the `servicesOverride` functionality in the `OpenStackDataPlaneDeployment` controller.
4. New jobs need to be created, ansibleee needs to mount the updated inventory and execute the required playbooks.
5. `edpm_ansible` enters the picture to execute the `edpm_kernel` role to apply the new KernelArgs
6. System needs to reboot and the end result should be new kernel args in `/proc/cmdline`. 

`edpm_ansible` makes up a small component of the workflow in this case, and `edpm_kernel` is just a random role chosen to test the overall capability. We could choose any arbitrary ansible variable and role for testing this overall workflow.

roles/validations/tasks/edpm/hugepages_and_reboot.yml
bshephar: Implemented:
https://github.com/openstack-k8s-operators/ci-framework/pull/1356/commits/e45dbe5dca1534209b813da681c1c7666d7347c8

roles/validations/tasks/edpm/hugepages_and_reboot.yml
```diff
@@ -0,0 +1,128 @@
...
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  cifmw.general.ci_script:
+    output_dir: "{{ cifmw_openshift_login_basedir }}/artifacts"
+    script: >-
+      oc get osdpns --no-headers -o custom-columns=":metadata.name"
+  register: deployed_nodeset_name
+
+  # collect initial confighash from the nodeset
```
lewisdenny, line 28: Nit: indention

lewisdenny (commented):
Looks good to me otherwise

roles/validations/tasks/edpm/hugepages_and_reboot.yml
```diff
@@ -0,0 +1,128 @@
...
+  register: initial_node_state
+
+# Since jobs use different names for their OpenStackDataPlaneNodeSet's, we need to dynamically
+# determine the name of the currently deployed one
+- name: Determine name of deployed NodeSet
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  cifmw.general.ci_script:
+    output_dir: "{{ cifmw_openshift_login_basedir }}/artifacts"
```
cjeanner, line 23: hmmm I'd avoid using that parameter from another role.

You should:

Create `cifmw_validations_basedir` parameter (like we do in other roles)
https://github.com/openstack-k8s-operators/ci-framework/blob/9869c491e6cff0512cc4f5bcad2246393b737947/roles/artifacts/defaults/main.yml#L20

Ensure the directories exist
```YAML
- name: Ensure directories exist
  ansible.builtin.file:
    path: "{{ cifmw_validations_basedir }}/{{ item }}"
    state: directory
    mode: "0755"
  loop:
    - artifacts
    - logs
```
`logs` is needed because the ci_scripts builds that path based on the `output_dir/../logs`

Care to make those 2 small corrections? I guess the "Hotfix" would also need it. Note that creating the directory tree should end in the main `main.yml` since it's called as an entrypoint for the role.

roles/validations/tasks/edpm/hugepages_and_reboot.yml
```diff
@@ -0,0 +1,128 @@
...
+
+# collect initial confighash from the nodeset
+- name: Collecting initial confighash from the nodeset
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  cifmw.general.ci_script:
+    output_dir: "{{ cifmw_validations_basedir }}/artifacts"
+    script: >-
+      set -o pipefail && oc get -n {{ cifmw_install_yamls_defaults['NAMESPACE'] }} osdpns "{{ deployed_nodeset_name.stdout }}" -o yaml | yq .status.configHash
```
cjeanner, line 36: @pablintino @rebtoor guess we shouldn't rely on `cifmw_install_yamls_defaults['NAMESPACE']` in an "architecture" deployment, should we?
I'm not sure the `openshift_login` exposes any namespace - is there any other trustful source for that data, or should we get yet another `cifmw_validations_namespace` parameter?

Any thoughts?

pablintino: Correct, if this is going to be used for VAs we cannot rely on that variable. Here I see two approaches:

- Add an extra var, like you suggest
- Assume that a deployment will have a single dataplane instance and fetch the NS that has that deployment beforehand.

bshephar:
recheck

bshephar:
recheck

bshephar:
recheck

cjeanner:
/approve

We can refine if needed later. Let's start getting actual validations and testing!

PR #1357, opened 2024-03-28
Support end to end run for deployed topologies.

psathyan:
Currently, default.yaml is being referenced as the default workflow steps. Here, we take it as user input via arch_workflow_file.

Fix an issue with nova_migration key path.

Fixes: https://github.com/openstack-k8s-operators/ci-framework/pull/1357

### DIFF
--- a/playbooks/06-deploy-architecture.yml
+++ b/playbooks/06-deploy-architecture.yml
@@ -32,55 +32,62 @@
       ansible.builtin.stat:
         path: "{{ cifmw_basedir }}/artifacts/nova_migration_key"
 
-    - name: Create nova migration keypair if does not exists
-      when:
-        - not _nova_key_file.stat.exists
-      community.crypto.openssh_keypair:
-        comment: "nova migration"
-        path: "{{ _nova_key_file.stat.path }}"
-        type: "{{ cifmw_ssh_keytype | default('ecdsa') }}"
-        size: "{{ cifmw_ssh_keysize | default(521) }}"
-
-    - name: Generate needed facts out of local files
+    - name: Ensure nova migration keypair details are propagated
       vars:
-        _ifaces_vars: >-
+        _ssh_file: >-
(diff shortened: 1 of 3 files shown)


playbooks/06-deploy-architecture.yml
```diff
@@ -37,7 +37,14 @@
         - not _nova_key_file.stat.exists
       community.crypto.openssh_keypair:
         comment: "nova migration"
-        path: "{{ _nova_key_file.stat.path }}"
+        path: >-
```
cjeanner, line 40: can't we make a block around those tasks (with the `ansible.builtin.stat` and a `vars` in order to avoid the repetition? Since `ansible.builtin.stat` seems to *remove* the `path` attribute if the file doesn't exist, that's the right way to avoid repetition all the way...

Thoughts?

```YAML
- name: Manage nova_migration key
  vars:
    _sshkey: "{{ <build path> }}"
  block:
    - name: Look for nova migration keypair file
    - name: Create nova migration keypair if does not exists
    - name: Generate needed facts out of local files

Comment (psathyan) [reply playbooks/06-deploy-architecture.yml]: Thank you. Yes, it make's sense.

Let me update it in my incoming commit.

File: roles/reproducer/tasks/configure_controller.yml [hunk 2]
```diff
@@ -291,10 +291,12 @@
...
             {{
-              (ansible_user_dir,
-               'src/github.com/openstack-k8s-operators',
-               'architecture/automation/vars/default.yaml') |
-               path_join
+              (
+                ansible_user_dir,
+                'src/github.com/openstack-k8s-operators',
+                'architecture/automation/vars'
+                cifmw_arch_automation_file | default('default.yaml')
```
cjeanner, line 298: care to add that new top-level parameter in the docs/sources/usage/01_usage.md ?

PR #1360, opened 2024-03-28
Use zuul_inventory to find out edpm vms ip

raukadah:
crc_ci_bootstrap_networks_out is defined in extracted crc EDPM Zuul job vars. crc_ci_bootstrap_networks_out var is used in artifacts edpm task files to find out the compute ip.
    
The value of crc_ci_bootstrap_networks_out is written in /etc/ci/env networking-info.yml file.
    
In Devscript reproducer environment, we donot pass crc_ci_bootstrap_networks_out var in the var file. that's why this
content is not available.
    
In VA HCI job, we run log collection from the controller. Since crc_ci_bootstrap_networks_out is not defined, that's why
edpm log collection does not happen.
    
In order to fix that, let' use zuul inventory to find out the edpm vms ip and perform log collection on that.

The format of extracted crc EDPM job inventory is like this
```
all:
    hosts:
        compute-0:
            ansible_host: x.x.x.x

```
and reproducer inventory is
```
all:
computes:
  hosts:
    compute-0:
      ansible_host: x.x.x.x
```
We are filtering out proper compute ips based on above format and setting the proper var.

### DIFF
--- a/roles/artifacts/tasks/edpm.yml
+++ b/roles/artifacts/tasks/edpm.yml
@@ -23,31 +23,29 @@
   when:
     - not cifmw_is_nested_virt | default(false) | bool
   block:
-    - name: Check for CI env directory
-      register: network_env_dir
-      ansible.builtin.stat:
-        path: /etc/ci/env
+    - name: Slurp zuul inventory file
+      ansible.builtin.slurp:
+        path: "{{ ansible_user_dir }}/ci-framework-data/artifacts/zuul_inventory.yml"
+      register: _inv
 
-    - name: Extract data from network env file if available
+    - name: Extract Compute from zuul mapping if any
       when:
-        - network_env_dir.stat.exists
-      block:
-        - name: Load network env file
-          ansible.builtin.include_vars:
-            dir: /etc/ci/env/
-
-        - name: Extract Compute from zuul mapping if any
-          when:
-            - crc_ci_bootstrap_networks_out is defined
-            - cifmw_edpm_deploy_extra_vars is defined
-          ansible.builtin.set_fact:
-            ssh_key_file: "{{ cifmw_edpm_deploy_extra_vars.SSH_KEY_FILE }}"
-            ssh_user: "{{ hostvars['compute-0'].ansible_user | default('zuul') }}"
-            edpm_vms: >-
+        - cifmw_edpm_deploy_extra_vars is defined
+      vars:
+        _inv_data: "{{ _inv['content'] | b64decode | from_yaml }}"
+        _edpm_vms_data: >-
+                      {{ _inv_data['computes']['hosts']
+                        if 'computes' in _inv_data
+                        else _inv_data['all']['hosts']
+                       }}
+      ansible.builtin.set_fact:
+        ssh_key_file: "{{ cifmw_edpm_deploy_extra_vars.SSH_KEY_FILE }}"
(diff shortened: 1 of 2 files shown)


roles/artifacts/tasks/edpm.yml
```diff
@@ -38,16 +38,15 @@
 
         - name: Extract Compute from zuul mapping if any
           when:
-            - crc_ci_bootstrap_networks_out is defined
```
cjeanner, line 41: makes 100% sense, but the explanation is wrong: that parameter comes from the upstream job definition - and we don't have it for the VA path.

Care to update the RCA in the comment + commit message?

raukadah: Done

raukadah:
recheck

roles/artifacts/tasks/edpm.yml
```diff
@@ -23,31 +23,32 @@
   when:
     - not cifmw_is_nested_virt | default(false) | bool
   block:
-    - name: Check for CI env directory
-      register: network_env_dir
-      ansible.builtin.stat:
-        path: /etc/ci/env
+    - name: Slurp zuul inventory file
+      ansible.builtin.slurp:
+        path: "{{ ansible_user_dir }}/ci-framework-data/artifacts/zuul_inventory.yml"
```
pablintino, line 28: FTR: This thing is not gonna work with BM nodes

roles/artifacts/tasks/edpm.yml
```diff
@@ -23,31 +23,32 @@
...
-      when:
-        - network_env_dir.stat.exists
-      block:
-        - name: Load network env file
-          ansible.builtin.include_vars:
-            dir: /etc/ci/env/
+    - name: Filter compute ips from reproducer/zuul generated inventory
+      vars:
+        _compute_data: "{{ _inv['content'] | b64decode | from_yaml }}"
+      ansible.builtin.set_fact:
```
pablintino, line 34: No need to create an entire new task to just set a fact that can be a variable of this task https://github.com/openstack-k8s-operators/ci-framework/pull/1360/files#diff-8bc3f0e951dc1302fb2cc860f49de1369e43a6c4fc7b55fca0da1df484cbf327R41

pablintino (changes requested):

roles/artifacts/tasks/edpm.yml
raukadah: For BM Jobs, We use [`cifmw_is_nested_virt`](https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/artifacts/tasks/edpm.yml#L4). The above change is not applied to BM job. I will update the BM job template to run BM job on artifacts role changes.

roles/artifacts/tasks/edpm.yml
raukadah: Done

roles/artifacts/tasks/edpm.yml
```diff
@@ -23,31 +23,32 @@
...
+        _compute_ips: >-
```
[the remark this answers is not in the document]
raukadah: Done

roles/artifacts/tasks/edpm.yml
pablintino: FTR: We will check if we can get rid off this and just use the sos-reports, embedded in must-gather.
I was referring to real BM jobs here.

PR #1363, opened 2024-03-28
Ensure we're using the right VM layout

cjeanner:
In the `libvirt_manager/deploy_layout.yml` we select the "layout" we
want to deploy. It generates a cached fact named `_layout`, and should
be using this one exclusively.

In parallel, the `reproducer` role should also use the correct layout,
as soon as the fact is available.

For the records, in the "zuul reproducer" path, the layout may be
updated by ansible in order to reflect the correct amount of computes,
using an internal parameter named
`cifmw_libvirt_manager_configuration_gen` - `libvirt_manager` then
selects either that parameter if exists, or defaults to the standard
`cifmw_libvirt_manager_configuration`, creating that `_layout` internal
fact.

This patch reconcile the parameter usage.

### DIFF
--- a/roles/libvirt_manager/tasks/create_vms.yml
+++ b/roles/libvirt_manager/tasks/create_vms.yml
@@ -174,10 +174,10 @@
       }}
     cifmw_virtualbmc_action: "add"
     _ipmi_host: >-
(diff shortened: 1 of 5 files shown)


pablintino:
recheck

PR #1364, opened 2024-03-28
Try to get a faster re-run for libvirt_manager

cjeanner:
By registering `overlay` state, we should be able to avoid a long list
of tasks that are useless against an already created VM.

We also slightly amend the way we generated the SSH configuration for
the ProxyJump in order to allow running multiple "labs" on multiple
hypervisor, while avoiding overriding the whole ssh configuration and
losing access to the VMs.

With this patch, in case you're deploying two labs on two distinct
machines, you'll be able to reach either of them by using, for instance,
`ssh controller-0.hypervisor-1` or `ssh compute-0.hypervisor-2`.

### DIFF
--- a/roles/libvirt_manager/molecule/deploy_layout/converge.yml
+++ b/roles/libvirt_manager/molecule/deploy_layout/converge.yml
@@ -47,7 +47,7 @@
           extra_disks_num: 1
           extra_disks_size: 1G
         compute:
-          amount: 2
+          amount: 1
           image_url: "{{ cifmw_discovered_image_url }}"
           sha256_image_name: "{{ cifmw_discovered_hash }}"
           image_local_dir: "{{ cifmw_basedir }}/images/"
@@ -153,7 +153,6 @@
             compare_vms: >-
               {{
                 ['cifmw-compute-0',
-                 'cifmw-compute-1',
                  'cifmw-baremetal-0'] | sort
               }}
           ansible.builtin.assert:
--- a/roles/virtualbmc/molecule/default/prepare.yml
+++ b/roles/virtualbmc/molecule/default/prepare.yml
@@ -32,10 +32,7 @@
           vms:
             compute:
               amount: 1
-              image_url: "{{ cifmw_discovered_image_url }}"
-              sha256_image_name: "{{ cifmw_discovered_hash }}"
-              image_local_dir: "{{ cifmw_basedir }}/images/"
-              disk_file_name: "centos-stream-9.qcow2"
+              disk_file_name: "blank"
               disksize: 15
               memory: 4
               cpus: 2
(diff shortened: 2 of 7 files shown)


roles/virtualbmc/tasks/main.yml
```diff
@@ -67,14 +67,14 @@
...
+  vars:
+    dest_dir: "{{ cifmw_virtualbmc_sshkey_path | dirname }}"
+  containers.podman.podman_container:
+    image: "{{ cifmw_virtualbmc_image }}"
+    name: "{{ cifmw_virtualbmc_container_name }}"
+    network: host
+    state: started
+    volumes:
+      - "{{ cifmw_virtualbmc_sshkey_path }}:{{ cifmw_virtualbmc_ipmi_key_path }}:ro,Z"
+      - "{{ dest_dir }}/virtualbmc.conf:/etc/virtualbmc/virtualbmc.conf:ro,Z"
```
fao89, line 80: tested, and it worked!

cjeanner: I moved that change in a dedicated patch, alongside another tweak to that container:
https://github.com/openstack-k8s-operators/ci-framework/pull/1405

cjeanner:
recheck

roles/virtualbmc/molecule/default/prepare.yml
```diff
@@ -32,10 +32,7 @@
           vms:
             compute:
               amount: 1
-              image_url: "{{ cifmw_discovered_image_url }}"
-              sha256_image_name: "{{ cifmw_discovered_hash }}"
-              image_local_dir: "{{ cifmw_basedir }}/images/"
-              disk_file_name: "centos-stream-9.qcow2"
+              disk_file_name: "blank"
```
marios, line 35: don't understand the relevance of this bit with respect to the commit message though?

roles/libvirt_manager/molecule/deploy_layout/converge.yml
```diff
@@ -47,7 +47,7 @@
           extra_disks_num: 1
           extra_disks_size: 1G
         compute:
-          amount: 2
+          amount: 1
```
marios, line 50: is this related to the 'skip tasks' or 'host specific inventory vars' or something else?

marios (commented):
makes sense as described in the commit, just couple questions inline (perhaps unrelated changes? not sure)

roles/virtualbmc/molecule/default/prepare.yml
cjeanner: it will make molecule faster on that role. Not directly related, but we can still consider it as an optimization

roles/libvirt_manager/molecule/deploy_layout/converge.yml
cjeanner: no, just that we don't need to create so many nodes "for nothing". Also improves a bit the speed of molecule itself.

cescgina (approved):

PR #1374, opened 2024-04-01
Set tenant MTU to 1500 for shift on stack

rlobillo:
For enabling the communication between VMs and considering that the MTU for the tenant network that openshift-installer creates is 1442, the tenant VLAN should be set to 1500 so the geneve tunnel works as expected.

This change is making the tenant vlan MTU configurable and it is setting it to 1500 on the shift-on-stack-overrides.yml

  - [x] README in the role

### DIFF
--- a/scenarios/reproducers/shift-on-stack-overrides.yml
+++ b/scenarios/reproducers/shift-on-stack-overrides.yml
@@ -4,3 +4,7 @@ cifmw_libvirt_manager_compute_cpus: 8
 cifmw_allow_vms_to_reach_osp_api: true
 # HCI requires bigger size to hold OCP on OSP disks
 cifmw_block_device_size: 100G
+cifmw_networking_mapper_definition_patches_01:
+  networks:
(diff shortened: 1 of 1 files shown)


scenarios/reproducers/shift-on-stack-overrides.yml
pablintino, line 1: The Networking Definition already support stuff like this. Just create a variable like this one in your scenario file:
https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/networking_mapper/README.md#networking-definition-patching
```yaml
cifmw_networking_mapper_definition_patches_01:
  networks:
     tenant:
        mtu: 1500
```

pablintino (changes requested):

scenarios/reproducers/shift-on-stack-overrides.yml
rlobillo: Fantastic, thank you!

PR #1376, opened 2024-04-02
Add missing default vars for compute disk,mem and cpu for va-hci

raukadah:
https://github.com/openstack-k8s-operators/ci-framework/pull/1259 adds the VA-HCI var for shiftstack. But it does not add the default values for compute disk,mem and cpu in va-hci scenario reproducer var file.

It broke the ci with following error:
```
You cannot get both OpenShift cluster types. Please chose between CRC and OCP cluster
```

After digging deep we found that, cifmw_libvirt_manager_configuration is undefined as cifmw_libvirt_manager_compute_disksize' is undefined.

Defining default vars of these vars fixes the issue.

### DIFF
--- a/scenarios/reproducers/va-hci.yml
+++ b/scenarios/reproducers/va-hci.yml
@@ -74,6 +74,9 @@ cifmw_root_partition_id: >-
     ternary(4, 1)
   }}
 cifmw_libvirt_manager_compute_amount: 3
+cifmw_libvirt_manager_compute_disksize: 20
+cifmw_libvirt_manager_compute_memory: 4
+cifmw_libvirt_manager_compute_cpus: 1
 cifmw_libvirt_manager_configuration:
   networks:
     osp_trunk: |


scenarios/reproducers/va-hci.yml
```diff
@@ -74,6 +74,9 @@ cifmw_root_partition_id: >-
     ternary(4, 1)
   }}
 cifmw_libvirt_manager_compute_amount: 3
+cifmw_libvirt_manager_compute_disksize: 200
+cifmw_libvirt_manager_compute_memory: 50
```
afazekas, line 78: 50 isn't too big ?

cjeanner:
/approve
/lgtm

The main issue is linked to the switch from `import_role` to `include_role` in the reproducer role.

IMHO, we should ensure either the `reproducer.yml` play or the `reproducer/main.yml` (role) is importing the defaults/main.yml from libvirt_manager in order to benefit from all the defaults.
We may also consider doing so for `rhol_crc` defaults and, why not, `devscripts` defaults.

Let's get this PR in first in order to unblock users, and work on a proper follow-up.

cjeanner:

Zuul won't check that path, since VA isn't running in upstream CI (resources, OCP, etc).

cjeanner:
/hold cancel
/lgtm

defaults now reflect the defaults set in the libvirt_manager role.

cjeanner:

again, that one doesn't test the change anyway.

PR #1377, opened 2024-04-02
Allow compound command

afazekas:
Use shell module instead of cmd in order to
allow compound command (pipes).

  - [ ] README in the role

### DIFF
--- a/roles/kustomize_deploy/tasks/execute_step.yml
+++ b/roles/kustomize_deploy/tasks/execute_step.yml
@@ -182,7 +182,7 @@
       environment:
         KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
         PATH: "{{ cifmw_path }}"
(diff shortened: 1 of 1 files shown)


afazekas:
Do you suggest spinning up busybox or similar containers for commands and kustomize it ?

cjeanner:
those validations are mostly to "wait for an action to be done". So mostly some `oc wait` and related commands. If you need deeper checks, please use the available hook capability, allowing to run a plain playbook (or side-load a CR, for what it may be worth):

https://github.com/openstack-k8s-operators/ci-framework/blob/ba36715ce8cc58bbacd937c9b21ae1c71cc638d8/roles/kustomize_deploy/tasks/execute_step.yml#L200-L208

We may even allow side-loading of a hook if wanted/needed, in the same way we allow injecting kustomizations from the user, in the framework job definition.

PR #1379, opened 2024-04-02
Prepare architecture automation file change

cjeanner:
We want the "validations" to get a better name, "wait_conditions". This
should avoid people trying to nudge other commands than "oc wait"
related ones.


cjeanner:
recheck though Zuul doesn't test that change....

PR #1384, opened 2024-04-03
nfs server: install also iptables on the node

tosky:
The iptables package on the target node is required by the ansible.builtin.iptables module.
On EL9 iptables does not exist but it is provided by iptables-nft, but keep iptables for now.


tosky:
(testing internally, please ask me for details)

PR #1393, opened 2024-04-04
Remove documentation fragment in application_credential plugin

lewisdenny:
The module was added in [1] and the doc build job didn't run in the PR.
I added the doc fragment file in [2] to try fix this but it still doesn't
work.

This patch removes the doc fragment and reference to it which will
unblock ci-framework patches failing like [3].

[1] https://github.com/openstack-k8s-operators/ci-framework/pull/1296
[2] https://github.com/openstack-k8s-operators/ci-framework/pull/1385
[3] https://github.com/openstack-k8s-operators/ci-framework/actions/runs/8537234033/job/23387381595?pr=1345#step:4:204


lewisdenny:
I tested this in a Ubuntu container just like the Github action, if we are okay to remove the doc frag we can merge this otherwise maybe @steveb can quickly fix :)

PR #1395, opened 2024-04-04
Correct spelling by ignoring some new strings

cjeanner:
Unfortunately, we can't get a nice pyspelling configuration ignoring
strings with numbers or underscrors/hyphens (tried it, failed it).

And since the ansible-doc building the modules documentation is kind of
dumb, there's no way to ignore standard `code` or some nice `css
class`...


cjeanner:
merging as-is: no feature, no code, no nothing, just strings to ignore in spelling.

PR #1396, opened 2024-04-04
Store nfs related facts to artifacts/parameters

cescgina:
Store the nfs server ip and network range to the parameters folder so it
can be later consumed by hooks.


playbooks/06-deploy-edpm.yml
```diff
@@ -1,14 +1,14 @@
 ---
+- name: Deploy NFS server on target nodes
```
cjeanner, line 2: Why are you moving it before the pre_deploy hooks?

PR #1398, opened 2024-04-04
Ensure install_yamls "download_tools.yml" play is run early

cjeanner:
This playbook is usually injected via a hook. We can make things
slightly faster by calling it once for all from the main deployment.

Since we don't import hooks in the nested ansible-playbooks call, we
should therefore get a slightly faster deployment.

### DIFF
--- a/roles/reproducer/tasks/main.yml
+++ b/roles/reproducer/tasks/main.yml
@@ -176,6 +176,18 @@
           -e @scenarios/reproducers/networking-definition.yml
           playbooks/01-bootstrap.yml
 
+    - name: Install dev tools from install_yamls on controller-0
+      delegate_to: controller-0
+      environment:
+        ANSIBLE_LOG_PATH: "~/ansible-bootstrap.log"
+      no_log: true
+      ansible.builtin.command:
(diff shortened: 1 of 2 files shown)


roles/reproducer/tasks/main.yml
```diff
@@ -176,6 +176,18 @@
...
 
+    - name: Install dev tools from install_yamls on controller-0
+      delegate_to: controller-0
+      environment:
+        ANSIBLE_LOG_PATH: "~/ansible-bootstrap.log"
+      no_log: true
+      ansible.builtin.command:
+        chdir: "/home/zuul/src/github.com/openstack-k8s-operators/install_yamls/devsetup"
+        cmd: >-
+          ansible-playbook -i ~/ci-framework-data/artifacts/zuul_inventory.yml
```
pablintino, line 187: I don't think install_yamls playbook is build to be able to consume our inventory. Given how simple it is, how we were consuming it and how install_yamls consumes it, why not pass `-i localhost, -c local` instead of our inventory there? Or, if preferred, directly consume the make target (not sure if the make module support the `creates` parameter.

cjeanner:
recheck

PR #1400, opened 2024-04-04
Generate and use scripts for CI runner

cjeanner:
This will make things easier to understand and follow, especially when
people want to run the job on their own.

Exposing the scripts, with "ordering", will ensure they know what to do
once they get on the controller-0.


cjeanner:
I'll rebase on #1401 to benefit from the new template - that will be more consistent.

cjeanner:
recheck

cjeanner:
recheck

lewisdenny (approved):
/lgtm

I can't test this because my local reproducer is broken and we don't test reproducer upstream but the code looks good to me.

PR #1401, opened 2024-04-04
Add a new parameter to inject extravars in nested plays

cjeanner:
It may happen we want to inject extra ansible variables in the nested
calls to ansible-playbook.

The new `cifmw_reproducer_play_extravars` parameter allows to list any
"extra-vars" you want.

By leveraging it, you can for instance inject a crafted scenario file
listing a series of hooks directly in the application deployment.

  - [X] README in the role


cescgina (commented):
/lgtm

PR #1402, opened 2024-04-04
[Test operator] Fix skiplist generation

lpiwowar:
The test_operator_cr is a copy of the cifmw_test_operator_tempest_config therefore it does not contain a reference to the cifmw_test_operator_exclude_list variable.

Because of that, we have to specify the value for excludeList directly in the test_operator_cr.

  - [x] README in the role


lpiwowar:
Testing here: https://github.com/openstack-k8s-operators/ci-framework/pull/1402

PR #1404, opened 2024-04-04
Support cifmw_ceph_client_cluster overrides

fultonj:
If a user overrides `cifmw_cephadm_cluster`, then the Ceph cluster name is something other than the default of "ceph" and the keyring and conf files are named from the override. However, without this patch the `cifmw_ceph_client` role does not respect this override. Update the role and the call to it from the Ceph playbook so that files consumed by Ceph clients get a name consistent with the server. This feature is useful for deployments with more than one Ceph server.

  - [x] README in the role


raukadah:
recheck after zuul job dequeue

katarimanojk (approved):

PR #1406, opened 2024-04-05
Ensure we create CRC and OCP first

cjeanner:
By sorting the libvirt_manager.vms content using dictsort, we have a way
to more or less ensure "crc" and "ocp" will be first.

Doing so, we give some more time to the OpenShift services to start
before hitting them - meaning we'll see less `retries` in the logs,
since we launched computes and controller.

This should allow to "lose less time".


psathyan (approved):

cjeanner:
recheck

PR #1407, opened 2024-04-05
Move again NFS server after EDPM deploy

tosky:
Otherwise the target EDPM node has no full networking setup, which means the choosen VLAN where to expose

This means that any usage of NFS should be moved to the post_deploy phase, possibly reconfiguring an existing control plane.

This mirror the similar mechanism used to configure and use Ceph on EDPM nodes.

### DIFF
--- a/playbooks/06-deploy-edpm.yml
+++ b/playbooks/06-deploy-edpm.yml
@@ -1,7 +1,4 @@
 ---
-- name: Deploy NFS server on target nodes
-  ansible.builtin.import_playbook: "nfs.yml"
-
 - name: Run pre_deploy hooks
   when:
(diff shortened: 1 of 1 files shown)


tosky:
Temporarily integrated https://github.com/openstack-k8s-operators/ci-framework/pull/1409 to check whether zuul is really able to combine two changes on the same github repository.

cjeanner:
@tosky hello! Is this still relevant? If so, care to mark it "ready for review" and advertise it? Thanks!

tosky:
It is still relevant but several unrelated issues prevented me from testing it, and now I have troubles understanding what is failing. But it is still definitely relevant, and I don't think I can mark it as ready yet.

PR #1408, opened 2024-04-05
Add a workaround to fix podman mkdir /run/user/1000: permission denied

raukadah:
On RHEL-9.4 with podman podman-4.9.4-0.1.el9.x86_64, we are facing following issue while pulling images using podman.
```
[zuul@titan102 ~]$ podman images
WARN[0000] RunRoot is pointing to a path (/run/user/1000/containers) which is not writable. Most likely podman will fail.
Error: creating events dirs: mkdir /run/user/1000: permission denied
```

Based on https://github.com/containers/podman/issues/5049#issuecomment-1348628646 running `loginctl enable-linger $USER` fixes the issue.


marios (approved):
lgtm based on commit message & explanation from last reviews call

reference for other reviewers:  https://www.man7.org/linux/man-pages/man1/loginctl.1.html

```
       enable-linger [USER...], disable-linger [USER...]
           Enable/disable user lingering for one or more users. If
           enabled for a specific user, a user manager is spawned for
           the user at boot and kept around after logouts. This allows
           users who are not logged in to run long-running services.
           Takes one or more user names or numeric UIDs as argument. If
           no argument is specified, enables/disables lingering for the
           user of the session of the caller.

```

cjeanner:
/approve

might make sense for podman to run while no active session exists for the user. Kind of a regression though imho - not sure if podman or systemd is the real issue there.

@raukadah we may have to consider a dedicated "podman" role once @lewisdenny has his sushy in?

lewisdenny (approved):
/lgtm

raukadah:
recheck

cjeanner:
recheck

PR #1409, opened 2024-04-05
Include install-certs dataplane service

karelyatin:
It's required with tls enabled, [1] already enabled tls by default so need to include the install-certs service before the services requiring the certs.
The issue should be visible once dataplane-operator is updated[2] or tlsEnabled flag is explicitly set in the nodeset.

Also update nova-custom-ceph dataplane service to
include missing tls params.

[1] openstack-k8s-operators/dataplane-operator#754
[2] openstack-k8s-operators/openstack-operator#732

  - [x] README in the role

### DIFF
--- a/roles/hci_prepare/tasks/phase2.yml
+++ b/roles/hci_prepare/tasks/phase2.yml
@@ -104,6 +104,7 @@
               - configure-os
               - run-os
               - reboot-os
+              - install-certs
               - ceph-client
               - ovn
               - neutron-metadata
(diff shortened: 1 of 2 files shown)


cjeanner:
/approve

We'll need something similar in the architecture repository I guess. This affects install_yamls based deployments only, as launched in upstream CI.

tosky:
@cjeanner architecture repo already got https://github.com/openstack-k8s-operators/architecture/pull/165

PR #1413, opened 2024-04-05
Add support for edpm-networker datatype

psathyan:
This change set adds support for edpm-networker-values datatype. Networker nodes are part of the external dataplane mgmt. They require a subset of the configuration needed by edpm-compute nodes.

Here, we create a common edpm-values template that can be leveraged across different edpm node sets.

  - [ ] README in the role


roles/ci_gen_kustomize_values/templates/common/edpm-values/values.yaml.j2
cjeanner, line 1: is there any differences with the `hci/edpm-values/values.yaml` ? If not, can we consider removing that hci dedicated template to avoid duplication?

roles/ci_gen_kustomize_values/templates/common/network-values/values.yaml.j2
```diff
@@ -78,8 +78,14 @@ data:
...
         "cniVersion": "0.3.1",
         "name": "{{ network.network_name }}",
+{%  if network.network_name == "octavia" %}
+        "type": "bridge",
+{%  else %}
         "type": "macvlan",
-{%  if network.vlan_id is defined%}
+{%  endif %}
+{%  if network.network_name == "octavia" %}
+        "bridge": "octbr",
```
rebtoor, line 87: why don't you put this line in the first `if` so we can avoid a second iteration?

roles/ci_gen_kustomize_values/tasks/edpm_networker_values.yml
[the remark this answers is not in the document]
psathyan: Yes @cjeanner that is correct. `edpm-networker-values` would always be a subset of `edpm-values` likewise for the case of `edpm-ceph-values`.

Thanks for the suggestion. 

Yes, I think it is possible to `edpm_common_values.yaml` 
- have a link to `edpm_network_values` as it is the datatype that we are looking at
- include the task `edpm_common_values.yaml` in `edpm_values.yml` with the added check for `migration_key`

Let  fine tune it a bit.

roles/ci_gen_kustomize_values/templates/common/edpm-values/values.yaml.j2
psathyan: The only difference with `hci/edpm-values/values.yaml` is the `if condition` we have for `migration`.

I believe we can have a template defined in `common` and remove the `hci` specific one...

psathyan: Yes,  `include` capability is definitely required as most of the pieces of code is the same.

cjeanner:
/approve

I could at least ensure HCI is still working fine with this patch.

PR #1415, opened 2024-04-06
[openshift_obs] Add support for deploying observability operator

psathyan:
Cluster Observability Operator is required for OpenStack Telemetry. This change set adds support for installing and configuring the observability operator. The default configuration is used as the Telemetry operator does not required any specific configuration.

  - [x] README in the role

### DIFF
--- a/roles/openshift_obs/tasks/main.yml
+++ b/roles/openshift_obs/tasks/main.yml
@@ -1,0 +1,46 @@
+---
+# Copyright Red Hat, Inc.
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+- name: Install cluster observability operator.
+  kubernetes.core.k8s:
+    definition: "{{cifmw_openshift_obs_definition }}"
+    kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
+    state: present
+
+- name: Wait for observability-operator pod
+  kubernetes.core.k8s_info:
+    kind: Pod
+    namespace: openshift-operators
+    label_selectors:
+      - app.kubernetes.io/name = observability-operator
+    wait: true
+    wait_timeout: 300
+    wait_condition:
+      type: Ready
+      status: "True"
+    kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
+
+- name: Wait for observability operator deployment
+  kubernetes.core.k8s_info:
+    kind: Deployment
+    namespace: openshift-operators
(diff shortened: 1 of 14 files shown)


hooks/playbooks/observability.yml
```diff
@@ -0,0 +1,56 @@
...
+            source: redhat-operators
+            sourceNamespace: openshift-marketplace
+        validate_certs: false
+        api_key: "{{ cifmw_openshift_token | default(omit) }}"
+        host: "{{ cifmw_openshift_api | default(omit) }}"
+        username: "{{ cifmw_openshift_user | default(omit) }}"
+        password: "{{ cifmw_openshift_password | default(omit) }}"
+        kubeconfig: "{{ cifmw_openshift_kubeconfig | default(omit) }}"
+
+    - name: Wait until observability operator deployment is successful.
```
jlarriba, line 47: I think that there is some time between the above Subscription is created and OpenShift creates the Deployment, as we had to implement custom waits in our CI: https://github.com/openstack-k8s-operators/telemetry-operator/blob/main/ci/create-coo-subscription.yaml#L28, but I never used the kubernetes.core collection so this might be already taking that in account.

psathyan: Thank you. I have include a check for the pod before waiting on deployment.

jlarriba (approved):
I have never used the k8s_info ansible module before, but this LGTM

psathyan:
recheck

cjeanner:
/approve

Thanks for that one @psathyan !

lewisdenny (approved):
Thanks, looks good!

roles/openshift_obs/tasks/main.yml
```diff
@@ -0,0 +1,61 @@
...
+          operators.coreos.com/observability-operator.openshift-operators: ""
+      spec:
+        channel: development
+        installPlanApproval: Automatic
+        name: cluster-observability-operator
+        source: redhat-operators
+        sourceNamespace: openshift-marketplace
+    kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
+    state: present
+    wait: true
```
pablintino, line 35: This wait here does nothing, as the k8s module only supports the waiting argument, without explicit conditions, for `Deployment`s, `Pod`s and `DaemonSet`.
https://docs.ansible.com/ansible/latest/collections/kubernetes/core/k8s_module.html#parameter-wait

This is how we did it for other role, maybe there is a better way, but we haven't been able to do so using the module.
https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/ci_nmstate/tasks/nmstate_k8s_install.yml#L34-L51

PS: As you are already waiting for the Pod and the Deployment this wait here being skipped is not an issue.

roles/openshift_obs/tasks/main.yml
```diff
@@ -0,0 +1,61 @@
...
+      metadata:
+        name: observability-operator
+        namespace: openshift-operators
+        labels:
+          operators.coreos.com/observability-operator.openshift-operators: ""
+      spec:
+        channel: development
+        installPlanApproval: Automatic
+        name: cluster-observability-operator
+        source: redhat-operators
```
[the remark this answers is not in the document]
psathyan: Thanks @lewisdenny @pablintino 

Let me update them in the next commit.

roles/openshift_obs/tasks/main.yml
psathyan: Done. thank you.

PR #1419, opened 2024-04-08
[libvirt_manager] Remove filter on list vms task

lewisdenny:
When using the libvirt_manager role to create `blank` vms with no image, in a shutoff state the filter this patch removes prevents the code following it from executing as the list returns no results leading to the `libvirt-uuids.yml` file not being generated for example.


danpawlik:
recheck

PR #1421, opened 2024-04-08
Allow to override libvirt_manager_configuration

rdiazcam:
This commit allows the user to modify the default/common configuration provided in `cifmw_libvirt_manager_configuration` by setting the `cifmw_libvirt_manager_configuration_overrides` variable.

  - [x] README in the role

### DIFF
--- a/roles/devscripts/README.md
+++ b/roles/devscripts/README.md
@@ -45,6 +45,8 @@ networks.
 * `cifmw_devscripts_external_net` (dict) Key/value pair containing information
   about the network infrastructure.
   Refer [section](#supported-keys-in-cifmw_devscripts_external_net).
+* `cifmw_libvirt_manager_configuration_overrides` (dict) key/value pairs to be
+  used to override cifmw_libvirt_manager_configuration.
 
 ### Secrets management
 
--- a/roles/devscripts/tasks/320_prepare_layout.yml
+++ b/roles/devscripts/tasks/320_prepare_layout.yml
@@ -22,6 +22,17 @@
     recurse: false
   register: _ctrl_vm_xml
 
+- name: Override libvirt manager configuration
+  when:
+    - cifmw_libvirt_manager_configuration_overrides is defined
+  ansible.builtin.set_fact:
+    cifmw_libvirt_manager_configuration_gen: >-
+      {{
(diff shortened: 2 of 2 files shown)


cjeanner:
/ok-to-test

Might be better than having per-settings overrides (like we have for the compute amount, or ram, or cpu, or....).

Could be worth overriding the existing override [here](https://github.com/openstack-k8s-operators/ci-framework/blob/main/scenarios/reproducers/shift-on-stack-overrides.yml) for instance?

pablintino:
> /ok-to-test
> 
> Might be better than having per-settings overrides (like we have for the compute amount, or ram, or cpu, or....).
> 
> Could be worth overriding the existing override [here](https://github.com/openstack-k8s-operators/ci-framework/blob/main/scenarios/reproducers/shift-on-stack-overrides.yml) for instance?

@cjeanner @rdiazcam While I'm ok with that take into account that with this strategy we can only create a single override. If we define the override downstream it will be imposible to "combine" the override in DS without redefining the variables.
I agree with @cjeanner in the approach, but to do so we need to tweak/change the approach to something like we did for the netoworking definition:
```yaml

- name: Patch the cifmw_libvirt_manager_configuration if needed
  vars:
    _libvirt_mngr_conf_patches: >-
      {{
        hostvars[inventory_hostname] |
        dict2items |
        selectattr("key", "match", "^cifmw_libvirt_manager_configuration_patches.*") |
        sort(attribute='key') |
        map(attribute='value') |
        list
      }}
  ansible.builtin.set_fact:
    cifmw_libvirt_manager_configuration_gen: >-
      {{
        cifmw_libvirt_manager_configuration_gen |
        combine(item, recursive=True)
      }}
  loop: "{{ [cifmw_libvirt_manager_configuration] + _libvirt_mngr_conf_patches }}"
```
Anyway, one thing from this PR that is kind of "meh" for me is the place where the patching is done and the vars we are using, cause we are exposing in the devscripts role a variable that seems to be related to libvirt. https://github.com/openstack-k8s-operators/ci-framework/pull/1421/files#diff-07fd5234ecc2db0ce4bc5f645f2104574591ce762a864279ffee819f09281541R48-R49

PR #1422, opened 2024-04-08
Append git version to cifmw galaxy collection

queria:
As setup_molecule is usually main installation method for installing cifmw.general galaxy collection,
append short git hash to its version for easier identification what code version is (or was) used,
otherwise it always states only:

  $ ansible-galaxy collection list |grep cifmw
  cifmw.general                 1.0.0

This change makes it slightly more useful
(as long as there is git command and git repo available, ignored otherwise):

  cifmw.general                 1.0.0-a44ca164


queria:
recheck

PR #1423, opened 2024-04-08
[libvirt_manager] Fix ssh config creation

lewisdenny:
When running from laptop i.e `inventory_hostname != 'localhost'` deploying the reproducer, the `Inject ssh jumpers` task runs delegated to localhost. This causes `ansible_host` var is to be overriden to `localhost` causing the generated ssh config to be broken and cause the deployment to fail.

### DIFF
--- a/roles/libvirt_manager/tasks/start_manage_vms.yml
+++ b/roles/libvirt_manager/tasks/start_manage_vms.yml
@@ -54,7 +54,6 @@
   delegate_to: localhost
   vars:
     extracted_ip: "{{ (vm_ip.stdout.split())[3] | ansible.utils.ipaddr('address') }}"
-    proxy_hostname: "{{ ansible_host | default(inventory_hostname) }}"
   ansible.builtin.blockinfile:
     create: true
     path: "{{ lookup('env', 'HOME') }}/.ssh/config"
@@ -62,7 +61,7 @@
     mode: "0600"
     block: |-
       Host {{ vm_ip.nic.host }} cifmw-{{ vm_ip.nic.host }} {{ extracted_ip }}
-        ProxyJump {{ ansible_user | default(lookup('env', 'USER')) }}@{{ proxy_hostname }}
+        ProxyJump {{ ansible_user | default(lookup('env', 'USER')) }}@{{ inventory_hostname }}
         Hostname {{ extracted_ip }}
         User {{ 'core' if vm_ip.nic.host is match('^(crc|ocp).*') else 'zuul' }}
         StrictHostKeyChecking no


lewisdenny:
Testing:
```
- name: Debug Lewis
  ansible.builtin.debug:
    msg: "{{ ansible_host | default('ansible_host not defined') }} {{ inventory_hostname | default('inventory_hostname not defined') }}"

- name: Debug Lewis delegate
  ansible.builtin.debug:
    msg: "{{ ansible_host | default('ansible_host not defined') }} {{ inventory_hostname | default('inventory_hostname not defined') }}"
  when:
    - inventory_hostname != 'localhost'
  delegate_to: localhost
```

Output:
```
TASK [libvirt_manager : Debug Lewis msg={{ ansible_host | default('ansible_host not defined') }} {{ inventory_hostname | default('inventory_hostname not defined') }}] **************************************************************************************************
task path: /Users/ldenny/Code/ci/upstream/ci-framework/roles/libvirt_manager/tasks/start_manage_vms.yml:51
Monday 08 April 2024  22:55:07 +1000 (0:00:00.021)       0:05:00.713 ********** 
ok: [titan105.lab] => {
    "msg": "titan105.lab titan105.lab"
}

TASK [libvirt_manager : Debug Lewis delegate msg={{ ansible_host | default('ansible_host not defined') }} {{ inventory_hostname | default('inventory_hostname not defined') }}] *****************************************************************************************
task path: /Users/ldenny/Code/ci/upstream/ci-framework/roles/libvirt_manager/tasks/start_manage_vms.yml:55
Monday 08 April 2024  22:55:07 +1000 (0:00:00.019)       0:05:00.732 ********** 
ok: [titan105.lab -> localhost] => {
    "msg": "localhost titan105.lab"
}
```

cjeanner:
/hold

Reason for the hold:
I'm slightly surprised by that error. I've been deploying dozen over dozen of times, and never hit that issue, no matter if my host was freshly provisioned or not.
The closest issue I faced way while working on another patch, with `molecule` and the fact that `instance` had to be treated exactly as `localhost` in that specific scenario.

I'd like to take the time to properly test it against various scenarios before removing the hold, if you don't mind.

roles/libvirt_manager/tasks/start_manage_vms.yml
```diff
@@ -54,15 +54,14 @@
...
-    proxy_hostname: "{{ ansible_host | default(inventory_hostname) }}"
   ansible.builtin.blockinfile:
     create: true
     path: "{{ lookup('env', 'HOME') }}/.ssh/config"
     marker: "## {mark} {{ vm_ip.nic.host }}"
     mode: "0600"
     block: |-
       Host {{ vm_ip.nic.host }} cifmw-{{ vm_ip.nic.host }} {{ extracted_ip }}
-        ProxyJump {{ ansible_user | default(lookup('env', 'USER')) }}@{{ proxy_hostname }}
+        ProxyJump {{ ansible_user | default(lookup('env', 'USER')) }}@{{ inventory_hostname }}
```
cjeanner, line 64: now I remember why there's that thing.
This is my inventory file:
```YAML
localhosts:
  hosts:
    localhost:
      ansible_connection: local

hypervisors:
  hosts:
    builder1:
      ansible_user: virtuser
      ansible_host: builder1.rh.lain.internux.ch
    builder2:
      ansible_user: virtuser
      ansible_host: builder2.rh.lain.internux.ch
```
As you can see, I'm using aliases, and leverage the `ansible_host` to allow proper DNS resolution (though I don't really need it in my LAN).

If we go straight to use the `inventory_hostname` we'll face issues when people are doing the same as I'm doing,in order to get an easier/faster way to name the `cifmw_target_host` (like I do) while not having DNS resolution for that alias (in my case, `builder-1` and `builder-2`)...

We may want to get some smarter detection of the name to use, buuut.... I'm not really sure how :/.

In your case, adding `ansible_host` to your inventory may solve your issue (alongside the `localhost` entry, to ensure you're using `local` connection)...

lewisdenny: Done and working now! Thanks Cédric, I'll look at adding an assertion to check for this in case someone else misses it :)

PR #1428, opened 2024-04-09
[reproducer] Add reproducer validations

lewisdenny:
This patch adds a common place for reproducer validates, allowing us to enforce, in code, thing steps documented and fail fast with helpful messages.

This patch:
- Adds reproducer validation task file
- Moves `cifmw_use_libvirt` check to validation task file
- Adds new check for ansible_host being defined
- Adds new check for ansible_connection being set to local
- Adds new check to ensure correct OS version is used for hypervisor

The ansible_host and ansible_connection checks can be tested locally by simply commenting them out in your inventory file.

### DIFF
--- a/roles/reproducer/README.md
+++ b/roles/reproducer/README.md
@@ -5,6 +5,7 @@ Role to deploy close to CI layout on a hypervisor.
 None
 
 ## Parameters
+
 * `cifmw_reproducer_basedir`: (String) Base directory. Defaults to `cifmw_basedir`, which defaults to `~/ci-framework-data`.
 * `cifmw_reproducer_compute_repos`: (List[mapping]) List of yum repository that must be deployed on the compute nodes during their creation. Defaults to `[]`.
 * `cifmw_reproducer_play_extravars`: (List[string]) List of extra-vars you want to pass down to the EDPM deployment playbooks. Defaults to `[]`.
@@ -17,6 +18,7 @@ None
 * `cifmw_reproducer_hp_rhos_release`: (Bool) Allows to consume rhos-release on the hypervisor. Defaults to `false`.
 * `cifmw_reproducer_dnf_tweaks`: (List) Options you want to inject in dnf.conf, both on controller-0 and hypervisor. Defaults to `[]`.
 * `cifmw_reproducer_skip_fetch_repositories`: (Bool) Skip fetching repositories from zuul var and simply copy the code from the ansible controller. Defaults to `false`.
+* `cifmw_reproducer_supported_hypervisor_os`: (List) List of supported hypervisor operating systems and their minimum version.
 
 ### run_job and run_content_provider booleans and risks.
 
--- a/roles/reproducer/defaults/main.yml
+++ b/roles/reproducer/defaults/main.yml
@@ -31,3 +31,8 @@ cifmw_reproducer_dnf_tweaks: []
 cifmw_reproducer_compute_repos: []
 cifmw_reproducer_repositories_path: "src"
 cifmw_reproducer_play_extravars: []
+cifmw_reproducer_supported_hypervisor_os:
+  CentOS:
+    minimum_version: 9
+  RedHat:
+    minimum_version: 9.4
--- a/roles/reproducer/tasks/validations.yml
+++ b/roles/reproducer/tasks/validations.yml
@@ -1,0 +1,40 @@
+---
+- name: Ensure we pass needed parameter
+  ansible.builtin.assert:
+    that:
+      - cifmw_use_libvirt is defined
+      - cifmw_use_libvirt | bool
+    msg: >-
+      Please ensure you pass "cifmw_use_libvirt: true" to the
+      deployment!
+    quiet: true
+
+- name: Verify ansible_host is defined in inventory
+  when: cifmw_target_host != 'localhost'
+  delegate_to: localhost
+  ansible.builtin.assert:
+    that: ansible_host != 'localhost'
+    msg: >-
+      Please ensure you define your ansible_host in your inventory file
+      https://ci-framework.readthedocs.io/en/latest/quickstart/03_deploy_infra.html#inventory-file
+    quiet: true
+
+- name: Verify ansible_connection is set to "local" in provided inventory
+  delegate_to: localhost
+  ansible.builtin.assert:
+    that: ansible_connection == 'local'
+    msg: >-
+      Please ensure you define your ansible_connection to local for localhost in your inventory file
+      https://ci-framework.readthedocs.io/en/latest/quickstart/03_deploy_infra.html#inventory-file
+    quiet: true
+
+- name: Verify we are running on a supported hypervisor OS version
+  ansible.builtin.assert:
+    that:
+      - ansible_distribution in cifmw_reproducer_supported_hypervisor_os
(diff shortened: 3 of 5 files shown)


roles/reproducer/tasks/validations.yml
```diff
@@ -0,0 +1,49 @@
...
+    msg: >-
+      Please ensure host: {{ cifmw_target_host }} is running a supported version of Centos
+      https://ci-framework.readthedocs.io/en/latest/quickstart/02_prepare_env.html#tested-environment
+    quiet: true
+
+- name: Verify we are running on supported RHEL version
+  when: ansible_distribution == 'RedHat'
+  ansible.builtin.assert:
+    that:
+      - ansible_distribution_version >= '9.2'
```
bshephar, line 45: Does this fail on CentOS stream 9 then?

Might be better to have a list of supported versions and then validate that `ansible_distribution_version` is in `"{{ supported_hypervisor_os_list }}"`?

bshephar (commented):
Nice one

roles/reproducer/tasks/validations.yml
lewisdenny: I don't think so due to the when clause preventing it running on Centos but let me verify.

bshephar: Ah yeah, you're right:
```
[m3@localhost ~]$ ansible localhost -m setup -a 'filter=ansible_distribution'
[WARNING]: No inventory was parsed, only implicit localhost is available
localhost | SUCCESS => {
    "ansible_facts": {
        "ansible_distribution": "CentOS"
    },
    "changed": false
}

```

lewisdenny: yeah:
```
TASK [Verify we are running on supported Centos version] *********************************************************************************************************************************
ok: [localhost]

TASK [Verify we are running on supported RHEL version] ***********************************************************************************************************************************
skipping: [localhost]

PLAY RECAP *******************************************************************************************************************************************************************************
localhost                  : ok=2    changed=0    unreachable=0    failed=0    skipped=1    rescued=0    ignored=0
```

raukadah, line 45: @lewisdenny We have moved all the downstream CI jobs to RHEL-9.4. Can we bump the RHEL conditional to RHEL-9.4?

lewisdenny: Good idea, done!

roles/reproducer/tasks/validations.yml
```diff
@@ -0,0 +1,40 @@
...
+- name: Verify ansible_connection is set to "local" in provided inventory
+  delegate_to: localhost
+  ansible.builtin.assert:
+    that: ansible_connection == 'local'
+    msg: >-
+      Please ensure you define your ansible_connection to local for localhost in your inventory file
+      https://ci-framework.readthedocs.io/en/latest/quickstart/03_deploy_infra.html#inventory-file
+    quiet: true
+
+- name: Verify we are running on supported Centos version
```
bshephar, line 31:
```suggestion
- name: Verify we are running on a supported hypervisor
```

roles/reproducer/README.md
```diff
@@ -17,6 +18,7 @@ None
 * `cifmw_reproducer_hp_rhos_release`: (Bool) Allows to consume rhos-release on the hypervisor. Defaults to `false`.
 * `cifmw_reproducer_dnf_tweaks`: (List) Options you want to inject in dnf.conf, both on controller-0 and hypervisor. Defaults to `[]`.
 * `cifmw_reproducer_skip_fetch_repositories`: (Bool) Skip fetching repositories from zuul var and simply copy the code from the ansible controller. Defaults to `false`.
+* `cifmw_reproducer_supported_hypervisor_os`: (List) List of supported hypervisor operating systems and their minimum version.
```
cjeanner, line 21: this seems more like a variable than a parameter. I'd not expose it, and move it to `vars/main.yml`. Thoughts?

roles/reproducer/defaults/main.yml
```diff
@@ -31,3 +31,8 @@ cifmw_reproducer_dnf_tweaks: []
 cifmw_reproducer_compute_repos: []
 cifmw_reproducer_repositories_path: "src"
 cifmw_reproducer_play_extravars: []
+cifmw_reproducer_supported_hypervisor_os:
+  CentOS:
+    minimum_version: 9
+  RedHat:
+    minimum_version: 9.4
```
cjeanner, line 38: hm, minimal is probably still 9.2 - we started with that, and "it works". Not sure we want to push 9.4.. Maybe we should stick to the major release ("9")?

roles/reproducer/tasks/validations.yml
cjeanner, line 1: I love that new file!
FTR, we have a dangling jira task about adding some validation role (name is now taken) in order to:
- ensure deprecated parameters aren't in use
- ensure we don't try to override internal variables set with `set_fact`
- ensure needed parameters are passed (and in the right format)

Maybe this is a first draw for that new role?

roles/reproducer/tasks/validations.yml
```diff
@@ -0,0 +1,40 @@
...
+      - cifmw_use_libvirt is defined
+      - cifmw_use_libvirt | bool
+    msg: >-
+      Please ensure you pass "cifmw_use_libvirt: true" to the
+      deployment!
+    quiet: true
+
+- name: Verify ansible_host is defined in inventory
+  when: cifmw_target_host != 'localhost'
+  delegate_to: localhost
```
cjeanner, line 14: in fact, we should loop on all the known inventory: we need the `ansible_host` for remote nodes as well.

roles/reproducer/defaults/main.yml
lewisdenny: I would go with either 9.2 or 9.4, I think CentOS Stream 9 is basically unversioned, it's all just 9 but with RHEL we have some granularity so we might as well use it. 

The reason for going with 9.4 is all downstream CI is using it, so it's the "tested" version. Thoughts?

roles/reproducer/README.md
lewisdenny: Sounds good :)

cjeanner:
/approve

just needs a rebase against `main` to get green molecule, and we should be able to nudge it in!

The role itself can come later, at least we have a nice tasks file we may call at will.

lewisdenny:
Thanks Cédric!

roles/reproducer/tasks/validations.yml
```diff
@@ -0,0 +1,40 @@
...
+    quiet: true
+
+- name: Verify we are running on a supported hypervisor OS version
+  ansible.builtin.assert:
+    that:
+      - ansible_distribution in cifmw_reproducer_supported_hypervisor_os
+      - ansible_distribution_version | float  >= cifmw_reproducer_supported_hypervisor_os[ansible_distribution].minimum_version
+    msg: >-
+      Please ensure host: {{ cifmw_target_host }} is running a supported OS.
+      {{ ansible_distribution_version }} is less than the minium supported version of
```
marios, line 38: minium

marios (commented):
nice

raukadah (approved):
Thank you @lewisdenny for adding it, It will make ci-framework much better!

PR #1429, opened 2024-04-09
[deploy_bmh] Add rootDeviceHints field support

pablintino:
  - [x] README in the role

### DIFF
--- a/roles/deploy_bmh/template/bmh.yml.j2
+++ b/roles/deploy_bmh/template/bmh.yml.j2
@@ -20,9 +20,8 @@ spec:
   bootMACAddress: {{ nic.mac }}
 {% endfor                                                                            %}
   bootMode: {{ node_data['boot_mode'] }}
-  online:
-{%- if node_data['status'] == "running"                                              %}
- true
-{% else                                                                              %}
- false
+  online: {{ 'true' if node_data['status'] | default("") == "running" else 'false' }}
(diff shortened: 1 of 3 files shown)


roles/deploy_bmh/template/bmh.yml.j2
```diff
@@ -20,9 +20,8 @@ spec:
   bootMACAddress: {{ nic.mac }}
 {% endfor                                                                            %}
   bootMode: {{ node_data['boot_mode'] }}
-  online:
-{%- if node_data['status'] == "running"                                              %}
- true
-{% else                                                                              %}
- false
+  online: {{ 'true' if node_data['status'] == "running" else 'false' }}
```
rebtoor, line 23: <3

rebtoor (approved):

roles/deploy_bmh/template/bmh.yml.j2
cjeanner: ah, you're using that data from the vbmc output? Good to know it's useful :). We'll need the same from sushy I guess?

pablintino: @abays Good to know, I'll amend this by allowing `status` to not be passed and simply default it to false. For real BM we won't pass the value and that will be enough.

rebtoor (approved):

PR #1431, opened 2024-04-09
Symlink skiplists between tempest and test-operator

kopecmartin:
The skiplist, as it contains tempest tests, should be the same for
all roles that run tempest. We will decrease efforts to keep the
lists in sync and avoid potential confusion as of what is skipped
in what tempest automation ansible role.

  - [x] README in the role


marios (approved):
makes sense, but is there already a plan to stop using the tempest role anywhere
i.e. will we eventually remove this altogether

PR #1434, opened 2024-04-09
Use ansible_user_id instead of ansible_user on podman fix

raukadah:


cjeanner:
/approve
/lgtm

I think we'll have to override zuul result again.... Note that this issue wasn't caught in molecule apparently.

cjeanner:

there are issues with libvirt_manager related tests in molecule, I'm investigating them now, and it's kind of weird...

PR #1435, opened 2024-04-09
Determine jump result dir based on the tempest role

kopecmartin:
If test_operator role is run, the results the jump tool is looking
for are under test_operator dir, not tempest one in case of the
tempest role. This adjusts the dir location based on the role that
is used to run tempest.

  - [x] README in the role

### DIFF
--- a/roles/polarion/defaults/main.yml
+++ b/roles/polarion/defaults/main.yml
@@ -20,6 +20,6 @@
 
 cifmw_polarion_basedir: "{{ cifmw_basedir | default(ansible_user_dir ~ '/ci-framework-data') }}"
 cifmw_polarion_jump_repo_dir: "{{ cifmw_polarion_basedir }}/polarion-jump"
-cifmw_polarion_jump_result_dir: "{{ cifmw_polarion_basedir }}/tests/tempest/"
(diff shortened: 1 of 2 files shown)


lpiwowar (approved):

roles/polarion/defaults/main.yml
```diff
@@ -20,6 +20,6 @@
 
 cifmw_polarion_basedir: "{{ cifmw_basedir | default(ansible_user_dir ~ '/ci-framework-data') }}"
 cifmw_polarion_jump_repo_dir: "{{ cifmw_polarion_basedir }}/polarion-jump"
-cifmw_polarion_jump_result_dir: "{{ cifmw_polarion_basedir }}/tests/tempest/"
+cifmw_polarion_jump_result_dir: "{{ cifmw_polarion_basedir }}/tests/{{ cifmw_run_test_role | default('tempest') }}/"
```
marios, line 23: is the idea to eventually only use the test operator?

roles/polarion/defaults/main.yml
```diff
@@ -20,6 +20,6 @@
...
 cifmw_polarion_jump_repo_branch: "master"
 cifmw_polarion_use_stage: false
```
marios, line 25: why don't we want to add the new var here?

cifmw_run_test_role: tempest

cjeanner: that variable comes from the `run_test` role - not a good idea to redefine it here.

PR #1436, opened 2024-04-09
Add new feature in the reproducer: molecule jobs

cjeanner:
This patch enables molecule job reproducer, at least for the ones
configured in the ci-framework (not sure if molecule is running in any
other places).

This patch allows the reproducer role to detect if we're reproducing a
molecule job or a "standard" deployment; it will update the layout
accordingly (it needs #1439 for the `amount: 0` support), and create the
needed playbook/script to run the whole stack accordingly.

Several changes were needed in the reproducer role in order to ensure
we're consistent with the data we fetch from Zuul, and some tasks had to
move in dedicated files to make things easier to read, as well as
leveraging the `include_tasks` (to avoid endless list of `skipping` log
lines).

This change also needs #1440 to ensure we're able to pass a few
parameters down in the job playbook.

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/1440
Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/1439

### DIFF
--- a/roles/reproducer/tasks/ci_deploy_data.yml
+++ b/roles/reproducer/tasks/ci_deploy_data.yml
@@ -1,0 +1,48 @@
+---
+- name: Fetch needed artifacts
+  ansible.builtin.get_url:
+    dest: "{{ _reproducer_basedir }}/{{ item }}"
+    url: "{{ data_baseurl }}/{{ item }}"
+    force: true
+  loop:
+    - parameters/custom-params.yml
+    - parameters/install-yamls-params.yml
+    - parameters/zuul-params.yml
+
+- name: Load parameters in current runtime
+  ansible.builtin.include_vars:
+    dir: "{{ _reproducer_basedir }}/parameters"
+
+- name: Extract compute count
+  vars:
+    zuul_inventory: >-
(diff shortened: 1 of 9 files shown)


roles/reproducer/tasks/ci_deploy_data.yml
```diff
@@ -0,0 +1,48 @@
...
+    ci_job_networking: "{{ zuul_inventory.all.hosts.crc.crc_ci_bootstrap_networking }}"
+    operator_content_provider: >-
+      {{
+        zuul_params.content_provider_registry_ip is defined
+      }}
+    openstack_content_provider: >-
+      {{
+        zuul_params.content_provider_dlrn_md5_hash is defined
+      }}
+    zuul: "{{ zuul_inventory.all.vars }}"
```
cjeanner, line 44: rename to `zuul_vars` (plus update calls/pointers)

cjeanner: Done.

cescgina (approved):

raukadah (approved):
/approve

PR #1437, opened 2024-04-09
Deploy bmh/support bm nmstate

pablintino:
  - [x] README in the role

### DIFF
--- a/roles/deploy_bmh/template/bmh-nmstate.yml.j2
+++ b/roles/deploy_bmh/template/bmh-nmstate.yml.j2
@@ -1,0 +1,10 @@
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ node_name }}-nmstate-secret
+  namespace: {{ cimfw_deploy_bmh_namespace }}
+type: Opaque
+stringData:
+ nmstate: |-
+  {{ node_nmstate | to_nice_yaml(indent=2) | indent(4)  }}
--- a/roles/deploy_bmh/template/bmh.yml.j2
+++ b/roles/deploy_bmh/template/bmh.yml.j2
@@ -25,3 +25,6 @@ spec:
   rootDeviceHints:
     {{}}: {{ node_data['root_device_hint'] }}
 {% endif                                                                             %}
+{%- if 'nmstate' in node_data %}
+  preprovisioningNetworkDataName: {{ node_name }}-netconfig-secret
+{% endif %}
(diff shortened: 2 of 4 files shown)


rebtoor (approved):

roles/deploy_bmh/template/bmh-nmstate.yml.j2
```diff
@@ -0,0 +1,10 @@
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ node_name }}-nmstate-secret
+  namespace: {{ cimfw_deploy_bmh_namespace }}
+type: Opaque
+stringData:
+ nmstate: |-
```
psathyan, line 9: nit: two indent currently it is one.

roles/deploy_bmh/template/bmh-nmstate.yml.j2
```diff
@@ -0,0 +1,10 @@
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ node_name }}-nmstate-secret
+  namespace: {{ cimfw_deploy_bmh_namespace }}
```
psathyan, line 6: AFAIK, this should be in `openshift-machine-api`

roles/deploy_bmh/template/bmh.yml.j2
```diff
@@ -25,3 +25,6 @@ spec:
   rootDeviceHints:
     {{}}: {{ node_data['root_device_hint'] }}
 {% endif                                                                             %}
+{%- if 'nmstate' in node_data %}
+  preprovisioningNetworkDataName: {{ node_name }}-netconfig-secret
```
psathyan, line 29: I believe this is wrong. It would need to be `{{ node_name }}-nmstate-secret`

Depending on https://github.com/openstack-k8s-operators/ci-framework/pull/1437/files#diff-9c7c344f6baca959deed26179a04746fdc0036f9d3e9944061c034c02efbd3b2R5

pablintino: Agree, fixing in a followup PR.

roles/deploy_bmh/template/bmh-nmstate.yml.j2
pablintino: We are passing that value in the downstream file I have, but I'll change the default to a more sane one.
In any case, and just for your information, it seems it's possible to watch all namespaces by setting `watchAllNamespaces: true` in the ` Provisioning` CR.

roles/deploy_bmh/template/bmh-nmstate.yml.j2
pablintino: Fixing.

PR #1438, opened 2024-04-09
Pin CentOS image used in molecule

cjeanner:
It seems latest CentOS image isn't booting properly.

By passing the proper parameter via the molecule configuration, we
should be able to get a global fix, instead of having to go in each
roles to set the appropriate parameter.

### DIFF
--- a/.config/molecule/config_local.yml
+++ b/.config/molecule/config_local.yml
@@ -16,6 +16,15 @@ platforms:
 
 provisioner:
   name: ansible
+  # Expose configuration to all jobs by default
+  # Useful when an fix requires to provide some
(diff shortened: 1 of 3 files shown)


cjeanner:
Before that PR, `libvirt_manager` molecule was failing because the CentOS image didn't properly boot, preventing to get the compute IP:
```
TASK [libvirt_manager : Grab IPs for nodes type compute] ***********************
failed: [instance] (item=compute-0) => changed=true
```

With this PR, it's green again: https://review.rdoproject.org/zuul/build/28fb579fca50468e9d07a5b2f5cbb0c5
```
cifmw-molecule-libvirt_manager SUCCESS
```

PR #1440, opened 2024-04-10
Extend "molecule" related playbooks

cjeanner:
In order to allow molecule job reproducer, we must slightly extend the
existing CI playbook, to alow setting a custom `hosts` as well as
passing down environment files/parameters to ensure we're reproducing
the right job.

### DIFF
--- a/ci/playbooks/molecule-prepare.yml
+++ b/ci/playbooks/molecule-prepare.yml
@@ -1,6 +1,6 @@
 ---
 - name: "Run ci/playbooks/molecule-prepare.yml"
-  hosts: controller
+  hosts: "{{ cifmw_zuul_target_host | default('controller') }}"
   gather_facts: true
   tasks:
     - name: Prepare workspace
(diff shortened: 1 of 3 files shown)


cjeanner:
recheck

cjeanner:
@pablintino proposal:
current run failed one molecule, rhol_crc, due to a collection to fail to install:
```
fatal: unable to access 'https://github.com/containers/ansible-podman-collections/': Could not resolve host: github.com
ERROR! Failed to clone a Git repository from `[https://github.com/containers/ansible-podman-collections`.](https://github.com/containers/ansible-podman-collections%60.)
```

Can we consider this patch as green and override `rdoproject.org/github-check` ?

pablintino:
> @pablintino proposal: current run failed one molecule, rhol_crc, due to a collection to fail to install:
> 
> ```
> fatal: unable to access 'https://github.com/containers/ansible-podman-collections/': Could not resolve host: github.com
> ERROR! Failed to clone a Git repository from `[https://github.com/containers/ansible-podman-collections`.](https://github.com/containers/ansible-podman-collections%60.)
> ```
> 
> Can we consider this patch as green and override `rdoproject.org/github-check` ?

Yeah, I 'm more than fine with that.

PR #1441, opened 2024-04-10
Add validation and assert before get all ip addresses

katarimanojk:
There can be a case where scale out compute nodes in HCI may not have any n/w interface configured in the storage network which may result in failures without proper info while fetching ip addresses.

This patch adds a task to validate and asset if any of the EDPM nodes doesn't have at least one n/w interface in the storage network.

  - [x] README in the role

### DIFF
--- a/playbooks/ceph.yml
+++ b/playbooks/ceph.yml
@@ -299,6 +299,13 @@
       vars:
         this_host: "{{ _target_hosts | first }}"
 
+    - name: Assert if any EDPM node's n/w interface is missing in storage network
+      ansible.builtin.assert:
+        that:
+          - hostvars[item][all_addresses] | ansible.utils.ipaddr(cifmw_cephadm_rgw_network) | length > 0
+        fail_msg: "node {{ item }} doesn't have any interface connected to network {{ cifmw_cephadm_rgw_network }}"
+      loop: "{{ _target_hosts }}"
+
     - name: Get already assigned IP addresses
       ansible.builtin.set_fact:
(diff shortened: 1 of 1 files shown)


playbooks/ceph.yml
```diff
@@ -299,6 +299,14 @@
       vars:
         this_host: "{{ _target_hosts | first }}"
 
+    - name: Assert if any EDPM node's n/w interface is missing in storage network
+      ansible.builtin.assert:
+        that:
+          - cifmw_cephadm_rgw_network is defined
+          - hostvars[item][all_addresses] | ansible.utils.ipaddr(cifmw_cephadm_rgw_network) | length > 0
+        fail_msg: "node {{ item }} doesn't have any interface connected to network {{ cifmw_cephadm_rgw_network }}"
```
cjeanner, line 307: there are two assertion there - the message "just" talks about the second one, and may very well fail while computing the `fail_msg` in case that `cifmw_cephadm_rgw_network` parameter is indeed undefined..

Thoughts? Can we expect this one to not be defined at that point of the play?

fultonj: Per line 294 we know `cifmw_cephadm_rgw_network` will be defined so I think line 305 can be removed.

katarimanojk: Yeah it will be always defined at L294, first assertion is removed in the latest patch.

cjeanner:
/approve

That was discussed while working on some issues in HCI - thank you for making the error more obvious now!

Regarding prow: it's dead for now, guess we'll have to wait a bit before rechecking jobs.

fmount:
Thank you @katarimanojk !!

pablintino:
Rebased to force zuul to trigger.

PR #1442, opened 2024-04-10
Wait for secret creation in install_openstack_ca

cescgina:
Wait for a few seconds to check if the combined-ca-bundle secret is
created, using retries. In molecule tests, it seems that the secret takes a bit to be
created, so the tests were failing because we were getting the secret
too quickly. The task has ignore_errors: true to match the previous
behaviour of ignore-errors=true.

### DIFF
--- a/roles/install_openstack_ca/tasks/main.yml
+++ b/roles/install_openstack_ca/tasks/main.yml
@@ -14,16 +14,20 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
-- name: Get CA bundle data
+- name: Get CA bundle data with retries
   environment:
     KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
     PATH: "{{ cifmw_path }}"
   ansible.builtin.command:
-    cmd: 'oc get secret combined-ca-bundle -n openstack -o "jsonpath={.data.tls-ca-bundle\.pem}" --ignore-not-found'
+    cmd: 'oc get secret combined-ca-bundle -n openstack -o "jsonpath={.data.tls-ca-bundle\.pem}"'
+  retries: 10
+  delay: 3
+  until: ca_bundle_data.rc == 0
   register: ca_bundle_data
+  ignore_errors: true
 
 - name: Get CA bundle
-  when: ca_bundle_data.stdout | length > 0
+  when: ca_bundle_data.rc == 0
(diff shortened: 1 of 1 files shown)


roles/install_openstack_ca/tasks/main.yml
```diff
@@ -14,12 +14,15 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
-- name: Get CA bundle data
+- name: Get CA bundle data with retries
   environment:
     KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
     PATH: "{{ cifmw_path }}"
   ansible.builtin.command:
-    cmd: 'oc get secret combined-ca-bundle -n openstack -o "jsonpath={.data.tls-ca-bundle\.pem}" --ignore-not-found'
```
lewisdenny, line 22: Maybe we should add `ignore_errors: true` to match the behavior of `--ignore-not-found`

Try get the secret a few times to allow it time to create, if it never shows up then allow the task failure to match `--ignore-not-found`

cescgina: thanks @lewisdenny  that's a great idea

cescgina:
recheck

cjeanner:
/approve

@cescgina I updated the PR title and description to match the actual commit information.

cescgina:
> /approve
> 
> @cescgina I updated the PR title and description to match the actual commit information.

thanks!

roles/install_openstack_ca/tasks/main.yml
```diff
@@ -14,13 +14,17 @@
...
+    cmd: 'oc get secret combined-ca-bundle -n openstack -o "jsonpath={.data.tls-ca-bundle\.pem}"'
+  retries: 10
+  delay: 3
+  until: ca_bundle_data.rc == 0
   register: ca_bundle_data
+  ignore_errors: true
 
 - name: Get CA bundle
   when: ca_bundle_data.stdout | length > 0
```
pablintino, line 30: Are we sure that if it fails the stdout is empty? (I saw oc sometimes using stdout for errors too, instead of stderr). Isn't safer to check here if `rc == 0`?

cescgina: you're right, that condition worked for the previous version, since it had `--ignore-not-found`, updating it now

PR #1445, opened 2024-04-10
Ensure we keep track of the kustomize builds

cjeanner:
We were missing artifacts in the flexible loop. This patch ensures we're
using the same location as the `install_operators.yml` tasks to keep the
various outputs of the `execute_steps.yml` kustomize calls.

Depends-On: https://github.com/openstack-k8s-operators/architecture/pull/173


rebtoor (approved):

cjeanner:
recheck

PR #1447, opened 2024-04-11
[deploy_bmh] Pass host credentials through a file

pablintino:
  - [x] README in the role

### DIFF
--- a/roles/deploy_bmh/tasks/main.yml
+++ b/roles/deploy_bmh/tasks/main.yml
@@ -44,6 +44,19 @@
       vars:
         node_name: "{{ item.key }}"
         node_data: "{{ item.value }}"
+        _credentials_content: >-
+          {{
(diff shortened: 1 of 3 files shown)


rebtoor (approved):

roles/deploy_bmh/tasks/main.yml
```diff
@@ -44,6 +44,19 @@
       vars:
         node_name: "{{ item.key }}"
         node_data: "{{ item.value }}"
+        _credentials_content: >-
+          {{
+            lookup('file', item.value['credentials_file']) | from_yaml
```
cjeanner, line 49: beware of the `lookup()` - it's running from the ansible controller. I think this is actually what you want, but I'd rather double-check that fact.

pablintino: Yes, that's the case. I'll copy that one from the jumpbox into the controller-0.

PR #1449, opened 2024-04-11
[Draft] Add horizon test role

rachael-george:
  - [ ] README in the role

### DIFF
--- a/roles/horizon/README.md
+++ b/roles/horizon/README.md
@@ -1,0 +1,10 @@
+# horizon
+Please explain the role purpose.
+
+## Privilege escalation
+If apply, please explain the privilege escalation done in this role.
+
+## Parameters
+* `param_1`: this is an example
+
+## Examples
--- a/roles/horizon/defaults/main.yml
+++ b/roles/horizon/defaults/main.yml
@@ -1,0 +1,32 @@
+---
+# Copyright Red Hat, Inc.
+# All Rights Reserved.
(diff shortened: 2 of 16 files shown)


tosky:
I would suggest to integrate with test-operator instead - we are moving into that direction and tempest and tobiko are already covered.

ashu-011:
> I would suggest to integrate with test-operator instead - we are moving into that direction and tempest and tobiko are already covered.

Hi @tosky  , Yes that's right . I had this discussion with Pragadeesh earlier . As discussed in hindsight this role is for short term only and for long term we are going to move in the direction test-operator itself .

tosky:
> > I would suggest to integrate with test-operator instead - we are moving into that direction and tempest and tobiko are already covered.
> 
> Hi @tosky , Yes that's right . I had this discussion with Pragadeesh earlier . As discussed in hindsight this role is for short term only and for long term we are going to move in the direction of test-operator itself .

Do we really need a short and long term? Tobiko users started with their own role, then quickly moved to test-operator and I think everything went fine. I'd argue that the cost of going directly with test operators is not too high and it's going to save significant time later.

rachael-george:
> @rachael-george @tosky heya! is this still relevant? I saw mention of integrating that test suit elsewhere... ?

Yes, we'll integrate it with the test-operator. But I want to keep this open for any changes in the ci-framework repo

PR #1453, opened 2024-04-11
[openshift_login] Only slurp install-yamls-params when update is needed

elfiesmelfie:
The install-yamls-params.yml param file is always slurped. The task that uses this slurped file is only run conditionally. Add the slurp to a conditional block with the file-update task.

The bootstrap playbook, which runs install_yamls and creates the file can skip running install_yamls when the bootstrap tag is skipped. This causes errors when the infra playbook is run and expects the params file to exist so it can be read.

This issue was discovered when trying to debug stf jobs failures.
The job failures are here:  https://review.rdoproject.org/zuul/build/043eab3a41c04c79ae46392100bc6607
The current workaround/fix is in progress here: https://github.com/infrawatch/service-telemetry-operator/pull/590

  - [x] README in the role

### DIFF
--- a/roles/openshift_login/tasks/main.yml
+++ b/roles/openshift_login/tasks/main.yml
@@ -98,22 +98,24 @@
     dest: "{{ cifmw_basedir|default(ansible_user_dir ~ '/ci-framework-data') }}/artifacts/parameters/openshift-login-params.yml"
     content: "{{ cifmw_openshift_login_params_content | from_yaml | to_nice_yaml }}"
 
-- name: Read the install yamls parameters file
-  ansible.builtin.slurp:
-    path: "{{ cifmw_basedir|default(ansible_user_dir ~ '/ci-framework-data') }}/artifacts/parameters/install-yamls-params.yml"
-  register: cifmw_openshift_login_install_yamls_artifacts_slurp
-
-- name: Append the KUBECONFIG to the install yamls parameters
+- name: Update the install-yamls-params with KUBECONFIG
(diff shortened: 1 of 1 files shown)


roles/openshift_login/tasks/main.yml
```diff
@@ -98,22 +98,23 @@
     dest: "{{ cifmw_basedir|default(ansible_user_dir ~ '/ci-framework-data') }}/artifacts/parameters/openshift-login-params.yml"
     content: "{{ cifmw_openshift_login_params_content | from_yaml | to_nice_yaml }}"
 
-- name: Read the install yamls parameters file
-  ansible.builtin.slurp:
-    path: "{{ cifmw_basedir|default(ansible_user_dir ~ '/ci-framework-data') }}/artifacts/parameters/install-yamls-params.yml"
-  register: cifmw_openshift_login_install_yamls_artifacts_slurp
+- when: cifmw_install_yamls_environment is defined
```
cjeanner, line 101: Care to add a name to the block?

roles/openshift_login/tasks/main.yml
```diff
@@ -98,22 +98,23 @@
...
+  block:
+  - name: Read the install yamls parameters file
```
cjeanner, line 103: This isn't properly aligned - you can push the whole block content by one more indentation (2 spaces).

elfiesmelfie: In principle, you're onboard with this change, right?

elfiesmelfie:
recheck

PR #1454, opened 2024-04-11
[copy-quay] Add quay api url parameter option

michaeltcoelho:
Makes quay's api url a parameter option for the quay-copy cli. 


cjeanner:
/approve

Seems good - ftr molecule is running the build + some dumb app call, I think we're covered. @arxcruz care to have a look as well?

arxcruz (commented):
/lgtm
lgtm

Now need to change the config.yaml template generation (in another repo)

PR #1459, opened 2024-04-12
[networking_mapper] Save instance's facts for troubleshooting

pablintino:
  - [x] README in the role


marios (approved):
seems sane, a little more context in the commit message would be nice (like what are you working on that makes this a useful thing specifically - i agree in the general case this is useful but was there something specific you're looking for today?)

cjeanner (commented):
/approve

keeping tracks of the middle steps is a common thing in the ci-framework - for instance we do the same for anything related to CR generation.

PR #1403, opened 2024-04-04
[install_yamls] Skip setting operators_repo var for repos not in openstack-k8s-operators org

elfiesmelfie:
https://logserver.rdoproject.org/85/585/6d95d205438e4a2a78f43f4606ae2efd0a29c678/github-check/stf-crc-ocp_414-local_build-index_deploy/a646dd6/controller/ci-framework-data/logs/ansible.log-2024-04-04-10:47 2024-04-04 10:47:09,970 p=28890 u=zuul n=ansible | TASK [install_yamls : Get environment structure base_path={{ cifmw_install_yamls_repo }}] ***
2024-04-04 10:47:09,970 p=28890 u=zuul n=ansible | Thursday 04 April 2024  10:47:09 -0400 (0:00:00.036)       0:00:56.974 ********
2024-04-04 10:47:10,143 p=28890 u=zuul n=ansible | fatal: [localhost]: FAILED! => {"changed": false, "module_stderr": "/bin/sh: line 1: SERVICE-TELEMETRY_REPO=/home/zuul/src/github.com/infrawatch/service-telemetry-operator: No such file or directory\n", "module_stdout": "", "msg": "MODULE FAILURE\nSee stdout/stderr for the exact error", "rc": 127}

As a pull request owner and reviewers, I checked that:
- [x] Appropriate testing is done and actually running
- [x] Appropriate documentation exists and/or is up-to-date:
  - [x] README in the role
  - [x] Content of the docs/source is reflecting the changes

PR #1461, opened 2024-04-12
[bootstrap] Add the packages tag to import_role/setup_ci

elfiesmelfie:
The oc client (and other packages) are installed set up in the ci_setup role. Intuitively, the task in 01-bootstrap.yml should have the `packages` tag

  - [ ] README in the role

### DIFF
--- a/playbooks/01-bootstrap.yml
+++ b/playbooks/01-bootstrap.yml
@@ -40,6 +40,7 @@
     - name: Run ci_setup role
       tags:
         - bootstrap
+        - packages
       ansible.builtin.import_role:
         role: ci_setup
 


playbooks/01-bootstrap.yml
```diff
@@ -40,6 +40,7 @@
     - name: Run ci_setup role
       tags:
         - bootstrap
+        - packages
```
cjeanner, line 43: we're using `import_role` here - the tags associated to this task are applied to all of the imported tasks under `ci_setup` - which isn't what we want to do, since there's not only "package related" tasks in there.

Not sure about the intended effect?

elfiesmelfie: I'm trying to resolve some issues in CI jobs using the ci framework. I am trying two paths at the moment.

First solution:
The issue is described here: https://github.com/openstack-k8s-operators/ci-framework/pull/1403 (first attempt to resolve the issue). The issue is occurring in all STF-related jobs, e.g. https://github.com/infrawatch/service-telemetry-operator/pull/589.

Second solution (this PR + [this ](https://github.com/openstack-k8s-operators/ci-framework/pull/1453)+[this](https://github.com/openstack-k8s-operators/ci-framework/pull/1461))
I came to this by trying to skip the install_yamls portion of the setup in bootstrap and infra playbooks:
https://github.com/infrawatch/service-telemetry-operator/pull/590 
This resulted in not having the oc binaries available, which is why I was trying to add the packages tag in this PR.

cjeanner: hmm after a quick look, the `oc` comes here:
https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/ci_setup/tasks/packages.yml#L18-L83

It seems to be missing the `packages` tag (as well as `bootstrap btw`) - I'd start with adding it here first.

elfiesmelfie: I think that it is covered with [the always tag in main](https://github.com/openstack-k8s-operators/ci-framework/blob/17b29222d62e8fe786d5bc9156feede80914e21e/roles/ci_setup/tasks/main.yml#L22-L25).

I think all the tasks _should_ be run each time,  unless a tag is specified.  
However, we want this to be run always, so the tasks should be tagged as `always`, right?

elfiesmelfie:
@cjeanner PR#1403 resolved the issue I had in STF's jobs. Does this still need doing?
I'm happy to continue with the fix, if you still think its necessary/nice to have

cjeanner:
> @cjeanner PR#1403 resolved the issue I had in STF's jobs. Does this still need doing? I'm happy to continue with the fix, if you still think its necessary/nice to have

I propose to close that one since you don't face the issue anymore.
I'll do some checks over this week to ensure the `packages` tag is accessible and properly applied/filtered. There was a CI check for that back then, but we dropped it (too complicated). I'll see if I can make a kind of linter to ensure any modules affecting packages (package, dnf, pip, etc) is properly tagged. Shouldn't be THAT hard...

PR #1465, opened 2024-04-15
Fix multiple IP addresses listing during collect

psathyan:
Primarily in the case of hybrid, we tend to see multiple IP address being listed when attempting to gather address for OCP type VMs.

The multiple lines are primarily due to api and ingress addresses being present in the arp cache.

  - [x] README in the role

### DIFF
--- a/roles/libvirt_manager/tasks/start_manage_vms.yml
+++ b/roles/libvirt_manager/tasks/start_manage_vms.yml
@@ -30,6 +30,12 @@
       virsh -c qemu:///system -q
       domifaddr --source arp
       cifmw-{{ nic.host }} | grep "{{ nic.mac }}"
+      {% if cifmw_openshift_api_ip_address | default(false) %}
+      | grep -v "{{ cifmw_openshift_api_ip_address }}"
+      {% endif %}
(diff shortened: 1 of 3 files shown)


common-requirements.txt
```diff
@@ -1,4 +1,5 @@
 ansible-core>=2.14.0
+dnspython
```
pablintino, line 2: Not needed anymore if we use the `dig` command directly.

roles/libvirt_manager/tasks/start_manage_vms.yml
```diff
@@ -30,6 +30,12 @@
       virsh -c qemu:///system -q
       domifaddr --source arp
       cifmw-{{ nic.host }} | grep "{{ nic.mac }}"
+      {% if cifmw_openshift_api_ip_address is defined %}
```
cjeanner, line 33: aren't both set, meaning you'll want to chain the greps, with probably different address, and eventually fail with no address?

danpawlik:
recheck

rebtoor (approved):

rlandy:
recheck

PR #1468, opened 2024-04-15
[devscripts] Fix patch failure of dev-scripts repo.

psathyan:
This changes fixes the error thrown when patching the upstream repo.


cjeanner:
/approve
/lgtm

this is needed because upstream code changed to support rocky9, making our patch deprecated.

PR #1471, opened 2024-04-15
[ci_gen_kustomize_values] Add support to va2 NFV related values.yml

Jaganathancse:
This changes to provide a support to configure VA2 NFV EDPM values via CIFW (ci_gen_kustomize_values).
Based on existing PR#1369.

  - [X] README in the role

### DIFF
--- a/roles/ci_gen_kustomize_values/README.md
+++ b/roles/ci_gen_kustomize_values/README.md
@@ -39,8 +39,10 @@ This configMap needs some more parameters in order to properly override the `arc
 * `cifmw_ci_gen_kustomize_values_ssh_public_key`: (String) SSH public key associated to `cifmw_ci_gen_kustomize_values_ssh_private_key`.
 * `cifmw_ci_gen_kustomize_values_migration_priv_key`: (String) SSH private key dedicated for the nova-migration services.
 * `cifmw_ci_gen_kustomize_values_migration_pub_key`: (String) SSH public key associated to `cifmw_ci_gen_kustomize_values_migration_priv_key`.
-
 Note that all of the SSH keys should be in `ecdsa` format to comply with FIPS directives.
+
+### Required parameters only when baremetal compute nodes are used.
+* `cifmw_ci_gen_kustomize_values_ctlplane_interface`: (String) Used to override default controlplane interface for OSP compute nodes.
 
 ## Examples
 
--- a/roles/ci_gen_kustomize_values/tasks/edpm_core_asserts.yml
+++ b/roles/ci_gen_kustomize_values/tasks/edpm_core_asserts.yml
@@ -25,3 +25,11 @@
       - cifmw_ci_gen_kustomize_values_ssh_private_key != ''
       - cifmw_ci_gen_kustomize_values_ssh_public_key is defined
       - cifmw_ci_gen_kustomize_values_ssh_public_key != ''
+
+- name: Ensure the required baremetal host parameters are configured.
+  when:
+    - cifmw_baremetal_hosts is defined
+  ansible.builtin.assert:
+    that:
+      - cifmw_ci_gen_kustomize_values_ctlplane_interface is defined
+      - cifmw_ci_gen_kustomize_values_ctlplane_interface != ''
--- a/roles/ci_gen_kustomize_values/templates/ovs-dpdk-sriov/edpm-values/values.yaml.j2
+++ b/roles/ci_gen_kustomize_values/templates/ovs-dpdk-sriov/edpm-values/values.yaml.j2
@@ -1,0 +1,36 @@
+{% set instances_names = []                                                            %}
+{% set _original_nodeset = (original_content.data | default({})).nodeset | default({}) %}
+{% set _original_nodes = _original_nodeset.nodes | default({})                         %}
+{% for _inst in cifmw_baremetal_hosts.keys()                                           %}
+{%   set _ = instances_names.append(_inst)                                             %}
+{% endfor                                                                              %}
+data:
+  baremetalSetTemplate:
+    ctlplaneInterface: {{ cifmw_ci_gen_kustomize_values_ctlplane_interface }}
+  ssh_keys:
+    authorized: {{ cifmw_ci_gen_kustomize_values_ssh_authorizedkeys | b64encode }}
+    private: {{ cifmw_ci_gen_kustomize_values_ssh_private_key | b64encode }}
+    public: {{ cifmw_ci_gen_kustomize_values_ssh_public_key | b64encode }}
+  nova:
+    migration:
+      ssh_keys:
+        private: {{ cifmw_ci_gen_kustomize_values_migration_priv_key | b64encode }}
+        public: {{ cifmw_ci_gen_kustomize_values_migration_pub_key | b64encode }}
+  nodeset:
+    ansible:
+      ansibleUser: "zuul"
+      ansibleVars:
+        edpm_fips_mode: "{{ 'enabled' if cifmw_fips_enabled|default(false)|bool else 'check' }}"
+        timesync_ntp_servers:
+          - hostname: "{{ cifmw_ci_gen_kustomize_values_ntp_srv | default('pool.ntp.org') }}"
+{% if cifmw_ci_gen_kustomize_values_sshd_ranges | default([]) | length > 0             %}
+        edpm_sshd_allowed_ranges:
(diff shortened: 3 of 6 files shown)


roles/ci_gen_kustomize_values/README.md
```diff
@@ -39,9 +39,26 @@ This configMap needs some more parameters in order to properly override the `arc
...
 * `cifmw_ci_gen_kustomize_values_migration_priv_key`: (String) SSH private key dedicated for the nova-migration services.
 * `cifmw_ci_gen_kustomize_values_migration_pub_key`: (String) SSH public key associated to `cifmw_ci_gen_kustomize_values_migration_priv_key`.
+* `cifmw_ci_gen_kustomize_values_provisioning_interface`: (String) Used to override default provisioning interface from OCP nodes.
+* `cifmw_ci_gen_kustomize_values_root_password`: (String) Used to override default baremetal compute root password.
+* `cifmw_ci_gen_kustomize_values_os_net_configs_mappings`: (Dictionary) Used to override default os-net-config nic mappings.
+* `cifmw_ci_gen_kustomize_values_network_config_template`: (List) Used to override default os-net-config network config template.
 
 Note that all of the SSH keys should be in `ecdsa` format to comply with FIPS directives.
 
+NFV parameters in order to properly override the `architecture` provided one based on compute node.
```
pablintino, line 49: We don't go that deep in this role, if the user wants to override those values, why not pathing the output using `cifmw_architecture_user_kustomize`?

roles/ci_gen_kustomize_values/templates/ovs-dpdk-sriov/edpm-values/values.yaml.j2
```diff
@@ -0,0 +1,74 @@
...
+{% for _inst in cifmw_networking_env_definition.instances.keys()                       %}
+{%   if _inst.startswith(_vm_type)                                                     %}
+{%     set _ = instances_names.append(_inst)                                           %}
+{%   endif                                                                             %}
+{% endfor                                                                              %}
+data:
+  root_password: {{ cifmw_ci_gen_kustomize_values_root_password | b64encode }}
+  preProvisioned: {{ cifmw_ci_gen_kustomize_values_pre_provisioned | b64encode }}
+  baremetalSetTemplate:
+    ctlplaneInterface: {{ cifmw_networking_env_definition.instances[instance].networks.ctlplane.interface_name }}
```
pablintino, line 14: Are you sure this is the value you want here? iiuc correctly this value is used in out OSP BMO to render [this](https://github.com/openstack-k8s-operators/openstack-baremetal-operator/blob/77086641d3fd316776dc9f771bd20b0f593cba29/templates/openstackbaremetalset/cloudinit/networkdata#L6)

roles/ci_gen_kustomize_values/templates/ovs-dpdk-sriov/edpm-values/values.yaml.j2
```diff
@@ -0,0 +1,74 @@
...
+        edpm_ovn_bridge_mappings: "{{ cifmw_ci_gen_kustomize_values_ovn_bridge_mapping }}"
+        edpm_neutron_sriov_agent_SRIOV_NIC_physical_device_mappings: "{{ cifmw_ci_gen_kustomize_values_sriov_physical_device_mapping }}"
+{% if cifmw_ci_gen_kustomize_values_sshd_ranges | default([]) | length > 0             %}
+        edpm_sshd_allowed_ranges:
+{%   for range in cifmw_ci_gen_kustomize_values_sshd_ranges                            %}
+          - "{{ range }}"
+{%   endfor                                                                            %}
+{% endif                                                                               %}
+    nodes:
+{% for instance in instances_names                                                     %}
```
pablintino, line 60: This list is empty in NFV cases, as the `cifmw_networking_env_definition` doesn't have computes.
The main source of information for this type of VAs is the `cifmw_baremetal_hosts` variable. Iterate over that dict to generate this. At the same time, if we are going to use OSP BMO and the IPAM integration, all the keys here but hostName are not needed, keep just that one.

roles/ci_gen_kustomize_values/templates/ovs-dpdk-sriov/edpm-values/values.yaml.j2
Jaganathancse: @pablintino 
We have covered the customization in architecture repo. This role generates VA2 architecture edpm values.yaml  format only right? not OSP BMO CR

roles/ci_gen_kustomize_values/templates/ovs-dpdk-sriov/edpm-values/values.yaml.j2
Jaganathancse: > This list is empty in NFV cases, as the `cifmw_networking_env_definition` doesn't have computes. The main source of information for this type of VAs is the `cifmw_baremetal_hosts` variable. Iterate over that dict to generate this. At the same time, if we are going to use OSP BMO and the IPAM integration, all the keys here but hostName are not needed, keep just that one.

Ack

roles/ci_gen_kustomize_values/README.md
Jaganathancse: @pablintino ack

roles/ci_gen_kustomize_values/tasks/edpm_core_asserts.yml
```diff
@@ -25,3 +25,5 @@
       - cifmw_ci_gen_kustomize_values_ssh_private_key != ''
       - cifmw_ci_gen_kustomize_values_ssh_public_key is defined
       - cifmw_ci_gen_kustomize_values_ssh_public_key != ''
+      - cifmw_ci_gen_kustomize_values_ctlplane_interface is defined
```
cjeanner, line 28: as I see things, the way we have to ensure we're having the needed parameters won't scale.

Wondering if we should:
- keep the check in the ci-framework (and find a way to make it scalable)
- drop it completely and just let the template fail (but that won't provide any nice message showing what's missing)
- add a new entry in the "stages" defined in the automation file, listing mandatory parameters for the template (maybe without the `cifmw_ci_gen_kustomize_` prefix to make it "generic" ?)

In any cases, we must have that discussion - the sooner the better

cc @abays @pablintino @psathyan @rebtoor @lewisdenny 

Guess we need a proper sync/call about that, an design a solution.

Jaganathancse: > as I see things, the way we have to ensure we're having the needed parameters won't scale.
> 
> Wondering if we should:
> 
> * keep the check in the ci-framework (and find a way to make it scalable)
> * drop it completely and just let the template fail (but that won't provide any nice message showing what's missing)
> * add a new entry in the "stages" defined in the automation file, listing mandatory parameters for the template (maybe without the `cifmw_ci_gen_kustomize_` prefix to make it "generic" ?)
> 
> In any cases, we must have that discussion - the sooner the better
> 
> cc @abays @pablintino @psathyan @rebtoor @lewisdenny
> 
> Guess we need a proper sync/call about that, an design a solution.

@cjeanner 
Moved cifmw_ci_gen_kustomize_values_ctlplane_interface validation only when cifmw_baremetal_hosts is deined.

cjeanner: Is this for "any" kind of host, or would this be only for computes?

Asking, because in such a case, that block should go in https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/ci_gen_kustomize_values/tasks/edpm_values.yml

in any cases... we'll have to find a better way.

Jaganathancse: @cjeanner edpm-values only for computes. this link is common and not specific to baremetal computes.

cjeanner: OK, so:
- we'll keep the core validations only, since they apply to any cases (we know for now)
- I'll do a follow-up to clean non-core validations
- I'll try to make a doc about those validations and why we probably won't add anything new (core vs specific)
- README will also show that situation: template will fail on its own if you don't provide the expected parameters

That's more sustainable than the current content, and we can't really rely on the `architecture` automation for those validations: it doesn't have  a clue about "what a parameter is", and it should stay as is.

PR #1481, opened 2024-04-16
Change default for cifmw_virtualbmc_daemon_port

karelyatin:
vbmcd is also setup by dev-scripts with default
server_port i.e 50891 and that conflicts with vbmc setup from cifmw.
This patch changes default for cifmw_virtualbmc_daemon_port to 50881 to avoid the conflict, since changed default dropped the overrides from scenario files.

Related-Issue: OSPRH-6324

  - [ ] README in the role


cjeanner:
I'll close this PR:

the issue is hit with DT that aren't inheriting from the established VA scenario files. After some discussions with @karelyatin we'll advertise this "good practice":

**for a given DT derived from an established VA, it is expected to include the VA scenario file, and just override/extend the needed bits to get the DT**

This should ensure defaults are good, and avoid a lot of duplicated content.

PR #1482, opened 2024-04-16
Delete/Recreate openshift marketplace pods

raukadah:
On crc Zuul reproducer job, cert-manager operator installation is failing with following error.
```
the cert-manager CRDs are not yet installed on the Kubernetes API server
```

The cert-manager operator get installed from openshift marketplace. After digging deep, we found that pods under openshift-marketplace namespace are hitting CrashLoopBackOff due to following error.
```
failed to populate resolver cache from source redhat-operators/openshift-marketplace:
failed to list bundles: rpc error: code = Unavailable desc = connection error: desc =
```

Based on https://github.com/crc-org/crc/issues/4109#issuecomment-2042497411, Delete and recreating openshift-marketplace pods fixes the issue.
    
Since OCP is deployed after pre_infra hook and cert_manager role iscalled before post_infra. There is no way to run this workaround as a hook.
    
It would be best to include under openshift_setup role.

### DIFF
--- a/roles/openshift_setup/tasks/fix_openshift_marketplace.yml
+++ b/roles/openshift_setup/tasks/fix_openshift_marketplace.yml
@@ -1,0 +1,22 @@
+---
+# Note(Chandan): This workaround is taken from
+# https://github.com/crc-org/crc/issues/4109#issuecomment-2042497411.
+# It will be removed once https://github.com/crc-org/crc/issues/4109
+# closed.
+- name: Delete the pods from openshift-marketplace namespace
+  kubernetes.core.k8s:
+    kind: Pod
+    state: absent
+    kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
+    namespace: openshift-marketplace
+
+- name: Wait for openshift-marketplace pods to be running
+  kubernetes.core.k8s_info:
+    kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
+    kind: Pod
+    namespace: openshift-marketplace
+    wait: true
+    wait_condition:
+      type: Ready
+      status: "True"
+    wait_timeout: 300
--- a/roles/openshift_setup/tasks/main.yml
+++ b/roles/openshift_setup/tasks/main.yml
@@ -166,3 +166,6 @@
           spec:
             additionalTrustedCA:
               name: "registry-cas"
+
+- name: Fix openshift-marketplace pods
+  ansible.builtin.import_tasks: fix_openshift_marketplace.yml


hooks/playbooks/disable_catalogsource.yml
```diff
@@ -0,0 +1,18 @@
+---
+- name: Disable/Enable default CatalogSource
```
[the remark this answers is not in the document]
raukadah: Done

hooks/playbooks/disable_catalogsource.yml
```diff
@@ -0,0 +1,52 @@
...
+      ansible.builtin.command:
+        cmd: >-
+          oc get pods -o name -n openshift-marketplace
+      register: _pod_list
+
+    - name: Get the redhat-operators pods having CrashLoopBackOff state
+      kubernetes.core.k8s_info:
+        kind: Pod
+        kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
+        name: "{{ pod_list.stdout | regex_search('^pod/redhat-operators-.*$', multiline=True) | split('/') | last }}"
```
pablintino, line 19: Uhmm I don't get this one, this will query a single pods as your are passing the name field. Isn't enough if you remove the previous task and the name? The field_selector here should do the rest.

raukadah: Yes, got it now! Updated it.

raukadah:
recheck

scenarios/centos-9/multinode-ci.yml
```diff
@@ -4,6 +4,12 @@ cifmw_use_libvirt: false
 
 cifmw_openshift_setup_skip_internal_registry_tls_verify: true
 
+pre_infra:
```
lewisdenny, line 7: We can't use `pre_infra` here as OCP isn't available at that stage. 

`post_infra` looks to be the correct hook to use but I didn't test.

You can see CI failing due to this[1]:
```
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Could not find or access '/home/zuul/.crc/machines/crc/kubeconfig' on the Ansible Controller.\nIf you are using a module and expect the file to exist on the remote, see the remote_src option"}
```
[1] https://logserver.rdoproject.org/82/1482/43267db1d1546643bcf8d8f5f0a5fc8cc69f8f2d/github-check/podified-multinode-edpm-deployment-crc/a1df9e4/controller/ci-framework-data/logs/ci_script_000_run_disable_enable_red_hat.log

lewisdenny (changes requested):
Interesting issue Chandan, I was going to suggest we add a todo to remove this "workaround" once fixed but seems from the RH solution you linked this is just how it works.

One request left

hooks/playbooks/disable_catalogsource.yml
```diff
@@ -0,0 +1,41 @@
...
+  gather_facts: false
+  tasks:
+    - name: Get the pods having CrashLoopBackOff state
+      kubernetes.core.k8s_info:
+        kind: Pod
+        kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
+        namespace: openshift-marketplace
+        field_selectors:
+          - status.phase=CrashLoopBackOff
+      register: _pod_status
```
karelyatin, line 13: just deleting the crash pod will also recover it https://github.com/crc-org/crc/issues/4109#issuecomment-2042497411 
Issue seen in 4.15.3 atleast, 4.15.8 didn't hit it.

scenarios/centos-9/multinode-ci.yml
raukadah: Done!

raukadah: Thank you @pablintino @lewisdenny for the review. Since OCP is deployed after pre_infra hook and cert_manager role iscalled before post_infra. There is no way to run this workaround
as a hook so we need to include in the cert_manager role itself. I have updated the same.

lewisdenny: Nice, let's see how CI likes it. Thanks for adding the comment too :)

arxcruz:
/lgtm
lgtm

rebtoor (approved):

PR #1483, opened 2024-04-16
Fix push_code and zuul_vars vs zuul

cjeanner:
With the addition of the molecule job reproducer, we had to take
parameters one level up in the environment. Those parameters were at the
same level as `zuul` - so we created a new `zuul_vars`, containing `zuul` itself.

This patch now checks for both `zuul` and `zuul_vars`, so that we can
match in CI runtime and in a job reproducer.

- [X] Working outside of zuul (@cjeanner)
- [x] Working with a molecule job reproducer (@cjeanner)
- [x] Working with a deploy job reproducer (@cjeanner)
- [x] Downstream testproject (@cescgina)
- [x] Downstream VA job (@cescgina)

### DIFF
--- a/roles/reproducer/tasks/push_code.yml
+++ b/roles/reproducer/tasks/push_code.yml
@@ -1,14 +1,14 @@
 ---
 - name: Sync zuul content if available
   when:
-    - zuul_vars.zuul is defined
-    - zuul_vars.zuul.items is defined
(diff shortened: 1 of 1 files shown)


cjeanner:

Zuul has some issues (a patch broke crc based jobs, and some hardware issues are hitting randomly).

The testing done on our side, covering more cases than the CI, should provide enough warranties.

PR #1492, opened 2024-04-17
Arbitrary hook type for rapid experiment

afazekas:
In case you have something simple to try out
in architecture/automation via this hook you can do it:
 - without cross repo change
 - without too many readings
 - less string escaping concerns

  - [ ] README in the role

### DIFF
--- a/roles/run_hook/tasks/arbitrary.yml
+++ b/roles/run_hook/tasks/arbitrary.yml
@@ -1,0 +1,6 @@
+---
+- name: Call code
+  ansible.builtin.shell:  # noqa: command-instead-of-module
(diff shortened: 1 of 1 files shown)


cjeanner:
This opens the door to all of the bad practices anyone can think of, and even beyond that. So, really, no.

rebtoor:
I agree with @cjeanner.

afazekas:
At least please have the framework to not have to playbook people deal with:
 https://github.com/openstack-k8s-operators/ci-framework/blob/main/hooks/playbooks/link2file.yml#L30
 
 In user facing yaml what looks like string "foo bar"  become "foo" .

PR #1493, opened 2024-04-17
Fix relative path for custom CR hook type

queria:
With last change renaming hook type from 'crd' to 'cr', there are two typos:
- path to builtin hook CR directory changed from `/hooks/crds` to `/hooks/crs`
- and typo which was there already before `hook.src` should be `hook.source`

  - [x] README in the role

### DIFF
--- a/roles/run_hook/tasks/cr.yml
+++ b/roles/run_hook/tasks/cr.yml
@@ -24,8 +24,8 @@
       {{
         (
           ansible.builtin.role_path,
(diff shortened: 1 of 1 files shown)


psathyan (approved):

cjeanner:
/approve

@psathyan guess we'd need some quick molecule - meaning switching the nodeset to crc, and try to load a CR via the hook. WDYT?

PR #1498, opened 2024-04-17
Don't trigger deployment job on reproducer role change

cjeanner:
This role isn't used in upstream deployment jobs, let's not trigger any.


lewisdenny (approved):
/lgtm Makes sense and regex looks correct

https://regex101.com/r/OC64H5/1

psathyan (approved):

PR #1501, opened 2024-04-18
[reproducer] Fix remaining zuul -> zuul_vars change

lewisdenny:

### DIFF
--- a/roles/reproducer/tasks/ci_job.yml
+++ b/roles/reproducer/tasks/ci_job.yml
@@ -84,7 +84,7 @@
           vars:
             zuul_params_filtered: >-
               {{
-                zuul | dict2items |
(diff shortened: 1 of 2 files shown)


roles/reproducer/tasks/ci_job.yml
```diff
@@ -84,7 +84,7 @@
           vars:
             zuul_params_filtered: >-
               {{
-                zuul | dict2items |
+                zuul_vars | dict2items |
```
cjeanner, line 87: I think it should be `zuul_vars.zuul` actually

weird I didn't see it :/

lewisdenny: Yeah I think your right, testing

PR #1503, opened 2024-04-18
Replace shell cmds with ansible in append repos task

raukadah:
Append repos in dlrn targets task removes """ from end of mock config file and move it to a local mock
config file.
    
Then, we read all the repo file content and append it in the local mock config file.
    
In the end, we need to add back """ in the end of the local mock config file.
    
The whole process is replaced by ansible.We no longer needs to enable or disble any repos.
repo-setup takes care of that.


cjeanner:
/approve

replacing a mighty shell with ansible sequence is nice, thanks!

marios (approved):
lgtm

PR #1506, opened 2024-04-18
Ensure the configuration patch works

cjeanner:
Modifying molecule to ensure the libvirt_manager_configuration patch is
properly working.

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/1507

### DIFF
--- a/roles/libvirt_manager/molecule/deploy_layout/converge.yml
+++ b/roles/libvirt_manager/molecule/deploy_layout/converge.yml
@@ -153,6 +153,7 @@
             compare_vms: >-
               {{
                 ['cifmw-compute-0',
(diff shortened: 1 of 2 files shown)


cjeanner:
recheck

psathyan (approved):
I was only concerned of `_layout` but molecule passed :)

Thank you.

cjeanner:
> I was only concerned of `_layout` but molecule passed :)
> 
> Thank you.

yeaaahh... that one is... well. I guess I'll have to rename it sooner or later. It's the updated layout used internally to the role.

lewisdenny (approved):

PR #1507, opened 2024-04-18
Add "reproducer" role ignore to the right place

cjeanner:
- [X] no test can be made at that point


cjeanner:
I added a Depends-On in another patch (#1506) and it works fine.

raukadah (approved):
/approve

PR #1509, opened 2024-04-18
Group common parameters for Architecture deployment

cjeanner:
A series of parameters are mandatory in all cases. Let's avoid
duplication and group them in a single file.

That single file is then loaded if and only if we set the
`cifmw_architecture_scenario` parameter to a non-empty string.

As noted in the va-common.yml file, ANY parameter there can be
overridden by a parameter file, or even using `-e foo=bar` directly.

This should make "creating new VA or DT" process slightly easier for
everyone, while providing a unified way to fix all the scenarios in case
of new parameter/feature/others.

### DIFF
--- a/scenarios/reproducers/va-common.yml
+++ b/scenarios/reproducers/va-common.yml
@@ -1,0 +1,67 @@
+---
+# This file groups common parameters for Architecture based deployment.
+# It is *automatically* included in any run having cifmw_architecture_scenario
+# parameter set.
+# You can of course manually pass it, it won't make any difference.
+#
+# Any of those parameter might be overridden from your own parameter file!
+#
+# This file should NOT be used for the CRC based deployment (3-nodes.yml).
+
+
+# Ensure some basic directories and parameter are set
+cifmw_install_yamls_repo: >-
+  {{
+    (ansible_user_dir,
+     'src/github.com/openstack-k8s-operators/install_yamls') |
+     path_join
+  }}
+cifmw_basedir: "{{ ansible_user_dir }}/ci-framework-data"
+cifmw_path: "{{ ansible_user_dir }}/bin:{{ ansible_env.PATH }}"
+
+# Ensure our vbmcd isn't conflicting with devscripts managed instance
+cifmw_virtualbmc_daemon_port: 50881
+
+# Ensure we pass libvirt backend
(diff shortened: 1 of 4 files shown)


scenarios/reproducers/va-common.yml
```diff
@@ -0,0 +1,67 @@
+---
+# This file groups common parameters for Architecture based deployment.
+# It is *automatically* included in any run having cifmw_architecture_scenario
+# parameter set.
+# You can of course manually pass it, it won't make any difference.
+#
+# Any of those parameter might be overridden from your own parameter file!
+#
+# This file should NOT be used for the CRC based deployment (3-nodes.yml).
```
lewisdenny, line 9: Can we use an assert checking for something like `cifmw_rhol_crc_config` which would be defined if 3-nodes.yml was being used to enforce this?

scenarios/reproducers/va-common.yml
```diff
@@ -0,0 +1,67 @@
...
+  {{
+    (cifmw_repo_setup_os_release is defined and
+     cifmw_repo_setup_os_release == 'rhel') |
+    ternary(4, 1)
+  }}
+
+cifmw_libvirt_manager_compute_amount: 3
+cifmw_libvirt_manager_compute_disksize: 20
+cifmw_libvirt_manager_compute_memory: 4
+cifmw_libvirt_manager_compute_cpus: 1
```
pablintino, line 58: Not the case for NFV and the bast majority of hybrid setups, that won't have virtual computes.

scenarios/reproducers/va-common.yml
cjeanner: including the file shouldn't break things either. Just... "kind of useless data for your 3-nodes". It's "should", not "must" :).

scenarios/reproducers/va-common.yml
cjeanner: NFV doesn't consume those parameter at all (no virtual compute). but a DT? or future VA? IMHO it doesn't hurt to move them in that file.

pablintino: As discussed in the review call, if this doesn't imply we are going to deploy those computes in NFV, I'm fine.

PR #1510, opened 2024-04-18
Pinning golang version for doc builds to 1.21.7

jpodivin:
There is an ongoing issue with doc builds under dataplane-operator-docs-preview https://review.rdoproject.org/zuul/builds?job_name=dataplane-operator-docs-preview The new version of go `1.22` is not compatible with some of the source, due to several modules moving from experimental to builtins.

  - [x] README in the role

### DIFF
--- a/ci/playbooks/pre-doc.yml
+++ b/ci/playbooks/pre-doc.yml
@@ -23,5 +23,22 @@
         name:
           - make
           - python3
-          - golang
           - ruby
+
+    - name: Download golang version 1.21.7
+      ansible.builtin.get_url:
+        source: https://go.dev/dl/go1.21.7.linux-amd64.tar.gz
+        dest: go1.21.7.linux-amd64.tar.gz
+        checksum: ffeabdfe10a7e84197ac7d28fda7484d0f855a92
+
+    - name: Install golang
+      become: true
+      ansible.builtin.unarchive:
+        src: go1.27.linux-amd64.tar.gz
+        dest: ~/bin
+        include:
+          - bin/go
+          - bin/gofmt
+        extra_opts:
+          - --strip-components=2
+        creates: ~/bin/go


ci/playbooks/pre-doc.yml
```diff
@@ -23,5 +23,19 @@
...
+        checksum: ffeabdfe10a7e84197ac7d28fda7484d0f855a92
+
+    - name: Install golang
+      become: true
+      ansible.builtin.command:
+        cmd: |
+          tar -C ~/bin -xzf go1.21.7.linux-amd64.tar.gz
+          export PATH=$PATH:~/bin/go/bin
+        args:
+          creates: /usr/local/bin/go
```
rebtoor, line 41:
```suggestion
          creates: ~/bin/go
```

ci/playbooks/pre-doc.yml
```diff
@@ -23,5 +23,19 @@
...
+      ansible.builtin.get_url:
+        source: https://go.dev/dl/go1.21.7.linux-amd64.tar.gz
+        dest: go1.21.7.linux-amd64.tar.gz
+        checksum: ffeabdfe10a7e84197ac7d28fda7484d0f855a92
+
+    - name: Install golang
+      become: true
+      ansible.builtin.command:
+        cmd: |
+          tar -C ~/bin -xzf go1.21.7.linux-amd64.tar.gz
```
rebtoor, line 38:
```suggestion
          tar -xzf go1.21.7.linux-amd64.tar.gz -C ~/bin go/bin/ --strip-components=2
```

ci/playbooks/pre-doc.yml
rebtoor, line 41:
```suggestion
    - name: Install golang
      ansible.builtin.unarchive:
        src: go1.21.7.linux-amd64.tar.gz
        dest: ~/bin/
        remote_src: true
        include: go/bin/
        extra_opts:
          - --strip-components=2
```

rebtoor: we could also use the unarchive module to download the tar, but we'd lose the checksum verification.

lewisdenny (commented):
Might something like this be simpler:
```
---
- name: Test
  hosts: localhost
  vars:
    _golang_version: "-1.21.7"
  tasks:
    - name: Install required packages
      become: true
      ansible.builtin.package:
        name:
          - make
          - python3
          - golang{{ _golang_version | default("") }}
          - ruby
```

Another thought is maybe this version should be defined outside this playbook in the projects job definition? 

For example here [1]. The reason being this change will effect 3 projects and the other two jobs are not affected by go1.22. They can continue with whatever is install by default and `dataplane-operator-docs-preview` can be pinned.

[1] https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/zuul.d/jobs.yaml#L5

jpodivin:
> Might something like this be simpler:
> 
> ```
> ---
> - name: Test
>   hosts: localhost
>   vars:
>     _golang_version: "-1.21.7"
>   tasks:
>     - name: Install required packages
>       become: true
>       ansible.builtin.package:
>         name:
>           - make
>           - python3
>           - golang{{ _golang_version | default("") }}
>           - ruby
> ```
> 
> Another thought is maybe `_golang_version` should be defined outside this playbook in the projects job definition? i.e: [1]
> 
> The reason being this change will effect 3 projects and the other two jobs are not affected by go1.22. They can continue with whatever is install by default and `dataplane-operator-docs-preview` can be pinned.
> 
> [1] https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/zuul.d/jobs.yaml#L5

In principle this could work. As long as repos carry the version in question. History has proven that we can not rely on that.
By downloading it straight from source we avoid the issue completely. It is true that other projects are not affected. 
But I'm not certain that will remain the case in the long run. GO does make breaking changes in minor versions.

Eventually we will hit this again.

rebtoor (approved):

PR #1528, opened 2024-04-22
Change to YAML stdout for Ansible

lewisdenny:
The current json output can be hard to read with large outputs from
errors.


cjeanner:
What "large output" are you referring to? Are we missing any dedicated ansible logging or calls to ci_script, allowing to get the task log directly in a nice and readable file?

lewisdenny:
> What "large output" are you referring to? Are we missing any dedicated ansible logging or calls to ci_script, allowing to get the task log directly in a nice and readable file?

Mainly during development work is when loops fail in a spectacular way. The `stdout` is for humans to read and IMO YAML is easier to read vs JSON.

cjeanner:
/approve

won't hurt - and if it makes life easier, let's go!

PR #1529, opened 2024-04-22
Extract podman related bits in its own role

cjeanner:
Since we'll see more podman containers in the near future, let's get a
deidcated role to group the needed actions.

This new `podman` role installs the package, and enable session
lingering if requestes (defaults `true`).

We can also override the package list to install in case we want more of
podman-related content, such as buildah, skopeo and so on.

  - [X] README in the role


raukadah (commented):
/approve

PR #1530, opened 2024-04-22
Exclude main.yml from found_validations

bshephar:
When we run all validations, we need to avoid executing tasks/main.yml, since that would result in an infinite loop. We only want to run named validation files under the tasks/team/ directories.

Signed-off-by: Brendan Shephard <bshephar@redhat.com>

  - [x] README in the role


lewisdenny (approved):
/lgtm

cjeanner (commented):
/approve

PR #1533, opened 2024-04-22
repro/validations: Fix undefined cifmw_target_host

queria:
In reproducer validations, error message for OS check references cifmw_target_host which is not mandatory parameter and this can result in:
```
  {"msg": "The task includes an option with an undefined variable.
   The error was: 'cifmw_target_host' is undefined. 'cifmw_target_host' is undefined

   The error appears to be in '/home/zuul/ci-framework/roles/reproducer/tasks/validations.yml': line 37, column 3
   ...
```

Avoid this by falling back to value of whatever is current ansible_host.

  - [ ] README in the role


roles/reproducer/tasks/validations.yml
```diff
@@ -36,7 +36,7 @@
       - ansible_distribution in cifmw_reproducer_supported_hypervisor_os
       - ansible_distribution_version | float  >= cifmw_reproducer_supported_hypervisor_os[ansible_distribution].minimum_version
     msg: >-
-      Please ensure host: {{ cifmw_target_host }} is running a supported OS.
+      Please ensure host: {{ cifmw_target_host | default(ansible_host) }} is running a supported OS.
```
cjeanner, line 39: I'd rather use `| default('localhost')`
reason:
- ansible_host may not be set (it comes from the inventory)
- at that point, the only reason to not have `cifmw_target_host` is a run against `localhost`
- default value for `cifmw_target_host` is `localhost` in all playbooks

queria: Ack, I wasnt sure about ansible_host being safe(r), thanks fixed.

PR #1534, opened 2024-04-22
[bop]Move dlrn-venv to a proper ansible var

raukadah:
This pull request moves the hardcoded dlrn venv path to a proper ansible var. It also modified the DLRN installation shell related tasks to ansible pip
tasks.

### DIFF
--- a/roles/build_openstack_packages/tasks/install_dlrn.yml
+++ b/roles/build_openstack_packages/tasks/install_dlrn.yml
@@ -74,7 +74,7 @@
 - name: Create dlrn-venv
   ansible.builtin.pip:  # noqa: package-latest
     name: pip
-    virtualenv: "{{ cifmw_bop_build_repo_dir }}/dlrn-venv"
+    virtualenv: "{{ cifmw_bop_dlrn_venv }}"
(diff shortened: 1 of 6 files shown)


roles/build_openstack_packages/tasks/install_dlrn.yml
```diff
@@ -93,29 +93,30 @@
...
+    - name: Install DLRN from source
+      ansible.builtin.command:
+        cmd: "{{ cifmw_bop_dlrn_venv }}/bin/python setup.py install"
+        chdir: "{{ cifmw_bop_build_repo_dir }}/DLRN"
+        creates: "{{ cifmw_bop_dlrn_venv }}/bin/dlrn"
 
 - name: Install DLRN from pip
   when: not cifmw_bop_dlrn_from_source|bool
   block:
     - name: Pip install rdopkg
```
pablintino, line 110: Why not use a single pip task? It accepts a list of packages. Should be a bit faster.

raukadah: Yes, correct, Done!

PR #1538, opened 2024-04-22
Fix ansible lint issue raised in pod job

cjeanner:
```
WARNING  Listing 1 violation(s) that are fatal
roles/shiftstack/molecule/default/cleanup.yml:30: fqcn[action]: Use FQCN for module actions, such `kubernetes.core.k8s`.
```

It seems the linter running in Zuul is stricter than the one in prow.


cjeanner:
recheck github clone issue.

PR #1539, opened 2024-04-22
Allow some retries for setup_molecule run

cjeanner:
#### Allow some retries for setup_molecule run

We see some unrelated failures while running the pod, during setup
stage: apparently, the DNS may fail to answer in time, and this crashes
the env setup.

With this patch, we should be in better position, since it would retry
to install the libs in case of failure, with a small delay.

Hopefully, this change will stabilize enough the job so that we may
consider it "voting" in a near future.

#### Ensure we don't run deploy jobs on unrelated modifications

A couple of files nested in ci/playbooks aren't used at all in any
deployment job.
Let's avoid useless resources usage whenever we modify them!

### DIFF
--- a/ci/playbooks/pre-commit.yml
+++ b/ci/playbooks/pre-commit.yml
@@ -10,19 +10,23 @@
           - python3
           - python3-pip
 
-    - name: Run makes target
+    - name: Run make targets
       vars:
         src_dir: >-
           {{
             zuul.projects['github.com/openstack-k8s-operators/ci-framework'].src_dir
           }}
       block:
-        - name: Instal dependencies
+        - name: Install dependencies
+          register: _setup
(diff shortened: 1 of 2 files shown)


ci/playbooks/pre-commit.yml
```diff
@@ -18,11 +18,15 @@
           }}
       block:
         - name: Instal dependencies
```
rebtoor, line 20: small typo here!

cjeanner: right - fixed with some others.

rebtoor (approved):

cjeanner:
Note: the pre-commit job in zuul WILL fail - it needs #1538 to pass correctly.

PR #1542, opened 2024-04-22
Create local PV directories with 0775 permissions

cschwede:
k8s changes permissions of the PV directories when starting eg. an StatefulSet. However, if the playbook is rerun after starting a deployment, the group permissions will be reset to read-only and services can't create new subdirectories.

Changing the group permissions to writeable should be ok for new directories, as the default group is root until the PV gets mounted and prevents service errors if the playbook is re-run.

### DIFF
--- a/roles/ci_local_storage/molecule/default/converge.yml
+++ b/roles/ci_local_storage/molecule/default/converge.yml
@@ -51,7 +51,7 @@
                 ] | path_join
               }}
             state: directory
-            mode: "0755"
+            mode: "0775"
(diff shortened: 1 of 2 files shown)


psathyan (approved):

karelyatin (commented):
/lgtm

raukadah (commented):
/approve

psathyan (approved):

cjeanner:
oh, thank you @cschwede ! I was wondering if we could avoid things to change upon re-run - you've done it :)

PR #1545, opened 2024-04-23
[bop] Lower the priority of DLRN repo while using gating repo

raukadah:
Currently DLRN and gating repo have priority 1. In order to install packages from gating repo, we need to decrease the DLRN repos priority.

It will allow us to install packages from gating.rep Lower the priority of DLRN repo while using gating repo

It will allow us to install packages from gating.repo.

### DIFF
--- a/roles/build_openstack_packages/tasks/create_repo.yml
+++ b/roles/build_openstack_packages/tasks/create_repo.yml
@@ -50,3 +50,15 @@
       gpgcheck=0
       priority=1
     dest: "{{ cifmw_bop_gating_repo_dest }}/gating.repo"
+
+- name: Check for DLRN repo at the destination
+  ansible.builtin.stat:
+    path: "{{ cifmw_bop_gating_repo_dest }}/delorean.repo"
+  register: _dlrn_repo
+
+- name: Lower the priority of DLRN repos to allow installation from gating repo
+  when: _dlrn_repo.stat.exists
+  ansible.builtin.replace:
+    path: "{{ cifmw_bop_gating_repo_dest }}/delorean.repo"
+    regexp: "priority=1"
+    replace: "priority=20"


roles/build_openstack_packages/tasks/create_repo.yml
```diff
@@ -50,3 +50,15 @@
...
+- name: Check for DLRN repo at the destination
+  ansible.builtin.stat:
+    path: "{{ cifmw_bop_gating_repo_dest }}/delorean.repo"
+  register: _dlrn_repo
+
+- name: Lower the priority of DLRN repos to allow installation from gating repo
+  when: _dlrn_repo.stat.exists
+  ansible.builtin.replace:
+    path: "{{ cifmw_bop_gating_repo_dest }}/delorean.repo"
+    regexp: "priority=1"
```
cjeanner, line 63: are we sure this priority is always set?

raukadah: Yes, it is always set to 1, https://trunk.rdoproject.org/centos9-antelope/current-podified/delorean.repo
```
[delorean-component-barbican]
name=delorean-python-barbican-tests-tempest-82b0e48639ad52838902dffbc3acaa5bc789b87e
baseurl=https://trunk.rdoproject.org/centos9-antelope/component/barbican/82/b0/82b0e48639ad52838902dffbc3acaa5bc789b87e_553bf219
enabled=1
gpgcheck=0
priority=1
```
As it is the tested repo.

cjeanner: OK, so the plan is:
- go with that current form for now
- consider [toiletwater|https://galaxy.ansible.com/ui/repo/published/sivel/toiletwater/docs/] collection and its advanced TOML and INI support

I'd rather be 100% sure we won't see any weird change in DLRN leading to the removal of that `priority=1` - or even a small change of the value, leading to a crash...

PR #1548, opened 2024-04-23
Correct import vs include, apply tags

cjeanner:
The 06-deploy-architecture.yml playbook exposes a series of useful tags,
but due to the differences between import and include, some tasks
weren't running when we "select" them with a tag.

There were also issues with import vs include in the play:
- include will dynamically load the content when the play hits that
  task; any tag you set there will be applied only to that "include"
  action, and not reported to the included content.

- import, on the contrary, load all the content during the catalog
  build, and any tag (or `when` condition) set on that import task will
  be applied down to any of the imported content.

In short, when there is no condition on the import, let's use
`import_*`.
If we want to apply the tags down included content, we have to use
`apply` directive.

In order to target precisely the "deploy rhoso" part, new tags were
needed:
- edpm_bootstrap
- edpm_post

With those new tags, in case you want to iterate only on the "generate
CRs" for instance, you can call this:
```Bash
[controller-0]$ ./deploy-architecture.sh --tags edpm \
        --skip-tags edpm_bootstrap,edpm_post \
        -e cifmw_kustomize_deploy_generate_crs_only=true
```
This will target only the necessary tasks to start generating the CRs.

### DIFF
--- a/playbooks/06-deploy-architecture.yml
+++ b/playbooks/06-deploy-architecture.yml
@@ -11,16 +11,22 @@
   tasks:
     # end_play will end only current play, not the main edpm-deploy.yml
     - name: Early end if not architecture deploy
+      tags:
+        - always
       when:
         - cifmw_architecture_scenario is not defined
       ansible.builtin.meta: end_play
 
     - name: Load Networking Environment Definition
+      tags:
+        - always
       ansible.builtin.include_role:
         name: networking_mapper
         tasks_from: load_env_definition.yml
 
     - name: Fetch network facts
+      tags:
+        - always
       when:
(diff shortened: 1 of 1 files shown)


cjeanner:
Need to refine it a bit, with the import_playbook, we need some more things to target just the deployment...

cjeanner:
> The patch looks good to me but these extra tags should be documented here: https://ci-framework.readthedocs.io/en/latest/usage/01_usage.html#ansible-tags IMO

not really sure: this is architecture only, and I really, really don't expect upstream to use architecture path: it's not even documented for many, many reasons. So it's more something that will require a downstream update, in the cookbook. Agree?

PR #1549, opened 2024-04-23
Discover name of the logs PVC

lpiwowar:
This test-operator patch [1] changes the name of the logs PVC. This change requires an update of the test-operator role so that it is capable of discovering the logs pod.

[1] https://github.com/openstack-k8s-operators/test-operator/pull/85

  - [ ] README in the role


roles/test_operator/tasks/run-test-operator-job.yml
```diff
@@ -57,6 +57,20 @@
       {{ testjob.attempts == (cifmw_test_operator_timeout / 10) | round | int }}
   when: not cifmw_test_operator_dry_run | bool
 
+- name: Get inormation about PVC that stores the logs
```
cjeanner, line 60: you can push that in a block in order to group the two actions under the same `when` conditions. That would be better imho :)

lewisdenny: also inormation -> information

lewisdenny (changes requested):

roles/test_operator/tasks/run-test-operator-job.yml
lpiwowar: Ack, thank you @cjeanner and @lewisdenny. Fixed.

lewisdenny (approved):

PR #1552, opened 2024-04-23
[shiftstack] Add shifstack-qa gerrithub repo change support

eurijon:
This commit adds a new parameter `cifmw_shiftstack_qa_gerrithub_change` and adds support for fetching a gerrithub change from the shiftstack-qa repository (i.e. 'refs/changes/29/1188429/50').

  - [x] README in the role

### DIFF
--- a/docs/dictionary/en-custom.txt
+++ b/docs/dictionary/en-custom.txt
@@ -161,6 +161,7 @@ gapped
 genericcloud
 genindex
 gerrit
+gerrithub
 gib
 githooks
 golang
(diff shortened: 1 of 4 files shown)


lewisdenny:
@raukadah Is this something that would be better added into content provider? Note we may not have Gerrit Hub configured.

raukadah:
> @raukadah Is this something that would be better added into content provider? Note we may not have Gerrit Hub configured.

@lewisdenny Nope, content provider is meant for building contents from opendev and operator changes. 

shiftstack role is used for validating the environment just like tempest/test-operator does. It is triggered once the edpm deployment finishes. So I think this changes is specific to this role. So I think this change looks ok to me.

PR #1556, opened 2024-04-24
Install cert_manager from release manifest

raukadah:
This pr add the support of installing cert-manager fromrelease manifest. It will be useful when we saw issue with
 OLm resources.

Note: It also add "INSTALL_CERT_MANAGER: false" in multinode-ci.yml scenario as it has higher prefernce otherwise
It will call install_yaml certmanager scripts.

### DIFF
--- a/roles/cert_manager/defaults/main.yml
+++ b/roles/cert_manager/defaults/main.yml
@@ -23,6 +23,9 @@ cifmw_cert_manager_manifests_dir: "{{ cifmw_manifests | default(cifmw_cert_manag
 cifmw_cert_manager_operator_namespace: cert-manager-operator
 cifmw_cert_manager_openshift_version: stable-v1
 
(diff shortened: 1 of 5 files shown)


raukadah:
install_yamls cert-manager Issue discovered here: https://logserver.rdoproject.org/56/1556/6c4d4f5287d1e135de47eacb52bf669481beb5f5/github-check/podified-multinode-edpm-deployment-crc/1b12ed7/controller/ci-framework-data/artifacts/ci_script_004_run.sh and logs : https://logserver.rdoproject.org/56/1556/6c4d4f5287d1e135de47eacb52bf669481beb5f5/github-check/podified-multinode-edpm-deployment-crc/1b12ed7/controller/ci-framework-data/logs/ci_script_004_run.log

cjeanner:
/approve

thanks for that splitting, it will make things less verbose and slightly faster instead of the long `skipped` tasks list.

PR #1559, opened 2024-04-24
role: config_drive

hjensas:
Add a new role `config_drive` that can build a ISO image with instance metadata for cloud-init.

Uses the simple `NoCloud` DataSource.

  - [ ] README in the role


roles/config_drive/defaults/main.yml
```diff
@@ -0,0 +1,65 @@
...
+  }}
+cifmw_config_drive_workdir: "{{ cifmw_config_drive_basedir }}/cifmw_config_drive"
+
+cifmw_config_drive_uuid: uuid-test
+cifmw_config_drive_instancedir: "{{ cifmw_config_drive_workdir }}/{{ cifmw_config_drive_uuid }}"
+cifmw_config_drive_iso_image: "{{ cifmw_config_drive_workdir }}/{{ cifmw_config_drive_uuid }}.iso"
+cifmw_config_drive_name: test
+cifmw_config_drive_hostname: test.example.com
+cifmw_config_drive_userdata:
+# user-data - example
```
rlandy, line 32: Is there a better place to put an example other than with comments?
Wouldn't block on this.

rlandy (commented):
/lgtm

roles/config_drive/defaults/main.yml
hjensas: There is an example in the README, I guess we can drop this. (If we merge before I have chance to update it I can do a follow up to remove examples from here.)

roles/config_drive/README.md
```diff
@@ -0,0 +1,60 @@
+# config_drive
+Role to build a `NoCloud` config-drive ISO image.
+
+## Privilege escalation
+No privilege escalation required.
```
lewisdenny, line 5: Could you please update this, you are using become for installing packages

lewisdenny (changes requested):

roles/config_drive/defaults/main.yml
lewisdenny: +1 to removing it, I don't mind leaving the empty vars but the full example should live only in the README to prevent having multiple places to update in the future.

roles/config_drive/README.md
lewisdenny: Would be happy to approve/lgtm after those two issues are resolved.

roles/config_drive/defaults/main.yml
hjensas: Done

roles/config_drive/README.md
hjensas: Done

lewisdenny (approved):
/approve
/lgtm

PR #1561, opened 2024-04-24
 Use cifmw_rhol_crc_version to 2.30.0 for crc reproducer

raukadah:
In CI, we are using crc version 2.30.0. By default the crc reproducer
job deploys latest crc which breaks while reproducing the job locally and in crc ci-reproducer job.
    
This pr sets the cifmw_rhol_crc_version to 2.30.0 so that crc ci-reproducer reproduces the job without any issue.

### DIFF
--- a/roles/rhol_crc/defaults/main.yml
+++ b/roles/rhol_crc/defaults/main.yml
@@ -28,7 +28,7 @@ cifmw_rhol_crc_kubeadmin_pwd: 12345678
 cifmw_rhol_crc_dryrun: false
 
 # Binary variables:
-cifmw_rhol_crc_version: latest
+cifmw_rhol_crc_version: 2.30.0
 cifmw_rhol_crc_tarball_name: crc-linux-amd64.tar.xz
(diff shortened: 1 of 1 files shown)


scenarios/reproducers/3-nodes.yml
```diff
@@ -7,6 +7,7 @@ cifmw_install_yamls_repo: "~/src/github.com/openstack-k8s-operators/install_yaml
 cifmw_basedir: "{{ ansible_user_dir }}/ci-framework-data"
 cifmw_path: "{{ ansible_user_dir }}/.crc/bin:{{ ansible_user_dir }}/.crc/bin/oc:{{ ansible_user_dir }}/bin:{{ ansible_env.PATH }}"
 
+cifmw_rhol_crc_version: 2.30.0
```
cjeanner, line 10: maybe better to bump the version in the role directly?

raukadah: Done!

raukadah:
Closes in favor of https://github.com/openstack-k8s-operators/ci-framework/pull/1562

PR #1563, opened 2024-04-25
Use base.yaml irrelevant file list

raukadah:
It will help to avoid zuul job freeze issue[1]. where[1] child jobs gets triggering without running content provider.

Podified base job is parented from cifmw-podified-multinode-edpm-base-crc and is defined in base.yaml. Let's use the irrelevant file list from cifmw-podified-multinode-edpm-base-crc instead of overidding it.

Links:
[1]. https://github.com/openstack-k8s-operators/ironic-operator/pull/421#issuecomment-2069087141

### DIFF
--- a/zuul.d/podified_multinode.yaml
+++ b/zuul.d/podified_multinode.yaml
@@ -15,15 +15,6 @@
     nodeset: centos-9-medium-crc-extracted-2-30-0-3xl
     run:
       - ci/playbooks/edpm/run.yml
-    irrelevant-files:
-      - ^roles/.*_build
-      - ^roles/build.*
-      - ^roles/local_env_vm
-      - ^ci/templates
-      - ^docs
-      - ^.*/*.md
-      - ^OWNERS
-      - ^.github
     extra-vars:
       crc_ci_bootstrap_cloud_name: "{{ nodepool.cloud | replace('-nodepool-tripleo','') }}"
       crc_ci_bootstrap_networking:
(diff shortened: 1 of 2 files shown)


zuul.d/podified_multinode.yaml
```diff
@@ -15,15 +15,6 @@
...
       - ci/playbooks/edpm/run.yml
-    irrelevant-files:
-      - ^roles/.*_build
-      - ^roles/build.*
-      - ^roles/local_env_vm
-      - ^ci/templates
-      - ^docs
-      - ^.*/*.md
-      - ^OWNERS
-      - ^.github
```
marios, line 26: add to this list test/functional don't remove it

marios (changes requested):

zuul.d/podified_multinode.yaml
raukadah: @marios Thank you for the review, This job cifmw-base-multinode-podified is parented from cifmw-podified-multinode-edpm-base-crc. cifmw-podified-multinode-edpm-base-crc job uses the[ `irrelevant-files: *ir_files`](https://github.com/openstack-k8s-operators/ci-framework/blob/main/zuul.d/base.yaml#L121) defined in [content provider job irrelevant file ](https://github.com/openstack-k8s-operators/ci-framework/blob/main/zuul.d/base.yaml#L30) .

The above irrelevant file list defined under cifmw-base-multinode-podified is already included in content provider job. That's why we are removing it.

marios: ack as discussed in slack we were missing the noop. indeed this irrelevant files match is working as intended (via inheritance) and the job freeze thing went away.

marios (approved):

PR #1566, opened 2024-04-25
Add edpm template for uni04delta

jirimacku:
Just a link to the hci template


psathyan:
ok-to-test

cjeanner:
I think that one would benefit of a rebase + test against the new tree; especially, the `hci` directory has been removed, and the related template is now located in the "common" directory.
Also, with the split dataplane, it may require some tiny updates in the architecture repository.

PR #1568, opened 2024-04-25
Use official way of installing cmtcl

rabi:

### DIFF
--- a/roles/cert_manager/tasks/validate_certs.yml
+++ b/roles/cert_manager/tasks/validate_certs.yml
@@ -20,12 +20,18 @@
     state: directory
     mode: '0755'
 
-- name: Install cert-manager cmctl CLI
-  ansible.builtin.unarchive:
-    src:
-      "https://github.com/cert-manager/cert-manager/releases/{{ cifmw_cert_manager_version }}/download/cmctl-linux-amd64.tar.gz"
-    dest: "{{ lookup('env', 'HOME') }}/bin/"
-    remote_src: true
+# jinja[invalid] is triggered by the call to the lookup_plugin.pipe
+# against go binary while it doesn't existin in the env running
+# ansible-lint
+- name: Install cert-manager cmctl CLI  # noqa: jinja[invalid]
+  become: true
+  vars:
+    _os: "{{ lookup('pipe', 'go env GOOS') }}"
+    _arch: "{{ lookup('pipe', 'go env GOARCH') }}"
+  ansible.builtin.get_url:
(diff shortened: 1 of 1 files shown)


roles/cert_manager/tasks/validate_certs.yml
```diff
@@ -21,11 +21,10 @@
     mode: '0755'
 
 - name: Install cert-manager cmctl CLI
-  ansible.builtin.unarchive:
-    src:
-      "https://github.com/cert-manager/cert-manager/releases/{{ cifmw_cert_manager_version }}/download/cmctl-linux-amd64.tar.gz"
-    dest: "{{ lookup('env', 'HOME') }}/bin/"
-    remote_src: true
+  ansible.builtin.shell: |
```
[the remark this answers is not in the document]
rabi: that sounds better thanks. Let's see if it works.

roles/cert_manager/tasks/validate_certs.yml
```diff
@@ -21,11 +21,14 @@
...
 
 - name: Install cert-manager cmctl CLI
-  ansible.builtin.unarchive:
-    src:
-      "https://github.com/cert-manager/cert-manager/releases/{{ cifmw_cert_manager_version }}/download/cmctl-linux-amd64.tar.gz"
-    dest: "{{ lookup('env', 'HOME') }}/bin/"
-    remote_src: true
+  become: true
+  vars:
+    _os: "{{ lookup('pipe', 'go env GOOS') }}"
```
cjeanner, line 26: humpf, we need to mock that lookup I think - the pod doesn't have `go` - hence the lookup is failing. At least that's my understanding of the issue. it means editing `.ansible-lint` to extend the mocked modules (?) by adding
`lookup_plugin.pipe`...

rlandy (commented):
approving

roles/cert_manager/tasks/validate_certs.yml
```diff
@@ -20,12 +20,18 @@
     state: directory
     mode: '0755'
 
-- name: Install cert-manager cmctl CLI
-  ansible.builtin.unarchive:
-    src:
-      "https://github.com/cert-manager/cert-manager/releases/{{ cifmw_cert_manager_version }}/download/cmctl-linux-amd64.tar.gz"
-    dest: "{{ lookup('env', 'HOME') }}/bin/"
-    remote_src: true
+# jinja[invalid] is triggered by the call to the lookup_plugin.pipe
```
rabi, line 23: Thanks! I guess it would have been better to use shell module like I had proposed initially.

PR #1570, opened 2024-04-25
Enable proper logging

cjeanner:
Since we're using zuul pod, we're losing some logs, masked by the
community.general.make module.

Adding the capacity to `tee` in a log file should ensure we keep track
of the potential issues.

This patch modifies all of the test targets in the Makefile to use the
new LOG_DIR parameter.

### DIFF
--- a/ci/playbooks/pre-commit.yml
+++ b/ci/playbooks/pre-commit.yml
@@ -10,6 +10,12 @@
           - python3
           - python3-pip
 
+    - name: Ensure zuul-output exists
+      ansible.builtin.file:
+        path: "{{ ansible_user_dir }}/zuul-output/logs"
+        state: directory
+        mode: "0755"
+
     - name: Run make targets
       vars:
         src_dir: >-
@@ -34,3 +40,12 @@
             target: pre_commit_nodeps
             params:
               USE_VENV: "no"
+              LOG_DIR: "{{ ansible_user_dir }}/zuul-output/logs"
+      always:
+        - name: Expose pre-commit log as artifact
+          zuul_return:
+            data:
+              zuul:
+                artifacts:
+                  - name: pre-commit logfile
+                    url: pre_commit.log
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,8 @@ CI_CTX_NAME ?= localhost/cifmw:latest
 MOLECULE_CONFIG ?= ${MOLECULE_CONFIG:-.config/molecule/config_podman.yml}
 # Run molecule against all tests
 TEST_ALL_ROLES ?= ${TEST_ALL_ROLES:-no}
+# Log output location
+LOG_DIR ?= /tmp
 
 # target vars for generic operator install info 1: target name , 2: operator name
 define vars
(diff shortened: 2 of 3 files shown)


cjeanner:
It works, provided we get this patch in:
https://review.rdoproject.org/r/c/config/+/53035

Here's the nice log for pre-commit:
https://logserver.rdoproject.org/70/1570/fbc95f6ac681613b394a35cd61605b86b7e94ca4/github-check/cifmw-pod-pre-commit/43fb3e7/pre_commit.log

cjeanner:
recheck

lewisdenny (changes requested):
Thanks Cédric, this is a massive improvement, much faster and even easier to read than the Prow pre-commit job.

Two points:

1. I tested this on a PR with lint errors(just to test your PR, not because I suck ;) ) and I can see the output could be improved by passing the `--nocolor` argument to ansible-lint. [1]
2. In the job output.log we should add a link to the log. I might push a patch for this today if I get time. 

[1] https://logserver.rdoproject.org/99/1499/90292a48b7e190190ed5677c3dfd82d16b19897f/github-check/cifmw-pod-pre-commit/f4e75ec/pre_commit.log

lewisdenny:
> Thanks Cédric, this is a massive improvement, much faster and even easier to read than the Prow pre-commit job.
> 
> Two points:
> 
>     1. I tested this on a PR with lint errors(just to test your PR, not because I suck ;) ) and I can see the output could be improved by passing the `--nocolor` argument to ansible-lint. [1]
> 
>     2. In the job output.log we should add a link to the log. I might push a patch for this today if I get time.
> 
> 
> [1] https://logserver.rdoproject.org/99/1499/90292a48b7e190190ed5677c3dfd82d16b19897f/github-check/cifmw-pod-pre-commit/f4e75ec/pre_commit.log

My idea for point 2 here: https://github.com/openstack-k8s-operators/ci-framework/pull/1579

cjeanner:
> 1. I tested this on a PR with lint errors(just to test your PR, not because I suck ;) ) and I can see the output could be improved by passing the `--nocolor` argument to ansible-lint. [1]
> 2. In the job output.log we should add a link to the log. I might push a patch for this today if I get time.
> [1] https://logserver.rdoproject.org/99/1499/90292a48b7e190190ed5677c3dfd82d16b19897f/github-check/cifmw-pod-pre-commit/f4e75ec/pre_commit.log
> 
> My idea for point 2 here: #1579

(a pity we can't get threads in here like gitlab...)

I saw #1579 - but I think we can leverage Zuul and its artifacts, allowing to get that data directly in the main/frontpage, and direct log access via the "artifacts" tab. Though I don't remember how to do this, there's a Zuul thing, we'll have to ask some other Zuul-savvy ppl.

For the --no-color: yes, but.... having colors when running locally is good imho. I thought `tee` had some option for that, but apparently it doesn't :/. Lemme find a way to keep the color in local shell but pipe cleaned content.

ci/playbooks/pre-commit.yml
```diff
@@ -34,3 +40,12 @@
...
               USE_VENV: "no"
+              LOG_DIR: "{{ ansible_user_dir }}/zuul-output/logs"
+      always:
+        - name: Expose pre-commit log as artifact
+          zuul_return:
+            data:
+              zuul:
+                artifacts:
+                  - name: pre-commit logfile
+                    url: pre_commit.log
```
marios, line 51: will you expand that to add the others? check_zuul_files.log etc

marios (commented):
seems reasonable - in two minds about the new requirement though. Nothing against ansi2txt more like just being careful about adding requirements especially if it is not functional but I think it is a matter of debate how useful the colorized logs are; maybe even invaluable to folks with problems seeing particular colors? but in those cases i also suppose they would already have means to convert their text files to the appropriate colors. 

anyway,  +1 just waiting to see if you will expand the artifacts?

ci/playbooks/pre-commit.yml
cjeanner: once we get more of those tests, yes. Have to consider how to do it "right" to avoid code duplication though. Maybe same playbook, with some "when" and appropriate parameter passed down the zuul job? Not sure yet, so I'd just go with that pre-commit alone for now.

lewisdenny (approved):
Thanks Cedric, that's great!

ci/playbooks/pre-commit.yml
cjeanner: FYI, here's what I'm thinking for adding the other jobs: #1584 
We'll need another one to start getting job dependency via #1573 - but things are coming nicely.

PR #1573, opened 2024-04-26
Prepare for job dependency

cjeanner:
In order to save resources and time, we want to make jobs dependent.
Since a patch won't merge if molecule (or pre-commit, or ansible-test)
don't pass, let's not run the full deploy if those smaller, faster tests
fail.

To get that dependency, we leverage the `provides` parameter for zuul
job definition. The next step is to modify the "big jobs" to `requires`
the listed job(s).

For now, we can depend on molecule and pre-commit
- ansible-test is still in Prow

### DIFF
--- a/ci/playbooks/noop.yml
+++ b/ci/playbooks/noop.yml
@@ -1,0 +1,8 @@
+---
+- name: Noop playbook
+  hosts: all
+  gather_facts: false
+  tasks:
+    - name: Hello world
+      ansible.builtin.debug:
+        msg: "Hello world"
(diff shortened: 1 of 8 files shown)


cjeanner:
> Build failed (check pipeline). Post `recheck` (without leading slash) to rerun all jobs. Make sure the failure cause has been resolved before you rerun jobs.
> 
> https://review.rdoproject.org/zuul/buildset/76e4eae37117483cb7233f1eda55c38a
> 
only valid failure is cifmw-molecule_pkg_build - correction will be shortly in, I'll rebase that PR.

cjeanner:

both failures aren't related to the change; all jobs were green at least once in the life-cycle of this patch.

lewisdenny (approved):
/lgtm

PR #1579, opened 2024-04-29
 Add helpful link to pre-commit logs 

lewisdenny:


cjeanner:
There's a better way, embedded in Zuul itself, with the "artifacts". Though I don't remember how to leverage this.

PR #1583, opened 2024-04-29
Integrate update_containers in validated architecture role

raukadah:
update_containers role is used to update container images
using openstack version interface. Let's use the same to
update the containers in va.

It will be used in downstream VA jobs to use RHEL openstack
services container. 
  - [x] README in the role


rebtoor (approved):

cjeanner:
/approve

makes sense, will improve QE work.

PR #1584, opened 2024-04-29
pod/extend pod jobs

cjeanner:
In order to get out of the Prow CI, we have to port some other jobs to
Zuul pods.

In order to avoid code ducplication, some changes were needed in the
`pre-commit.yml` playbook - for instance, it got renamed with a more
generic name, and it's now getting the `make` target directly from the
job definition.

In addition, two checks that were in pre-commit are now standalone:
- check_zuul_files
- check_k8s_snippets_comment

This ensures we get consistent logging and related artifacts in Zuul, as
well as ensuring we trigger them only when needed (those are small jobs,
but still - let's save resources)

### DIFF
--- a/zuul.d/pods.yaml
+++ b/zuul.d/pods.yaml
@@ -1,10 +1,40 @@
 ---
 - job:
-    name: cifmw-pod-pre-commit
+    name: cifmw-pod-base
     nodeset:
       nodes:
         - name: container
           label: pod-centos-9-stream
     description: |
-      Run pre-commit against ci-framework
-    run: ci/playbooks/pre-commit.yml
+      Run lightweight jobs in pods
+    run: ci/playbooks/pod-jobs.yml
+
+- job:
+    name: cifmw-pod-pre-commit
+    parent: cifmw-pod-base
+    vars:
+      run_test: "pre_commit_nodeps"
+
+- job:
+    name: cifmw-pod-ansible-test
+    parent: cifmw-pod-base
+    vars:
+      run_test: "ansible_test_nodeps"
+
+- job:
+    name: cifmw-pod-zuul-files
+    parent: cifmw-pod-base
+    vars:
+      run_test: "check_zuul_files"
+    files:
+      - ^zuul.d/.*
+      - ^ci/templates/.*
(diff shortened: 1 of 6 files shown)


cjeanner:
ansible-test is failing (hence the `voting: false`) in zuul env while it's working in Prow, and locally. I'll need to debug this in order to switch that job to "voting". Seems a checksum isn't matching :(.

lewisdenny:
recheck

lewisdenny:
> ansible-test is failing (hence the `voting: false`) in zuul env while it's working in Prow, and locally. I'll need to debug this in order to switch that job to "voting". Seems a checksum isn't matching :(.

yeah something in `/tmp/logs/ci_script_012_run_using_chdir.log` is different, size of the file too. I guess something different in the environments but I'll run a recheck and check that file live for clues.

lewisdenny:
If we are getting rid of the prow job and the file looks fine we should just update the checksum

lewisdenny (commented):
This is looking pretty good to me once the hash issue is sorted.

cjeanner:
Thank you @lewisdenny for your check! I could make tests converge, so we're green on both end!

zuul.d/pods.yaml
```diff
@@ -1,10 +1,40 @@
...
+      - ^ci/templates/.*
+      - ^ci/config/.*
+
+- job:
+    name: cifmw-pod-k8s-snippets-source
+    parent: cifmw-pod-base
+    vars:
+      run_test: "check_k8s_snippets_comment"
+    files:
+      - ^roles/ci_gen_kustomize_values/templates/.*
```
pablintino, line 40: Why did it run in this PR? I wasn't expecting this job to report in this PR.

cjeanner: hmm good question. Not that it takes too long, but.. indeed, it shouldn't run.

I can investigate this in a follow-up. The job itself is a matter of 5-6 minutes max, not taking too much time... But that whole patch is blocking further job on saving resources with actual nodeset :(.

lewisdenny (approved):
Thanks Cédric, this is a nice improvement.

PR #1591, opened 2024-04-29
Fix VM base image machine-id

hjensas:
Issue, **all** VMs using the same base image had identintical machine-id in /etc/machine-id. Root cause is that virt-sysprep both removes and sets the machine-id. Logs show:

```
[   4.3] Performing "machine-id" ...
< -- SNIP -- >
[   4.6] Setting the machine ID in /etc/machine-id

```
To fix the issue, run virt-sysprep again and specify that it should only run `--operation machine-id`.

  - [x] README in the role

### DIFF
--- a/roles/libvirt_manager/tasks/deploy_layout.yml
+++ b/roles/libvirt_manager/tasks/deploy_layout.yml
@@ -178,6 +178,7 @@
       {% endif %}
       --root-password "password:{{ item.value.password | default('fooBar') }}"
       --ssh-inject {{ _admin_user }}:file:{{ ansible_user_dir }}/.ssh/authorized_keys
+      --truncate /etc/machine-id
       | tee -a {{ item.value.image_local_dir }}/{{ item.value.disk_file_name }}.log
     creates: "{{ item.value.image_local_dir }}/{{ item.value.disk_file_name }}.log"
   loop: "{{ _layout.vms | dict2items }}"


lewisdenny (changes requested):
This is a really annoying, weird behavior of `virt-sysprep`? Checking the man page I see there might be an easier way: 
```
 $ man virt-sysprep | grep -A1 'truncate FILE'
       --truncate FILE (see "customize" below)
           This command truncates "FILE" to a zero-length file. The file must exist already.
```

So just updating the task above your new one with this one line marked with "HERE" (Apologies if you've already considered this):
```
- name: Prepare disk image with SSH and growing volume  # noqa: risky-shell-pipe
  when:
    - item.value.disk_file_name != 'blank'
    - item.key is not match('^(crc|ocp).*$')
  vars:
    _admin_user: "{{ item.value.admin_user | default('root') }}"
    _root_part: "{{ item.value.root_part_id | default('1') }}"
    _static_ip_script: >-
      {{
        (
          cifmw_libvirt_manager_ip_script_basedir,
          'static-ip-' + item.key + '.sh'
        ) | ansible.builtin.path_join
      }}
  ansible.builtin.shell:
    cmd: >-
      set -o pipefail;
      virt-sysprep -a "{{ item.value.image_local_dir }}/{{ item.value.disk_file_name }}"
      --selinux-relabel
      --firstboot-command "growpart /dev/sda {{ _root_part }}"
      --firstboot-command "xfs_growfs /"
      {% if item.value.ip_address is defined %}
      --firstboot {{ _static_ip_script }}
      {% endif %}
      --root-password "password:{{ item.value.password | default('fooBar') }}"
      --ssh-inject {{ _admin_user }}:file:{{ ansible_user_dir }}/.ssh/authorized_keys
      --truncate /etc/machine-id <-- HERE
      | tee -a {{ item.value.image_local_dir }}/{{ item.value.disk_file_name }}.log
    creates: "{{ item.value.image_local_dir }}/{{ item.value.disk_file_name }}.log"
  loop: "{{ _layout.vms | dict2items }}"
  loop_control:
    label: "{{ item.key }}"
```

lewisdenny:
I tested this locally and it seems to work:
```
$ grep Truncating ci-framework-data/images/centos-stream-9.qcow2.log
[   9.4] Truncating: /etc/machine-id

$ sudo  guestmount -a ci-framework-data/images/centos-stream-9.qcow2 -i --ro /qcow_mount  # safe, read only
$ sudo ls -l /qcow_mount/etc/machine-id
-r--r--r--. 1 root root 0 Apr 30 08:24 /qcow_mount/etc/machine-id
```

hjensas:
@lewisdenny, yes I think this behavior is weird as well ...

Using --truncate works as well, I have updated the patch to do that.

cjeanner:
/approve

Duh...  that machine-id thing hitting back again.

Thanks @lewisdenny for the `--truncate` - I was about to propose something like that. Thanks @hjensas for the update - imho we're good now.

lewisdenny (approved):

lewisdenny:
> @lewisdenny, yes I think this behavior is weird as well ...
> 
> Using --truncate works as well, I have updated the patch to do that.

Thanks mate, I tested your method worked as well :) lgtm +1

PR #1592, opened 2024-04-30
Use bootstrap_cmd for downstream repo-setup

rabi:

### DIFF
--- a/roles/edpm_deploy_baremetal/tasks/main.yml
+++ b/roles/edpm_deploy_baremetal/tasks/main.yml
@@ -101,6 +101,7 @@
   when:
     - not cifmw_edpm_deploy_baremetal_dry_run | bool
   vars:
+    osp_trunk_repo: "/etc/yum.repos.d/osptrunk-candidate-deps.repo"
     cifmw_edpm_deploy_openstack_crs_path: >-
       {{
         [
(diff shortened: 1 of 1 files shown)


roles/edpm_deploy_baremetal/tasks/main.yml
```diff
@@ -131,10 +132,25 @@
                   path: /spec/nodeTemplate/ansible/ansibleVars/edpm_container_registry_insecure_registries
                   value: ["{{ content_provider_registry_ip }}:5001"]
           {% endif %}
-
+          {% cifmw_edpm_deploy_baremetal_repo_setup_override | bool %}
+                - op: add
+                  path: /spec/nodeTemplate/ansible/ansibleVars/edpm_bootstrap_command
+                  value: |
+                    curl -O http://download.devel.redhat.com/rcm-guest/puddles/OpenStack/rhos-release/rhos-release-latest.noarch.rpm
```
cjeanner, line 139: this is internal URI - we shouldn't expose it.

roles/edpm_deploy_baremetal/tasks/main.yml
```diff
@@ -131,10 +132,25 @@
...
+                    dnf -y localinstall rhos-release-latest.noarch.rpm
+                    rhos-release rhel-9.2
+                    curl -o /etc/yum.repos.d/delorean.repo https://osp-trunk.hosted.upshift.rdu2.redhat.com/rhel9-osp18/current-podified/delorean.repo
```
cjeanner, line 142: ditto - internal URI, please don't expose it

roles/edpm_deploy_baremetal/tasks/main.yml
```diff
@@ -131,10 +132,25 @@
...
+                    echo "[osptrunk-candidate-deps]" >> "{{ osp_trunk_repo }}"
+                    echo "name=osptrunk-candidate-deps" >> "{{ osp_trunk_repo }}"
+                    echo "baseurl=http://download.eng.bos.redhat.com/brewroot/repos/rhos-18.0-rhel-9-trunk-candidate/latest/x86_64/" >> "{{ osp_trunk_repo }}"
```
cjeanner, line 145: ditto

cjeanner (changes requested):
You shouldn't expose internal resource URI....

roles/edpm_deploy_baremetal/tasks/main.yml
cjeanner: This project doesn't expose internal things - and getting a downstream-only resource file is usual in here. You may think "it's just uri", right. but it provides far more data than you may think about internal layout. Using parameters and providing them via a config file is easy enough at this point.

PR #1597, opened 2024-04-30
Cleanup commented out defaults

lewisdenny:


psathyan (approved):

raukadah (commented):
@lewisdenny Thank you for the pr. These vars are used in downstream. We have kept these vars commented out so that end user should know these vars,  exists in the role and they can use it whenever they need it. These vars are already mentioned in README. But I would keep it in the role defaults for awareness.

raukadah (approved):
/approve

PR #1598, opened 2024-04-30
Remove cacheable: true from ci_data.yml

lewisdenny:
Follow up from https://github.com/openstack-k8s-operators/ci-framework/pull/1499#discussion_r1584333119


cjeanner:
/approve

the less we use cached facts, the better imho. Thanks for that one, @lewisdenny

psathyan (approved):

PR #1600, opened 2024-04-30
Move ansi2txt from molecule requirements to common

lewisdenny:
`ansi2txt` is required as of #1570 however the requirement was added to `test-requirements.txt` which only gets called for Molecule jobs. This patch adds the requirement to `common-requirements.txt` which is called in both scenarios.

This has broken `make run_ctx_pre_commit` for example.

### DIFF
--- a/common-requirements.txt
+++ b/common-requirements.txt
@@ -4,6 +4,7 @@ kubernetes
 kubernetes-validate
 openstacksdk
 jsonschema>=4.20.0
+ansi2txt
 
 # Allows to unpin cryptography
 pyOpenSSL>=22.1.0
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -22,6 +22,3 @@ pre-commit # MIT
 yamllint
 pyspelling
 mkdocs-pymdownx-material-extras
-
-# Common requirements
-ansi2txt


cjeanner:
hmm I just had to rebuild the container to get around that issue you faced: the container build process is done once, and is actually getting the molecule env ready, since it's the most complete.

I'm therefore not sure this is the appropriate solution?

lewisdenny:
> hmm I just had to rebuild the container to get around that issue you faced: the container build process is done once, and is actually getting the molecule env ready, since it's the most complete.
> 
> I'm therefore not sure this is the appropriate solution?

Ah I missed the second build call running `containerfiles/Containerfile.tests`

So in the case of `make run_ctx_pre_commit` rebuilding the container first would have fixed it but `make pre_commit` is broken by this change then. 

I still think `common-requirements.txt` is a better home for this because as your comment in `test-requirements.txt` suggests, this is a common requirement and `common-requirements.txt` is called for both `Molecule` and `pre_commit` tests.

marios (approved):

lewisdenny:
recheck

cjeanner:
`ansi2txt` is really for testing - so I think it's more about correcting the `setup_tests` target (and associated script)...?
The `Common requirements` was clumsy. It's more "generic requirements for tests" ? I.e not only for one kind of test.

lewisdenny:
> `ansi2txt` is really for testing - so I think it's more about correcting the `setup_tests` target (and associated script)...? The `Common requirements` was clumsy. It's more "generic requirements for tests" ? I.e not only for one kind of test.

Yes that's correct, when running `make pre_commit` it will fail as `ansi2txt` isn't installed. That's what I'm fixing in this patch. 

However this isn't an issue if you use the ctx make targets and from the docs we only expect people to use the ctx make targets [1] so maybe we should be clear in the Makefile what make targets are for end users?

[1] https://ci-framework.readthedocs.io/en/latest/development/01_contributing.html#testing

cjeanner:
/hold

ansi2txt is for testing, so it really should stay in the test-requirements.txt. IMHO we should make user consume the container (so doc update), and/or ensure test targets are now using the test-requirements.txt for their dependency install.

lewisdenny:
> /hold
> 
> 
> 
> ansi2txt is for testing, so it really should stay in the test-requirements.txt. IMHO we should make user consume the container (so doc update), and/or ensure test targets are now using the test-requirements.txt for their dependency install.

Cool, I think we have an understanding now. I'll close this patch and open another that makes the user path more clear for local testing.

PR #1604, opened 2024-04-30
Split dataplane deployment in edpm_deploy role

viroel:
OpenStackDataPlaneDeployment and OpenStackDataPlaneNodeSet need to be deployed in different phases, since one depends on the other now. This PR tries to split the deployment in two steps, ignoring a specific kind of resource.

  - [x] README in the role

### DIFF
--- a/roles/edpm_deploy/defaults/main.yml
+++ b/roles/edpm_deploy/defaults/main.yml
@@ -30,3 +30,5 @@ cifmw_edpm_deploy_timeout: 60
 cifmw_edpm_deploy_registry_url: "{{ cifmw_install_yamls_defaults['DATAPLANE_REGISTRY_URL'] }}"
 cifmw_edpm_deploy_prepare_run: true
 cifmw_edpm_deploy_nova_compute_extra_config: ""
+# Defines the kind of resources that should be applied only in step 2 of the EDPM deployment
+cifmw_edpm_deploy_step2_kind: "OpenStackDataPlaneDeployment"
--- a/roles/edpm_deploy/tasks/main.yml
+++ b/roles/edpm_deploy/tasks/main.yml
@@ -136,13 +136,15 @@
       ansible.builtin.debug:
         var: cifmw_edpm_deploy_crs_kustomize_result
 
-    - name: Apply the OpenStackDataPlaneNodeSet CR
-      when: not cifmw_edpm_deploy_dryrun
-      cifmw.general.ci_script:
-        output_dir: "{{ cifmw_edpm_deploy_basedir }}/artifacts"
-        script: "oc apply -f {{ cifmw_edpm_deploy_crs_kustomize_result.output_path }}"
+    - name: Apply dataplane resources but ignore DataPlaneDeployment
+      kubernetes.core.k8s:
+        kubeconfig: "{{ cifmw_openshift_kubeconfig  }}"
+        api_key: "{{ cifmw_openshift_token | default(omit) }}"
+        context: "{{ cifmw_openshift_context | default(omit) }}"
+        state: present
+        definition: "{{ lookup('file', cifmw_edpm_deploy_crs_kustomize_result.output_path) | from_yaml_all | rejectattr('kind', 'search', cifmw_edpm_deploy_step2_kind) }}"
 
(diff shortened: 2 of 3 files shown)


roles/edpm_deploy/defaults/main.yml
```diff
@@ -30,3 +30,5 @@ cifmw_edpm_deploy_timeout: 60
 cifmw_edpm_deploy_registry_url: "{{ cifmw_install_yamls_defaults['DATAPLANE_REGISTRY_URL'] }}"
 cifmw_edpm_deploy_prepare_run: true
 cifmw_edpm_deploy_nova_compute_extra_config: ""
+# Defines the kind of resources that should be applied only in step 2 of the EDPM deployment
+cifmw_edpm_deploy_step2_kind: "OpenStackDataPlaneDeployment"
```
viroel, line 34: need to update README afterwards

roles/edpm_deploy/tasks/main.yml
```diff
@@ -136,34 +136,35 @@
...
+      vars:
+        _timeout_min_string: "{{cifmw_edpm_deploy_timeout + 'm'}}"
+      kubernetes.core.k8s:
+        kubeconfig: "{{ cifmw_openshift_kubeconfig  }}"
+        api_key: "{{ cifmw_openshift_token | default(omit) }}"
+        context: "{{ cifmw_openshift_context | default(omit)  }}"
+        state: present
+        definition: "{{ lookup('file', cifmw_edpm_deploy_crs_kustomize_result.output_path) | from_yaml_all | rejectattr('kind', 'search', cifmw_edpm_deploy_step2_kind }}"
+        wait: true
+        wait_condition:
```
viroel, line 149: not sure if is goint to work, since we are creating multiple resources here, not all have this condition type.
the other approach is to use another task with k8s_info + wait_condition.

viroel:
recheck

roles/edpm_deploy/tasks/main.yml
```diff
@@ -136,13 +136,15 @@
       ansible.builtin.debug:
         var: cifmw_edpm_deploy_crs_kustomize_result
 
-    - name: Apply the OpenStackDataPlaneNodeSet CR
-      when: not cifmw_edpm_deploy_dryrun
-      cifmw.general.ci_script:
-        output_dir: "{{ cifmw_edpm_deploy_basedir }}/artifacts"
-        script: "oc apply -f {{ cifmw_edpm_deploy_crs_kustomize_result.output_path }}"
+    - name: Apply dataplane resources but ignore DataPlaneDeployment
```
fao89, line 139:
```suggestion
    - name: Apply dataplane resources but ignore "{{ cifmw_edpm_deploy_step2_kind }}"
```

fao89 (approved):
lgtm

roles/edpm_deploy/tasks/main.yml
viroel: I will change in a follow up PR, adding also baremetal, thanks

roles/edpm_deploy/tasks/main.yml
```diff
@@ -162,7 +164,38 @@
...
+    - name: Apply DataPlaneDeployment resource
+      kubernetes.core.k8s:
+        kubeconfig: "{{ cifmw_openshift_kubeconfig  }}"
+        api_key: "{{ cifmw_openshift_token | default(omit) }}"
+        context: "{{ cifmw_openshift_context | default(omit) }}"
+        state: present
+        definition: "{{ lookup('file', cifmw_edpm_deploy_crs_kustomize_result.output_path) | from_yaml_all | selectattr('kind', 'search', cifmw_edpm_deploy_step2_kind) }}"
+
+    - name: Wait for OpenStackDataPlaneDeployment become Ready
+      when: not cifmw_edpm_deploy_dryrun
```
pablintino, line 179: The entire block is conditioned to the dry-run flag, this, and all the occurrences of the flag inside the block, is not needed.

viroel: True, I can improve that in a next PR, which should include this change for baremetal deploy too.

PR #1605, opened 2024-04-30
[NFV] Remove hardcoded zuul edpm user

pablintino:
In NFV deployments, where BM nodes are used, we do not handle the deployed image, so we cannot warranty the user to be used. To avoid handling what we don't know, let the VA set the value and let the user modify it if he/she changes the deployed image.

  - [x] README in the role


lewisdenny (approved):

fultonj:
+2

PR #1606, opened 2024-05-01
Add repo-setup to the edpm services if predeployed

pablintino:
  - [x] README in the role


roles/ci_gen_kustomize_values/templates/common/edpm-nodeset-values/values.yaml.j2
```diff
@@ -51,6 +51,15 @@ data:
 {%        endif                                                                        %}
 {%      endfor                                                                         %}
 {% endfor                                                                              %}
+{% if ('repo-setup' not in (_original_nodeset['services'] | default([]))) and
+      ('repo-setup' in ci_gen_kustomize_edpm_nodeset_predeployed_services)             %}
+    services:
+      - "repo-setup"
+{%   for svc in ci_gen_kustomize_edpm_nodeset_predeployed_services                     %}
+      - "{{ svc }}"
+{%   endfor                                                                            %}
```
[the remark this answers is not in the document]
pablintino: I used the wrong variable in the loop.
The idea is, check, before deploying if the repo-setup is already deployed (the user/job may have deployed it, as it's not part of the VA), and, if it's deployed add it to the services list in the first position.

roles/ci_gen_kustomize_values/tasks/edpm_core_facts.yml
```diff
@@ -0,0 +1,37 @@
...
+# License for the specific language governing permissions and limitations
+# under the License.
+
+
+- name: Fetch the already deployed services for further usage
+  when: ci_gen_kustomize_fetch_ocp_state
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  ansible.builtin.command:
```
cjeanner, line 23: can't we use `kubernetes.core.k8sinfo` instead?

pablintino: Short answer: no
Long answer: yes, but, it's a mess, why? API Versions. The dumb `kubernetes.core.k8sinfo` is not able to properly guess custom CRs versions, so we need to explicitly pass them, meaning hard-coding it. I don't want to hard-code that info in the framework, cause a change in the API Version of the CR is going to need a CI-framework change.

lewisdenny (approved):
/approve

PR #1609, opened 2024-05-01
Replace use of domifaddr with ip neighbour - v2

hjensas:
virsh domifaddr does not work with IPv6, replace the usage and instead use the ip neighbour command to get VM ip's.

This change s roles/libvirt_manager/tasks/start_manage_vms.yml only, there is also use of domifaddr in
roles/libvirt_manager/tasks/deploy_edpm_compute.yml.

Let's look at that in a follow up.

  - [ ] README in the role


hjensas:
I put this in draft because I want to add a molecule test.

hjensas:
Closing this - discovering the address of a node is fragile / unstable.
For one, it depends on the node emitting traffic on boot - an other approach is in development.

PR #1614, opened 2024-05-02
Expose CRC kubeadmin-password in controller-0

cjeanner:
As a pull request owner and reviewers, I checked that:
- [X] Appropriate testing is done and actually running

PR #1615, opened 2024-05-02
Correct import vs include

cjeanner:
The `import_role` in 06-deploy-edpm.yml play was leading to an uncaught
issue in architecture driven deployment: with `import`, the content of
the role was directly in the playbook and, while we exit it early via
the ansible.builtin.meta, it was still "too late" for some undefined
variable hit (install_yamls related).

The `include_role` in 06-deploy-architecture.yml play was leading to an
error when we were trying to iterate on `edpm_deploy` tag: the `always`
tag wasn't applied to the subsequent list of tasks, leading to an error
with undefined variable. Switchint to `import` allows to get all the
tasks tagged with `always`.

### DIFF
--- a/playbooks/06-deploy-architecture.yml
+++ b/playbooks/06-deploy-architecture.yml
@@ -20,7 +20,7 @@
     - name: Load Networking Environment Definition
       tags:
         - always
-      ansible.builtin.include_role:
+      ansible.builtin.import_role:
         name: networking_mapper
         tasks_from: load_env_definition.yml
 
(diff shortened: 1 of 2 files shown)


cjeanner:
recheck

cjeanner:
/hold

it also has #1614 - I'll rebase once that one is in.

son-vyas:
@cjeanner is this MR ready now?

cjeanner:
@son-vyas yep, rebased and hold removed :)

cjeanner:
recheck

PR #1616, opened 2024-05-03
Use ad-hoc reboot play

bshephar:
This change implements the supported user interface for initiating reboots. We were using the reboot-os service, but it's expected that this wouldn't work consistently for day 2 operations. Nothing in the edpm_kernel role creates a reboot_required file and therefore the default reboot-os service could potentially skip rebooting for changes to kernel args. This instead implements a custom reboot service that will ensure the reboot is executed against the defined set of nodes.

  - [x] README in the role


bshephar:
recheck

rebtoor (approved):

PR #1618, opened 2024-05-03
Don't trigger deployment on specific files

cjeanner:
There's no need to trigger a deployment when we modify ansible-lint,
pre-commit and spellcheck configurations.


son-vyas:
/lgtm
certainly we dont need deployment  on these files - .ansible-lint, .pre-commit-config.yaml,  .spellcheck.yml

PR #1620, opened 2024-05-03
[DNM] Set bmo_setup false

son-vyas:
  - [x] README in the role


lewisdenny (changes requested):
I'm confused.
- The title has DNM but it's marked ready for review with lgtm?
- The commit message is setting bmo_setup false but the content of the patch is setting it to true?

PR #1623, opened 2024-05-03
[reproducer] Fix undefined devscripts_kubeadm

psathyan:
Missing `_` hence the undefined error. This change fixes the issue

### DIFF
--- a/roles/reproducer/tasks/configure_controller.yml
+++ b/roles/reproducer/tasks/configure_controller.yml
@@ -272,13 +272,13 @@
     - name: Inject kubeadmin-password if exists
       when:
         - _devscripts_kubeadm.content is defined or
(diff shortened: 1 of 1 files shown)


roles/reproducer/tasks/configure_controller.yml
```diff
@@ -278,7 +278,7 @@
         content: >-
```
karelyatin, line 278: need to fix also _crc_kubeconfig -> _crc_kubeadm

cjeanner:
/approve
/lgtm

I'll merge it once molecule is OK for the reproducer role - that's the only place where that path is "tested" - and for CRC only...

karelyatin (commented):
/lgtm

PR #1624, opened 2024-05-06
Drop cifmw_repo_setup_component var

raukadah:
Remove cifmw_repo_setup_component var in favor of
cifmw_repo_setup_component_name var. The correct var for specifying the component name is cifmw_repo_setup_component_name.

It is causing issue while running dlrn reporting in the component line. Let's drop cifmw_repo_setup_component var and use cifmw_repo_setup_component_name in artifacts to get the proper hashes.


rlandy (commented):
/approve

PR #1625, opened 2024-05-06
[test-operator] Fix tempest workflows

lpiwowar:
Currently, the tempest_workflow var is not passed to the run-test-operator-job.yml thanks to this the role does not wait for the last test pod and finishes prematurely.

  - [x] README in the role


kopecmartin (approved):
oh, good catch

PR #1626, opened 2024-05-06
Dump dlrn md5 hash from current-podified while using component repo

raukadah:
DLRN md5 hash files are generated when we promote all components in promoted-components, podified-ci-testing and current-podified.

in repo_setup role, we generate the dlrn_md5 hash file based on full_hash. full_hash is returned by all dlrn tags.

In component repo, there is no dlrn md5 hash and repo_setup role is generating wrong dlrn_md5 hash file.

In component pipeline, tcib job as a parent job generating the wrong dlrn md5 hash file and passed it child edpm/podified job.

The child job repo-setup will receive wrong dlrn md5 hash tag and the repo-setup tool will fail.

In component pipeline, we generate current-podified repos then populate the component repos. So, we need to write dlrn_md5 hash from current-podified in the component jobs.

This pull request fixes the same.

  - [x] README in the role

### DIFF
--- a/roles/repo_setup/tasks/artifacts.yml
+++ b/roles/repo_setup/tasks/artifacts.yml
@@ -12,7 +12,7 @@
       {% else -%}
       --tag {{cifmw_repo_setup_promotion }}
       {% endif -%}
-      {% if cifmw_repo_setup_dlrn_hash_tag | length > 0 -%}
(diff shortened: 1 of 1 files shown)


cjeanner:
/approve
/lgtm

all green, I don't see obvious issues.

PR #1628, opened 2024-05-06
Allow to run a layout without any CRC

cjeanner:
When running a molecule reproducer, we may not have any CRC. While this
case was mostly covered, it wasn't properly addressed when the user
doesn't have any pre-deployed CRC.

This patch ensures we really-really want CRC before bootstrapping it.

### Manual testing
- [X] va-hci could properly deploy the layout
- [X] zuul reproducer for molecule (NO CRC) is working as expected
- [X] working even if no CRC is present on node, with or without CRC in the layout (`amount: 0` set by the zuul reproducer)


eduolivares (approved):

raukadah (commented):
/approve

PR #1629, opened 2024-05-06
Ensure we run custom, injected hooks in all cases

cjeanner:
Until now we had to get at least one hook in the automation file to run
custom, injected hooks.
This is leading to issues when a user wants to inject a hook on their
own, without any pre-existing hooks from the automation.


pablintino:
recheck

pablintino:
recheck

PR #1633, opened 2024-05-06
Add a (post-deploy) hook to set a multiattach volume type

tosky:
This is required by a few tempest tests,
see https://review.opendev.org/c/openstack/tempest/+/875372 .

This is the ci-framework port of this devstack change: https://review.opendev.org/c/openstack/devstack/+/877337

  - [x] README in the role

### DIFF
--- a/hooks/playbooks/cinder_multiattach_volume_type.yml
+++ b/hooks/playbooks/cinder_multiattach_volume_type.yml
@@ -1,0 +1,20 @@
+---
+- name: Create cinder resources needed for tempest run
+  hosts: "{{ cifmw_target_hook_host | default('localhost') }}"
+  gather_facts: false
+  tasks:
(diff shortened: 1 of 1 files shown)


tosky:
Tested internally, ready for review

SeanMooney (approved):
i have been testing this as part of https://github.com/openstack-k8s-operators/nova-operator/pull/753 and this work correctly.

this looks good to me over all the only think i will say is if your are using shell its can be good to ensure you add  
`set -euxo pipefail`

the ci framework whoever is not generally harding the shell usage that way so this is at least consistent with the current repo so this can be treated as a nit

raukadah:
/approve

Thank you @SeanMooney, May be we can do a followup to add  `set -euxo pipefail` to the all hooks wherever shell module is used.

PR #1636, opened 2024-05-07
[edpm_prepare] Ensure the ctrlplane CR has latest content

pablintino:
Initially the edpm_prepare role apply logic wasn't designed to be called more than once, but now it's possible, leading to the unexpected situation of ci_kustomize picking always the initial install_yamls CR as the input instead of the last applied content. Forcing the ci_kustomize output to the same file always will ensure we consume always the same CR instead of creating the ci_kustomize default one.

  - [x] README in the role


pablintino:
recheck

pablintino:
Not needed anymore.

PR #1637, opened 2024-05-07
Fix(?) CRC DNS

cjeanner:
In CRC, we can't really configure existing interfaces - they get
overridden by some other process, meaning we can't ensure it won't get
any public DNS.

In order to ensure we use only the local dnsmasq , the best way is
apparently to unmanage the /etc/resolv.conf from within NetworkManager,
and hardcode the content.
Even with that, the comment seems to be removed and replaced by
`# Generated by NetworkManager` - which is... weird. Since it shouldn't
do anything in there...

We use the CRC "public" IPv4 - apparently 127.0.0.1 doesn't work, at
least from the openshift-dns/dns-default pod (throws connections
refused).

In any cases, this change seems to work (for now), ensuring DNS are
properly working.

We might want to amend the DNS configuration on the controller-0 to use
the openshift-dns/dns-default service instead - but that's for a
follow-up.

### DIFF
--- a/roles/reproducer/tasks/configure_crc.yml
+++ b/roles/reproducer/tasks/configure_crc.yml
@@ -25,6 +25,24 @@
         never_default4: true
         state: present
 
+    - name: Ensure NetworkManager does not override /etc/resolv.conf
+      become: true
+      ansible.builtin.copy:
+        dest: "/etc/NetworkManager/conf.d/no-resolvconf.conf"
+        mode: "0644"
+        content: |
+          [main]
+          rc-manager = unmanaged
+
+    - name: Ensure we use local dnsmasq only
+      become: true
+      ansible.builtin.copy:
+        dest: "/etc/resolv.conf"
+        mode: "0644"
+        content: |
+          # Managed by ansible/cifmw
+          nameserver {{ _crc.ip_v4 }}
+
     - name: Check which dnsmasq config we must edit
       register: _dnsmasq
       ansible.builtin.stat:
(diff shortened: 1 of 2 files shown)


roles/reproducer/tasks/configure_crc.yml
```diff
@@ -25,6 +25,24 @@
...
         state: present
 
+    - name: Ensure NetworkManager does not override /etc/resolv.conf
+      become: true
+      ansible.builtin.copy:
+        dest: "/etc/NetworkManager/conf.d/no-resolvconf.conf"
+        mode: "0644"
+        content: |
+          [main]
+          rc-manager = unmanaged
```
marios, line 35: had to go find that https://www.networkmanager.dev/docs/api/latest/NetworkManager.conf.html 

seems legit ;) and yeah I don't think we want NetworkManager doing anything to resolv.conf!

```
Set the resolv.conf management mode. This option is about how NetworkManager writes to /etc/resolv.conf, if at all. The default value depends on NetworkManager build options, and this version of NetworkManager was build with a default of "auto". Regardless of this setting, NetworkManager will always write its version of resolv.conf to its runtime state directory as /run/NetworkManager/resolv.conf. 
...

resolvconf: NetworkManager will run resolvconf to update the DNS configuration.

netconfig: NetworkManager will run netconfig to update the DNS configuration.

unmanaged: don't touch /etc/resolv.conf.

none: deprecated alias for symlink.

```

PR #1638, opened 2024-05-07
Avoid ansibleee-operator hardcoded version

cescgina:
When preparing to test beta operator images, version is bumped to v1.0.0
instead of v0.0.1, so the edpm_prepare roles fails to find the
ansibleee-operator, since it looks for the v0.0.1 version. Instead we
get the full name first and then use it to patch the csv.


cescgina:
recheck

cescgina:
recheck

PR #1560, opened 2024-04-24
Add support for Sushy Emulator backed virtual baremetal

lewisdenny:
This patch enables Redfish BMC control of VMs deployed via the reproducer
layout.

Sushy Emulator is deployed onto the Ansible controller as a Podman Pod
and access' the hypervisor's libvirt socket via SSH.

- `cifmw_use_sushy_emulator` parameter is added to control the
deployment of Sushy Emulator in the reproducer and set to true.
- `cifmw_use_vbmc` parameter is added to control the
deployment of Virtual BMC in the reproducer and behavoir changed to
default to false.
- sushy_emulator role now creates and uses it's own ssh keypair to
to communicate with the hypervisor
- `baremetal-info.yml` creation has now been extracted into it's own
task file and enhanced to support adding new baremetal hosts entries
resulting from scaling or additional hosts for baremetal instances.
- Sushy Emulator verification has now moved to it's own task file and
tests both Sushy Emulator connection and connection to each host.
- sushy_emulator role now uses the `cifmw_libvirt_manager_uuids` fact
to filter vms rather than using an additional `virsh` call. This means
`libvirt-client` can be removed from the packages list to install.
- Add `cifmw_sushy_redfish_bmc_protocol` param to control the bmc protocol Sushy Emulator uses. [1][2][3]

[1] https://github.com/openstack-k8s-operators/install_yamls/blob/6f43dd716e8d1ab7b38926ca29c45699a874e164/devsetup/scripts/edpm-compute-baremetal.sh#L76C14-L76C35
[2] https://github.com/metal3-io/baremetal-operator/blob/b87793e1e394bbc4e4b93fda967e2ff9f1bbbe79/docs/api.md?plain=1#L37-L38
[3] https://book.metal3.io/quick-start.html?highlight=redfish-virtualmedia#create-baremetalhosts

Jira: [OSPRH-6497](https://issues.redhat.com//browse/OSPRH-6497)

As a pull request owner and reviewers, I checked that:
- [x] Appropriate testing is done and actually running
- [x] Appropriate documentation exists and/or is up-to-date:
  - [x] README in the role
  - [x] Content of the docs/source is reflecting the changes

PR #1639, opened 2024-05-07
Remove cifmw_discover_latest_image_qcow_prefix from molecule config

eduolivares:
Previous value CentOS-Stream-GenericCloud-9-20240401 is not available
anymore.
This patch comments it out from the molecule config, because the default
upstream image can be used instead.

### DIFF
--- a/.config/molecule/config_local.yml
+++ b/.config/molecule/config_local.yml
@@ -20,10 +20,10 @@ provisioner:
   # Useful when an fix requires to provide some
   # CIFMW parameter to many roles, such as a broken
   # CentOS image.
-  inventory:
-    group_vars:
-      all:
-        cifmw_discover_latest_image_qcow_prefix: "CentOS-Stream-GenericCloud-9-20240401"
+  # inventory:
+  #   group_vars:
+  #     all:
+  #       cifmw_discover_latest_image_qcow_prefix: "CentOS-Stream-GenericCloud-9-20240506"
 
   config_options:
     defaults:


.config/molecule/config_local.yml
```diff
@@ -23,7 +23,7 @@ provisioner:
   inventory:
     group_vars:
       all:
-        cifmw_discover_latest_image_qcow_prefix: "CentOS-Stream-GenericCloud-9-20240401"
+        cifmw_discover_latest_image_qcow_prefix: "CentOS-Stream-GenericCloud-9-20240506"
```
cjeanner, line 26: I think we can remove that altogether, the upstream image is fixed afaik.

Care to comment out the `inventory` block so that we know how to override that kind of parameter in the future?

eduolivares: done, but I'm afraid there will be other unrelated molecule failures:
- https://review.rdoproject.org/zuul/build/8082f1cd117d41df82be84d55ed04795
- https://review.rdoproject.org/zuul/build/cb761caa529d4bf898c03134d5e11903

cjeanner: `kustomize_deploy` will be fixed in #1641
`sushy_emulator`.... maybe better check with @lewisdenny

cjeanner:
/approve
/lgtm

I'll check Zuul and override its result if only sushy is "missing" - this patch's needed to fix other molecule jobs, and sushy error isn't related to this patch here.

cjeanner:
(though sushy passed that time - so it's some random issue - kustomize_deploy has failed, but it's "expected" now)

.config/molecule/config_local.yml
lewisdenny: sushy_emulator molecule test is slightly flaky atm but completely reworked in https://github.com/openstack-k8s-operators/ci-framework/pull/1560

PR #1640, opened 2024-05-07
Fix allow/skip list generation in test-operator

kopecmartin:
There was an issue with the allow/skip list generation in the test-operator role. The generated lists of tests contained the literal interpretation of \n.

We want to get from:

```
excludeList: test1\ntest2
```
to:

```
excludeList: |
  test1
  test2
```

  - [x] README in the role

### DIFF
--- a/roles/test_operator/tasks/tempest-tests.yml
+++ b/roles/test_operator/tasks/tempest-tests.yml
@@ -33,12 +33,17 @@
         list_allowed
 
     - name: Set variable
+      vars:
+        allowed_tests: |
+          {% for item in list_allowed.allowed_tests %}
+          {{ item }}
+          {% endfor %}
       ansible.builtin.set_fact:
         test_operator_cr: >-
           {{
               test_operator_cr |
               combine({'spec': {'tempestRun': { 'includeList':
-                      list_allowed.allowed_tests | join('\n')
+                      allowed_tests
                       }}}, recursive=true)
           }}
 
@@ -60,12 +65,17 @@
         list_skipped
 
     - name: Set variable
+      vars:
+        skipped_tests: |
+          {% for item in list_skipped.skipped_tests %}
+          {{ item }}
+          {% endfor %}
       ansible.builtin.set_fact:
         test_operator_cr: >-
           {{
               test_operator_cr |
               combine({'spec': {'tempestRun': { 'excludeList':
-                      list_skipped.skipped_tests | join('\n')
+                      skipped_tests
                       }}}, recursive=true)
           }}
 


roles/test_operator/tasks/tempest-tests.yml
```diff
@@ -32,13 +32,20 @@
       register:
         list_allowed
 
+    - name: Prepare includeList for Tempest CR
+      set_fact:
```
pablintino, line 36: Please, use FQN module names: `ansible.builtin.set_fact`. It's encourage by Ansible guidelines and our linting.

roles/test_operator/tasks/tempest-tests.yml
```diff
@@ -32,13 +32,20 @@
...
+          {{ item }}
+          {% endfor %}
+
     - name: Set variable
       ansible.builtin.set_fact:
         test_operator_cr: >-
           {{
               test_operator_cr |
               combine({'spec': {'tempestRun': { 'includeList':
-                      list_allowed.allowed_tests | join('\n')
```
pablintino, line 41: Isn't this entire change the same as the simple removal of `| join('\n')` or, at most, change it to `| list` ?

roles/test_operator/tasks/tempest-tests.yml
```diff
@@ -59,13 +66,20 @@
...
+          {% endfor %}
+
     - name: Set variable
       ansible.builtin.set_fact:
         test_operator_cr: >-
           {{
               test_operator_cr |
               combine({'spec': {'tempestRun': { 'excludeList':
-                      list_skipped.skipped_tests | join('\n')
+                      skipped_tests
```
pablintino, line 82: ditto.

roles/test_operator/tasks/tempest-tests.yml
kopecmartin: there is but it doesn't work in the form as is and as i didn't create that one, i had to create a new PR

kopecmartin: I think that `| list` won't work cause the end result should not be a list, we want:
`
includeList: |
  test1
  test2
`
and not
`
includeList:
  - test1
  - test2
`

roles/test_operator/tasks/tempest-tests.yml
```diff
@@ -60,12 +65,17 @@
         list_skipped
 
     - name: Set variable
```
lpiwowar, line 67: This is really a nit :see_no_evil: . Maybe a more specific title for the task: `Set includeList variable`?

lpiwowar:
LGTM, waiting for the job results.

lpiwowar (approved):
LGTM

PR #1642, opened 2024-05-07
Extend ip ranges and regexes to allow deploying additional ocp workers

lmiccini:
Without the additional ips we would have errors like:

  fatal: [hypervisor-1 -> controller-0(192.168.111.50)]: FAILED! => changed=false
  message: IP Pool for 192.168.122.10-192.168.122.14 exhausted

Similarly, we need to expand the regex to allow ssh to get the right key for ssh'ing to ocp_worker nodes.


cjeanner:
Note: molecule jobs are currently broken for both libvirt_manager and reproducer roles, they should be fixed with #1639

PR #1643, opened 2024-05-07
Generic nat to custom defined nets

afazekas:
In case a net defined with cifmw_ci_nmstate_instance_config
it should be allowed to be also routed to external
to the hypervisor network.
Routing only happens if has address on the hypervisor ns0 .

  - [x] README in the role

### DIFF
--- a/roles/reproducer/tasks/main.yml
+++ b/roles/reproducer/tasks/main.yml
@@ -142,6 +142,33 @@
       }}
   failed_when: false
 
+- name: Ensure custom networks can be routed outside to the hypervisor
+  when:
+    - cifmw_reproducer_nat_public_interfaces | default ('false') | bool
+  become: true
+  block:
+    - name: Enable masquered over public interfaces
+      ansible.posix.firewalld:
+        masquerade: true
+        zone: public
+        permanent: true
+        immediate: true
+        state: enabled
+
+    - name: Ensure network is in libvirt zone
+      when:
+        - cifmw_ci_nmstate_instance_config is defined
+      ansible.posix.firewalld:
+        zone: libvirt
+        interface: "{{ item.name }}"
+        state: enabled
+        permanent: true
+        immediate: true
+      loop: >-
+        {{ cifmw_ci_nmstate_instance_config[inventory_hostname].interfaces | default([]) }}
+      loop_control:
+        label: "{{ item.name }}"
+
 - name: Run only on hypervisor with controller-0
   when:
     - (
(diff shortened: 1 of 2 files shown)


roles/reproducer/tasks/main.yml
```diff
@@ -142,6 +142,17 @@
       }}
   failed_when: false
 
+- name: Ensure custom networks can be routed outside to the hypervisor
```
cjeanner, line 145: 2 things:

you probably want to ensure firewalld is used, like we do here for instance:
https://github.com/openstack-k8s-operators/ci-framework/blob/52261a3c2e260af2379c6fb0f45120dd204c66b5/roles/libvirt_manager/tasks/create_networks.yml#L84-L88

and you probably should just ensure the created networks/ports are in the `libvirt` zone, like we do here:
https://github.com/openstack-k8s-operators/ci-framework/blob/52261a3c2e260af2379c6fb0f45120dd204c66b5/roles/libvirt_manager/tasks/create_networks.yml#L90-L102

You can loop on the nmstate networks I guess - and reproduce that same behavior.

afazekas:
recheck

roles/reproducer/tasks/main.yml
```diff
@@ -142,6 +142,33 @@
...
+  when:
+    - cifmw_reproducer_nat_public_interfaces | default ('false') | bool
+  become: true
+  block:
+    - name: Enable masquered over public interfaces
```
hjensas, line 150: FYI, this part can be dopped. It is managed elsewhere. See: https://github.com/openstack-k8s-operators/ci-framework/pull/1928

roles/reproducer/tasks/main.yml
```diff
@@ -142,6 +142,33 @@
...
+      ansible.posix.firewalld:
+        masquerade: true
+        zone: public
+        permanent: true
+        immediate: true
+        state: enabled
+
+    - name: Ensure network is in libvirt zone
```
hjensas, line 158: I think it would be better to change the `ci_nmstate`role - add a `zone` parameter where interfaces are defined there. So you set:

```
- name: Configure a couple of nets using ci_nmstate in two Ansible hosts
      vars:
        cifmw_ci_nmstate_network_layout:
            instance-1:
              default:
                connection: ci-private-network
                zone: _your_zone_of_choice_
```

You can even add that in the more advanced "direct" thing - cifmw_ci_nmstate_instance_config - it should be trivial to filter out the zone before passing the data to nmstate.

hjensas (changes requested):

pablintino:
Closing as it a long time stale PR.

PR #1645, opened 2024-05-07
Do not trigger deployment on molecule related change

cjeanner:
We may modify molecule config or zuul job from time to time. Such
changes shouldn't trigger a full deployment, since it doesn't affect it
in any way.


cjeanner:
recheck

PR #1646, opened 2024-05-07
Allow to deploy architecture against CRC

cjeanner:
This patch allows to deploy at least va-hci.yml scenario against CRC.

It's not enough, crafted parameter files are also needed to ensure
proper variables are passed in the right order, but mostly: it works.

It could be used to run architecture validation jobs, at least when VMs
are enough to test things up (for now, HCI, mostly).
When it comes to "baremetal provisioning", sushy may be leveraged to act
as the IPMI/related interface for metal3, allowing to go further down
the path of architecture testing.

This capability isn't "the end" - it's just offering a light way to test
things up.

Documentation will follow to show what specific parameter files are
needed (hence the DRAFT here).

- [X] For now there's no real testing for that patch. Running Architecture driven deploy isn't a thing yet.

### DIFF
--- a/playbooks/06-deploy-architecture.yml
+++ b/playbooks/06-deploy-architecture.yml
@@ -126,6 +126,15 @@
         name: kustomize_deploy
         tasks_from: check_requirements.yml
 
+    - name: Reduce OCP cluster size in architecture
+      when:
+        - groups['ocps'] | length == 1
+      ansible.builtin.import_role:
+        name: kustomize_deploy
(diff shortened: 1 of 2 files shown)


cjeanner:
recheck

lewisdenny (approved):
The code looks good to me and safe due to only running when `cifmw_architecture_scenario` is defined and when there is one node in the OCP group on the Zuul job definition.

PR #1649, opened 2024-05-07
DRAFT network management changes

cjeanner:
We need to generate all the MAC addresses before deploying anything:
this will allow us to stop relying on libvirt to get the IP in further
steps.

The goal here is to generate the "interfaces-info.yml" content before
starting to deploy any VM, and re-use the generated content when we
associate network interfaces to the VMs.

The generator takes care to not override existing data if we're
re-running a deployment, but only adds new content (for instance if you
add new nodes, or new network intefaces).

There's also a list of TODO for the future work on that networking
change.


roles/libvirt_manager/tasks/deploy_layout.yml
```diff
@@ -201,6 +201,66 @@
...
+        cifmw_libvirt_manager_bm_mac_mapping: >-
+          {{ _existing_nic_info.content | b64decode | from_yaml }}
+
+    - name: Generate VM list
+      vars:
+        vm_list: >-
+          {% set _vm_list = [] -%}
+          {% set _range = [] -%}
+          {% for _type in _layout.vms.keys() -%}
+          {%   set _range = range(0, lcm.vms[_type].amount | default(1)) -%}
```
hjensas, line 246: `s/lcm.vms[_type].amount/_layout.vms[_type].amount/` ?

hjensas (commented):
This change is too big to review - I see there are individual patches proposed as separate PRs. :+1:  Let's focus on getting those merged, so that this can get smaller and hopefully easier to grasp.

roles/libvirt_manager/tasks/create_networks.yml
```diff
@@ -81,6 +82,112 @@
...
+      - "option:dns-server,{{ ansible_facts[_name].ipv4.address }}"
+      {% endif -%}
+      {% if ansible_facts[_name].ipv6 is defined -%}
+      - "option6:dns-server,[{{ ansible_facts[_name].ipv6.address }}]"
+      {% endif -%}
+    _dns_listener:
+      - "{{ ansible_facts[_name].ipv4.address | default('') }}"
+      - "{{ ansible_facts[_name].ipv6.address | default('') }}"
+    _dns_addr: |-
+      {% if _no_prefix_name is match('^[a-z0-9]bm$')%}
```
hjensas, line 116: Why are we adding these in `create_networks.yml`? These are DNS entries for OpenShift, I don't think it belongs here. We only need one entry for each of these right? Can we not make it simple and just require this in the input?

roles/libvirt_manager/tasks/create_networks.yml
```diff
@@ -81,6 +82,112 @@
...
+    _name: "{{ item | replace('-', '_') }}"
+    _no_prefix_name: "{{ _name | regex_replace('cifmw[-_]','') }}"
+    _fixed_nets: >-
+      {{
+        cifmw_libvirt_manager_fixed_networks_defaults +
+        cifmw_libvirt_manager_fixed_networks
+      }}
+    _dns_options: |-
+      {% if ansible_facts[_name].ipv4 is defined -%}
+      - "option:dns-server,{{ ansible_facts[_name].ipv4.address }}"
```
hjensas, line 107: This assumes the DNS server for a network is the address of the network itself. I am not convinced this is always what we want.

roles/libvirt_manager/tasks/create_networks.yml
```diff
@@ -81,6 +82,112 @@
...
+      {% endif -%}
+      {% if ansible_facts[_name].ipv6 is defined -%}
+      - "option6:dns-server,[{{ ansible_facts[_name].ipv6.address }}]"
```
hjensas, line 110: This causes error `'list object' has no attribute 'address'.` - because ansible fact for ipv4 and ipv6 follow different schema - wee need to do something to select the IPv6 entry where `scope == global` - and then pick the address.

```
  "ipv6": [
    {
      "address": "fc00:abcd:abcd:fc00::1",
      "prefix": "64",
      "scope": "global"
    },
    {
      "address": "fe80::5054:ff:fe43:2504",
      "prefix": "64",
      "scope": "link"
    }
  ],
```

```
  "ipv4": {
    "address": "172.31.255.1",
    "broadcast": "172.31.255.255",
    "netmask": "255.255.255.0",
    "network": "172.31.255.0"
  },
```

cjeanner: ah, good to know - I'll check that out, couldn't test actual v6 on my end :/.

roles/libvirt_manager/tasks/create_networks.yml
cjeanner: I'll see if I can move them out, for instance in the reproducer role instead.

cjeanner:
All features (and even more) are in proper PRs. Closing that draft/test/hack.

PR #1655, opened 2024-05-08
[kustomize_deploy] Combine multiple user kustomizations as one

pablintino:
In automation it may happen that the jobs have already defined a cifmw_architecture_user_kustomize but the user wants to add his own kutomization or just concat multiple ones as part of his job parenting strategy. With the previous approach only one cifmw_architecture_user_kustomize may survive, forcing the user to be aware of the already defined kustomization and making him duplicate all the content.
With this approach we can consume multiple kustomization dicts that will be combined all together before consuming them as a single one.

  - [x] README in the role

### DIFF
--- a/docs/source/usage/01_usage.md
+++ b/docs/source/usage/01_usage.md
@@ -44,7 +44,7 @@ are shared among multiple roles:
   in the architecture repository. Defaults to `default.yaml`
 - `cifmw_architecture_scenario`: (String) The selected VA scenario to deploy.
 - `cifmw_architecture_wait_condition`: (Dict) Structure defining custom wait_conditions for the automation.
-- `cifmw_architecture_user_kustomize`: (Dict) Structure defining user provided kustomization for automation.
+- `cifmw_architecture_user_kustomize.*`: (Dict) Structures defining user provided kustomization for automation. All these variables are combined together.
 - `cifmw_ceph_target`: (String) The Ansible inventory group where ceph is deployed. Defaults to `computes`.
 - `cifmw_run_tests`: (Bool) Specifies whether tests should be executed.
   Defaults to false.
--- a/roles/kustomize_deploy/tasks/_compute_user_kustomize.yml
+++ b/roles/kustomize_deploy/tasks/_compute_user_kustomize.yml
@@ -1,0 +1,29 @@
+---
+- name: Set the final cifmw_architecture_user_kustomize based on its patches
+  vars:
+    _kustomize_user_patches: >-
+      {{
+        hostvars[inventory_hostname] |
+        dict2items |
+        selectattr("key", "match", "^cifmw_architecture_user_kustomize.+") |
+        sort(attribute='key')
+      }}
+  ansible.builtin.set_fact:
+    _cifmw_kustomize_deploy_user_kustomize: >-
+      {{
+        _cifmw_kustomize_deploy_user_kustomize |
+        default({}) |
+        combine(item.value, recursive=True)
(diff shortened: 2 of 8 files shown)


docs/source/usage/01_usage.md
```diff
@@ -44,7 +44,7 @@ are shared among multiple roles:
   in the architecture repository. Defaults to `default.yaml`
 - `cifmw_architecture_scenario`: (String) The selected VA scenario to deploy.
 - `cifmw_architecture_wait_condition`: (Dict) Structure defining custom wait_conditions for the automation.
-- `cifmw_architecture_user_kustomize`: (Dict) Structure defining user provided kustomization for automation.
+- `cifmw_architecture_user_kustomize.*`: (Dict) Structures defining user provided kustomization for automation. All these variables are combined together in reverse order.
```
cjeanner, line 47: why `reverse` order ?

roles/kustomize_deploy/tasks/execute_step.yml
```diff
@@ -125,16 +128,16 @@
            }}
         cifmw_ci_gen_kustomize_values_userdata: >-
           {{
-            (cifmw_architecture_user_kustomize is defined and
-             cifmw_architecture_user_kustomize[_stage_name_id][_name] is defined
+            (_cifmw_kustomize_deploy_user_kustomize is defined and
```
cjeanner, line 131: `_cifmw_kustomize_deploy_user_kustomize` would always be defined I think. We can probably make those checks smaller then.

roles/kustomize_deploy/tasks/install_operators.yml
```diff
@@ -14,6 +14,9 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+- name: Compute the user kustomize dict
```
cjeanner, line 17: hmmmm I'd put that import once, in the `check_requirements.yml` instead?

roles/kustomize_deploy/tasks/execute_step.yml
```diff
@@ -110,6 +110,9 @@
       ansible.builtin.include_role:
         name: run_hook
 
+    - name: Compute the user kustomize dict
```
cjeanner, line 113: (check note in install_operators.yml)

roles/kustomize_deploy/tasks/execute_step.yml
rebtoor: Yup, should it fallback to `{}` right?

rebtoor (approved):

rebtoor (approved):

docs/source/usage/01_usage.md
cjeanner: sounds good!

roles/kustomize_deploy/tasks/_compute_user_kustomize.yml
```diff
@@ -0,0 +1,29 @@
...
+  loop: >-
+    {{
+      _kustomize_user_patches +
+      [
+        {
+          'key': 'cifmw_architecture_user_kustomize',
+          'value': (cifmw_architecture_user_kustomize | default({}))
+        }
+      ]
+    }}
```
cjeanner, line 27: That's smart! 👍

cjeanner:
Not sure if ready to be approved? There's an unticked checkbox about testing - but since it's not seen as a checkbox, check doesn't fail.

rebtoor (approved):

PR #1658, opened 2024-05-08
[libvirt_manager] Increase the number of retries for SSH configure

psathyan:
In some environment, configure ssh access on ocp is failing with rc = 255 meaning the system is not completely booted. We also reach the maximum retries by then.

This patch increases the number of retries to allow ocp-0 to fully boot and allow remote connections.

### DIFF
--- a/roles/libvirt_manager/tasks/start_manage_vms.yml
+++ b/roles/libvirt_manager/tasks/start_manage_vms.yml
@@ -177,7 +177,7 @@
       set -o pipefail;
       cat ~/.ssh/authorized_keys |
       ssh -v {{ _user }}@{{ vm_ip.nic.host }} "cat >> ~/.ssh/authorized_keys"
-  retries: 5
+  retries: 18
   delay: 10
   register: _ssh_access
   until: _ssh_access.rc == 0


roles/libvirt_manager/tasks/start_manage_vms.yml
```diff
@@ -177,7 +177,7 @@
       set -o pipefail;
       cat ~/.ssh/authorized_keys |
       ssh -v {{ _user }}@{{ vm_ip.nic.host }} "cat >> ~/.ssh/authorized_keys"
-  retries: 5
+  retries: 18
```
cjeanner, line 180: why 18 and not 20? I know, sounds stupid question, but... why? :)

psathyan: :) total wait time is 180 seconds which would translate to 3 minutes.

cjeanner: should be enough - maybe we can even set a parameter for the future?

psathyan: this patch did not work in the failing environment. Looks like the problem is elsewhere.

psathyan:
closing this PR as it didn't help resolve the issue at hand.

PR #1660, opened 2024-05-08
Replace noop molecule job parent

cjeanner:
Seems we don't have the label downstream - let's use a parent that seems
to be small enough, while existing in both locations


rlandy (commented):
/lgtm

cjeanner:

Molecule passed, it's hitting the deployment which is useless for that change.

PR #1662, opened 2024-05-09
Fix dlrn_reporting role for component promotion

raukadah:
For integration promotion, Only aggregate dlrn hash is needed but for component promotion, we need to commit, distro and extended hash.

This pr fixes the same by adding proper conditions for components.

  - [x] README in the role


raukadah:
recheck

PR #1666, opened 2024-05-09
Move the config of ceph backends to a template file, gating++

tosky:
Use 'template' instead of 'copy' to generate the Kustomization which will be used to configure the services with ceph as backend. This is needed because, according the ansible documentation, "If you need variable interpolation in copied files, use the ansible.builtin.template module. Using a variable with the content parameter produces unpredictable results." https://docs.ansible.com/ansible/latest/collections/ansible/builtin/copy_module.html#synopsis

Remove an old template, previously used for the same purpose, then removed and readded by mistake in a rebase.

Finally, make sure that a job which deploys ceph with HCI and use all these files actually is triggered when these files are changed.


tosky:
Updated again to fix the name of the template file (which has been moved around in the meantime) in the zuul files list

raukadah (approved):

PR #1667, opened 2024-05-09
Architecture repository test playbook

cjeanner:
The architecture repository needs some more testing. This playbook
allows to use the ci-framework "flexible loop" in order to ensure we
can:
- load the files (all of them)
- loop on the automation steps
- generate all the needed content (via ci_gen_kustomize_values)
- build final CRs with all the data

This is mostly a "read-only", "linting-like" check, and does NOT apply
the CRs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,6 +94,28 @@ check_zuul_files: role_molecule ## Regenerate zuul files and check if they have 
 check_k8s_snippets_comment: ## Check template snippets in ci_gen_kustomize_values to ensure proper source is set
 	./scripts/check_k8s_snippets_comment.sh 2>&1 | ansi2txt | tee $(LOG_DIR)/check_k8s_snippets_comment.log
 
+##@ Architecture tests
+.PHONY: architecture_test
+architecture_test: setup_tests architecture_test_nodeps ## Runs architecture-test with dependencies install
+
+.PHONY: architecture_test_nodeps
+architecture_test_nodeps: export PATH=/tmp/bin:/usr/bin:/usr/sbin:/bin:/sbin:/usr/local/bin:/usr/local/sbin
+architecture_test_nodeps:
+	$(call check-var-defined,SCENARIO_NAME)
+	$(call check-var-defined,ARCH_REPO)
+	$(call check-var-defined,NET_ENV_FILE)
+	cp -r ${ARCH_REPO} /tmp/local-arch-repo;
+	ansible-playbook -i localhost, -c local \
+		/install_yamls/devsetup/download_tools.yaml \
+		--tags kubectl,kustomize;
+	ansible-playbook -i localhost, -c local \
+		ci/playbooks/validate-architecture.yml \
+		-e cifmw_zuul_target_host=localhost \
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -146,6 +168,25 @@ run_ctx_ansible_test: ci_ctx ## Run molecule check in a container
 		-e ANSIBLE_REMOTE_TMP=/tmp \
 		${CI_CTX_NAME} bash -c "make ansible_test_nodeps" ;
 
+.PHONY: run_ctx_architecture_test
+run_ctx_architecture_test: ## Run architecture_test in a container
+	$(call check-var-defined,SCENARIO_NAME)
```
queria, line 173: Maybe alternatively the 'check-var' could be splitted into separate first 'dependency', and then ci_ctx target can be also as explicit dependency?

e.g.:

```
## Run architecture_test in a container
.PHONY: run_ctx_architecture_test
run_ctx_architecture_test: checkvars_architecture_test ci_ctx
    podman run --rm ...
```

queria: Possibly (if the install_yamls is act not needed here - as its not above) the `checkvars_architecture_test` target could be then reused by both `architecture_test_nodeps` and `run_ctx_architecture_test`

Makefile
```diff
@@ -146,6 +168,25 @@ run_ctx_ansible_test: ci_ctx ## Run molecule check in a container
...
+	@$(MAKE) ci_ctx
+	podman run --rm --security-opt label=disable -v .:/opt/sources \
+		-e HOME=/tmp \
+		-e ANSIBLE_LOCAL_TMP=/tmp \
+		-e ANSIBLE_REMOTE_TMP=/tmp \
+		-e SCENARIO_NAME=${SCENARIO_NAME} \
+		-e ARCH_REPO=/architecture \
+		-e NET_ENV_FILE=/net-env-file.yml \
+		-v ${ARCH_REPO}:/architecture:ro \
+		-v ${NET_ENV_FILE}:/net-env-file.yml:ro \
```
queria, line 186: For ET_ENV_FILE and others (_REPO), it might be worth mentioning somewhere they need to be absolute paths (or its treated as existing podman volume) - or maybe `$(readlink -f $...)` them here?

(I know about this podman behavior, just did not thinked about it when trying, it results correctly in:
```Error: creating named volume "ci/playbooks/files/networking-env-definition.yml": running volume create option: names must match [a-zA-Z0-9][a-zA-Z0-9_.-]*: invalid argument```)

queria:
/lgtm

my comments above are just possible improvements/tweaks, aside that seems working for me

only one question maybe - is it not in current version - possible to get the generated snippets outside of the container?

Makefile
cjeanner: actually you can use relative path, but must use `./foo/bar` or `../foo/bar` :)

cjeanner:
> /lgtm
> 
> my comments above are just possible improvements/tweaks, aside that seems working for me
> 
> only one question maybe - is it not in current version - possible to get the generated snippets outside of the container?

Just a matter to edit the podman run command to add some new volume, yes.

PR #1674, opened 2024-05-10
Fix Save repo-setup hash content task

raukadah:
`Save repo-setup hash content` task dumps the proper dlrn md5 hash for component/intergration/check pipeline.

Currently this task depends on cifmw_repo_setup_component_promotion_tag whose default value is defined in role defaults.

In this case, this task is working wrongly. Let's use cifmw_repo_setup_component_name length as a param to dump the proper hash.


roles/repo_setup/tasks/artifacts.yml
```diff
@@ -24,14 +24,14 @@
     _repo_setup_json: "{{ _get_hash.stdout | from_json }}"
   block:
     - name: Dump full hash in delorean.repo.md5 file
-      when: cifmw_repo_setup_component_promotion_tag is not defined
+      when: cifmw_repo_setup_component_name | length <= 0
```
pablintino, line 27: nit: Length of a list returning a negative value?

marios (approved):

PR #1676, opened 2024-05-10
Allow passing base64 encoded patches to ci_gen_kustomize

pablintino:
Some content passed to the ci_gen_kustomize role can be interpreted by Ansible as a jinja template (edpm_network_config_template). That content, when ingested by the framework, can lead to a crash as Ansible may try to use it as a template. Mechanisms like raw/endraw or !unsafe don't work, as those tags are removed in the first write of the variable to a file (and the framework performs one write of those variables before edpm-deploy).
To allow easily passing those kind of variables this mechanism can be handy, as the content is safely encoded as a plain string before getting applied.

  - [x] README in the role

### DIFF
--- a/roles/ci_gen_kustomize_values/README.md
+++ b/roles/ci_gen_kustomize_values/README.md
@@ -30,10 +30,14 @@ with a message.
   Defaults to `[]`.
 * `cifmw_ci_gen_kustomize_values_userdata`: (Dict) Data structure you want to combine in the generated output.
   Defaults to `{}`.
+* `cifmw_ci_gen_kustomize_values_userdata_b64`: (List) Base64 encoded list of data to combine in the generated output.
+  Defaults to `[]`.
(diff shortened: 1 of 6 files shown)


roles/ci_gen_kustomize_values/filter_plugins/ci_gen_kustomize_create_path_patch.py
rebtoor, line 1: Should the file be called `ci_gen_kustomize_create_b64_patch.py`?

rebtoor:
@pablintino can you update the commit message and the pr topic/description to reflect the fact that we are adding a new (filter) plugin?

cjeanner:
As a quick note, that kind of feature would benefit of some (new) job based on what's coming via #1667 - with some crafted environment files, we would be able to ensure things are building accordingly :).

cjeanner:
/approve

I love the idea, and it will help ppl when they are passing templates to be interpreted within openshift.

rebtoor (approved):

PR #1678, opened 2024-05-10
[os_net_setup] Update public allocation pool and network configuration

fyanac:
Extend public allocation pool

### DIFF
--- a/roles/os_net_setup/defaults/main.yml
+++ b/roles/os_net_setup/defaults/main.yml
@@ -6,15 +6,16 @@ cifmw_os_net_setup_osp_calls_delay: 5
 cifmw_os_net_setup_config:
   - name: public
     external: true
+    shared: true
     provider_network_type: flat
     provider_physical_network: datacentre
     subnets:
       - name: public_subnet
         cidr: 192.168.122.0/24
-        allocation_pool_start: 192.168.122.190
+        allocation_pool_start: 192.168.122.171
         allocation_pool_end: 192.168.122.250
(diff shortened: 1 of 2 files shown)


roles/os_net_setup/defaults/main.yml
```diff
@@ -6,15 +6,17 @@ cifmw_os_net_setup_osp_calls_delay: 5
 cifmw_os_net_setup_config:
   - name: public
     external: true
+    shared: true
+    mtu: 1330
```
karelyatin, line 10: why mtu reduction?

Apart from that all looks good.

karelyatin (changes requested):

roles/os_net_setup/defaults/main.yml
rsafrono: This is for whitebox path mtu discovery tests (the ones under test_mtu.py). For this coverage our tripleo jobs had external network mtu 1500 and tenant mtu 9000. When we were planning osp18 jobs I suggested for simplification to use normal tenant networks mtu but just reduce external network mtu. So this is it.
Maybe we should have set this only for neutron jobs, not for everybody.

karelyatin: +1 to only override where it's really needed, not in default

fyanac: I removed mtu reduction

rsafrono (approved):
lgtm

karelyatin (approved):

PR #1679, opened 2024-05-10
libvirt_manager - enable sushy_emulator

hjensas:
Add a new boolean option:
 `cifmw_libvirt_manager_enable_sushy_emulator` (defaults to: false).

When enabled - include the sushy_emulator role in deploy_layout.yml to set-up virtual RedFish BMC.

Also:
- Change hos sushy_emulator role builds the libvirt_uri, instead of a lookup from hostvars require the hostname to be defined in a variable when calling the role. This allows use of the role even if the libvirt hypervisor is not managed by ansible.
- Create a firewall rule to allow traffic to the sushy-emulator port when installed as a `podman` container.

  - [ ] README in the role


hjensas:
I see there is already a patch to do this : https://github.com/openstack-k8s-operators/ci-framework/pull/1560

PR #1682, opened 2024-05-13
Reload crc NetworkManager in reproducer

cescgina:
The fix for reproducer dns[1] does not seem to work in all instances.
NetworkManager was still overwriting the /etc/resolv.conf contents, so
this change reloads the service after changing the config to ensure it
does not modify the resolv.conf on reboot.

[1] https://github.com/openstack-k8s-operators/ci-framework/pull/1637


cescgina:
recheck

lewisdenny (approved):
/lgtm

PR #1683, opened 2024-05-13
Add nodeset and job config for multicell

bogdando:
Add nodeset for multicell adoption jobs that replicates
the multinode adoption nodeset, but custom networks and hostnames.

Add a non-voting WIP job for multicells development (via node
autoholds, as there is no other way to set up dev env)

  - [x] README in the role

Required-By: https://github.com/openstack-k8s-operators/install_yamls/pull/826
Required-By: https://review.rdoproject.org/r/c/rdo-jobs/+/53192

JIRA [OSPRH-6548](https://issues.redhat.com/browse/OSPRH-6548)

### DIFF
--- a/zuul.d/adoption.yaml
+++ b/zuul.d/adoption.yaml
@@ -141,11 +141,11 @@
     nodeset: centos-9-multinode-rhel-9-2-crc-extracted-2.30-3xl
     roles:
       - zuul: github.com/openstack-k8s-operators/ci-framework
-    pre-run:
+    pre-run: &multinode-prerun
       - ci/playbooks/multinode-customizations.yml
       - ci/playbooks/e2e-prepare.yml
       - ci/playbooks/dump_zuul_data.yml
-    post-run:
+    post-run: &multinode-postrun
       - ci/playbooks/e2e-collect-logs.yml
       - ci/playbooks/collect-logs.yml
       - ci/playbooks/multinode-autohold.yml
@@ -173,7 +173,7 @@
             vlan: 44
             range: 172.21.0.0/24
         instances:
-          controller:
+          controller: &multinode-crlr
             networks:
               default:
                 ip: 192.168.122.11
@@ -185,7 +185,7 @@
                 ip: 172.19.0.4
               storage_mgmt:
                 ip: 172.20.0.4
-          crc:
+          crc: &multinode-crc
             networks:
               default:
                 ip: 192.168.122.10
@@ -199,7 +199,7 @@
                 ip: 172.20.0.5
               external:
                 ip: 172.21.0.5
-          undercloud:
+          undercloud: &multinode-uc
             networks:
               default:
                 ip: 192.168.122.100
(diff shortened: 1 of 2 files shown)


bogdando:
recheck retry lmit

zuul.d/adoption.yaml
```diff
@@ -239,63 +239,87 @@
...
                 config_nm: false
-          overcloud-controller-2:
+          overcloud-controller-2: &multinode-crlr-leaf2
             networks:
               default:
                 ip: 192.168.122.105
                 config_nm: false
               internal-api:
-                ip: 172.17.0.105
+                ip: 172.17.2.105
```
marios, line 268: since you are adding a new base job below can you please keep these changes confined there?
i.e. avoid changes to cifmw-adoption-base-source-multinode and keep the changes in cifmw-adoption-base-source-multinode-novacells

marios (changes requested):

zuul.d/adoption.yaml
bogdando: the plan is to reuse the same networks "leafs" for both jobs, to reduce the code duplication and to simplify maintaining changes across them (see in the commit message)

bogdando:
recheck cifmw-pod-ansible-test

lewisdenny (commented):
The config looks good to me, re-requesting @marios to review. 

@cc/ @bshewale in case this merges mid 4.15 upgrade

zuul.d/adoption.yaml
marios: i would prefer not to make more changes to the multinode config while we (me and @cescgina) are still working on that as already discussed.

If this is really a blocking issue for you then perhaps we can consider it but you would have to ensure this is working/doesn't break the existing job (i.e. please make sure you are running the periodic-data-plane-adoption-multinode-osp-17-to-extracted-crc-no-ceph at least, and lets coordinate with Joan regarding the ceph job when you are closer to merging this).

However for now, can you please mark this as draft to prevent accidental merge?

marios (changes requested):

zuul.d/adoption.yaml
marios: forgot to add... note the

```
config_nm: false
```

which means the CI framework network code will not apply this configuration. For tripleo nodes, instead we have os-net-config, so you'll have to update that config instead to create the networks with the ranges you want https://github.com/openstack-k8s-operators/install_yamls/blob/c154ee06caecb856ad910979da218bd3b47f22e2/devsetup/tripleo/overcloud_net_config.j2#L2

bogdando: done

bogdando:
@marios PTAL, I've addressed you concerns

marios (approved):
yes thanks, this way it is cleaner and you can work independently of the other multinode job

bogdando:
many thanks for reviews, can we merge it now?

cescgina (commented):
/approve

PR #1689, opened 2024-05-14
Switch away from `lookup()`

cjeanner:
Since we want to run some validations against `architecture` repository,
we end calling the role against a remote node instead of localhost.

The usage of `lookup()` breaks the loop, especially on Zuul: the
`lookup()` is launched onto the zuul-executor node, where we don't
manage anything, and may lead to multiple issues such as:
- collection not present (it seems to be present, but still)
- kustomize being missing from the path (seems to be present for now)
- 
Calling plain `command` ensures we're running the build onto an
environment we 100% manage and know.

This patch also fixes molecule tests on kustomize_deploy.

### DIFF
--- a/roles/kustomize_deploy/tasks/execute_step.yml
+++ b/roles/kustomize_deploy/tasks/execute_step.yml
@@ -192,17 +192,25 @@
              path_join | realpath
           }}
       block:
+        # We have to use plain `kustomize build`: lookup would
+        # be executed locally, on the ansible-controller. In CI,
+        # that would be the zuul-executor, and we don't manage them,
+        # leading to a risk to either NOT have kustomize, or an old,
+        # deprecated version.
         - name: "Build kustomized content for {{ stage.path }}"
+          register: _kustomize_output
+          environment:
+            PATH: "{{ cifmw_path | default(ansible_env.PATH) }}"
+          ansible.builtin.command:
+            chdir: "{{ _chdir }}"
+            cmd: kustomize build
+
+        - name: Output kustomize build in final file
           ansible.builtin.copy:
+            backup: true
             dest: "{{ _output }}"
+            content: "{{ _kustomize_output.stdout }}"
             mode: "0644"
-            content: >-
-              {{
-                lookup(
-                  'kubernetes.core.kustomize',
-                  dir=_chdir
-                )
-              }}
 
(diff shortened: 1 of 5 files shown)


roles/kustomize_deploy/tasks/execute_step.yml
```diff
@@ -192,17 +192,19 @@
...
-                  dir=_chdir
-                )
-              }}
+          environment:
+            PATH: "{{ cifmw_path | default(ansible_env.PATH) }}"
+          ansible.builtin.shell:
+            chdir: "{{ _chdir }}"
+            cmd: >-
+              set -o pipefail;
+              kubectl kustomize > {{ _output }}
```
rebtoor, line 207:
```suggestion
          ansible.builtin.command:
            chdir: "{{ _chdir }}"
            cmd: kubectl kustomize -o {{ _output }}
```

rebtoor: i was thinking about adding `creates` but probably we want to enforce the file creation each time we run the role, right?

cjeanner: checking in a few, I want a green run first, molecule seems to be broken on that role.

roles/kustomize_deploy/tasks/execute_step.yml
```diff
@@ -192,17 +192,25 @@
...
+          ansible.builtin.command:
+            chdir: "{{ _chdir }}"
+            cmd: >-
+              kubectl kustomize .
+
+        - name: Output kustomize build in final file
           ansible.builtin.copy:
+            backup: true
             dest: "{{ _output }}"
-            mode: "0644"
```
rebtoor, line 198: let's keep it please :)

rebtoor (approved):

rebtoor:
recheck

rebtoor (approved):

psathyan:
This may have broken jobs based on OCP deployed by devscripts as `kustomize` binary is not installed.

```
2024-05-14 21:48:25,261 p=22441 u=zuul n=ansible | fatal: [localhost]: FAILED! => changed=false 
  cmd: kustomize build
  msg: '[Errno 2] No such file or directory: b''kustomize'''
  rc: 2
  stderr: ''
  stderr_lines: <omitted>
  stdout: ''
  stdout_lines: <omitted>
  ```

PR #1693, opened 2024-05-14
Add become in bootstrap-hypervisor.yml

hjensas:
Let's not assume it is possible to login as root.
Add's "become: true" to the tasks that require priviliges.

- [X] No actual test for that playbook - it's provided as an extra.

### DIFF
--- a/docs/source/files/bootstrap-hypervisor.yml
+++ b/docs/source/files/bootstrap-hypervisor.yml
@@ -4,6 +4,7 @@
     _user: "{{ remote_user | default('zuul') }}"
   tasks:
     - name: Create user on remote host
+      become: true
       ansible.builtin.user:
         name: "{{ _user }}"
         state: present
@@ -34,6 +35,7 @@
 
     - name: Add RSA pub key to authorized keys
       when: rsa_pub_key.stat.exists
+      become: true
       ansible.posix.authorized_key:
         user: "{{ _user }}"
         state: present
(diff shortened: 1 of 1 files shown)


docs/source/files/bootstrap-hypervisor.yml
```diff
@@ -4,6 +4,7 @@
     _user: "{{ remote_user | default('zuul') }}"
   tasks:
     - name: Create user on remote host
```
cjeanner, line 6: sounds weird, since:
- the doc mentions "password root access"
- that playbook is the one creating the rootless user + sudo

so basically.... you mostly don't use that playbook... ? I had the same reflection with a DS equivalent patch...

hjensas: The usecases are:
- you have a baremetal "cloud" 
- you run nested virt in some cloud
 ... and the server you requested is provisioned - you are given access via the "cloud-user" of the image deployed. You have SSH access, you have sudo - but you cannot login as root.

It is possible to just use the `cloud-user`. But with that follows the editing of stuff in notes/chat with co-workers that uses the default `zuul` user.

cjeanner: oh. ok. heh, makes sense.
Note that my own baremetal doesn't have "zuul" user - that's how I catch (many) issues with paths, facts and so on.

PR #1694, opened 2024-05-14
Ceph settings: more modular manila configuration

tosky:
Add more logic to the manila configuration so that both a cephfs and an NFS/ganesha backend can be configured.

  - [x] README in the role


fmount (approved):
/lgtm

PR #1695, opened 2024-05-14
wip: check disposition of testing of ironic tests

juliakreger:
We've put in fixes for tests, and we likely have tests we should be fixing if they are broken, but don't know why they were disabled.

So this change is to bring some more clarity so the team can figure out the correct course of action.

  - [ ] README in the role


juliakreger:
recheck

PR #1697, opened 2024-05-15
Add redfish bmc protocol to BMH CR

lewisdenny:
Currently the connection string is set to `http://$IP:8000/redfish/v1/Systems/$UUID`, the BMH CR expects a bmc protocol of `ipmi, redfish or redfish-virtualmedia`.

This patch adds the `cifmw_deploy_bmh_redfish_bmc_protocol` prefixed to the connection string if the connection type is `redfish`

`cifmw_deploy_bmh_redfish_bmc_protocol` is set to `redfish-virtualmedia` by default to match [1] but is now configurable if needed.

More details can be found [2] [3]

[1] https://github.com/openstack-k8s-operators/install_yamls/blob/6f43dd716e8d1ab7b38926ca29c45699a874e164/devsetup/scripts/edpm-compute-baremetal.sh#L76C14-L76C35
[2] https://github.com/metal3-io/baremetal-operator/blob/b87793e1e394bbc4e4b93fda967e2ff9f1bbbe79/docs/api.md?plain=1#L37-L38
[3] https://book.metal3.io/quick-start.html?highlight=redfish-virtualmedia#create-baremetalhosts

  - [x] README in the role

### DIFF
--- a/roles/deploy_bmh/template/bmh.yml.j2
+++ b/roles/deploy_bmh/template/bmh.yml.j2
@@ -13,7 +13,7 @@ metadata:
     workload: {{ node_name.split('-')[0] }}
 spec:
   bmc:
-    address: {{ node_data['connection'] }}
+    address: {{ cifmw_deploy_bmh_redfish_bmc_protocol ~ '+' if 'redfish' in node_data['connection'] }}{{ node_data['connection'] }}
     credentialsName: {{ node_name }}-bmc-secret
     disableCertificateVerification: {{ cifmw_deploy_bmh_disable_certificate_validation }}
 {% for nic in (node_data['nics'] | default([])) if nic['network'] == cifmw_deploy_bmh_boot_interface %}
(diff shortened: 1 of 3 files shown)


roles/deploy_bmh/template/bmh.yml.j2
```diff
@@ -13,7 +13,7 @@ metadata:
     workload: {{ node_name.split('-')[0] }}
 spec:
   bmc:
-    address: {{ node_data['connection'] }}
+    address: {{ cifmw_deploy_bmh_redfish_bmc_protocol ~ '+' if 'redfish' in node_data['connection'] }}{{ node_data['connection'] }}
```
pablintino, line 16: We are using `idrac` in most of the DS hardware, this approach will blow the entire NFV job.

lewisdenny: I checked the D/S job and only seen ipmi addresses in the rendered bmh resource? If you could point me to a reference on Slack I would be grateful. 

You're right though, this test is too broad and would affect anyone using redfish, I'll have a think and submit a revision, thanks :)

PR #1703, opened 2024-05-15
Switch from `kustomize` to `oc kustomize`

rebtoor:
We are restoring the "original" behavior but using `oc` instead of `kubectl` since it should be available everywhere.

### DIFF
--- a/roles/kustomize_deploy/tasks/execute_step.yml
+++ b/roles/kustomize_deploy/tasks/execute_step.yml
@@ -192,7 +192,7 @@
              path_join | realpath
           }}
       block:
-        # We have to use plain `kustomize build`: lookup would
+        # We have to use plain `oc kustomize`: lookup would
         # be executed locally, on the ansible-controller. In CI,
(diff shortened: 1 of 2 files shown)


roles/kustomize_deploy/tasks/execute_step.yml
```diff
@@ -192,25 +192,29 @@
...
         # leading to a risk to either NOT have kustomize, or an old,
         # deprecated version.
         - name: "Build kustomized content for {{ stage.path }}"
-          register: _kustomize_output
           environment:
             PATH: "{{ cifmw_path | default(ansible_env.PATH) }}"
           ansible.builtin.command:
             chdir: "{{ _chdir }}"
-            cmd: kustomize build
+            cmd: "oc kustomize -o {{ _output }}"
```
cjeanner, line 205: why don't we just capture the output and use the copy to inject it? Everything would be automated, and backup would be in the "right" order/naming... ?

raukadah:
/lgtm
/approve

tested downstream!

PR #1704, opened 2024-05-15
Don't hardcode device image file in unit template

cescgina:
The cifmw_block_device role creatas a systemd unit that harcodes the
device image file to user for ceph-osd-losetup. Instead it should use
the cifmw_block_device_image_file var, since that var will always point
to correct path for the file (as it has a default).

### DIFF
--- a/roles/cifmw_block_device/templates/ceph-osd-losetup.service.j2
+++ b/roles/cifmw_block_device/templates/ceph-osd-losetup.service.j2
@@ -9,7 +9,7 @@ After=syslog.target
 [Service]
 Type=oneshot
 ExecStart=/bin/bash -c '/sbin/losetup {{ cifmw_block_device_loop }} || \
-/sbin/losetup {{ cifmw_block_device_loop }} /var/lib/ceph-osd.img ; partprobe {{ cifmw_block_device_loop }}'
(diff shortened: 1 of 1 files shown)


fultonj (approved):
+2 Thank you for fixing this bug.

roles/cifmw_block_device/templates/ceph-osd-losetup.service.j2
```diff
@@ -9,7 +9,7 @@ After=syslog.target
 [Service]
 Type=oneshot
 ExecStart=/bin/bash -c '/sbin/losetup {{ cifmw_block_device_loop }} || \
-/sbin/losetup {{ cifmw_block_device_loop }} /var/lib/ceph-osd.img ; partprobe {{ cifmw_block_device_loop }}'
+/sbin/losetup {{ cifmw_block_device_loop }} {{ cifmw_block_device_image_file }} ; partprobe {{ cifmw_block_device_loop }}'
```
fultonj, line 12: Yes, we have the same default for this value and should use the variable in case someone overrides it.

https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/cifmw_block_device/defaults/main.yml#L21

PR #1705, opened 2024-05-15
Fix loop for jump

kopecmartin:
item.path does not need the map(attribute='path'), that's a relict from #1675 PR.

  - [x] README in the role


kopecmartin:
we've verified it works in the ci-framework-testproject repo , check PR 224 - the latest build failed but it did for a different reason (tldr, dfg name is case sensitive in jump) - we can see there though that the polarion call/task got rendered ok:
`jump.py --testrun-id=create-new-test-run --xml-file=/home/zuul/ci-framework-data/tests/test_operator/tempest-tests/tempest_results.xml --update_testcases=true `

raukadah (approved):

PR #1535, opened 2024-04-22
Adding nfv ovs dpdk & sriov scenario's.

Jaganathancse:
This changes to add NFV ovs dpdk and sriov reprodicers scenario's.

As a pull request owner and reviewers, I checked that:
- [x] Appropriate testing is done and actually running
- [x] Appropriate documentation exists and/or is up-to-date:
  - [x] README in the role
  - [x] Content of the docs/source is reflecting the changes

PR #1706, opened 2024-05-15
Updating baremetal provision validations

Jaganathancse:
This change is to validate provisioning related variables based on pre provisiond or not.

  - [x] README in the role

### DIFF
--- a/roles/ci_gen_kustomize_values/defaults/main.yml
+++ b/roles/ci_gen_kustomize_values/defaults/main.yml
@@ -16,6 +16,7 @@
 
 # Top-level parameter shared with deploy_kustomize role
 cifmw_architecture_scenario: null
+cifw_bmh_pre_provisioned: true
 
 cifmw_ci_gen_kustomize_values_basedir: >-
   {{
--- a/roles/ci_gen_kustomize_values/tasks/edpm_core_asserts.yml
+++ b/roles/ci_gen_kustomize_values/tasks/edpm_core_asserts.yml
@@ -29,6 +29,7 @@
 - name: Ensure the required baremetal host parameters are configured.
   when:
     - cifmw_baremetal_hosts is defined
+    - cifw_bmh_pre_provisioned == false
   ansible.builtin.assert:
     that:
       - cifmw_ci_gen_kustomize_values_ctlplane_interface is defined


roles/ci_gen_kustomize_values/defaults/main.yml
```diff
@@ -16,6 +16,7 @@
 
 # Top-level parameter shared with deploy_kustomize role
 cifmw_architecture_scenario: null
+cifw_bmh_pre_provisioned: false
```
lewisdenny, line 19:
```suggestion
cifmw_bmh_pre_provisioned: true
```

lewisdenny: I think best to default to true are more jobs use pre_provisioned nodes (provided by libvirt_manager role)

So NFV job definition or vars file will need to be updated to add `cifmw_bmh_pre_provisioned: false`

lewisdenny (changes requested):

roles/ci_gen_kustomize_values/defaults/main.yml
Jaganathancse: @lewisdenny ack

Jaganathancse: > @lewisdenny ack

done

lewisdenny: From what I can see, `scenarios/reproducers/va-nfv-ovs-dpdk-sriov.yml` needs to have `cifw_bmh_pre_provisioned: false` added in this PR so current behavior is retained for SRIOV jobs.

Jaganathancse: @lewisdenny yes will add it once this PR merged.
also will update this PR also: https://github.com/openstack-k8s-operators/ci-framework/pull/1535

pablintino:
Can I ask for more context about why is this needed? I'd prefer to avoid adding another extra variable.

roles/ci_gen_kustomize_values/defaults/main.yml
lewisdenny: Okay sure, I guess the only thing is this validation won't run until your follow up patch merges but if you are okay with that than I am as well.

lewisdenny:
> Can I ask for more context about why is this needed? I'd prefer to avoid adding another extra variable.

The VA-NFV job uses preprovisioned VMs and provisions baremetal nodes, other VA jobs just use preprovisioned nodes via libvirt_manager role. 

The when clause is too broad and triggers on the HCI VA job for example, we only care about `cifmw_ci_gen_kustomize_values_ctlplane_interface` when provisioning nodes like in the VA-NFV job.

This is my understanding from chatting with Jaganathancse but I might be mistaken and would be open to a better idea. I'm not super happy about needed a new var but at this stage I need a quick simple fix so I can merge [1560](https://github.com/openstack-k8s-operators/ci-framework/pull/1560)

pablintino:
Entirely replaced by https://github.com/openstack-k8s-operators/ci-framework/pull/1719
Now the user should kustomize `ctlplaneInterface` like any other variable.

PR #1707, opened 2024-05-15
Introduce ci_lvms_storage role

fultonj:
The `ci_lvms_storage` role aims to be a drop in replacement for `ci_local_storage`. It uses LVMS (Logical Volume Manager Storage) based on the TopoLVM CSI driver to dynamically provision local storage built from block devices on OCP nodes.

This role requires `cifmw_lvms_disk_list` to contain a list of paths to block devices which already exist on OCP nodes. If the `cifmw_devscripts_config_overrides.vm_extradisks_list` list contains `"vda vdb"`, then it will create these block devices and `cifmw_lvms_disk_list` should be set to `['/dev/vda', '/dev/vdb']`. This role will then pass this list to the `deviceSelector` list used by the `LVMCluster` CRD.

This patch introduces the role but `ci_local_storage` is still used by the `06-deploy-{architecture,edpm}.yml` playbooks unless `cifmw_use_lvms` is true. The `ci_gen_kustomize_values` role has been updated to set the `storageClass` to `lvms-local-storage` only if `cifmw_use_lvms` is true. Otherwise it keeps its default `local-storage`. A similar change has been made to the `test_operator` role so that Tempest uses the chosen storage class.

In a follow up patch we can default `cifmw_use_lvms` to true so that LVMS becomes opt out. For now it is opt in for early testing.

Related: https://issues.redhat.com/browse/OSPRH-2749

  - [x] README in the role

### DIFF
--- a/roles/ci_gen_kustomize_values/README.md
+++ b/roles/ci_gen_kustomize_values/README.md
@@ -33,6 +33,8 @@ with a message.
 * `cifmw_ci_gen_kustomize_values_userdata_b64`: (List) Base64 encoded list of data to combine in the generated output.
   Defaults to `[]`.
 * `ci_gen_kustomize_fetch_ocp_state`: (Boolean) If true it enables generating CI templates based on the OCP state. Defaults to `true`.
+* `cifmw_ci_gen_kustomize_values_storage_class_prefix`: (String) Prefix for `storageClass` in generated values.yaml files. Defaults to `"lvms-"` only if `cifmw_use_lvms` is True, otherwise it defaults to `""`. The prefix is prepended to the `cifmw_ci_gen_kustomize_values_storage_class`. It is not recommended to override this value, instead set `cifmw_use_lvms` True or False.
+* `cifmw_ci_gen_kustomize_values_storage_class`: (String) Value for `storageClass` in generated values.yaml files. Defaults to `"lvms-local-storage"` only if `cifmw_use_lvms` is True, otherwise it defaults to `"local-storage"`.
 
 ### Specific parameters for edpm-values
 This configMap needs some more parameters in order to properly override the `architecture` provided one.
--- a/roles/ci_lvms_storage/README.md
+++ b/roles/ci_lvms_storage/README.md
@@ -1,0 +1,153 @@
+# ci_lvms_storage
+
+The `ci_lvms_storage` role aims to be a drop in replacement
+for `ci_local_storage`. It uses LVMS (Logical Volume Manager
+Storage) based on the TopoLVM CSI driver to dynamically
+provision local storage built from block devices on OCP
+nodes.
+
+This role requires `cifmw_lvms_disk_list` to contain a list of
+paths to block devices which already exist on OCP nodes. If
+the `cifmw_devscripts_config_overrides.vm_extradisks_list` list
+contains `"vda vdb"`, then the `devscripts` role will create these
+block devices and `cifmw_lvms_disk_list` should be set to
+`['/dev/vda', '/dev/vdb']`. This role will then pass the disk list
+to the `deviceSelector` list used by the `LVMCluster` CRD.
+
+## Privilege escalation
+
+It does not have any tasks which use become, but it does need to use
+`roles/ci_lvms_storage/templates/lvms-namespace.yaml.j2` to create
+create an OpenShift namespace where it then sets up a subscription
+as per `roles/ci_lvms_storage/templates/subscription.yaml.j2`. It
+uses the `cifmw_lvms_disk_list` list to find disks which it then wipes
+clean and adds to an LVMS cluster.
+
+## Parameters
+
+### ci-framework parameters
+
+* `cifmw_use_lvms`: (Boolean) Whether or not to use LVMS (default: `false`)
+
+If the ci-framework is called and `cifmw_use_lvms` is true, then
+the playbooks `06-deploy-architecture.yml` and `06-deploy-edpm.yml`
+call the `ci_lvms_storage` role to create a storage class called
+`lvms-local-storage` and the `ci_gen_kustomize_values` role will
+set the `storageClass` to `lvms-local-storage` in the generated
+values.yaml files used to build architecture CRs. The Tempest
+CR file, created by the `test_operator` role, will also set its
+`storageClass` value to `lvms-local-storage`.
+
+If the ci-framework is called and `cifmw_use_lvms` is false, then the
+playbooks `06-deploy-architecture.yml` and `06-deploy-edpm.yml`
+call the `ci_local_storage` role to create a storage class called
+`local-storage` and the `ci_gen_kustomize_values` role will set
+the `storageClass` to `local-storage` in the generated values.yaml
+files used to build architecture CRs. The Tempest CR file, created by
+the `test_operator` role, will also set its `storageClass` value to
+`local-storage`.
+
+* `cifmw_lvms_basedir`: (String) Installation base directory. Defaults to `cifmw_basedir` which defaults to `~/ci-framework-data`.
+* `cifmw_lvms_manifests_dir`: (String) Directory in where OCP manifests will be placed. Defaults to `"{{ cifmw_manifests | default(cifmw_cls_basedir ~ '/artifacts/manifests') }}/lvms"`
+
+### LVMCluster CRD Overrides
+
+* `cifmw_lvms_disk_list`: (List) A the list of pre-created block devices on OCP nodes used as PVs of the LVMS cluster.(default `[]`)
+* `cifmw_lvms_cluster_name`: (String) The `LVMCluster CRD` template override for the meta name (default `lvmcluster`)
+* `cifmw_lvms_force_wipe_devices_and_destroy_all_data`: (Boolean) The `LVMCluster CRD` template override for `deviceSelector` `forceWipeDevicesAndDestroyAllData` (default `true`)
+* `cifmw_lvms_fstype`: (String) The `LVMCluster CRD` template override for `deviceClasses` `fstype` (default `xfs`)
+* `cifmw_lvms_thin_pool_name`: (String) The `LVMCluster CRD` template override for `thinPoolConfig` name (default `cifmw_lvms_thin_pool`)
+* `cifmw_lvms_thin_pool_overprovision_ratio`: (Int) The `LVMCluster CRD` template override for `thinPoolConfig` overprovisionRatio (default `10`)
+* `cifmw_lvms_thin_pool_size_percent`: (Int) The `LVMCluster CRD` template override for `thinPoolConfig` sizePercent (default `90`)
+* `cifmw_lvms_storage_class`: (String) The `LVMCluster CRD` template override for the `deviceClasses` `name`. In this Ansible role it defaults to `local-storage` though the LVMS operator will prepend `lvms-` to this value, so the resultant default storage class will be `lvms-local-storage`.
+
+### Kubernetes parameters
+
+* `cifmw_lvms_namespace`: (String) The Kubernetes namespace where the LVMS cluster and operator pods will run (default `openshift-storage`)
+
+### kubernetes.core.k8s_info parameters
+
+After a Kubernetes manifest is placed in `cifmw_lvms_manifests_dir` by
+the `ansible.builtin.template` module, it is applied to the cluster
+with the `kubernetes.core.k8s` module. The `kubernetes.core.k8s_info`
+module then polls Kubernetes for the status of the created resource
+before the role goes on to apply other manifests. The polling can
+be controlled with the following parameters.
+
+* `cifmw_lvms_delay`: (Int) Ansible `delay` passed to tasks which wait for `kubernetes.core.k8s_info` (default `10`)
+* `cifmw_lvms_retries`: (Int) Ansible `retries` passed to tasks which wait for `kubernetes.core.k8s_info` (default `60`)
+
+## Examples
+
+The example playbook below will create an LVMS cluster using the disks
+from the `cifmw_lvms_disk_list` list. An example var for
+`cifmw_devscripts_config_overrides` is also set with a matching
+`vm_extradisks_list`. This parameter will result in the `devscripts`
+role (not called below) creating the devices when it provisions OCP.
+
+The playbook will use the `kubernetes.core.k8s_info` Ansible module
+to ensure the LVMS operator and storage cluster are ready and then
+report on their status. It will then create a test PVC and test pod
+to use the PVC. The last task in the playbook will remove the test
+resources as well as delete the entire LVMS cluster, operator and
+namespace.
+
+```yaml
+- name: LVMS playbook
+  gather_facts: false
+  hosts: all
+  vars:
+    cifmw_openshift_kubeconfig: "~/.kube/config"
+    cifmw_lvms_disk_list:
+      - /dev/vda
+      - /dev/vdb
+    cifmw_devscripts_config_overrides:
+      vm_extradisks: "true"
+      vm_extradisks_list: "vdb vda"
+      vm_extradisks_size: "10G"
+  tasks:
+    - name: Create Storage Class
+      ansible.builtin.include_role:
+        name: ci_lvms_storage
+
+    - name: Report on Storage Class
+      ansible.builtin.include_role:
+        name: ci_lvms_storage
+        tasks_from: status.yml
+
+    - name: Test the Storage Class
+      ansible.builtin.include_role:
+        name: ci_lvms_storage
+        tasks_from: test.yml
+
+    - name: Clean up Storage Class
+      ansible.builtin.include_role:
+        name: ci_lvms_storage
+        tasks_from: cleanup.yml
+```
+
+## Testing the `ci_lvms_storage` role
+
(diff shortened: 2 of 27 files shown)


pablintino:
@fultonj thanks for this contribution, as it's a critical work he are going to need really soon.

fultonj:
1. When `cifmw_lvms=false` (default), automated va1 deploys without an error (using local-storage as usual)
3. When `cifmw_lvms=true`, automated va1 deploys fine but the tempest test is blocked by https://issues.redhat.com/browse/OSPRH-7062 (testing pod requests RWX from LVMS, not an OpenStack on LVMS problem).

Given 1, there is no risk in merging this.

People can then easily test after this merges by setting `cifmw_lvms=true` and we can wait for fix from OSPRH-7062 before we default `cifmw_lvms=true`.

roles/ci_gen_kustomize_values/README.md
```diff
@@ -33,6 +33,8 @@ with a message.
 * `cifmw_ci_gen_kustomize_values_userdata_b64`: (List) Base64 encoded list of data to combine in the generated output.
   Defaults to `[]`.
 * `ci_gen_kustomize_fetch_ocp_state`: (Boolean) If true it enables generating CI templates based on the OCP state. Defaults to `true`.
+* `cifmw_ci_gen_kustomize_values_storage_class_prefix`: (String) Prefix for `storageClass` in generated values.yaml files. Defaults to `"lvms-"` only if `cifmw_lvms` is True, otherwise it defaults to `""`. The prefix is appended to the `cifmw_ci_gen_kustomize_values_storage_class`. It is not recommended to override this value, instead set `cifmw_lvms` True or False.
```
cescgina, line 36: do we have any use-case where we would want to change the prefix? If not we probably do not need to expose this as a parameter and just define it internally when `cifmw_lvms` is true

fmount:
+2 from me

roles/ci_lvms_storage/templates/lvms-cluster.yaml.j2
```diff
@@ -0,0 +1,22 @@
+---
+apiVersion: lvm.topolvm.io/v1alpha1
+kind: LVMCluster
+metadata:
+  name: lvmcluster
```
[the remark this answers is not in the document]
fultonj: I've deployed  >1 LVMS cluster on the same OpenShift by calling this role >1 time and overriding the `cifmw_lvms_namespace` each time. However, I didn't test them both at the same time. In that scenario you'd want to override the `cifmw_lvms_disk_list` each time too and it would make sense to override the name `lvmcluster` so I could make that a variable.

roles/ci_gen_kustomize_values/README.md
fultonj: We do need `cifmw_ci_gen_kustomize_values_storage_class`. Previously it was not there and `local-storage` was hard-coded in multiple spots.

I'm relying on its prefix variable to always being defined. Do you want to provide another way to solve the problem of getting the right storage class name?

Or do you think I should not document the variable in the readme though I'd keep it in the defaults file? I could prefix it with a `_` to imply that it's private while at the same time keeping it. We could also spend a little time using some other method of getting the name we want. It's mainly there in order to have one flag control different outcomes in multiple roles.

cescgina: I think it's fine just not documenting it in the README if we don't expect users to need to set it. I see why it's needed, and the current approach makes sense to me, I just wanted to clarify if we expected it to have a value different than the default, thanks for the explanation

roles/ci_lvms_storage/README.md
```diff
@@ -0,0 +1,152 @@
...
+paths to block devices which already exist on OCP nodes. If
+the `cifmw_devscripts_config_overrides.vm_extradisks_list` list
+contains `"vda vdb"`, then the `devscripts` role will create these
+block devices and `cifmw_lvms_disk_list` should be set to
+`['/dev/vda', '/dev/vdb']`. This role will then pass the disk list
+to the `deviceSelector` list used by the `LVMCluster` CRD.
+
+## Privilege escalation
+
+It does have any tasks which use become, but it does need to use
```
cescgina, line 19: Should this read `It does not have any tasks` ? I find it confusing otherwise

roles/ci_gen_kustomize_values/README.md
fultonj: we do have the prefix pattern already...

- cifmw_ci_gen_kustomize_values_snippets_dir_prefix
- cifmw_ci_gen_kustomize_values_dest_fname_prefix

so I thought it better to be transparent about how it works, especially if someone uses `cifmw_ci_gen_kustomize_values_storage_class` and is surprised to see a prefix there.

roles/ci_lvms_storage/README.md
fultonj: fixed.

roles/ci_lvms_storage/templates/lvms-cluster.yaml.j2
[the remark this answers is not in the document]
fultonj: fixed

https://github.com/openstack-k8s-operators/ci-framework/compare/ce29d5f1c119885a3f605f70d7fc36737417ef97..3054c17e7a4411b885ed79c4948755d56ebf0589

roles/ci_gen_kustomize_values/README.md
cescgina: that's a good point about transparency, and the explanation is thorough, I'm not opposed to keep it as it is now, just wanted to clarify

fultonj: https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/ci_gen_kustomize_values/README.md?plain=1#L23-L28

fultonj: OK, I think if we're going to keep the prefix we should document it for completeness.

fultonj:
> 2. When `cifmw_lvms=true`, automated va1 deploys fine but the tempest test is blocked by https://issues.redhat.com/browse/OSPRH-7062 (testing pod requests RWX from LVMS, not an OpenStack on LVMS problem).
> People can then easily test after this merges by setting `cifmw_lvms=true` and we can wait for fix from [OSPRH-7062](https://issues.redhat.com//browse/OSPRH-7062) before we default `cifmw_lvms=true`.

If you use `quay.io/openstack-k8s-operators/test-operator:latest` then you will not have this issue.

The latest test-operator on quay has fix https://github.com/openstack-k8s-operators/test-operator/pull/96 

I did a fully automated VA1 w/ tempest and LVMS and did not reproduce OSPRH-7062.

```
[zuul@controller-0 ~]$ oc get pvc
NAME                                              STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS         AGE
glance-conversion-glance-default-external-api-0   Bound    pvc-5f1d35f0-ac14-4964-b6f1-148345fe5393   10Gi       RWO            lvms-local-storage   13m
<snip>
persistence-rabbitmq-server-2                     Bound    pvc-26768eed-9f6b-4b6a-bc93-01c003b27720   10Gi       RWO            lvms-local-storage   33m
tempest-tests-38419                               Bound    pvc-52d4ef53-dfe4-47b7-8cde-d5ce929ab271   1Gi        RWO            lvms-local-storage   2m44s
[zuul@controller-0 ~]$ oc get pods | grep tempest
tempest-tests-ddxsr                                               1/1     Running     0             3m36s
[zuul@controller-0 ~]$ oc get sc
NAME                           PROVISIONER   RECLAIMPOLICY   VOLUMEBINDINGMODE      ALLOWVOLUMEEXPANSION   AGE
lvms-local-storage (default)   topolvm.io    Delete          WaitForFirstConsumer   true                   36m
[zuul@controller-0 ~]$ 
```

fultonj:
I recommend we merge this as (per my testing) it doesn't break anything. Then we can ask people to test (including tempest) by setting `cifmw_lvms=true`. Once there is enough comfort we can default `cifmw_lvms` true (for this PR can stay false).

fmount (commented):
/lgtm it can be merged and testing can be triggered via `cifmw_lvms=true` until we switch to `LVMS` as the default `storageClass`

roles/ci_gen_kustomize_values/README.md
lewisdenny, line 36:
```suggestion
* `cifmw_ci_gen_kustomize_values_storage_class_prefix`: (String) Prefix for `storageClass` in generated values.yaml files. Defaults to `"lvms-"` only if `cifmw_lvms` is True, otherwise it defaults to `""`. The prefix is prepended to the `cifmw_ci_gen_kustomize_values_storage_class`. It is not recommended to override this value, instead set `cifmw_lvms` True or False.
```

roles/ci_lvms_storage/README.md
```diff
@@ -0,0 +1,153 @@
...
+create an OpenShift namespace where it then sets up a subscription
+as per `roles/ci_lvms_storage/templates/subscription.yaml.j2`. It
+uses the `cifmw_lvms_disk_list` list to find disks which it then wipes
+clean and adds to an LVMS cluster.
+
+## Parameters
+
+### ci-framework parameters
+
+* `cifmw_lvms`: (Boolean) Whether or not to use LVMS (default: `false`)
```
lewisdenny, line 30: This should be added to `docs/source/usage/01_usage.md`

lewisdenny: [Suggestion] Rather than using a bool for this behavior it would be clearer to use  something like one of these suggestions:
- cifmw_storage_class: "lvms-local-storage" | cifmw_storage_class: "local-storage"
- cifmw_storage_class_provisioner: "ci_lvms_storage" | cifmw_storage_class_provisioner: "ci_local_storage"

roles/ci_lvms_storage/defaults/main.yml
```diff
@@ -0,0 +1,33 @@
...
+
+# All variables intended for modification should be placed in this file.
+# All variables within this role should have a prefix of "cifmw_lvms"
+
+cifmw_lvms_disk_list: []
+cifmw_lvms_cluster_name: lvmcluster
+cifmw_lvms_namespace: openshift-storage
+cifmw_lvms_basedir: "{{ cifmw_basedir | default(ansible_user_dir ~ '/ci-framework-data') }}"
+cifmw_lvms_manifests_dir: "{{ cifmw_manifests | default(cifmw_lvms_basedir ~ '/artifacts/manifests') }}/lvms"
+# The "lvms-" prefix is added to the cifmw_lvms_storage_class by the lvm-operator
```
lewisdenny, line 25:
```suggestion
# The "lvms-" prefix is prepended to the cifmw_lvms_storage_class by the lvm-operator
```

roles/ci_lvms_storage/files/.gitkeep
lewisdenny, line 1: [nit] Is this needed?

roles/ci_lvms_storage/handlers/main.yml
```diff
@@ -0,0 +1,15 @@
+---
```
lewisdenny, line 1: [nit] Please remove this file

roles/ci_lvms_storage/vars/main.yml
```diff
@@ -0,0 +1,22 @@
+---
```
lewisdenny, line 1: [nit] Please remove this file

roles/test_operator/README.md
```diff
@@ -18,6 +18,8 @@ Execute tests via the [test-operator](https://openstack-k8s-operators.github.io/
 * `cifmw_test_operator_controller_ip`: (String) An ip address of the controller node. Default value: `ansible_default_ipv4.address` which defaults to (`""`).
 * `cifmw_test_operator_tolerations`: (Dict) `tolerations` value that is applied to all pods spawned by the test-operator and to the test-operator-controller-manager and test-operator-logs pods. Default value: `{}`
 * `cifmw_test_operator_node_selector`: (Dict) `nodeSelector` value that is applied to all pods spawned by the test-operator and to the test-operator-controller-manager and test-operator-logs pods. Default value: `{}`
+* `cifmw_test_operator_storage_class_prefix`: (String) Prefix for `storageClass` in generated Tempest CRD file. Defaults to `"lvms-"` only if `cifmw_lvms` is True, otherwise it defaults to `""`. The prefix is appended to the `cifmw_test_operator_storage_class`. It is not recommended to override this value, instead set `cifmw_lvms` True or False.
```
lewisdenny, line 21:
```suggestion
* `cifmw_test_operator_storage_class_prefix`: (String) Prefix for `storageClass` in generated Tempest CRD file. Defaults to `"lvms-"` only if `cifmw_lvms` is True, otherwise it defaults to `""`. The prefix is prepended to the `cifmw_test_operator_storage_class`. It is not recommended to override this value, instead set `cifmw_lvms` True or False.
```

lewisdenny (changes requested):
Thank you for the great role and fantastic documentation!

lewisdenny:
/hold
I'm holding this for now while I clarify if we can merge this during beta testing.

@viroel @pablintino @rlandy This doesn't change the default behavior, IMO it's safe but I wanted to confirm.

pablintino:
> /hold I'm holding this for now while I clarify if we can merge this during beta testing.
> 
> @viroel @pablintino @rlandy This doesn't change the default behavior, IMO it's safe but I wanted to confirm.

Hey, this one is in my TODO list, I think it should be safe, but I'd prefer to hold it till @viroel has a full green pipeline for beta.

lewisdenny:
> > /hold I'm holding this for now while I clarify if we can merge this during beta testing.
> > @viroel @pablintino @rlandy This doesn't change the default behavior, IMO it's safe but I wanted to confirm.
> 
> Hey, this one is in my TODO list, I think it should be safe, but I'd prefer to hold it till @viroel has a full green pipeline for beta.

Perfect, glad I held and checked :)

roles/test_operator/README.md
fultonj: done

roles/ci_lvms_storage/handlers/main.yml
fultonj: removed

roles/ci_lvms_storage/files/.gitkeep
fultonj: removed

roles/ci_lvms_storage/defaults/main.yml
fultonj: done.

roles/ci_lvms_storage/README.md
fultonj: > This should be added to `docs/source/usage/01_usage.md`

done.

roles/ci_gen_kustomize_values/README.md
fultonj: done

roles/ci_lvms_storage/README.md
fultonj: > [Suggestion] Rather than using a bool for this behavior it would be clearer to use something like one of these suggestions:
> 
> * cifmw_storage_class: "lvms-local-storage" | cifmw_storage_class: "local-storage"
> * cifmw_storage_class_provisioner: "ci_lvms_storage" | cifmw_storage_class_provisioner: "ci_local_storage"

It's the LVMS operator which is prepending the prefix. I.e. a prefix of `"lvms-"` is prepended on whatever variable I pass for `cifmw_storage_class` by the LVMs operator. So I default the variable to `local-storage` and then it creates a storage class called `lvms-local-storage`.

  https://github.com/fultonj/ci-framework/blob/ci_lvms_storage/roles/ci_lvms_storage/defaults/main.yml#L26

I do want a single boolean to make it easy for people to use the new feature or not.

cjeanner, line 30: hm, late to the party, sorry - but: to match other "boolean" like that, I'd rather go with `cifmw_use_lvms` - that would match `cifmw_use_libvirt`, `cifmw_use_devscripts` and some others we already have through the framework.

Would it make sense? It would require a mighty `sed` in this PR...

cjeanner:
@fultonj care to run `make role_molecule` ? I think some automated steps were missed while creating the new role, leading to a missing "noop" configuration for molecule jobs in CI (needed for job dependency support - and I see I probably was too strict on the files triggering the zuul config check -.-' ).

Thank you!

roles/ci_lvms_storage/README.md
fultonj: I've made the update and we now have boolean variable `cifmw_use_lvms` in place of `cifmw_lvms`

cjeanner:
/approve

This change doesn't change the default behavior - so once we're good with the DS side of the CI (midstream and al) we should be able to merge.

roles/ci_lvms_storage/README.md
fultonj: This actually caused a problem.

I need this var to be in `reproducer-variables.yml` on the controller node it didn't get there because of this.

https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/reproducer/tasks/configure_controller.yml#L358

fultonj: I have a plan to fix it tracked in https://issues.redhat.com/browse/OSPRH-7424

PR #1708, opened 2024-05-15
Ensure we clean dnsmasq lease file upon cleanup

cjeanner:
Else, we hit an issue where dnsmasq won't give an IP because its lease
is still running.


lewisdenny (approved):
/lgtm

PR #1711, opened 2024-05-16
Allow to enable DNS listener in dnsmasq

cjeanner:
It might happen we want DNS on specific address, usually one associated
to a specific DHCP network.

Using the new parameter, we allow to pass a list of listeners, per
network. Dnsmasq will then listen and answer to DNS queries on those
addresses.

Note that you want to pass dhcp-options to ensure the DNS resolver is
properly passed, for instance:

```YAML
options:
  - "option:dns-server,192.168.0.1"
```
in the network definition.

  - [X] README in the role


hjensas (approved):
Looks good to me.

PR #1717, opened 2024-05-16
Prepare networking as early as possible

cjeanner:
In order to change how we provision networking, especially DHCP, we have
to get the virtual networks as early as possible in place.

Doing so, we'll be able to inject some specific VM, such as the nat64,
and properly manage networks, even for devscripts layout.

@cjeanner testing notes:
- [X] CRC layout is passing fine, networks are present and working as expected
- [X] 3-OCP layout is passing fine as well, networks are present and working as expected

### DIFF
--- a/roles/devscripts/tasks/133_host_network.yml
+++ b/roles/devscripts/tasks/133_host_network.yml
@@ -56,49 +56,6 @@
         list
       }}
   block:
-    - name: Ensure firewalld is installed
-      ansible.builtin.package:
-        name:
-          - firewalld
-
-    - name: Ensure br_netfilter module is loaded.
-      community.general.modprobe:
-        name: br_netfilter
-        state: present
-
-    - name: Ensure IP forwarding is enabled.
-      ansible.posix.sysctl:
-        name: net.ipv4.ip_forward
-        value: 1
-        state: present
-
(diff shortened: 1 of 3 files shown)


lewisdenny (commented):
If I understand correctly the tasks that we're copied from `roles/devscripts/tasks/133_host_network.yml` into `roles/reproducer/tasks/prepare_networking.yml` were protected by a when clause:
```
    - cifmw_devscripts_config.manage_br_bridge is defined
    - cifmw_devscripts_config.manage_br_bridge == 'n'
```

Is this no longer needed or being applied somewhere else?

cjeanner:
> If I understand correctly the tasks that we're copied from `roles/devscripts/tasks/133_host_network.yml` into `roles/reproducer/tasks/prepare_networking.yml` were protected by a when clause:
> 
> ```
>     - cifmw_devscripts_config.manage_br_bridge is defined
>     - cifmw_devscripts_config.manage_br_bridge == 'n'
> ```
> 
> Is this no longer needed or being applied somewhere else?

It was never really tested in molecule, and I happened to trigger that path by passing that specific override in the molecule test, leading to.... well.. "some" issues with that untested track, mostly due to lacking data that are usually fed by a job definition.... So I changed the override to keep it simple in the other PR.

PR #1718, opened 2024-05-16
Fix polarion role for "update test runs task"

son-vyas:
The task is encountering an issue because it expects a list of objects with a path attribute, but instead, it is receiving a
string.
The map(attribute='path') filter is being applied to a string, which causes the error.

To resolve this, ensure that xml_files.files is a list of dictionaries, each containing a path key.

  - [x] README in the role


arxcruz (commented):
https://github.com/openstack-k8s-operators/ci-framework/pull/1705/ fixes it, this patch is no longer required

PR #1720, opened 2024-05-16
[run_hook] Fixed `crd_path` content and added CR hook example

rebtoor:
* `role_path` is an ansible magic variable rather than a module, so calling it with `ansible.builtin.` prefix doesn't work as expected

* A proper example of CR hook has been added in the `README.md` file

  - [x] README in the role

### DIFF
--- a/roles/run_hook/tasks/cr.yml
+++ b/roles/run_hook/tasks/cr.yml
@@ -23,7 +23,7 @@
       {%- if hook.source is not ansible.builtin.abs -%}
       {{
         (
-          ansible.builtin.role_path,
+          role_path,
           '../../hooks/crs',
           hook.source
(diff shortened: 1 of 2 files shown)


roles/run_hook/tasks/cr.yml
```diff
@@ -23,8 +23,8 @@
       {%- if hook.source is not ansible.builtin.abs -%}
       {{
         (
-          ansible.builtin.role_path,
-          '../../hooks/crs',
+          role_path,
+          '../../hooks/crs/',
```
cjeanner, line 27: no need to `/` at the end, we're using `path_join` filter on purpose :).

cjeanner:
/approve

Thanks for the proper example addition!

cjeanner:
/approve

... let's click the right button now?

PR #1721, opened 2024-05-16
Allow to set fixed DNS

cjeanner:
  - [X] README in the role


cjeanner:
will work differently - closing.

PR #1722, opened 2024-05-16
Allow to patch devscripts config overrides

cjeanner:
The Framework (or a job) might need to inject specific configurations in
devscripts.

This pach proposes an interface to inject as many configuration snippets
as needed - in the exact same way other roles expose such feature.

As the molecule test has been updated, it now hits a specific path in
the deployment that wasn't tested before. In molecule, it leads to a
failure, because `firewalld` isn't present.
This patch also ensures this package is installed.

We also expose the "build devscripts config" as a dedicated tasks file,
allowing to get it "on-demand", and avoid code duplication:
it might happen someone wants to just know if there's an OCP cluster or
not, so patching the 110_check_ocp.yml to call the "config builder" if
needed allows to call it directly.

Added fix:
this patch also fix a missed renamed from `_layout` to
`_cifmw_libvirt_manager_layout`.

  - [X] README in the role

### DIFF
--- a/roles/devscripts/tasks/133_host_network.yml
+++ b/roles/devscripts/tasks/133_host_network.yml
@@ -56,6 +56,11 @@
         list
       }}
   block:
+    - name: Ensure firewalld is installed
+      ansible.builtin.package:
+        name:
+          - firewalld
+
     - name: Ensure br_netfilter module is loaded.
       community.general.modprobe:
         name: br_netfilter
@@ -88,7 +93,7 @@
     - name: Create the virtual networks.
       vars:
         cifmw_libvirt_manager_net_prefix_add: false
-        _layout:
+        _cifmw_libvirt_manager_layout:
           networks: "{{ cifmw_libvirt_manager_configuration.networks }}"
       ansible.builtin.include_role:
         name: libvirt_manager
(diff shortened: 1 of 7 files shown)


cjeanner:
recheck

cjeanner:
recheck

cjeanner:
recheck

roles/devscripts/tasks/133_host_network.yml
```diff
@@ -88,7 +93,7 @@
     - name: Create the virtual networks.
       vars:
         cifmw_libvirt_manager_net_prefix_add: false
-        _layout:
+        _cifmw_libvirt_manager_layout:
```
cjeanner, line 96: @psathyan apparently I missed that one while working on the renaming - not sure how to get through that path though... IIRC that's for hybrid?

roles/devscripts/tasks/133_host_network.yml
```diff
@@ -56,6 +56,11 @@
         list
       }}
   block:
+    - name: Ensure firewalld is installed
```
cjeanner, line 59: that one was needed - in molecule I was changing the value of the `manage_br_bridge` and hit a failure in the molecule job because nothing installs firewalld.

Better install it twice than never - this is why it's also in this patch.

roles/devscripts/tasks/133_host_network.yml
psathyan: This path is taken when user has choosen to set `manage_br_bridge: 'n'` ... generally the path is taken during hybrid scenarios for ensuring a physical port is attached to the bridge.

Thanks @cjeanner for fixing it. With the latest patch set, I did hit the issue with public networks.

psathyan (approved):

lewisdenny (approved):
/lgtm

PR #1725, opened 2024-05-17
Set nfs network vars from ipset

rabi:

jira: https://issues.redhat.com/browse/OSPRH-7107

### DIFF
--- a/playbooks/nfs.yml
+++ b/playbooks/nfs.yml
@@ -26,6 +26,15 @@
     nftables_path: /etc/nftables
     nftables_conf: /etc/sysconfig/nftables.conf
   tasks:
+    - name: Set custom cifmw PATH reusable fact
+      tags:
+        - always
+      when:
+        - cifmw_path is not defined
+      ansible.builtin.set_fact:
+        cifmw_path: "{{ ansible_user_dir }}/.crc/bin:{{ ansible_user_dir }}/.crc/bin/oc:{{ ansible_user_dir }}/bin:{{ ansible_env.PATH }}"
+        cacheable: true
+
     - name: Install required packages
       ansible.builtin.package:
         name:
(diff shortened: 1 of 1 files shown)


playbooks/nfs.yml
```diff
@@ -26,6 +26,15 @@
     nftables_path: /etc/nftables
     nftables_conf: /etc/sysconfig/nftables.conf
   tasks:
+    - name: Set custom cifmw PATH reusable fact
+      tags:
+        - always
+      when:
+        - cifmw_path is not defined
+      ansible.builtin.set_fact:
+        cifmw_path: "{{ ansible_user_dir }}/.crc/bin:{{ ansible_user_dir }}/.crc/bin/oc:{{ ansible_user_dir }}/bin:{{ ansible_env.PATH }}"
```
marios, line 35: so this is going to be used somewhere that is not running the bootstrap play https://github.com/openstack-k8s-operators/ci-framework/blob/89bdefa600aa5c6b4c86fc070d4656899efed094/playbooks/01-bootstrap.yml#L11 
?

marios (commented):
code seems reasonable but commit message is bare so hard to understand why/where this is used

playbooks/nfs.yml
rabi: This playbook/hook runs on compute hosts[1] and not on ansible controller and this var is not available for it.

[1] https://github.com/openstack-k8s-operators/ci-framework/blob/main/playbooks/nfs.yml#L19

playbooks/nfs.yml
```diff
@@ -56,19 +65,31 @@
...
+      environment:
+        KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+        PATH: "{{ cifmw_path }}"
+      vars:
+        _nfs_network_name: "{{ cifmw_nfs_network | default('storage') }}"
+        _nfs_host: "{{ [groups[cifmw_nfs_target | default('computes')][0], ansible_domain] | join('.') | default('') }}"
+        _ipset_namespace: "{{ cifmw_install_yamls_defaults['NAMESPACE'] | default('openstack') }}"
+      ansible.builtin.command:
+        cmd: oc get ipset {{ _nfs_host }} -n {{ _ipset_namespace }} -o jsonpath='{.status.reservations[?(@.network=="{{ _nfs_network_name }}")].address}'
+      register: cifmw_nfs_ip_out
```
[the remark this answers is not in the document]
rabi: I assume you're suggesting to make mutiple api calls like now but in a loop. I would prefer to keep it multiple tasks rather than using indexes from results to avoid complexity. This is part of a hook only used when using nfs backend for glance and could be optimized, if required later.

lewisdenny (changes requested):
Thank you @rabi, this looks like a nice improvement. Please update the commit message with why this change is needed, a few lines about the change and I would be happy to add my +1

rabi:
> Thank you @rabi, this looks like a nice improvement. Please update the commit message with why this change is needed, a few lines about the change and I would be happy to add my +1

This is not an improvement but a bug fixed for nfs job. You can't assume ips for the nodes from the IPAM to be predictable like we were doing before. Someone should create jira for the nfs issue. I was pinged about the issue by CI team without a jira.

rabi:
> > Thank you @rabi, this looks like a nice improvement. Please update the commit message with why this change is needed, a few lines about the change and I would be happy to add my +1
> 
> This is not an improvement but a bug fixed for nfs job. You can't assume ips for the nodes from the IPAM to be predictable like we were doing before. Someone should create jira for the nfs issue. I was pinged about the issue by CI team with a jira.

Here is the jira: https://issues.redhat.com/browse/OSPRH-7107

lewisdenny (approved):
/lgtm

Great, thanks for the context Rabi

bshephar (approved):

playbooks/nfs.yml
```diff
@@ -57,18 +66,17 @@
...
-            crc_ci_bootstrap_networking.networks[_nfs_network_name].range
-          }}
-        cifmw_nfs_ip: >-
-          {{
-            crc_ci_bootstrap_networking.instances[_nfs_host].networks[_nfs_network_name].ip
-          }}
+        _nfs_host: "{{ [groups[cifmw_nfs_target | default('computes')][0], ansible_domain] | join('.') | default('') }}"
+        _ipset_namespace: "{{ cifmw_install_yamls_defaults['NAMESPACE'] | default('openstack') }}"
+      ansible.builtin.command:
+        cmd: oc get ipset {{ _nfs_host }} -n {{ _ipset_namespace }} -o jsonpath='{.status.reservations[?(@.network=="{{ _nfs_network_name }}")]}'
```
bshephar, line 78: Very nice.

bshephar (approved):

PR #1726, opened 2024-05-17
[networking_mapper] Allow to speed up run

cjeanner:
Introduces a new parameter to the role, to *not* gather networking facts
under certain circumstances.

This is for advanced usage only, and shouldn't be changed in most of the
cases.

  - [X] README in the role


cjeanner:
recheck

PR #1728, opened 2024-05-17
Trick ansible in loading mock data

cjeanner:
Fetch the remote mock file, inject it locally, and call include_vars to
load the mocked data.


cjeanner:
/hold

having some issues with the fact loading - need to debug that using the architecture repo patch.

cjeanner:
Got green on the architecture repository with a mock file!

PR #1729, opened 2024-05-18
[shiftstack] Templates for rendering the right values to be used in the DT

psathyan:
Adding the templates required for rendering the config map that will be used for deploying OpenStack Services.

*edpm-nodeset-values*
- key difference from common is the support for external network's fixed IP address incase of Ceph nodes.

*network-values*
- key difference from common is the support for Octavia bridge

*logs*
https://review.rdoproject.org/zuul/buildset/771618eb32c94bb99b1b0b11025389d9


rlobillo:
Is this change strictly related to titan03? I think it is not right? Can we put this patch on review and reduce the dependencies?

PR #1733, opened 2024-05-20
Use dlrnapi repo-status for finding component reported jobs

raukadah:
We report the job status by using aggregate hash in integration line but in component jobs, there is no aggregate hash. We use commit, distro and extended hash to report the job status in component pipeline.

We need to use same distro, commit and extended hash to fetch the status of reported job to match the criteria otherwise agg_hash will alwats return empty.

This pr seperates the checking of repos jobs for component and integration lines.


rlandy (commented):
/lgtm

PR #1738, opened 2024-05-21
Update the repo_setup role readme

raukadah:


lewisdenny (commented):
Thank you @raukadah, this is a big improvement, could you please mention this role installs repo_setup from `cifmw_repo_setup_src` and expand on the relationship with that repo? For example why is there also a repo_setup role [1]

[1] https://github.com/openstack-k8s-operators/repo-setup/tree/main/roles/repo_setup

raukadah:
> Thank you @raukadah, this is a big improvement, could you please mention this role installs repo_setup from `cifmw_repo_setup_src` and expand on the relationship with that repo? For example why is there also a repo_setup role [1]
> 
> [1] https://github.com/openstack-k8s-operators/repo-setup/tree/main/roles/repo_setup

Hey @lewisdenny , thank you for reviewing the pr. repo-setup/roles/repo_setup is an old import from tripleo. It is never used. We rewrote the whole tooling based on our needs. https://github.com/openstack-k8s-operators/repo-setup/pull/25 will drop the old role.

I will update the role how repo-setup uiltity is getting installed.

PR #1741, opened 2024-05-21
[dnsmasq] Refactor the role to be consumed by the `ci_network` role

rebtoor:
Since the `ci_network` role has already its own task(s) to manage dns configuration, this PR will try to fill the gap in terms of functionality between both roles, so in the future `dnsmasq` role can be consumed by `ci_network` role with low effort. 

  - [x] README in the role

### DIFF
--- a/roles/dnsmasq/README.md
+++ b/roles/dnsmasq/README.md
@@ -29,11 +29,11 @@ supported in libvirt).
 * `cifmw_dnsmasq_listen_addresses`: (List) List of IP addresses on which dnsmasq should be enabled. Defaults to `[]`.
 * `cifmw_dnsmasq_enable_dns`: (Bool) Toggle to enable DNS features of dnsmasq. Defaults to `false`.
 * `cifmw_dnsmasq_addresses`: (List) Specify a list of IP addresses to return for any host in the given domains. Defaults to `[]`.
-* `cifmw_dnsmasq_domain_name`: (String) DNS domain name used for the DHCP server. Defaults to `''`.
 
 #### Ranges mapping
 
 * `label`: (String) Network label ("tag" in dnsmasq manual).
+* `domain`: (String) domain name associated to the dhcp range.
 * `start_v4`: (String) IPv4 starting IP.
 * `start_v6`: (String) IPv6 starting IP.
 * `prefix_length_v4`: (Int) IPv4 prefix length. Defaults to `24`.
--- a/roles/dnsmasq/README.md
+++ b/roles/dnsmasq/README.md
@@ -27,7 +27,7 @@ supported in libvirt).
 * `cifmw_dnsmasq_listen_addresses`: (List) List of IP addresses on which dnsmasq should be enabled. Defaults to `[]`.
 * `cifmw_dnsmasq_enable_dns`: (Bool) Toggle to enable DNS features of dnsmasq. Defaults to `false`.
 * `cifmw_dnsmasq_addresses`: (List) Specify a list of IP addresses to return for any host in the given domains. Defaults to `[]`.
-* `cifmw_dnsmasq_domainname`: (String) DNS domain name used for the DHCP server. Defaults to `''`.
+* `cifmw_dnsmasq_domain_name`: (String) DNS domain name used for the DHCP server. Defaults to `''`.
 
 #### Ranges mapping
 
--- a/roles/dnsmasq/README.md
+++ b/roles/dnsmasq/README.md
@@ -16,7 +16,7 @@ supported in libvirt).
 * `cifmw_dnsmasq_basedir`: (String) Configuration directory location. Defaults to `/etc/cifmw-dnsmasq.d`.
 * `cifmw_dnsmasq_dns_config_file`: (String) DNS related settings configuration file path. Defaults to `{{ cifmw_dnsmasq_basedir }}/dns.conf`.
 * `cifmw_dnsmasq_listener_config_file`: (String) listener related settings configuration file path. Defaults to `{{ cifmw_dnsmasq_basedir }}/listener.conf`.
-* `cifmw_dnsmasq_global_options`: (Dict) Global options for dnsmasq. Defaults to `{}`.
+* `cifmw_dnsmasq_raw_config`: (String) Raw configure options for dnsmasq. Should be passed as a (multiline) string. Defaults to `""`.
 
 ### New network parameters
 
--- a/roles/dnsmasq/README.md
+++ b/roles/dnsmasq/README.md
@@ -22,7 +22,12 @@ supported in libvirt).
 * `cifmw_dnsmasq_network_state`: (String) Network status. Must be either `present` or `absent`.
 * `cifmw_dnsmasq_network_definition`: (Dict) Mapping representing the network definition.
 * `cifmw_dnsmasq_network_definition.ranges`: (List[mapping]) List of ranges associated to the network.
-* `cifmw_dnsmasq_network_listen_dns`: (List) List of IP addresses to listen to for DNS queries.
+* `cifmw_dnsmasq_forwarders`: (List) List of upstream DNS servers used as forwarders. Defaults to `[]`
+* `cifmw_dnsmasq_interfaces`: (List) List of interfaces on which dnsmasq should be enabled. Defaults to `[]`.
+* `cifmw_dnsmasq_listen_addresses`: (List) List of IP addresses on which dnsmasq should be enabled. Defaults to `[]`.
+* `cifmw_dnsmasq_enable_dns`: (Bool) Toggle to enable DNS features of dnsmasq. Defaults to `false`.
+* `cifmw_dnsmasq_addresses`: (List) Specify a list of IP addresses to return for any host in the given domains. Defaults to `[]`.
+* `cifmw_dnsmasq_domainname`: (String) DNS domain name used for the DHCP server. Defaults to `''`.
 
 #### Ranges mapping
 
--- a/roles/dnsmasq/README.md
+++ b/roles/dnsmasq/README.md
@@ -14,6 +14,7 @@ supported in libvirt).
 ## Common Parameters
 
 * `cifmw_dnsmasq_basedir`: (String) Configuration directory location. Defaults to `/etc/cifmw-dnsmasq.d`.
+* `cifmw_dnsmasq_enable_dns`: (Bool) Toggle to enable DNS features of dnsmasq. Defaults to `false`.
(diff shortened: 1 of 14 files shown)


roles/dnsmasq/molecule/default/converge.yml
```diff
@@ -18,6 +18,17 @@
...
+    cifmw_dnsmasq_interfaces:
+      - mol-test-1
+    cifmw_dnsmasq_addresses:
+      - fqdn: "api.ocp.openstack.lab"
+        address: "192.168.122.2"
+      - fqdn: ".apps.ocp.openstack.lab"
+        address: "192.168.122.3"
+    cifmw_dnsmasq_listen_addresses:
+      - "192.168.254.9"
+    cifmw_dnsmasq_domainname: "ocp.openstack.lab"
```
cjeanner, line 31: this reads better maybe?
```suggestion
    cifmw_dnsmasq_domain_name: "ocp.openstack.lab"
```

roles/dnsmasq/defaults/main.yml
```diff
@@ -19,3 +19,9 @@
 
 cifmw_dnsmasq_basedir: "/etc/cifmw-dnsmasq.d"
 cifmw_dnsmasq_global_options: {}
+cifmw_dnsmasq_forwarders: []
+cifmw_dnsmasq_interfaces: []
+cifmw_dnsmasq_listen_addresses: []
+cifmw_dnsmasq_enable_dns: false
+cifmw_dnsmasq_addresses: []
+cifmw_dnsmasq_domainname: ''
```
cjeanner, line 27: this ready better maybe?
```suggestion
cifmw_dnsmasq_domain_name: ''
```

roles/dnsmasq/README.md
```diff
@@ -22,7 +22,12 @@ supported in libvirt).
 * `cifmw_dnsmasq_network_state`: (String) Network status. Must be either `present` or `absent`.
 * `cifmw_dnsmasq_network_definition`: (Dict) Mapping representing the network definition.
 * `cifmw_dnsmasq_network_definition.ranges`: (List[mapping]) List of ranges associated to the network.
-* `cifmw_dnsmasq_network_listen_dns`: (List) List of IP addresses to listen to for DNS queries.
+* `cifmw_dnsmasq_forwarders`: (List) List of upstream DNS servers used as forwarders. Defaults to `[]`
+* `cifmw_dnsmasq_interfaces`: (List) List of interfaces on which dnsmasq should be enabled. Defaults to `[]`.
+* `cifmw_dnsmasq_listen_addresses`: (List) List of IP addresses on which dnsmasq should be enabled. Defaults to `[]`.
+* `cifmw_dnsmasq_enable_dns`: (Bool) Toggle to enable DNS features of dnsmasq. Defaults to `false`.
+* `cifmw_dnsmasq_addresses`: (List) Specify a list of IP addresses to return for any host in the given domains. Defaults to `[]`.
+* `cifmw_dnsmasq_domainname`: (String) DNS domain name used for the DHCP server. Defaults to `''`.
```
cjeanner, line 30: this reads better maybe?
```suggestion
* `cifmw_dnsmasq_domain_name`: (String) DNS domain name used for the DHCP server. Defaults to `''`.
```

roles/dnsmasq/molecule/default/converge.yml
```diff
@@ -100,18 +111,9 @@
              'artifacts') | path_join
           }}
         fname: "{{ item | basename }}"
```
cjeanner, line 113: I think `| basename` acts weirdly in ansible when applied to a directory:
```
The result of this filter is different from the Unix basename program;
where basename for /foo/bar/ returns bar, the basename filter returns an empty string ('').
```

[source](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/basename_filter.html#notes)

roles/dnsmasq/tasks/configure.yml
```diff
@@ -60,6 +60,12 @@
     src: "cifmw-dnsmasq.conf.j2"
     validate: "/usr/sbin/dnsmasq -C %s --test"
 
+- name: Render listener configuration
```
cjeanner, line 63: so this means we're expected to know everything before we start to deploy dnsmasq? It may get slightly ugly pretty fast with the way we'll call things:
- deploy dnsmasq basics
- create libvirt networks (and add configuration snippets to dnsmasq for the DHCP)
- get to devscripts parts, where we have to inject yet other snippets, some related to DNS, listeners and so on

We therefore want to set target file names to ensure we can add content "on the fly".

roles/dnsmasq/tasks/dns.yml
```diff
@@ -0,0 +1,34 @@
...
+# License for the specific language governing permissions and limitations
+# under the License.
+
+- name: Create dns configuration file
+  become: true
+  when:
+    - _act == 'install'
+  notify: Restart dnsmasq
+  ansible.builtin.template:
+    dest: "{{ cifmw_dnsmasq_basedir }}/dns.conf"
```
cjeanner, line 23: so if we could have something like `cifmw_dnsmasq_dns_config_file` or the like?

roles/dnsmasq/tasks/listener.yml
```diff
@@ -0,0 +1,34 @@
...
+# License for the specific language governing permissions and limitations
+# under the License.
+
+- name: Create listener configuration file
+  become: true
+  when:
+    - _act == 'install'
+  notify: Restart dnsmasq
+  ansible.builtin.template:
+    dest: "{{ cifmw_dnsmasq_basedir }}/listener.conf"
```
cjeanner, line 23: ditto, something like `cifmw_dnsmasq_listener_config_file` may be handy

roles/dnsmasq/templates/cifmw-dnsmasq.conf.j2
```diff
@@ -1,17 +1,16 @@
...
+
 user=dnsmasq
 group=dnsmasq
 pid-file=/var/run/cifmw-dnsmasq.pid
 dhcp-leasefile=/var/lib/dnsmasq/cifmw-dnsmasq.leases
-{% for key,val in cifmw_dnsmasq_global_options.items() %}
-{%   if val is not none                                %}
+
+{% for key,val in cifmw_dnsmasq_global_options.items() -%}
+{%-   if val is not none                               -%}
```
cjeanner, line 9: so @hjensas thought it might be nicer to use have a list of strings?
that way, we could pass anything we like:
```YAML
cifmw_dnsmasq_global_options:
  - 'port=0'
  - log-queries
  - exclude-interface=lo
  - exclude-interface=nat64-v4
  - exclude-interface=nat64-v6
```
and the like. That would allow to chose between both supported syntax (coma-separated values, or multiple option occurrences) in an easy way.

Thoughts?

roles/dnsmasq/templates/dns.conf.j2
```diff
@@ -0,0 +1,14 @@
+# Managed by ci-framework/dnsmasq
+
+{% for forwarder in cifmw_dnsmasq_forwarders -%}
+server={{ forwarder }}
+{% endfor                                     %}
+
+{% if cifmw_dnsmasq_domainname != ''         -%}
```
cjeanner, line 7: I heard the "right" way is to use `| length > 0` - but I don't really care.

roles/libvirt_manager/README.md
```diff
@@ -46,7 +46,7 @@ cifmw_libvirt_manager_configuration:
   vms:
     type_name:  # (string, such as "compute", "controller"
       start: (boolean, toggle if the VM should be strated or not. Optional, defaults to true)
-      manage: (boolean, toogle if the VM should be added to the inventory. Managed VM's must be on the cifmw_libvirt_manager_pub_net network. Optional, defaults to true)
+      manage: (boolean, toggle if the VM should be added to the inventory. Managed VM's must be on the cifmw_libvirt_manager_pub_net network. Optional, defaults to true)
```
cjeanner, line 49: that doesn't really belong to this patch I think :)

cjeanner (changes requested):
@hjensas care to have a look? The goal:
- modify dnsmasq role so that it can be used within ci_network role
- ensure it exposes the needed entry-points to nudge content "at will" while avoiding to rely on plain `ansible.builtin.copy` calls

roles/dnsmasq/README.md
```diff
@@ -22,7 +22,12 @@ supported in libvirt).
 * `cifmw_dnsmasq_network_state`: (String) Network status. Must be either `present` or `absent`.
 * `cifmw_dnsmasq_network_definition`: (Dict) Mapping representing the network definition.
 * `cifmw_dnsmasq_network_definition.ranges`: (List[mapping]) List of ranges associated to the network.
-* `cifmw_dnsmasq_network_listen_dns`: (List) List of IP addresses to listen to for DNS queries.
+* `cifmw_dnsmasq_forwarders`: (List) List of upstream DNS servers used as forwarders. Defaults to `[]`
+* `cifmw_dnsmasq_interfaces`: (List) List of interfaces on which dnsmasq should be enabled. Defaults to `[]`.
+* `cifmw_dnsmasq_listen_addresses`: (List) List of IP addresses on which dnsmasq should be enabled. Defaults to `[]`.
+* `cifmw_dnsmasq_enable_dns`: (Bool) Toggle to enable DNS features of dnsmasq. Defaults to `false`.
```
hjensas, line 28: So, we have 1 instance of dnsmasq running. We cannot toggle DNS on/off everytime we do "manage_network"? How do we know if antother network already enabled DNS?
I don't think we do - but this bool should be in "Common parameters"?

roles/libvirt_manager/README.md
rebtoor: correct, but documentation action was failing...

cjeanner:
/approve

I could run the fixed IP part with 32721e93b6bdb4742a922a6133b2158048b107e1 as last commit of this PR - and it's working fine apparently, at least for #1767 PR.

I propose to go with that version, and do the needed follow-ups with the needed changes.

rebtoor:
> /approve
> 
> I could run the fixed IP part with [32721e9](https://github.com/openstack-k8s-operators/ci-framework/commit/32721e93b6bdb4742a922a6133b2158048b107e1) as last commit of this PR - and it's working fine apparently, at least for #1767 PR.
> 
> I propose to go with that version, and do the needed follow-ups with the needed changes.

@hjensas what do you think?

hjensas (commented):
/lgtm

I think we can merge this, and iterate on it.

IMO we should move listening address and interface to the "manage_network" tasks - i.e we start with installing a basic service that do nothing - then as we add networks/hosts/dns records  with `manage_network`,`manage_host`,`manage_address` etc we add the interface and listen address, DNS records wildcards etc via those "api" task's.

PR #1745, opened 2024-05-22
Allow to modify VM prefix and index separator

cjeanner:
This change introduces two new parameters - usually for advanced usage
only. They allow to tweak the VM (*not* network!) final name by changing
the prefix (usually `cifmw-`) and the index separator (usually `-`) so
that we can end with VMs named `foo_compute_0` instead of
`cifmw-compute-0`.

  - [X] README in the role


cjeanner:
recheck

cjeanner:
Going far too deep for no benefit, this feature is needed for a very limited scope (OCP cluster management) - closing for now. Will reconsider once hitting that specific spot of the whole rework.

PR #1749, opened 2024-05-22
Avoid failures on cleaning

cjeanner:
While working on integrating dnsmasq role in the reproducer, a random
issue, related to the facts, might happen.

It seems to be caused by the way ansible manages the handler - the
`ansible_facts.packages` isn't available anymore, and the condition
triggering the libvirt cleanup fails. While it was already done.

This condition arrises when dnsmasq role is in the game, and the handler
kicks at the end of the playbook.

In the same fashion, the handler might be triggered when the service is
already removed, leading to another crash of the cleanup playbook. The
added conditions ensure the handler will succeed even if the service is
already removed, but would still fail if there's a real issue with the
service configuration.

- [X] getting those tested would be hard for now - dnsmasq isn't integrated yet

### DIFF
--- a/roles/dnsmasq/handlers/main.yml
+++ b/roles/dnsmasq/handlers/main.yml
@@ -16,6 +16,11 @@
 
 - name: Restart dnsmasq
   become: true
+  register: _dnsmasq
   ansible.builtin.systemd_service:
     name: cifmw-dnsmasq.service
     state: restarted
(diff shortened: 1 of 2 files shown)


cjeanner:
recheck

cescgina (commented):
/approve

psathyan (approved):

PR #1751, opened 2024-05-22
Generate all MAC addresses beforehand

cjeanner:
Pre-generating all the MAC, for all the instances, on all networks,
ensures we can stop relying on libvirt to learn about that data.

This also changes a bit how we gather logs from the reproducer molecule.
It's needed here in order to properly debug failures occuring only
upstream (not reproduceable on a local lab).

!!! WARNING !!!
this breaks the devscripts based deploy (for now). Users should really consume the `beta` branch!


cjeanner:
recheck

cjeanner:
/hold
- need to merge the 2 other patches embedded in this one
- need to consider if we wait for the follow-up patchES that will converge everything and merge everything in sequence.

Using `beta` branch should ensure things are stable for both users and downstream CI.

cjeanner:
Situation:
- merge the 2 other patches (no breakage)
- wait for the two follow-up patches to be ready (dnsmasq+fixed IPs, ocp provisioning)

cjeanner:
https://github.com/openstack-k8s-operators/ci-framework/pull/1777 will get in instead - one PR to rule them all!

PR #1752, opened 2024-05-23
Add reproducer scenario for PIDONE VA

lmiccini:
This commit adds a scenario file with the parameters necessary to deploy the "PIDONE VA", consisting in a 3 masters + 4 workers OpenShift cluster with network isolation (ipv4).

See https://github.com/openstack-k8s-operators/architecture/pull/245

### DIFF
--- a/scenarios/reproducers/va-pidone.yml
+++ b/scenarios/reproducers/va-pidone.yml
@@ -1,0 +1,130 @@
+---
+# Automation section. Most of those parameters will be passed to the
+# controller-0 as-is and be consumed by the `deploy-va.sh` script.
+# Please note, all paths are on the controller-0, meaning managed by the
+# Framework. Please do not edit them!
+_arch_repo: "/home/zuul/src/github.com/openstack-k8s-operators/architecture"
+cifmw_architecture_scenario: pidone
+cifmw_arch_automation_file: pidone.yaml
+
+# HERE if you want to overload kustomization, you can uncomment this parameter
+# and push the data structure you want to apply.
+# cifmw_architecture_user_kustomize:
+#   stage_0:
+#     'network-values':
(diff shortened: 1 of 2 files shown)


scenarios/reproducers/va-pidone.yml
```diff
@@ -0,0 +1,130 @@
...
+      image_local_dir: "/home/dev-scripts/pool"
+      disk_file_name: "ocp_master"
+      disksize: "105"
+      xml_paths:
+        - /home/dev-scripts/ocp_master_0.xml
+        - /home/dev-scripts/ocp_master_1.xml
+        - /home/dev-scripts/ocp_master_2.xml
+      nets:
+        - osp_trunk
+    ocp_worker:
```
cjeanner, line 54: I think you want to remove those entries here - the XML won't be generated, and therefore the deploy will crash.

As I understand things, the computes are what you really want? Maybe bumping the `3` to `4` to match your needs?

cjeanner: neverminds - I mixed workers and extra_workers

lmiccini: We want (need) 4 ocp workers because we want/need to test rebooting/crashing them and check that the control plane survives. 3 of them are regular workers, 1 is tainted so it only runs the test operator.
Computes are also in scope, we deploy 3 that will be used for instance ha and other tests.

abays:
Since the `do-not-merge/work-in-progress` label was removed, I'm going to remove this PR from `Draft` state.  It might be preventing the depends-on in https://github.com/openstack-k8s-operators/architecture/pull/245 from working properly.

abays (approved):
/approve
/lgtm

PR #1755, opened 2024-05-23
Silent the assertion in sushy_emulator

cjeanner:
The output is really huge, and I'm not really sure it provides useful
information as-is.


lewisdenny (approved):
/lgtm

PR #1758, opened 2024-05-23
Ensure we're targeting the right VMs

cjeanner:
If a user has multiple deployments, they therefore have mutliple blocks
in their `~/.ssh/config`.

While the ssh configuration allows to chose the right VM using the
hypervisor name (for instance controller-0.builder2), the dynamic
inventory built using `ansible.builtin.add_host` doesn't target the
specific host.

With `delegate_to`, it may happen ansible connects to a wrong VM from a
previous deployment that's still running, and the user will face some
really weird situations - for instance, facts being wrongs (this hit the
networking_mapper and the fetched MAC addresses).

With this patch, we now have specific host targeting, and we also get it
in the logs, for example:
```
changed: [builder2 -> controller-0(controller-0.builder2)]
```

### DIFF
--- a/roles/libvirt_manager/tasks/manage_vms.yml
+++ b/roles/libvirt_manager/tasks/manage_vms.yml
@@ -105,12 +105,10 @@
 
 - name: Inject nodes in the ansible inventory
   delegate_to: localhost
-  vars:
-    extracted_ip: "{{ (vm_ip.stdout.split())[3] | ansible.utils.ipaddr('address') }}"
   ansible.builtin.add_host:
     name: "{{ vm_ip.nic.host }}"
     groups: "{{ (vm_type == 'crc') | ternary('ocp', vm_type) }}s"
-    ansible_host: "{{ extracted_ip }}"
+    ansible_host: "{{ vm_ip.nic.host }}.{{ inventory_hostname }}"
     ansible_ssh_user: "{{ _cifmw_libvirt_manager_layout.vms[vm_type].admin_user | default(_init_admin_user) }}"
   loop: "{{ vm_ips.results }}"
   loop_control:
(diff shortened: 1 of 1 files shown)


roles/libvirt_manager/tasks/manage_vms.yml
```diff
@@ -173,7 +173,7 @@
     cmd: |-
       test -d /home/zuul && exit 0;
       set -xe -o pipefail;
-      echo "{{ vm_ip.nic.host }}" | sudo tee /etc/hostname;
+      echo "{{ vm_ip.nic.host }}.{{ inventory_hostname }}" | sudo tee /etc/hostname;
```
cjeanner, line 176: this is needed for sushy, since it will use the """fqdn""" - we need to ensure the host knows its own name - especially controller-0.

marios (commented):
seems sensible just wondering if anyone could (currently) be relying on the extracted_ip in the generated inventory? but I am not working in this area at all perhaps this is not a valid concern (and we are talking about the in memory inventory for add_host so... should be fine).

holding approve until the reviews call for any other comments

cjeanner:
recheck

cjeanner:
> seems sensible just wondering if anyone could (currently) be relying on the extracted_ip in the generated inventory? but I am not working in this area at all perhaps this is not a valid concern (and we are talking about the in memory inventory for add_host so... should be fine).
> 
> holding approve until the reviews call for any other comments

It's set in the `.ssh/config` - so there's nothing lost in there. This is really affecting only the dynamic, in-memory inventory, but does NOT affect written files that are then used on controller-0 for instance.

marios (approved):
/lgtm

PR #1760, opened 2024-05-23
Use '::' for sushy emulator default listen address

hjensas:
Using `0.0.0.0` makes the sushy emulator only listen on IPv4. 
When we use `::` it will listen on all IPv4 and IPv6 addresses.

With `0.0.0.0` we get:
```
  tcp   LISTEN 0      128          0.0.0.0:8000      0.0.0.0:*    users:(("python",pid=20353,fd=4),("python",pid=20353,fd=3),("sushy-emulator",pid=20333,fd=3))
```

With `::` we get:
```
  tcp   LISTEN 0      128                *:8000            *:*    users:(("python",pid=21650,fd=4),("python",pid=21650,fd=3),("sushy-emulator",pid=21628,fd=3))
```

  - [x] README in the role


lewisdenny:
recheck

lewisdenny (commented):
/lgtm

Thanks @hjensas, I incorrectly assumed :: would be for IPV6 and there for exposed it to be configurable with IPV4 as the default. Great to see we can just use :: for both :)

PR #1730, opened 2024-05-20
Update some jobs to CRC 2.36 (OCP4.15)

lewisdenny:
Update the following jobs to OCP 4.15:
- cifmw-base-multinode-kuttl
- cifmw-base-multinode-tempest
- cifmw-base-multinode-podified

Marks centos-9-medium-crc-extracted-2-34-1-3xl nodeset for deletion once migration is complete due to dependency from Nova operator repo.

JIRA: [OSPRH-6868](https://issues.redhat.com//browse/OSPRH-6868)

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/1761

As a pull request owner and reviewers, I checked that:
- [x] Appropriate documentation exists and/or is up-to-date:

PR #1761, opened 2024-05-27
Add additional listen-address to CRC dnsmasq service

lewisdenny:
Due to this change [1] introduced in CRC 2.32.0[2] the dnsmasq service is now running as a systemd service rather than in a container.

The behavior of the dnsmasq listen-address has now changed, the CRC dnsmasq only listens on the default address provided by Zuul. The Ansible controller is configured to query the interface we control on CRC, `ci-private-network`.

This breaks domain name resolution as dns queries are blocked.

This patch:
- Updates the CRC dnsmasq listen-address configuration to include the `ci-private-network` address allowing the Ansible controller to query it.

- Removes the hard coded `ci-private-network` address from the Ansible controllers default connection dns configuration to ensure they always match.

[1] https://github.com/crc-org/crc/commit/5f49891dc43b104e303fe49583068a40bd55825a
[2] https://github.com/crc-org/crc/releases/tag/v2.32.0

### DIFF
--- a/ci/playbooks/multinode-customizations.yml
+++ b/ci/playbooks/multinode-customizations.yml
@@ -100,6 +100,16 @@
                 replace: >-
                   {{ _crc_default_net_ip | ansible.utils.ipaddr('address') }}
 
+            # Note(Lewis): Only needed for CRC => 2.32.0-4.14.8
+            - name: Configure dnsmasq listen-address to listen on both br-ex and ci-private-network
+              when:
+                - not _dnsmasq.stat.exists
+              become: true
+              ansible.builtin.lineinfile:
+                path: "{{ _dnsmasq_config }}"
+                insertafter: '^listen-address='
+                line: "listen-address={{ _crc_default_net_ip | ansible.utils.ipaddr('address') }}"
+
     - name: Restart dnsmasq service if used
       become: true
       when:
@@ -196,6 +206,11 @@
       ansible.builtin.set_fact:
         persistent_ssh_key: "{{ pub_key_slurp['content'] | b64decode }}"
 
+    - name: Load network parameters
+      register: _cifmw_multinode_customizations_crc_net_env_slurp
+      ansible.builtin.slurp:
+        src: "/etc/ci/env/networking-info.yml"
+
     - name: Check if cloud-init is overriding NM settings
       become: true
       ansible.builtin.stat:
@@ -223,13 +238,26 @@
       ansible.builtin.command:
         cmd: "nmcli -g general.connection device show eth0"
 
-    - name: Add the deployment DNS in the controller resolv.conf
+    - name: Prepend CRC DNS server in the controllers default Network Manager connection configuation
       vars:
-        dns_servers_string: "192.168.122.10 {{ ansible_facts['dns']['nameservers'][0:1] | join(' ') }}"
+        _decoded_net_env: >-
+          {{
+            _cifmw_multinode_customizations_crc_net_env_slurp['content'] |
+            b64decode | from_yaml
+          }}
+        _crc_default_net_ip: >-
+          {{
+            _decoded_net_env.crc_ci_bootstrap_networks_out.crc.default.ip |
+            ansible.utils.ipaddr('address')
+          }}
+        dns_servers_string: >-
+          {{
+           ( _crc_default_net_ip, ansible_facts['dns']['nameservers'][0:1] | join(' ') ) | list
+          }}
       become: true
       ansible.builtin.shell:
         cmd: |-
-          nmcli con mod '{{ controller_default_connection_out.stdout | trim }}' ipv4.dns '{{ dns_servers_string }}' ipv4.route-metric 100;
+          nmcli con mod '{{ controller_default_connection_out.stdout | trim }}' ipv4.dns '{{ dns_servers_string | join(' ') }}' ipv4.route-metric 100;
           nmcli con up '{{ controller_default_connection_out.stdout | trim }}';
 
 - name: "Ensure persistent SSH access"


ci/playbooks/multinode-customizations.yml
```diff
@@ -100,6 +100,15 @@
                 replace: >-
                   {{ _crc_default_net_ip | ansible.utils.ipaddr('address') }}
 
+            - name: Configure dnsmasq listen-address to listen on both br-ex and ci-private-network
+              become: true
+              vars:
+                _original_listen_address: "listen-address={{ hostvars['crc'].ansible_host }}"
+              ansible.builtin.replace:
```
cjeanner, line 107: iirc, we could just use `ansible.builtin.lineinfile` and add the new listen-address - dnsmasq should listen on both. Might be even safer?

lewisdenny: So in the end have two `listen-address=` lines defined? I can quickly test that

cescgina: that's what we did for the reproducer a while ago, and it seems to work https://github.com/openstack-k8s-operators/ci-framework/blob/a1e6798dc98caae36b1c2881b77d84d20aebe965/roles/reproducer/tasks/configure_crc.yml#L87

cjeanner: yup, imho that's probably better and safer than editing the default interface.

lewisdenny: Sorry guys, could you expand on this?

The link from @cescgina seems to match what I'm doing here. As an example we search for `listen-address=192.168.25.178` and replace it with `listen-address=192.168.25.178,192.168.122.10`

lewisdenny:
recheck

ci/playbooks/multinode-customizations.yml
lewisdenny: thanks for the clarification on Slack, I confirmed lineinfile works so if your happy to go that direction I'll update this patch

lewisdenny:
Testing here: https://github.com/openstack-k8s-operators/ci-framework/pull/1730

rlandy:
recheck

lewisdenny:
Testing passed here and https://github.com/openstack-k8s-operators/ci-framework/pull/1730 meaning it fixes 4.15 and doesn't break 4.14, we should be fine to merge now :)

cescgina (approved):

cescgina:
> Testing here: #1730
> 
> Proof adding two lines for `listen-address` in config for works:
> 
> ```
> [core@crc ~]$ sudo ss -tpnl | grep dnsmasq
> LISTEN 0      32     192.168.122.10:53         0.0.0.0:*    users:(("dnsmasq",pid=32452,fd=7))
> LISTEN 0      32      192.168.26.86:53         0.0.0.0:*    users:(("dnsmasq",pid=32452,fd=5))
> 
> [core@crc ~]$ grep listen /etc/dnsmasq.d/crc-dnsmasq.conf
> listen-address=192.168.26.86
> listen-address=192.168.122.10
> ```

@lewisdenny I thought crc by default configures dnsmasq to listen only to localhost, do we set the `192.168.26.86` ip ?

lewisdenny:
> > Testing here: #1730
> > Proof adding two lines for `listen-address` in config for works:
> > ```
> > [core@crc ~]$ sudo ss -tpnl | grep dnsmasq
> > LISTEN 0      32     192.168.122.10:53         0.0.0.0:*    users:(("dnsmasq",pid=32452,fd=7))
> > LISTEN 0      32      192.168.26.86:53         0.0.0.0:*    users:(("dnsmasq",pid=32452,fd=5))
> > 
> > [core@crc ~]$ grep listen /etc/dnsmasq.d/crc-dnsmasq.conf
> > listen-address=192.168.26.86
> > listen-address=192.168.122.10
> > ```
> 
> @lewisdenny I thought crc by default configures dnsmasq to listen only to localhost, do we set the `192.168.26.86` ip ?

I'm not sure mate, before we add `192.168.122.10`, `192.168.26.86` is already there in the config. I'm not sure where else in the pipeline we are setting the listen-address.

lewisdenny:
> > > Testing here: #1730
> > > Proof adding two lines for `listen-address` in config for works:
> > > ```
> > > [core@crc ~]$ sudo ss -tpnl | grep dnsmasq
> > > LISTEN 0      32     192.168.122.10:53         0.0.0.0:*    users:(("dnsmasq",pid=32452,fd=7))
> > > LISTEN 0      32      192.168.26.86:53         0.0.0.0:*    users:(("dnsmasq",pid=32452,fd=5))
> > > 
> > > [core@crc ~]$ grep listen /etc/dnsmasq.d/crc-dnsmasq.conf
> > > listen-address=192.168.26.86
> > > listen-address=192.168.122.10
> > > ```
> > 
> > 
> > @lewisdenny I thought crc by default configures dnsmasq to listen only to localhost, do we set the `192.168.26.86` ip ?
> 
> I'm not sure mate, before we add `192.168.122.10`, `192.168.26.86` is already there in the config. I'm not sure where else in the pipeline we are setting the listen-address.

Yeah so CRC sets the listen-address to the IP of the VM or in our case "instance" of crc.

https://github.com/crc-org/crc/blob/813c80a473046247eb232bb7e9e3da79a954118b/pkg/crc/machine/start.go#L399C1-L403C65

cescgina:
> > > > Testing here: #1730
> > > > Proof adding two lines for `listen-address` in config for works:
> > > > ```
> > > > [core@crc ~]$ sudo ss -tpnl | grep dnsmasq
> > > > LISTEN 0      32     192.168.122.10:53         0.0.0.0:*    users:(("dnsmasq",pid=32452,fd=7))
> > > > LISTEN 0      32      192.168.26.86:53         0.0.0.0:*    users:(("dnsmasq",pid=32452,fd=5))
> > > > 
> > > > [core@crc ~]$ grep listen /etc/dnsmasq.d/crc-dnsmasq.conf
> > > > listen-address=192.168.26.86
> > > > listen-address=192.168.122.10
> > > > ```
> > > 
> > > 
> > > @lewisdenny I thought crc by default configures dnsmasq to listen only to localhost, do we set the `192.168.26.86` ip ?
> > 
> > 
> > I'm not sure mate, before we add `192.168.122.10`, `192.168.26.86` is already there in the config. I'm not sure where else in the pipeline we are setting the listen-address.
> 
> Yeah so CRC sets the listen-address to the IP of the VM [1] or in our case "instance" of crc. We need to access it on the internal network we create as part of the network bootstrap which is why we need this change.
> 
> So [1] is passed through a bunch of functions and ends up being `.IP` here in the template[2], you can trace back from [2] to [1].
> 
> I think your thought about CRC listening on localhost comes from the reproducer scenario, I guess there we start crc differently and it picks up 127.0.0.1 as it's IP? :)
> 
> [1] https://github.com/crc-org/crc/blob/813c80a473046247eb232bb7e9e3da79a954118b/pkg/crc/machine/start.go#L399C1-L403C65 [2] https://github.com/crc-org/crc/blob/813c80a473046247eb232bb7e9e3da79a954118b/pkg/crc/services/dns/template.go#L11

yes, you're right, I was thinking about the reproducer, where I looked at the stuff, thanks for digging!

PR #1762, opened 2024-05-27
nit: Ensure we focus on CI job rerun

cjeanner:
Until now, re-running a Zuul job using the reproducer lead to a "final"
error due to missing data for script generation.

This patch ensures we only get the CI job rerun related content,
avoiding that kind of issues.


lewisdenny (approved):
/lgtm

PR #1763, opened 2024-05-27
[DNM] Save ssh_config so it can be reused later

rlobillo:
  - [ ] README in the role


roles/reproducer/tasks/main.yml
```diff
@@ -286,3 +286,8 @@
         mode: "0755"
         owner: "zuul"
         group: "zuul"
+
+    - name: Save ssh config on the hypervisor
```
rlobillo, line 290: I can do this with the hook "post_deploy"

rlobillo:
not needed

PR #1765, opened 2024-05-27
sushy-emulator - use listening addr in _sushy_url

hjensas:
If listining address is set to something other than `::` or `0.0.0.0`, then use the listening address as _ip_address in the _sushy_url fact.

Also use ansible.utils.ipwrap to ensure address is wrapped in brackets when needed.

  - [x] README in the role

### DIFF
--- a/roles/sushy_emulator/tasks/create_container.yml
+++ b/roles/sushy_emulator/tasks/create_container.yml
@@ -41,9 +41,13 @@
       - "{{ cifmw_sushy_emulator_sshkey_path }}:/root/.ssh/id_rsa:ro,Z"
       - "{{ cifmw_sushy_emulator_sshkey_path }}.pub:/root/.ssh/id_rsa.pub:ro,Z"
 
-# We only support deploying Sushy Emulator Podman container on controller-0
 - name: Set Sushy Emulator URL for Podman installation
   vars:
-    _ip_address: "{{ hostvars['controller-0']['ansible_host'] }}"
+    _ip_address: >-
+      {%- if cifmw_sushy_emulator_listen_ip in ['::', '0.0.0.0'] -%}
+      {{ hostvars['controller-0']['ansible_host'] }}
+      {%- else -%}
+      {{ cifmw_sushy_emulator_listen_ip }}
+      {%- endif -%}
   ansible.builtin.set_fact:
-    _sushy_url: "http://{{ _ip_address }}:8000"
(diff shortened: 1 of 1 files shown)


hjensas:
recheck

roles/sushy_emulator/tasks/create_container.yml
```diff
@@ -41,9 +41,13 @@
       - "{{ cifmw_sushy_emulator_sshkey_path }}:/root/.ssh/id_rsa:ro,Z"
       - "{{ cifmw_sushy_emulator_sshkey_path }}.pub:/root/.ssh/id_rsa.pub:ro,Z"
 
-# We only support deploying Sushy Emulator Podman container on controller-0
```
lewisdenny, line 44: Hi Harald, thank you for the enhancement, where are you planning on running the sushy emulator container? The comment was added somewhat as a design decision but very open to hearing your use case

cjeanner: We may end with sushy on the hypervisor, in the same way we may have vbmc I guess?

hjensas: I kind of like the approach of installing it on controller-0 - it makes cleanup easy. Delete the VM and it's all gone. :+1: And in the case we want to use it in openstack the controller-0 is a nice place to put it. But I also think it is good if we allow some choice - currently I am running it as a pod on the hypervisor using the IP address of one of the virt bridges as listening address.

Also I think in an IPv6 job we may need to have hostvars['controller-0']['ansible_host'] be an IPv4 interface - and we want to reach sushy on a different interface with IPv6. It will by default listen on all interfaces - but the URL in the baremetal-info.yml the reproducer dumps is wrong in that case. this change allows me to control that too - i.e run on the controller but use a different interface from the one in that ansible fact/hostvar.

I think we should change this to write a host-record in the dnsmasq - sushy-emulator.cifmw.lab or possibly even one for each interface it is listening on like ocpbm-sushy.cifmw.lab and ocppr-sushy.cifmw.lab - but we can iterate on that as the changes @rebtoor is doing to the dnsmasq role is still in progress.

lewisdenny: That sounds really nice, ease of cleanup was one of our considerations but you're right, we can default to deploying on the controller but not enforce it. Thanks for the clear explanation mate.

PR #1767, opened 2024-05-27
Fix all IP addresses for all known VMs on all known networks

cjeanner:
In order to get there, we have to deploy dnsmasq service on the
reproducer, while ensuring it's not already running via NetworkManager
(this happens when CRC has been setup on the hypervisor).

Then, we generate specific network configuration in dnsmasq for each
managed interface - we can also NOT use dnsmasq for selected interfaces,
such as `ocppr`, used by devscripts to deploy OCP cluster.

To get the fixed IPs, we're using the networking_mapper, ensuring we're
getting IPs in the selected networks, without any collision, and just
inject a generated mapping IP, hostname and MAC in dnsmasq
configuration.

Note that dnsmasq is initialized in
libvirt_manager/generate_netwkring_data since that's the place we get
all the virtual network data, among them the interfaces to listen to.


roles/libvirt_manager/tasks/generate_networking_data.yml
```diff
@@ -0,0 +1,172 @@
...
+    cifmw_networking_mapper_gather_facts: false
+  ansible.builtin.import_role:
+    name: networking_mapper
+
+- name: Manage networks if needed
+  when:
+    - _cifmw_libvirt_manager_layout.networks is defined
+  ansible.builtin.include_tasks: create_networks.yml
+
+- name: Craft patch for cifmw_networking_definition
```
cjeanner, line 85: @pablintino That's so ugly I'm ashamed to ask you, but... is there any better way to automatically inject a network?

In the end, the patch usually looks like this:
(in molecule jobs)
```YAML
    group-templates:
      baremetals:
        networks:
          public:
            range: 20-24
      computes:
        networks:
          public: null
    instances: {}
    networks:
      public:
        network-v4: 192.168.110.0/24
```

or in "actual" job it may look like this:
```YAML
    group-templates: {}
    instances:
      controller-0:
        networks:
          public:
            ip-v4: 192.168.100.9
    networks:
      public:
        network-v4: 192.168.100.0/24
```

Then it gets combined in the networking_mapper.

roles/libvirt_manager/tasks/generate_networking_data.yml
```diff
@@ -0,0 +1,172 @@
...
+        network-v6: {{ net_6 }}
+      {% endif                                                                                                          %}
+    _gr_templates: |
+      {% for group in (_cifmw_libvirt_manager_layout.vms.keys() | sort) if group != 'controller' and
+                                                                           (_cifmw_libvirt_manager_layout.vms[group].amount | default(1) | int) > 0 %}
+      {% set _gr = (group == 'crc') | ternary('ocp', group)                                                             %}
+        {{ _gr }}s:
+          networks:
+            {{ _lnet_data.name | replace('cifmw_', '') }}:
+      {% if cifmw_networking_definition['group-templates'][_gr ~ 's']['network-template'] is not defined                %}
```
cjeanner, line 110: @pablintino as I understand we have to segment the network in "some way". My idea here is to either benefit of any existing `network-template` set for the group, or fallback on the "segment" used by `internalapi`
But it really look bad depending on the format of that "segment", since it may be an actual range with the whole network hardcoded in it...

No better idea for now, maybe you have a trick?

roles/libvirt_manager/tasks/generate_networking_data.yml
```diff
@@ -0,0 +1,172 @@
...
+            ip-v4: "{{ _lnet_data.ranges[0].start_v4 | ansible.utils.ipsubnet(_lnet_data.ranges[0].prefix_length_v4) | ansible.utils.nthhost(9) }}"
+      {% endif %}
+      {% if _lnet_data.ranges[0].start_v6 is defined and _lnet_data.ranges[0].start_v6 | length > 0 %}
+            ip-v6: "{{ _lnet_data.ranges[0].start_v6 | ansible.utils.ipsubnet(_lnet_data.ranges[0].prefix_length_v6) | ansible.utils.nthhost(9) }}"
+      {% endif %}
+      {% endif %}
+  ansible.builtin.set_fact:
+    cifmw_networking_mapper_definition_patch_01_libvirt:
+      networks: "{{ _networks | from_yaml }}"
+      'group-templates': "{{ _gr_templates | from_yaml | default({}, true) }}"
```
cjeanner, line 129: so here we may get an empty set for `group-templates` and `instances` - but I don't expect to get any for the `networks` one...

cjeanner (commented):
@pablintino if you have some time - I'd love to hear you yelling at me for that bad idea ^^'.

cjeanner:
recheck  unable to reproduce the issue locally.

cjeanner:
/hold

Waiting for coming Monday

PR #1768, opened 2024-05-27
nit: Ensure sushy is installed after base packages

cjeanner:
Base packages are installed using `async` to allow generating most of
not all the relevant content instead of waiting.

This leads to a race condition, where sushy_emulator role wants to
install dependency packages and fails, because the RPM database is still
locked with the dependency async run.

This patch moves the sushy bootstrap and installation to a later step,
ensuring we're not coliding with the async task.


hjensas (approved):
LGTM

lewisdenny (approved):
/approve

I will follow this patch up with a small contributor doc note explaining the behaviour. Good catch Cédric, crazy I never hit this.

lewisdenny:
> /approve
> 
> I will follow this patch up with a small contributor doc note explaining the behaviour. Good catch Cédric, crazy I never hit this.

Actually I don't think it's needed in this instance, the issue is quite specific.

PR #1777, opened 2024-05-29
Allow to deploy OCP using baremetal

cjeanner:

 (not in commit message)

Testing:
- :heavy_check_mark:  uni01alpha
- :heavy_check_mark: uni02beta
- :heavy_check_mark: uni04delta
- :heavy_check_mark: uni05epsilone
- :heavy_check_mark: uni06zeta
- :heavy_check_mark: uni07eta
- :heavy_check_mark: va-pidone
- :heavy_check_mark: bgp scenarios
- :heavy_check_mark: extra volumes support
- :heavy_check_mark: worker instances
- :question: NFV (hitting issues, not sure if related)

#### Testing information have been move to a proper comment, due to zuul doing some sad parsing attempts.


roles/libvirt_manager/README.md
```diff
@@ -20,6 +20,8 @@ Used for checking if:
 * `cifmw_libvirt_manager_images_url`: (String) Location basedir for the image URI. Defaults to `https://cloud.centos.org/centos/9-stream/x86_64/images`.
 * `cifmw_libvirt_manager_configuration`: (Dict) Structure describing the libvirt layout (networking and VMs).
 * `cifmw_libvirt_manager_vm_template`: (String) Template name to use to define the virtual machines. Defaults to `domain.xml.j2`. Advanced use only.
+* `cifmw_libvirt_manager_vm_separator`: (Char) Separator character between VM and its index. Defaults to `-`. Advanced use onlyg
```
hjensas, line 23: nit: typo `onlyg`

roles/libvirt_manager/README.md
```diff
@@ -20,6 +20,8 @@ Used for checking if:
...
+gcifmw_libvirt_manager_vm_prefix`: (String) VM name prefix. Defaults to `cifmw-`. Advanced use only.
```
hjensas, line 24: nit: typo? `gcifmw` ?

roles/libvirt_manager/vars/main.yml
```diff
@@ -42,3 +42,6 @@ cifmw_libvirt_manager_storage_pool: "cifmw-pool"
 cifmw_libvirt_manager_fixed_networks_defaults:
   - ocpbm
   - ocppr
+
+cifmw_libvirt_manager_no_dnsmasq_nets_defaults:
```
hjensas, line 46: Can we not turn this around?
 Have a variable or an additional property where we define networks to enable dnsmasq, and only set up dnsmasq if enabled for the network.

roles/devscripts/tasks/133_host_network.yml
```diff
@@ -23,13 +23,6 @@
     name: libvirt_manager
     tasks_from: delete_network.yml
 
-- name: Ensure the required network configurations are applied.
```
psathyan, line 26: I strongly believe this entire set of tasks i.e. `133_host_network` should be moved outside

roles/reproducer/tasks/devscripts_layout.yml
```diff
@@ -1,6 +1,223 @@
...
+            _libvirt_manager_networking.networks[cifmw_libvirt_manager_pub_net]
+          }}
+      block:
+        - name: Generate patched overrides for devscripts
+          vars:
+            _bootstrap_mac: "52:54:ab:83:31:87"
+          ansible.builtin.set_fact:
+            cifmw_devscripts_config_overrides_patch_01_baremetal:
+              external_bootstrap_mac: "{{ _bootstrap_mac }}"
+              manage_br_bridge: 'n'
```
psathyan, line 171: Please add `manage_int_bridge: 'n'`... there was a check for this leading to conflicts.

roles/reproducer/tasks/devscripts_layout.yml
```diff
@@ -47,6 +47,7 @@
       local=/{{ _cluster }}.openstack.lab/
       domain={{ _cluster }}.openstack.lab
       address=/api.{{ _cluster }}.openstack.lab/{{ _api_addr }}
+      address=/api-int.{{ _cluster }}.openstack.lab/{{ _api_addr }}
```
hjensas, line 50: @psathyan - is this needed? (I think you said this is only required for Single-node-Openshift the other day?)

roles/libvirt_manager/tasks/generate_networking_data.yml
```diff
@@ -0,0 +1,160 @@
...
+  ansible.builtin.import_role:
+    name: networking_mapper
+# END networking_mapper
+
+- name: Manage networks if needed
+  when:
+    - _cifmw_libvirt_manager_layout.networks is defined
+  ansible.builtin.include_tasks: create_networks.yml
+
+- name: Craft patch for cifmw_networking_definition
```
hjensas, line 86: This task fails for me:

```
fatal: [hypervisor-1]: FAILED! => 
  msg: |-
    The task includes an option with an undefined variable. The error was: networks:
      {{ _lnet_data.name | replace('cifmw_', '') }}:
    {% if _lnet_data.ranges[0].start_v4 is defined and _lnet_data.ranges[0].start_v4 | length > 0                     %}
      {% set net_4 = _lnet_data.ranges[0].start_v4 | ansible.utils.ipsubnet(_lnet_data.ranges[0].prefix_length_v4)    %}
        network-v4: {{ net_4}}
    {% endif                                                                                                          %}
    {% if _lnet_data.ranges[0].start_v6 is defined and _lnet_data.ranges[0].start_v6 | length > 0                     %}
        {% set net_6 = _lnet_data.ranges[0].start_v6 | ansible.utils.ipsubnet(_lnet_data.ranges[0].prefix_length_v6)  %}
        network-v6: {{ net_6 }}
    {% endif %}
    group-templates:
    {% for group in _cifmw_libvirt_manager_layout.vms.keys() if group != 'controller' %}
      {{ (group == 'crc') | ternary('ocp', group) }}s:
        networks:
          {{ _lnet_data.name | replace('cifmw_', '') }}: {}
    {% endfor                           %}
    instances:
      controller-0:
        networks:
          {{ _lnet_data.name | replace('cifmw_', '') }}:
    {% if net_4 is defined %}
            ip-v4: "{{ net_4 | ansible.utils.nthhost(9) }}"
    {% endif %}
    {% if net_6 is defined %}
            ip-v6: "{{ net_6 | ansible.utils.nthhost(9) }}"
    {% endif %}
    : {{
      _network_data |
      selectattr('name', 'match', _match) | first
    }}: No first item, sequence was empty.. No first item, sequence was empty.. {{
      _network_data |
      selectattr('name', 'match', _match) | first
    }}: No first item, sequence was empty.. No first item, sequence was empty.. networks:
      {{ _lnet_data.name | replace('cifmw_', '') }}:
    {% if _lnet_data.ranges[0].start_v4 is defined and _lnet_data.ranges[0].start_v4 | length > 0                     %}
      {% set net_4 = _lnet_data.ranges[0].start_v4 | ansible.utils.ipsubnet(_lnet_data.ranges[0].prefix_length_v4)    %}
        network-v4: {{ net_4}}
    {% endif                                                                                                          %}
    {% if _lnet_data.ranges[0].start_v6 is defined and _lnet_data.ranges[0].start_v6 | length > 0                     %}
        {% set net_6 = _lnet_data.ranges[0].start_v6 | ansible.utils.ipsubnet(_lnet_data.ranges[0].prefix_length_v6)  %}
        network-v6: {{ net_6 }}
    {% endif %}
    group-templates:
    {% for group in _cifmw_libvirt_manager_layout.vms.keys() if group != 'controller' %}
      {{ (group == 'crc') | ternary('ocp', group) }}s:
        networks:
          {{ _lnet_data.name | replace('cifmw_', '') }}: {}
    {% endfor                           %}
    instances:
      controller-0:
        networks:
          {{ _lnet_data.name | replace('cifmw_', '') }}:
    {% if net_4 is defined %}
            ip-v4: "{{ net_4 | ansible.utils.nthhost(9) }}"
    {% endif %}
    {% if net_6 is defined %}
            ip-v6: "{{ net_6 | ansible.utils.nthhost(9) }}"
    {% endif %}
    : {{
      _network_data |
      selectattr('name', 'match', _match) | first
    }}: No first item, sequence was empty.. No first item, sequence was empty.. {{
      _network_data |
      selectattr('name', 'match', _match) | first
    }}: No first item, sequence was empty.. No first item, sequence was empty.
  
    The error appears to be in '/home/hjensas/src/github.com/openstack-k8s-operators/ci-framework/roles/libvirt_manager/tasks/generate_networking_data.yml': line 86, column 3, but may
  
    - name: Craft patch for cifmw_networking_definition
      ^ here
```

hjensas: ok, figured this one out.
We now have to add ocpbm and ocppr networks to the cifmw_libvirt_manager_configuration.

roles/reproducer/tasks/devscripts_layout.yml
```diff
@@ -1,6 +1,224 @@
...
+
+- name: Ensure we use dnsmasq as resolver
+  become: true
+  vars:
+    _cluster: "{{ cifmw_devscripts_config['cluster_name'] }}"
+  ansible.builtin.copy:
+    dest: /etc/resolv.conf
+    content: |
+      search {{ _cluster }}.openstack.lab
+      nameserver {{ _libvirt_manager_networking.networks.ctlplane.gw_v4 }}
```
hjensas, line 79: When deploying IPv6 there is no `gw_v4`.

roles/reproducer/tasks/devscripts_layout.yml
```diff
@@ -1,6 +1,224 @@
...
+        tasks_from: "ocp_layout.yml"
+
+    - name: Refresh VBMC data
+      ansible.builtin.import_role:
+        name: "virtualbmc"
+        tasks_from: "list_hosts.yml"
+
+    - name: Generate baremetal content fact
+      vars:
+        _hypervisor_ip4: >-
```
hjensas, line 103: How is this supposed to work for IPv6?
If the bootstrap node is IPv6 only it cannot talk to the hypervisor or virtualbmc over IPv4?

roles/reproducer/tasks/prepare_networking.yml
```diff
@@ -46,7 +56,46 @@
...
+    _content: |-
+      {{ cifmw_target_host | default('localhost') }}:
+        interfaces:
+          - name: ocppr
+            type: linux-bridge
+            state: up
+            ipv4:
+              enabled: true
+              address:
+                - ip: 172.22.0.1
```
hjensas, line 82: So, this needs to be configurable.

roles/reproducer/tasks/devscripts_layout.yml
```diff
@@ -1,6 +1,224 @@
...
+          }}
+        _node:
+          name: "openshift-master-{{ vm_id  }}"
+          driver: "ipmi"
+          resource_class: "baremetal"
+          driver_info:
+            username: "admin"
+            password: "password"
+            address: "ipmi://{{ _hypervisor_ip4 }}:{{ _ipmi_port }}"
+            deploy_kernel: "http://172.22.0.2/images/ironic-python-agent.kernel"
```
hjensas, line 128: This needs to be configurable.

roles/reproducer/tasks/devscripts_layout.yml
```diff
@@ -1,6 +1,224 @@
...
+            deploy_ramdisk: >-
+              http://172.22.0.2/images/ironic-python-agent.initramfs
```
hjensas, line 130: Ditto, must be able set IPv6 address.

roles/reproducer/tasks/devscripts_layout.yml
```diff
@@ -1,6 +1,224 @@
...
+    content: |
+      [main]
+      rc-manager=none
+      dns=none
+    validate: "NetworkManager --print-config %s"
+
+- name: Ensure services are reloaded
+  ansible.builtin.meta: flush_handlers
+
+- name: Ensure we use dnsmasq as resolver
```
hjensas, line 71: So after changing `gw_v4` to `gw_v6` my server lost the ability to resolve names.
The reason for this is that ocppr is configured with an IPv4 address `172.22.0.1`.

```
[zuul@host26 ~]$ cat /etc/resolv.conf
search ocp.openstack.lab
nameserver fd00:aaaa:0000:0000:0000:0000:0000:0001
[zuul@host26 ~]$ ping google.com
ping: google.com: Name or service not known
```

```
[zuul@host26 ~]$ sudo ss -natulp | grep 53
udp   UNCONN 0      0         172.22.0.1:53         0.0.0.0:*    users:(("dnsmasq",pid=74001,fd=7))     
tcp   LISTEN 0      32        172.22.0.1:53         0.0.0.0:*    users:(("dnsmasq",pid=74001,fd=8))  
```

I think we need to change the approach - we should leave NetworkManager's dns config as is. Detect if it is using dnsmasq or resolvd - and then add forwarding via the cifmw-dnsmasq for the internal domains only.

roles/reproducer/tasks/prepare_networking.yml
cjeanner: You can, by just providing the parameter with the same name.
`set_fact` does NOT override the parameter, meaning you have full control over its value/content.

So if you build `cifmw_ci_nmstate_instance_config` - it will just work. More over, there's a `when` condition as a safety net to NOT override the parameter.

roles/libvirt_manager/vars/main.yml
cjeanner: Not sure about the benefit of either approach to be fair. With that one here (disallow-list) we have a working thing without any edition for "the cases I know". We can override/append networks using the parameter without `_defaults` so it's easy enough to set up for jobs.

roles/reproducer/tasks/devscripts_layout.yml
cjeanner: I have a proposal in another draft PR #1817 we'll use to test NFV - apparently there are some needs in there since the range isn't the same.

roles/devscripts/tasks/133_host_network.yml
cjeanner: maybe in a follow-up - for now, I want to target minimal changes.

roles/reproducer/tasks/devscripts_layout.yml
```diff
@@ -1,6 +1,224 @@
...
```
hjensas, line 70: This just does not work. Unless you want to start messing with creating dummy interfaces the interface will remain DOWN until you start hooking things up to the bridge.

```
9: cifmw-osp_trunk: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc noqueue state DOWN group default qlen 1000
    link/ether 52:54:00:68:ae:92 brd ff:ff:ff:ff:ff:ff
    inet6 fd00:aaaa::1/64 scope global tentative 
       valid_lft forever preferred_lft forever
```

We need to set dns=dnsmasq above in NetworkManager and configure conditional forwarding for our domain.

hjensas: If we add both IPv4 and IPv6 - or condition it on wheater it is defined or not we fix the issue below as well.
server=/{{ _cluster }}.openstack.lab/{{ _libvirt_manager_networking.networks.ctlplane.gw_v4 }}
server=/{{ _cluster }}.openstack.lab/{{ _libvirt_manager_networking.networks.ctlplane.gw_v6 }}

cjeanner: I'm modifying the previous PR to configure dnsmasq in NM with forwarders. I'll update that one to consume it properly.

roles/libvirt_manager/tasks/ocp_layout.yml
```diff
@@ -0,0 +1,66 @@
...
+  vars:
+    cifmw_virtualbmc_daemon_port: 51881
+  ansible.builtin.import_role:
+    name: "virtualbmc"
+
+- name: Manage virtual resources
+  vars:
+    cifmw_libvirt_manager_storage_pool: "oooq_pool"
+    action: "create"
+    cifmw_libvirt_manager_pool_dir: "/home/dev-scripts/pool"
```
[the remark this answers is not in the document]
cjeanner: not yet - that's an invasive change. That location is created during the bootstrap.

roles/libvirt_manager/tasks/ocp_layout.yml
```diff
@@ -0,0 +1,66 @@
...
+    - name: Craft layout and pass it to create_vms.yml
+      vars:
+        cifmw_libvirt_manager_vm_template: "ocp_domain.xml.j2"
+        cifmw_libvirt_manager_vm_prefix: ''
+        cifmw_libvirt_manager_vm_separator: '_'
+        vm_type: 'ocp_master'
+        _ocp_layout:
+          vms:
+            ocp_master:
+              disksize: "{{ _cifmw_libvirt_manager_layout.vms.ocp.disksize }}"
```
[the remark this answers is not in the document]
cjeanner: the image isn't at that size on the disk - so we shouldn't lose too much space But we might consider a smaller size, so that it's easier to grow later.

roles/reproducer/tasks/devscripts_layout.yml
cjeanner: we now rely on dnsmasq within networkmanager, with zone delegation to our own instance. Should be sufficient and far, far better.

roles/reproducer/tasks/devscripts_layout.yml
cjeanner: that's for follow-up - current patch is a 1:1 match with `main` features.

roles/libvirt_manager/tasks/ocp_layout.yml
```diff
@@ -0,0 +1,66 @@
...
+
+    - name: Craft layout and pass it to create_vms.yml
+      vars:
+        cifmw_libvirt_manager_vm_template: "ocp_domain.xml.j2"
+        cifmw_libvirt_manager_vm_prefix: ''
+        cifmw_libvirt_manager_vm_separator: '_'
+        vm_type: 'ocp_master'
+        _ocp_layout:
+          vms:
+            ocp_master:
```
hjensas, line 34: I think needs to be a copy of `cifmw_libvirt_manager_layout.vms.ocp` then remove the xml_paths and re-write the `disk_file_name` to `blank`.
See: OSPRH-7489

cjeanner: hmm, lemme see if that'd be easy.

cjeanner: could get it right. Note that the current (and coming) way to manage those extra volumes for OCP nodes isn't good, since they aren't recreated when you reuse an env.... Now that we get out of the devscripts world, a follow-up can simplify and make that thing better in many ways (with overlays for OCP extra volumes, for instance).

roles/reproducer/tasks/devscripts_layout.yml
hjensas: Closing this - I guess it does not hurt.

roles/libvirt_manager/tasks/create_networks.yml
```diff
@@ -81,6 +81,104 @@
...
+      {{
+        cifmw_libvirt_manager_no_dnsmasq_nets_defaults +
+        cifmw_libvirt_manager_no_dnsmasq_nets
+      }}
+    cifmw_dnsmasq_enable_dns: true
+    cifmw_dnsmasq_interfaces: "{{ _dns_listeners | difference(_no_dnsmasq) }}"
+    cifmw_dnsmasq_exclude_lo: true
+    cifmw_dnsmasq_listen_addresses:
+      - "127.0.0.2"
+      - "::2"
```
hjensas, line 154: We can drop `::2` - and just use the `127.0.0.2`.
```
Jun 05 08:02:19 host26.beaker.tripleo.ral3.eng.rdu2.redhat.com dnsmasq[135877]: failed to create listening socket for ::2: Cannot assign requested address
Jun 05 08:02:19 host26.beaker.tripleo.ral3.eng.rdu2.redhat.com dnsmasq[135877]: failed to create listening socket for ::2: Cannot assign requested address
```

roles/libvirt_manager/templates/ocp_domain.xml.j2
```diff
@@ -0,0 +1,119 @@
...
+            <source pool="oooq_pool" volume="{{ _prefix }}_master_{{ vm_id }}.qcow2"/>
+            <target dev="sda" bus="scsi"/>
+        </disk>
+{% set disk_suffix='abcdefghijklmnopqrstuvwxyz' %}
+{% if vm_data.value.extra_disks_num is defined %}
+{% for n in range(vm_data.value.extra_disks_num) %}
+        <disk type='volume' device='disk'>
+          <driver name='qemu' type='qcow2' />
+          <source pool='oooq_pool' volume='{{ vm_type }}{{ cifmw_libvirt_manager_vm_separator }}{{ vm_id }}-vol-{{ n }}' />
+          <target dev='vd{{ disk_suffix[n] }}' bus='scsi' />
```
cjeanner, line 42:
```suggestion
          <target dev='vd{{ disk_suffix[n] }}' bus='virtio' />
```

roles/reproducer/tasks/prepare_networking.yml
```diff
@@ -46,7 +64,62 @@
...
+  ansible.builtin.include_role:
     name: libvirt_manager
-    tasks_from: create_networks.yml
+    tasks_from: generate_networking_data.yml
+
+- name: Delegate our domains to our own instance
+  become: true
+  notify: Restart NetworkManager
+  vars:
+    _ip_4: >-
```
hjensas, line 112: these vars `_ip_4` and `_ip_6` are not used?

roles/reproducer/tasks/devscripts_layout.yml
```diff
@@ -1,6 +1,337 @@
...
+            _num_workers: >-
+              {{
+                _cifmw_libvirt_manager_layout.vms.ocp_worker.amount | default(0)
+              }}
+            _num_masters: >-
+              {{
+                _cifmw_libvirt_manager_layout.vms.ocp.amount | default(3)
+              }}
+          ansible.builtin.set_fact:
+            cifmw_devscripts_config_overrides_patch_02_baremetal:
```
hjensas, line 256: I think somewhere along the way we lost something setting the MAC address for the external (ocpbm) interfaces to `_bootstrap_mac` here?

https://github.com/openshift-metal3/dev-scripts/blob/master/config_example.sh#L392

```
[zuul@host26 ci-framework-data]$ virsh domiflist ocp-7fhmz-bootstrap
 Interface   Type     Source   Model    MAC
-----------------------------------------------------------
 vnet0       bridge   ocpbm    virtio   52:54:00:66:56:37
 vnet1       bridge   ocppr    virtio   52:54:00:a6:be:1e
```

```
[zuul@host26 ci-framework-data]$ grep ocp-boostrap /etc/cifmw-dnsmasq.d/ocpbm-hosts.conf 
dhcp-host=52:54:ab:83:31:87,[fd00:1:2::4],ocp-boostrap
```

The MACs don't match up, so the bootstrap nodes boots into no networking and fails `Unable to pull OpenShift release image`.

cjeanner: right, I lost it during some content moving -.-. It's back in place with latest push - sorry for that one!

roles/devscripts/vars/main.yml
```diff
@@ -37,16 +37,12 @@ cifmw_devscripts_config_defaults:
   cluster_name: "ocp"
   base_domain: "openstack.lab"
   ntp_servers: "clock.corp.redhat.com"
-  ip_stack: "v4"
-  network_type: "OVNKubernetes"
```
hjensas, line 41: Removal of `network_type` causes an error on L20 in `roles/devscripts/tasks/137_custom_install.yml`.

```
The conditional check 'cifmw_devscripts_config['network_type'] == 'OVNKubernetes'' failed. The error was: error while evaluating conditional (cifmw_devscripts_config['network_type'] == 'OVNKubernetes'): 'dict object' has no attribute 'network_type'. 'dict object' has no attribute 'network_type'
```

roles/devscripts/vars/main.yml
```diff
@@ -37,16 +37,12 @@ cifmw_devscripts_config_defaults:
...
   cluster_subnet_v4: "192.168.16.0/20"
   cluster_host_prefix_v4: "22"
   service_subnet_v4: "172.30.0.0/16"
   external_subnet_v4: "192.168.111.0/24"
   num_masters: 3
-  master_memory: 32768
-  master_disk: 80
-  master_vcpu: 10
   num_workers: 0
+  ip_stack: v4  # TODO(Harald) Check if we can get rid of it - see 133_host_network.yml
```
hjensas, line 48: So, the condition to select the `ansible_default_ipv6.interface` and then set up firewall rules in the ipv4 table to do masquerading on traffic going out that interface does not make any sense to me.

But is'nt this also used to set the ENV vars for the dev-scripts installer? See: `roles/devscripts/templates/conf_ciuser.j2` there is a snippet converting these to "shell vars" used by devscripts.

```
{% for item in cifmw_devscripts_config %}
export {{ item.upper() }}="{{ cifmw_devscripts_config[item] }}"
{% endfor %}
```

I think `NETWORK_TYPE` and `IP_STACK` is something we need to define for dev-scripts.

cjeanner: IP_STACK and NETWORK_TYPE have "good" default values:
https://github.com/openshift-metal3/dev-scripts/blob/3c655f4653f575009b6c92796f726d75e89027aa/config_example.sh#L220-L227

https://github.com/openshift-metal3/dev-scripts/blob/3c655f4653f575009b6c92796f726d75e89027aa/config_example.sh#L283-L294

I propose we keep that one in - I tried to loop on the interfaces but faced some weird failure. "it's working", we may work with some follow-up.. ?

roles/devscripts/vars/main.yml
cjeanner: yep, saw it, that's why I re-added it (with a comment)

lmiccini:
FWIW we tested this PR against the 3 x masters and 4 x workers architecture and we did not face any regression. :+1:

roles/devscripts/vars/main.yml
hjensas: Yeah - let's just keep them in. And we can try to clean up later.
Marking resolved.

roles/libvirt_manager/tasks/deploy_layout.yml
```diff
@@ -214,6 +215,50 @@
...
+      {{ hostvars[_ipmi_host].ansible_default_ipv4.address }}
+  ansible.builtin.include_role:
+    name: virtualbmc
+    tasks_from: manage_host
+  loop: "{{ _libvirt_manager_networking.instances | dict2items }}"
+  loop_control:
+    index_var: index
+    label: "{{ item.key }}"
+
+- name: Create group inventories
```
hjensas, line 247: This should filter on the vms being `managed: false` - i.e don't add the ironic nodes that break the neutron whitebox tests?
Ref: https://github.com/openstack-k8s-operators/ci-framework/pull/1860

cjeanner:
#### How to test
- Get the PR content locally (`gh pr checkout 1777`)
- Update the collections calling `ansible-galaxy collection install --force .` from the repository root
- Ensure you set `cpus`, `memory`, `disksize`, `uefi: true` and `root_part_id: 4` in the ocp/ocp_worker nodes
- Ensure you're passing the proper parameter file to sync the repository content onto controller-0 (check bellow)
- BE SURE to disable DHCP and DNS services in the libvirt networks, if you're not consuming the provided `scenarios/reproducers/*.yml` (check the changes in the 3-nodes.yml for instance)
- It is strongly advised to pass `--flush-cache` when you deploy
- It is strongly advised to run against an empty hypervisor (no pre-existing OCP images)

Ensure you copy the repository code onto controller-0 (consider updating `local_base_dir` accordingly to your env)
```YAML
---
local_home_dir: "{{ lookup('env', 'HOME') }}"
local_base_dir: "{{ local_home_dir }}/src/github.com/openstack-k8s-operators"
remote_base_dir: "/home/zuul/src/github.com/openstack-k8s-operators"
cifmw_reproducer_repositories:
  - src: "{{ local_base_dir }}/ci-framework"
    dest: "{{ remote_base_dir }}/"
```

#### Migrate cifmw_devscripts_config_overrides
The following parameter must be moved to another dict:
```YAML
cifmw_devscripts_config_overrides:
  vm_extradisks: "true"
  vm_extradisks_list: "vda vdb"
  vm_extradisks_size: "20G"
```
would become, in your scenario file (for instance scenarios/reproducers/my-scenario.yml):
```YAML
cifmw_libvirt_manager_configuration:
  vms:
    ocp:
      extra_disks_num: 2
      extra_disks_size: 20G
[ rest of the usual content for the whole scenario]
```
(this is in case you edit an existing scenario!)

You can also *patch* the libvirt_manager_configuration:
```YAML
cifmw_libvirt_manager_configuration_patch_01_add_ocp_volumes:
  vms:
    ocp:
      extra_disks_num: 2
      extra_disks_size: 20G
```

The `external_subnet_v4: 192.168.32.0/20` must be integrated in the network configuration:
```YAML
cifmw_libvirt_manager_configuration:
  networks:
    ocpbm: |-
      <network>
        <name>ocpbm</name>
        <forward mode='nat'/>
        <bridge name='ocpbm' stp='on' delay='0'/>
        <dns enable="no"/>
        <ip family='ipv4' address='192.168.32.1' prefix='20'>
        </ip>
      </network>
[...]
```

cjeanner:
recheck

cjeanner:
/hold

waiting for final go.

roles/libvirt_manager/tasks/deploy_layout.yml
cjeanner: added the `when` condition

PR #1779, opened 2024-05-29
[networking_mapper] Fix the way ranges are combined in group templates

pablintino:


psathyan (approved):

cjeanner:
recheck

cjeanner:
/approve

I could test with my current iterations - it will also be put to test with #1777 and the reproducer molecule test - that one should switch green.

cjeanner:
Molecule test for reproducer is passing in #1777 with this patch - so we can go with it!

rebtoor (approved):

PR #1780, opened 2024-05-29
Re-add the host IPs in controller-0 .ssh/config, revert hostname

cjeanner:
There's some magical things between .ssh/config and ansible that were
broken for HCI, when time comes to deploy ceph using the ceph.yml
playbook.

In parallel, this patch also reverts the hostname of the VMs to its
former "short" format (compute-0, etc) instead of a more FQDN-like
one (compute-1.hypervisor-1, etc).
That change was mostly needed for sushy since it consumes, by default,
the `ansible_host`.
Injecting that value in the /etc/hosts makes it easier to consume.

### DIFF
--- a/roles/libvirt_manager/tasks/manage_vms.yml
+++ b/roles/libvirt_manager/tasks/manage_vms.yml
@@ -171,7 +171,7 @@
     cmd: |-
       test -d /home/zuul && exit 0;
       set -xe -o pipefail;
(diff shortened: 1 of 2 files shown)


cjeanner:
recheck  issue was with infra, not the PR.

pablintino:
/lgtm
HCI VA job ran without issues with the fix

PR #1783, opened 2024-05-29
Fix some hardcoded repository locations

viroel:
While deploying with a different repository location we found these 2 places that have code location hardcoded. This PR allow these commands to check for cifmw known vaiables for these specific repositories.

  - [X] README in the role

### DIFF
--- a/roles/reproducer/tasks/configure_controller.yml
+++ b/roles/reproducer/tasks/configure_controller.yml
@@ -379,9 +379,8 @@
           cifmw_architecture_automation_file: >-
             {{
               (
-                ansible_user_dir,
-                'src/github.com/openstack-k8s-operators',
-                'architecture/automation/vars',
+                cifmw_architecture_repo | default(ansible_user_dir+'/src/github.com/openstack-k8s-operators/architecture'),
+                'automation/vars',
                 cifmw_arch_automation_file | default('default.yaml')
               ) | ansible.builtin.path_join
             }}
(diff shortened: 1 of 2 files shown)


roles/reproducer/tasks/configure_controller.yml
```diff
@@ -379,9 +379,8 @@
           cifmw_architecture_automation_file: >-
             {{
               (
-                ansible_user_dir,
-                'src/github.com/openstack-k8s-operators',
-                'architecture/automation/vars',
+                cifmw_architecture_repo | default(ansible_user_dir+'/src/github.com/openstack-k8s-operators/architecture'),
```
cjeanner, line 382: ansible_user_dir will reflect the value from the hypervisor - not a big deal in CI since we're using `zuul` everywhere, but in custom labs.... that will crash.

cjeanner (changes requested):

roles/reproducer/tasks/configure_controller.yml
viroel: so it is already broken for them? ansible_user_dir is already in the code. no?

cjeanner: oh - dang, I missed the fact it was between {% raw %} - meaning it's all good. Sorry for that one!

cjeanner (approved):

lewisdenny (approved):
/approve

Thanks Doug!

PR #1786, opened 2024-05-29
Fix '_sushy_emulator_available' var when cifmw_use_sushy_emulator=false

abays:
  - [x] README in the role

Without this, if  `cifmw_use_sushy_emulator` is set to `false`, then the `_sushy_emulator_available` var is undefined and leads to an error because [1] causes Ansible to skip setting the latter variable.

[1] https://github.com/openstack-k8s-operators/ci-framework/blob/eaff04f2e592f98da8f38a84ffddf909d5dddaab/roles/reproducer/tasks/configure_controller.yml#L314


hjensas (approved):

lewisdenny (approved):
/approve

Thanks @abays <3

PR #1787, opened 2024-05-30
nit: Remove redundant ansible-galaxy_install

lewisdenny:
We aimed to fix issues like [1] in https://github.com/openstack-k8s-operators/ci-framework/pull/1539 but e install the dependency twice.

Once in setup_molecule[2] (which has great retries thanks to @Cédric Jeanneret++) but then again in run_ansible_test [3] due to the --force.

The two paths we support are using `make run_ctx_ansible_test` locally and `cifmw-pod-ansible-test` job in CI.

Both of which already install dependencies before calling  `run_ansible_test`

[1] ERROR! Failed to clone a Git repository from `https://github.com/ansible-collections/community.general`.
[2] https://github.com/openstack-k8s-operators/ci-framework/blob/main/scripts/setup_molecule#L51
[3] https://github.com/openstack-k8s-operators/ci-framework/blob/main/scripts/run_ansible_test#L41

### DIFF
--- a/containerfiles/Containerfile.ci
+++ b/containerfiles/Containerfile.ci
@@ -4,6 +4,6 @@ LABEL summary="Provides root image for openstack-k8s-operators projects in Prow"
 USER root
 # workaround for RHBZ#2184640
 RUN rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-centosofficial
-RUN dnf update -y && dnf install -y git python3-pip python3-netaddr make gcc sudo rsync && yum clean all
+RUN dnf update -y && dnf install -y git python3-pip python3-netaddr make gcc sudo rsync && dnf clean all
 RUN mkdir -p /home/zuul/src/github.com/openstack-k8s-operators
 RUN git clone https://github.com/openstack-k8s-operators/install_yamls /home/zuul/src/github.com/openstack-k8s-operators/install_yamls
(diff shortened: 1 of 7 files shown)


lewisdenny:
setup_molecule and ansible test are installing things in different locations due to $HOME being overridden

containerfiles/Containerfile.ci
```diff
@@ -7,3 +7,6 @@ RUN rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-centosofficial
 RUN dnf update -y && dnf install -y git python3-pip python3-netaddr make gcc sudo rsync && yum clean all
 RUN mkdir -p /home/zuul/src/github.com/openstack-k8s-operators
 RUN git clone https://github.com/openstack-k8s-operators/install_yamls /home/zuul/src/github.com/openstack-k8s-operators/install_yamls
+RUN dnf clean all
```
pablintino, line 10: Just asking, but we have a `yum clean all`, that in most new RHEL based distros  is a symlink to dnf, why not just change the previous `yum clean all` with this one?

lewisdenny: Yeah no reason, I missed that `yum clean all` on line 7 and will just update it to dnf thanks :)

lewisdenny:
eh, this whole thing is a mess the more I look. Going to purpose a simple change to improve CI and revisit this issue in another patch.

PR #1791, opened 2024-05-30
Add override-checkout option to cifmw_reproducer_default_repositories

raukadah:
In downstream, Devs uses cifmw_reproducer_default_repositories var to clone downstream repositories on the controller-0. The default branch `main` of these downstream repos does not contains any repo/code files, which causes issues while running VA deployment due to missing files.

Defining cifmw_reproducer_default_repositories with override-checkout will allow users to checkout specific version/branch code on controller-0.

override-checkout key is optional.

Extended syntax will look like this:
cifmw_reproducer_default_repositories:
  - src: <repo_url>
    dest: <repo_dest>
    override-checkout: <branch name>

### DIFF
--- a/roles/reproducer/tasks/default_repositories.yml
+++ b/roles/reproducer/tasks/default_repositories.yml
@@ -13,3 +13,4 @@
       ansible.builtin.git:  # noqa: latest[git]
         dest: "{{ repository.dest }}"
         repo: "{{ repository.src }}"
+        version: "{{ repository['override-checkout'] | default(omit) }}"


roles/reproducer/tasks/default_repositories.yml
```diff
@@ -13,3 +13,4 @@
       ansible.builtin.git:  # noqa: latest[git]
         dest: "{{ repository.dest }}"
         repo: "{{ repository.src }}"
+        version: "{{ omit if 'override-checkout' not in repository else repository['override-checkout'] }}"
```
cjeanner, line 16: can't we use `{{ repository['override-checkout'] | default(omit) }}` ?

rebtoor (approved):

PR #1795, opened 2024-05-30
Add missing 'https://' to cifmw_ceph_rgw_keystone_ep

fultonj:
The default value for the `cifmw_ceph_rgw_keystone_ep` variable in the role `cifmw_cephadm` was missing the `https://` part of the keystone endpoint URL so RGW was not getting configured with a correct keystone endpoint.

  - [x] README in the role

### DIFF
--- a/roles/cifmw_cephadm/defaults/main.yml
+++ b/roles/cifmw_cephadm/defaults/main.yml
@@ -66,7 +66,7 @@ cifmw_cephadm_pacific_filter: "16.*"
 # The path of the rendered rgw spec file
 cifmw_ceph_rgw_spec_path: /tmp/ceph_rgw.yml
 cifmw_ceph_mds_spec_path: /tmp/ceph_mds.yml
-cifmw_ceph_rgw_keystone_ep: "keystone-internal.openstack.svc:5000"
+cifmw_ceph_rgw_keystone_ep: "https://keystone-internal.openstack.svc:5000"
 cifmw_ceph_rgw_keystone_psw: 12345678
 cifmw_ceph_rgw_keystone_user: "swift"
 cifmw_ceph_rgw_config:


fultonj:
recheck

roles/cifmw_cephadm/defaults/main.yml
```diff
@@ -66,7 +66,7 @@ cifmw_cephadm_pacific_filter: "16.*"
 # The path of the rendered rgw spec file
 cifmw_ceph_rgw_spec_path: /tmp/ceph_rgw.yml
 cifmw_ceph_mds_spec_path: /tmp/ceph_mds.yml
-cifmw_ceph_rgw_keystone_ep: "keystone-internal.openstack.svc:5000"
+cifmw_ceph_rgw_keystone_ep: "https://keystone-internal.openstack.svc:5000"
```
fmount, line 69: `https` here is added to the internal keystone endpoint: if tls is not enabled at Pod level (`podLevel: true` in the ctlplane) but the ingress is configured as an edge router [1] this will fail.
While I think https should be the default, we should isolate the prefix in it's own variable.
I'm ok to follow up on that, but I would try something like:

```
cifmw_ceph_rgw_keystone_ep: "{{ cifmw_ceph_rgw_keystone_urischeme | default("https") }}:keystone-internal.openstack.svc:5000"
```
so we can make it configurable w/o introducing extra complexity.

[1] https://www.redhat.com/architect/encryption-secure-routes-openshift

fmount: However, considering you can override the entire variable, my comment is not a blocker for the patch, more of a consideration.
So let me know if you want to keep it as it is and I can approve.

fmount (approved):
/lgtm

fmount:
Let's rebase/recheck and merge this patch to and see how CI behaves. I'm not sure it's enough to solve the current problem but given we have jobs w/ tls enabled at pod level I think we need this.

mandre:
I have tested this patch in my local environment and I confirm it allows me to use RGW with `enableSecureRBAC: true`. This fixes https://issues.redhat.com/browse/OSPRH-6199.

recheck

fmount:
recheck

PR #1798, opened 2024-05-31
Add tasks to serve gating repo over content provider

raukadah:
The build_openstack_packages role will build the packages, creates the gating repo and then serves the gating repo over simple http server.
    
The repo-setup role will run on the dependent job and populate gating repo on the controller. Gating repo will be used to pull build rpms from content provider.

Tested here: https://github.com/os-net-config/os-net-config/pull/20#issuecomment-2146841767

### DIFF
--- a/roles/build_openstack_packages/tasks/create_repo.yml
+++ b/roles/build_openstack_packages/tasks/create_repo.yml
@@ -51,6 +51,9 @@
       priority=1
     dest: "{{ cifmw_bop_gating_repo_dest }}/gating.repo"
 
+- name: Serve gating repo
+  ansible.builtin.import_tasks: serve_gating_repo.yml
+
 - name: Check for DLRN repo at the destination
   ansible.builtin.stat:
     path: "{{ cifmw_bop_gating_repo_dest }}/delorean.repo"
--- a/roles/build_openstack_packages/tasks/serve_gating_repo.yml
+++ b/roles/build_openstack_packages/tasks/serve_gating_repo.yml
@@ -1,0 +1,35 @@
+---
+- name: Install python-psutil
+  become: true
+  ansible.builtin.package:
+    name: python-psutil
+
+- name: Getting process IDs of the python http server
+  community.general.pids:
+    pattern: "python -m http.server {{ cifmw_bop_gating_port }}"
+  register: _pids_of_python
+
+- name: Printing the process IDs obtained
+  ansible.builtin.debug:
+    msg: "{{ _pids_of_python.pids }}"
+
+- name: Force kill the running process
+  ansible.builtin.command: "kill -9 {{ item }}"
+  loop: "{{ _pids_of_python.pids }}"
+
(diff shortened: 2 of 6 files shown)


raukadah:
rrecheck

raukadah:
recheck

roles/build_openstack_packages/tasks/serve_gating_repo.yml
```diff
@@ -0,0 +1,35 @@
+---
+- name: Install python-psutil
+  become: true
+  ansible.builtin.package:
+    name: python-psutil
+
+- name: Getting process IDs of the python http server
+  community.general.pids:
+    pattern: "python -m http.server"
```
marios, line 9: haven't played with that before, but wondering can we make it more specific like include the port number here, so we can be sure to only kill the one we are interested in?

```
pattern: "python -m http.server 8766"

```

marios (approved):
seems reasonable, just a thought about the pids search but maybe unnecessary

roles/build_openstack_packages/tasks/serve_gating_repo.yml
raukadah: Done

roles/build_openstack_packages/tasks/create_repo.yml
```diff
@@ -51,6 +51,9 @@
       priority=1
     dest: "{{ cifmw_bop_gating_repo_dest }}/gating.repo"
 
+- name: Serve gating repo
+  ansible.builtin.import_tasks: serve_gating_repo.yml
+
```
marios, line 56: for future maintenance we could consider adding vars: to this import so we can add 8766 as param avoiding repetition there in serve_gating_repo.yaml
```
- name: Serve gating repo
  ansible.builtin.import_tasks: serve_gating_repo.yml
  vars:
      gating_port: 8766
```
```

File: roles/build_openstack_packages/tasks/serve_gating_repo.yml [hunk 3]
```diff
@@ -0,0 +1,35 @@
+---
+- name: Install python-psutil
+  become: true
+  ansible.builtin.package:
+    name: python-psutil
+
+- name: Getting process IDs of the python http server
+  community.general.pids:
+    pattern: "python -m http.server 8766"
```
marios, line 9: could be 

```
"python -m http.server {{ gating_port }}"
```

with my other comment

marios (approved):
lgtm and clicking approve

just a thought inline and really up to you. Maybe overkill because we probably won't ever change 8766, but, maybe a bit cleaner for future generations of engineers to maintain ;)

roles/build_openstack_packages/tasks/create_repo.yml
raukadah: Thank you @marios for the suggestion!. Sounds good.

rebtoor (approved):

PR #1799, opened 2024-05-31
Move and add retries to ci_local_storage

pablintino:
SSH connection when fetching hostnames seems to be lost randomly. Add SSH retries and moving the role before OSP operators deployment may help the connection role to succeed.

  - [x] README in the role


cjeanner:
/approve
/lgtm

no real way to assert it's fixing things - it's a random issue, and only "many runs" will show if it's fixed or not.

PR #1801, opened 2024-05-31
dnsmasq - use dhcp-hostdir

hjensas:
The restart done by the handler happens only at the very end of the play. By using dhcp-hostsdir we get the benefit of dnsmasq reading any new or changed file in the hosts dir, automatically applying the new dhcp host config on the fly.

Since adding a host is loaded on-the-fly - we don't need to notify the handler when adding hosts.

Removing hosts however, reqires s SIGHUP - so for that case notify the handler.

Also change the computation of _host_entry - it should include both ipv4 and ipv6 address if both are configured.

  - [x] README in the role

### DIFF
--- a/roles/dnsmasq/templates/cifmw-dnsmasq.conf.j2
+++ b/roles/dnsmasq/templates/cifmw-dnsmasq.conf.j2
@@ -1,5 +1,4 @@
 # Managed by ci-framework/dnsmasq
-
 user=dnsmasq
 group=dnsmasq
 pid-file=/var/run/cifmw-dnsmasq.pid
@@ -10,3 +9,4 @@ dhcp-leasefile=/var/lib/dnsmasq/cifmw-dnsmasq.leases
 {% endif %}
 
 conf-dir={{ cifmw_dnsmasq_basedir }},*.conf
+dhcp-hostsdir="{{ cifmw_dnsmasq_basedir }}/dhcp-hosts.d"
--- a/roles/dnsmasq/tasks/configure.yml
+++ b/roles/dnsmasq/tasks/configure.yml
@@ -38,7 +38,7 @@
   become: true
   ansible.builtin.file:
     mode: "0755"
-    path: "{{ cifmw_dnsmasq_basedir }}"
+    path: "{{ cifmw_dnsmasq_basedir }}/dhcp-hosts.d"
     state: "{{ (_act == 'cleanup') | ternary('absent', 'directory') }}"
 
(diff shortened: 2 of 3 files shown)


roles/dnsmasq/templates/cifmw-dnsmasq.conf.j2
```diff
@@ -10,3 +10,4 @@ dhcp-leasefile=/var/lib/dnsmasq/cifmw-dnsmasq.leases
 {% endif %}
 
 conf-dir={{ cifmw_dnsmasq_basedir }},*.conf
+dhcp-hostsdir="{{ cifmw_dnsmasq_basedir }}/dhcp-hosts.d"
```
cjeanner, line 13: we can trigger the reload whenever we want/need using `ansible.builtin.meta: flush_handlers`
This is what's done in my branches.

hjensas: I am aware of that. But that is something that will be easy for consumers to forget to do. And I think it will make up for some hard-to-diagnose issues. For example if you add a dhcp-host in your play, then start the VM and expect it to get an address and try to perform operations against it - the play fails but soon after the play fails and the handler ran the VM become accessible - to the user that is not going to be easy to understand without a black belt in Ansible.

I actually think we should add the `flush_handler` thing at the end of all the `manage` task files. To enable adding multiple networks and do one restart we can change the "manage_network" interface to use a list like the `manage_host_record.yml`, `manage_forwarder.yml` etc does. But I don't want to do that now - because changing the interface will break your patches.

I am adding reload functionality in https://github.com/openstack-k8s-operators/ci-framework/pull/1801. Essentially "manage_host" can notify the "Reload" handler, while "manage_network", "manage_host_record", "manage_address" etc notify the "Restart" handler.

hjensas:
recheck

roles/dnsmasq/templates/cifmw-dnsmasq.conf.j2
```diff
@@ -1,5 +1,5 @@
 # Managed by ci-framework/dnsmasq
-
+log-debug
```
cjeanner, line 2: we may want to toggle this instead of having it on by default? You can easily pass it as a parameter in your env.
Or do you think debug logging all the time is really needed?

hjensas: Removed the log-debug.

PR #1804, opened 2024-05-31
polarion: update the usage of cifmw_polarion_update_testcases

tosky:
cifmw_polarion_update_testcases is named after a parameter of a tool and it doesn't match its behavior: it is about adding missing test cases when they don't exist. But the "normal" workflow is to pre-populate the test cases and only update (for real) their results, so that variable should almost never be enabled.

Also, remove the check about the variable being enabled, as its only usage is guarded by a default() filter.

  - [x] README in the role

### DIFF
--- a/roles/polarion/tasks/main.yml
+++ b/roles/polarion/tasks/main.yml
@@ -19,7 +19,6 @@
     that:
       - cifmw_polarion_jump_repo_url is defined and cifmw_polarion_jump_repo_url != ""
       - cifmw_polarion_testrun_id is defined and cifmw_polarion_testrun_id != ""
-      - cifmw_polarion_update_testcases is defined and cifmw_polarion_update_testcases | bool
 
 - name: Ensure deps are installed
   become: true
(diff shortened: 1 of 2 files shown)


jparoly (approved):

roles/polarion/tasks/main.yml
kopecmartin, line 1: there is one more occurrence of cifmw_polarion_update_testcases on line 99

tosky: if it's about 
I'm leaving it there just in case it may be needed for some special case, but of course I can remove it just to avoid any confusion (there is always cifmw_polarion_jump_extra_vars in case)

kopecmartin: right i see, let's leave it there, it's more practical if we need it later

kopecmartin (approved):
lgtm

tosky:
Internally tested, nothing seems to explode, move to "ready for review"

PR #1819, opened 2024-06-03
Allow cifmw_use_lvms in the reproducer variables file

fultonj:
Update the `reproducer` role to allow the `cifmw_use_lvms` variable to be in the reproducer-variables.yml file.

The `cifmw_use_lvms` variable needs to be in this file so that when the `deploy-edpm.yml` playbook is run by the `deploy-architecture.sh` script it opts to use LVMS and not local-storage. The filter in the `reproducer` role, without this patch, prevents LVMS from getting deployed.

The `rejectattr('key', 'match', '^cifmw_use_(?!lvms).*')` regex rejects all keys starting with `cifmw_use_` except `cifmw_use_lvms`. This fixes LVMS without adding other variables matching `cifmw_use_` to the file.

Jira: https://issues.redhat.com/browse/OSPRH-7424

  - [x] README in the role

### DIFF
--- a/roles/reproducer/tasks/configure_controller.yml
+++ b/roles/reproducer/tasks/configure_controller.yml
@@ -355,7 +355,7 @@
             rejectattr('key', 'equalto', 'cifmw_openshift_kubeconfig') |
             rejectattr('key', 'equalto', 'cifmw_openshift_token') |
             rejectattr('key', 'equalto', 'cifmw_set_openstack_containers_registry') |
-            rejectattr('key', 'match', '^cifmw_use.*') |
+            rejectattr('key', 'match', '^cifmw_use_(?!lvms).*') |
             rejectattr('key', 'match', '^cifmw_reproducer.*') |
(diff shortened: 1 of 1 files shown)


sdatko:
Just for the record, there are the following variables I found in CI-Framework and jobs repositories:
```
[sdatko@polluks RedHat]$ grep -Erio 'cifmw_use[^ `:]*' . | cut -d ':' -f 2- | sort | uniq
cifmw_use_crc
cifmw_use_devscripts
cifmw_use_hive
cifmw_use_libvirt
cifmw_use_libvirt_manager
cifmw_use_lvms
cifmw_use_opn
cifmw_use_reproducer
cifmw_use_sushy_emulator
cifmw_use_uefi
cifmw_use_vbmc
```
(I was curious if there is anything other than `_` after `use`, like `user_something`, etc.)

abays (approved):
/lgtm

PR #1828, opened 2024-06-04
Skip validations in edpm deploy for arch deployment

bshephar:
When running an architecture deployment, most of the control plane and dataplane setup are skipped in 06-edpm-deploy.yml. Instead, it all gets configured in the 06-deploy-architecture.yml playbook. This means that we need to also skip validations in the 06-epdm-deploy.yml since there is nothing to validate at this stage. We will instead fall back to the validations role execution in the architecture playbook.

  - [x] README in the role


lewisdenny (approved):
/approve

bshephar:
recheck

PR #1833, opened 2024-06-05
Newer CentOS image converge with RHEL image

cjeanner:
Newer images are all exposing UEFI, meaning the root partition ID is the
fourth for both families.


lewisdenny (approved):
/lgtm

PR #1815, opened 2024-06-03
Get OpenStackControlPlane to update the OpenStackVersion

fao89:
As a pull request owner and reviewers, I checked that:
- [x] Appropriate testing is done and actually running
- [x] Appropriate documentation exists and/or is up-to-date:
  - [x] README in the role
  - [x] Content of the docs/source is reflecting the changes

Depends-On: https://github.com/openstack-k8s-operators/data-plane-adoption/pull/480

PR #1835, opened 2024-06-05
Use the same OpenStackVersion CR name

rabi:
When building containers if we don't use the same OpenStackVersion name as OpenStackControlPlane then multiple OpenStackVersion CRs can be created and subsequent dataplane deployment would fail.


rabi:
Looks like we already have a fix https://github.com/openstack-k8s-operators/ci-framework/pull/1815 for it and it seems better as there is no assumption for the OpenStackControlPlane/OpenStackVersion name.

PR #1836, opened 2024-06-05
Introduce new cifmw_nolog global parameter

cjeanner:
Allow to toggle the no_log value for selected tasks (mostly command,
shell and related modules).

Ceph related content are NOT following that parameter for security
reasons.


lewisdenny (approved):
/lgtm
Thanks for the clarification on Ceph tasks.

cescgina (commented):
/approve

PR #1837, opened 2024-06-05
[networking_mapper] Make null templates behave like an empty one

pablintino:
Now, if the data inside a template network is null the mapper will raise an unhandled exception, that may be fine, but for this particular case making null and empty behave the same way will allow the mapper to raise a meaningful error letting him know the field cannot be empty as some child elements are mandatory.

  - [x] README in the role


cjeanner (commented):
/approve

rebtoor (approved):

PR #1842, opened 2024-06-05
Fix sushy_emulator integration in reproducer

jcapiitao:
It's a followup of #1753.
We get rid of realpath filter which resolves the path using the controller's filesystem [1]. When the filesystems don't match between the controller and the remote machine, we might hit somme issues. For instance, when running controller on a Fedora CoreOS system, the `/home` gets resolved as `/var/home` by realpath, leading to incorrect path in remote machine.
Without `realpath`, we end up with the `..` in the path which is a bit uggly but we keep the consistency.

Also, we set the hypervisor ansible user used as the emulator libvirt user in order to have a more dynamic configuration, otherwise it fails by default if hypervisor ansible user is not `zuul` (i.e default user).

[1] https://docs.ansible.com/ansible/latest/collections/ansible/builtin/realpath_filter.html#synopsis

  - [x] README in the role

### DIFF
--- a/roles/reproducer/tasks/configure_controller.yml
+++ b/roles/reproducer/tasks/configure_controller.yml
@@ -25,9 +25,10 @@
     cifmw_sushy_emulator_sshkey_path: >-
       {{
         [_ctl_reproducer_basedir, '../.ssh/sushy_emulator-key'] |
-        path_join | realpath
+        path_join
       }}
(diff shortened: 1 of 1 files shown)


lewisdenny (approved):
/lgtm

abays (approved):
/lgtm
/approve

PR #1843, opened 2024-06-05
Fix image pushing to local registry

sshnaidm:
With new podman_image one must specify image in push_args in order to push it to local registry. It doesn't work currently because of podman trying to push it to
"docker.io/library/localhost" taking "localhost" as an image name.
Specify exact image name in push args when registry doens't have any org, only IP.

https://issues.redhat.com/browse/OSPCIX-319

  - [x] README in the role


son-vyas:
lgtm

PR #1852, opened 2024-06-07
Add steps to run minor version update

ciecierski:
Integrate set_openstack_container role and install_yaml openstack_update_run target to run openstack minor update.

This pr introduces following changes:
 - Adjust update playbook to use new vars from set_openstack_containers role
 - Run openstack minor update using openstack_update_run target from install_yaml
 - Add molecule testing to test update role in dryrun mode

  - [x] README in the role

### DIFF
--- a/roles/repo_setup/tasks/configure.yml
+++ b/roles/repo_setup/tasks/configure.yml
@@ -5,6 +5,7 @@
   when:
     - content_provider_dlrn_md5_hash is defined
     - content_provider_dlrn_md5_hash | length > 0
+    - (not cifmw_run_update|default(false)) or (update_playbook_run is defined and cifmw_run_update|default(false))
 
 - name: Run repo-setup
   become: "{{ not cifmw_repo_setup_output.startswith(ansible_user_dir) }}"
--- a/roles/update/README.md
+++ b/roles/update/README.md
@@ -3,5 +3,6 @@ Role to run update
 
 ## Parameters
 * ``cifmw_update_extras`: (hash) Hold job variable that get set when running the update playbook.
-
-## Examples
+* ``cifmw_update_openstack_update_run_operators_updated`: (Boolean) Set if openstack_update_run make target should not modify openstack-operator csv to fake openstack services container change. Default to `True`.
+* ``cifmw_update_openstack_update_run_target_version`: (String) Define openstack target version to run update to.
+* `cifmw_update_run_dryrun`: (Boolean) Do a dry run on make openstack_update_run command. Defaults to `False`.
--- a/roles/update/defaults/main.yml
+++ b/roles/update/defaults/main.yml
@@ -17,3 +17,10 @@
 
 # All variables intended for modification should be placed in this file.
 # All variables within this role should have a prefix of "cifmw_update"
+
+cifmw_update_openstack_update_run_target_version: "0.0.2"
+cifmw_update_openstack_update_run_operators_updated: true
+cifmw_update_openstack_update_run_containers_namespace: "podified-antelope-centos9"
+cifmw_update_openstack_update_run_containers_target_tag: "current-podified"
+
+cifmw_update_run_dryrun: false
--- a/roles/update/tasks/main.yml
+++ b/roles/update/tasks/main.yml
@@ -14,6 +14,25 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
-- name: Hello World
-  ansible.builtin.debug:
-    msg: "Hello World from update"
+- name: Set openstack_update_run Makefile environment variables
+  tags:
+    - always
+  ansible.builtin.set_fact:
+    _make_openstack_update_run_params: |
+      {% if not cifmw_update_openstack_update_run_operators_updated -%}
+      FAKE_UPDATE: true
+      CONTAINERS_NAMESPACE: {{ cifmw_update_openstack_update_run_containers_namespace }}
+      CONTAINERS_TARGET_TAG: {{ cifmw_update_openstack_update_run_containers_target_tag }}
+      OPENSTACK_VERSION: {{ cifmw_update_openstack_update_run_target_version }}
+      {% else -%}
+      OPENSTACK_VERSION: {{ cifmw_update_openstack_update_run_target_version }}
+      {% endif -%}
+
+- name: Run make openstack_update_run
+  vars:
+    make_openstack_update_run_env: "{{ cifmw_install_yamls_environment | combine({'PATH': cifmw_path }) }}"
+    make_openstack_update_run_params: "{{ _make_openstack_update_run_params | from_yaml }}"
+    make_openstack_update_run_dryrun: "{{ cifmw_update_run_dryrun | bool }}"
+  ansible.builtin.include_role:
+    name: 'install_yamls_makes'
+    tasks_from: 'make_openstack_update_run'
--- a/playbooks/update.yml
+++ b/playbooks/update.yml
@@ -11,11 +11,12 @@
       ansible.builtin.include_role:
         name: repo_setup
 
-    - name: Update all openstack services containers env vars in meta operator with tag from delorean
+    - name: Update all openstack services containers env vars in meta operator with tag from delorean and update OPENSTACK_RELEASE_VERSION
       vars:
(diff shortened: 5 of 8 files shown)


ciecierski:
Testing here: https://review.rdoproject.org/r/c/testproject/+/47164 
Test is failing

roles/update/tasks/main.yml
```diff
@@ -14,6 +14,12 @@
...
-  ansible.builtin.debug:
-    msg: "Hello World from update"
+- name: Run make openstack_update_run
+  ansible.builtin.include_role:
+    name: 'install_yamls_makes'
+    tasks_from: 'make_openstack_update_run'
+  vars:
+    make_openstack_update_run_params:
+      FAKE_UPDATE: false
+      VERSION: "0.0.2"
```
sathlan, line 24: We should be able to reach those parameters from the job definition and as well those:

```
CONTAINERS_NAMESPACE=${CONTAINERS_NAMESPACE:-podified-antelope-centos9}
CONTAINERS_TARGET_TAG=${CONTAINERS_TARGET_TAG:-current-podified}
FAKE_UPDATE=${FAKE_UPDATE:-false}
OPENSTACK_VERSION=${OPENSTACK_VERSION:-0.0.2}
TIMEOUT=${TIMEOUT:-300s}
```

Hum, and it seems the install_yaml [review](https://github.com/openstack-k8s-operators/install_yamls/pull/843) is  using OPENSTACK_VERSION in the install_yaml

So either new variables (maybe not the timeout one ...) for each those in the update role or would be best as they are kinda important variable and not related to deployment.

Maybe `CONTAINERS_TARGET_TAG` can be consolidated with what we have for `cifmw_repo_setup_promotion` in `cifmw_update_extras` in the job definition ?

Also look at [that comment in the ci-framework review](https://github.com/openstack-k8s-operators/ci-framework/pull/1963#pullrequestreview-2151312090)

WDYT?

sathlan (changes requested):
We should make sure that all the install yaml variables are reachable from the job definition I think.

roles/repo_setup/tasks/configure.yml
```diff
@@ -5,6 +5,7 @@
   when:
     - content_provider_dlrn_md5_hash is defined
     - content_provider_dlrn_md5_hash | length > 0
+    - not cifmw_run_update|default(false)
```
sathlan, line 8: Hum... so we don't want to consume the content_provider hash during update.  So what about jobs that might use content provider, won't that be a problem ?

roles/update/defaults/main.yml
```diff
@@ -17,3 +17,8 @@
 
 # All variables intended for modification should be placed in this file.
 # All variables within this role should have a prefix of "cifmw_update"
+
+cifmw_update_openstack_update_run_target_version: "0.0.2"
+cifmw_update_openstack_update_run_operators_updated: false
+cifmw_update_openstack_update_run_containers_namespace: "podified-antelope-centos9"
+cifmw_update_openstack_update_run_containers_target_tag: "current-podified"
```
sathlan, line 24: Please add them to the readme.

Maybe we can add `cifmw_update_openstack_update_run_operators_updated` as well to the readme even if it's a calculated variable that assert if we need to fake the operator update or not.

sathlan (changes requested):
We need to beef up the readme I think so that we can understand this in 6 months.

roles/repo_setup/tasks/configure.yml
ciecierski: For some jobs we may want to use content provider, but not during deployment phase.
For deployment we need to fetch stable content published and use content provider only during update playbook run. cifmw_run_update is defined in update job under vars, so I need to move it to new update extras vars

raukadah (approved):
/approve

Looks good, not going to affect existing repo-setup functionality.

raukadah:
recheck

roles/repo_setup/tasks/configure.yml
```diff
@@ -5,6 +5,7 @@
   when:
     - content_provider_dlrn_md5_hash is defined
     - content_provider_dlrn_md5_hash | length > 0
+    - (not cifmw_run_update|default(false)) or (update_playbook_run is defined and cifmw_run_update|default(false))
```
marios, line 8: its a bit confusing

if you are not running an update OR
if you ARE running update AND the update playbook is defined
?
like that?

but what is the case that you are running an update and the update playbook is not defined?

roles/update/defaults/main.yml
```diff
@@ -17,3 +17,8 @@
 
 # All variables intended for modification should be placed in this file.
 # All variables within this role should have a prefix of "cifmw_update"
+
+cifmw_update_openstack_update_run_target_version: "0.0.2"
+cifmw_update_openstack_update_run_operators_updated: true
+cifmw_update_openstack_update_run_containers_namespace: "podified-antelope-centos9"
+cifmw_update_openstack_update_run_containers_target_tag: "current-podified"
```
marios, line 24: update_openstack_update

is that really necessary?

but this is totally subjective and this is just me thinking out loud not blocking on any naming preference here...

marios (approved):
i have not much context and was not involved in discussions around this minor update, beyond some familiarity with the OpenStackVersion resource.

some comments inline but setting approve as this is not wired up yet to any CI job and is needed to allow further testing and refinement.

roles/repo_setup/tasks/configure.yml
ciecierski: cifmw_run_update - is defined on job level, meaning this job is testing update. In that case during deployment phase don't use images from content_provider, but stable/promoted containers instead. 

update_playbook_run is defined - when the role is triggered from within ci/update_playbook, if content_provider is available use during update those containers

Hope it helps

ciecierski:
recheck

ciecierski:
recheck

raukadah:
@ciecierski thank you for updating the pr. I assume `[cifmw-molecule-update](https://review.rdoproject.org/zuul/build/62cdefb09b44438f96bfa438a7505495)` needs a crc nodeset to run the install_yamls target.

In order to use crc nodeset in the molecule job, we need to change the parent of cifmw-molecule-update from `cifmw-molecule-base` to `cifmw-molecule-base-crc`. I think this will fix the existing molecule job failure.

roles/update/tasks/main.yml
ciecierski: it is all passed from ci-framework to install_yaml as parameters 
 CONTAINERS_TARGET_TAG is cifmw_update_openstack_update_run_containers_target_tag

roles/repo_setup/tasks/configure.yml
sathlan: So 

 | run_update | update_playbook_run | result | comment |
 | ----- | ---- | ----- | ---- |
| undefined or false | undefined | true: use content provider if defined as usual | the normal deployment where those variables are just not there, so nothing change|
| true | undefined | false: don't use content provider during deployment | a job that is an update job, but running the deployment playbook |
| true | defined | true : use content provider during update | a job that is an update job and is running the update playbook |

It definitively looks correct to me.

raukadah (approved):

roles/update/README.md
```diff
@@ -3,5 +3,6 @@ Role to run update
 
 ## Parameters
 * ``cifmw_update_extras`: (hash) Hold job variable that get set when running the update playbook.
-
-## Examples
+* ``cifmw_update_openstack_update_run_operators_updated`: (Boolean) Set if openstack_update_run make target should not modify openstack-operator csv to fake openstack services container change. Default to `True`.
+* ``cifmw_update_openstack_update_run_target_version`: (String) Define openstack target version to run update to.
```
sathlan, line 7: Hum ... the double "`" is strange and is introduced in L5, maybe we have to fix them ... in another review ?

roles/update/defaults/main.yml
```diff
@@ -17,3 +17,10 @@
...
```
sathlan, line 24: Shouldn't we push those variables in the readme as well ?

sathlan (approved):
This look good to me as it stands and the comment can be left for another review if deemed necessary.

PR #1855, opened 2024-06-10
Use tlse in adoption standalone no ceph job

cescgina:
Enable tlse for the standalone no ceph job which runs in cifmw,
install_yamls, openstack-operator and edpm-ansible, so we have
the exact same coverage as we do in check.


marios (approved):

raukadah:
recheck

edpm jobs are fixed now!

cescgina:
recheck

cescgina:
recheck

marios:
recheck

should be green now

PR #1857, opened 2024-06-10
Don't run nfs.yml when architecture is defined

rlobillo:
  - [x] README in the role

### DIFF
--- a/playbooks/nfs.yml
+++ b/playbooks/nfs.yml
@@ -21,6 +21,7 @@
     - name: End play early if no NFS is needed
       when:
         - not cifmw_edpm_deploy_nfs | default('false') | bool
+        - cifmw_architecture_scenario is defined
       ansible.builtin.meta: end_play
   vars:
     nftables_path: /etc/nftables


rlobillo:
@tosky Would you mind to check this one? We are hitting problems deploying va-hci because of this.

tosky:
> @tosky Would you mind to check this one? We are hitting problems deploying va-hci because of this.
-2.

Why would there be a problem? The code has been in place for a while, and if the NFS server is not requested, it should be a noop.
Moreover, there are cases where the locally deployed NFS server may be used with the architecture repository.
This should not be tied to the usage of architecture.

tosky (changes requested):
This should not be connected or conditioned based on the usage of architecture, and moreover nfs.yml should be a noop if it's not explicitly requested. This change should not go in.

tosky:
So, it seems the ci-framework code is more complex than I thought and the same condition needs to be repeated more and more times. I'd let the ci-framework maintainers chime in but this may be correct in the end.

cjeanner (changes requested):
Review:
Not OK: this condition should go in this block here:
https://github.com/openstack-k8s-operators/ci-framework/blob/b6823fef04a0527173fe8d6770edd511e9d7725a/playbooks/nfs.yml#L20-L24

It will ensure the playbook is really skipped.

Context:
We can't really apply a `when` to `import_playbook` - it will "just" append that condition to all of the imported tasks, and it might lead to fun situations where the imported play will fail without even being launched...

More over, the `end_play` meta action applies only to the current play - so if we're running a series of playbook, only the one with that action will end, and the run will just jump to the next play in list.

tosky:
recheck

playbooks/nfs.yml
```diff
@@ -21,6 +21,7 @@
     - name: End play early if no NFS is needed
       when:
         - not cifmw_edpm_deploy_nfs | default('false') | bool
```
pablintino, line 23: @rlobillo care to change this by this?
```yaml
 (not cifmw_edpm_deploy_nfs | default('false') | bool) or (cifmw_architecture_scenario is defined)
```
I know the solution is **far** from perfect, but it will do the trick for now till we reorganize the code.

pablintino (changes requested):
I agree with @tosky here: https://github.com/openstack-k8s-operators/ci-framework/pull/1857#pullrequestreview-2108188851

Is this still problematic for you @rlobillo? If not, drop the change, if it's the case, have a look at my comment to fix the error the jobs are showing.

rlobillo:
/close

Not needed right now

PR #1863, opened 2024-06-12
Hooks for bootstrapping ironic

hjensas:
Adds three hook playbooks:
* `ironic_flavor.yml` - to create baremetal flavor
* `ironic_network.yml` - to create provisioning network and subnet
* `ironic_enroll_nodes.yml` - Uses baremetal-info.yml to create a nodes YAML and enrolls nodes in Ironic.

  - [x] README in the role

### DIFF
--- a/hooks/playbooks/ironic_network.yml
+++ b/hooks/playbooks/ironic_network.yml
@@ -1,0 +1,31 @@
+---
+- name: Create network for ironic provisioning
+  hosts: "{{ cifmw_target_hook_host | default('localhost') }}"
+  gather_facts: false
+  vars:
+    _namespace: openstack
+    _subnet_range: '172.20.1.0/24'
+    _subnet_gateway: '172.20.1.1'
+    _subnet_alloc_pool_start: '172.20.1.100'
+    _subnet_alloc_pool_end: '172.20.1.200'
+    _provider_physical_network: ironic
+    _provider_network_type: flat
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  tasks:
+    - name: Create baremetal network
+      ansible.builtin.shell: |
+        set -xe -o pipefail
+        oc project {{ _namespace }}
+        oc rsh openstackclient \
+          openstack network create provisioning \
+            --share \
+            --provider-physical-network {{ _provider_physical_network }} \
+            --provider-network-type {{ _provider_network_type }}
+        oc rsh openstackclient \
+          openstack subnet create provisioning-subnet \
+            --network provisioning \
+            --subnet-range {{ _subnet_range }} \
+            --gateway {{ _subnet_gateway }} \
+            --allocation-pool start={{ _subnet_alloc_pool_start }},end={{ _subnet_alloc_pool_end }}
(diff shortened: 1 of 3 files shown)


tosky:
We have other hooks but they don't  usually hardcode the stage name (in fact they could be used in pre_admin_setup or post_admin_setup too). My suggestion is to call them simply
ironic_enroll_nodes
ironic_create_network
ironic_create_flavor

Also, is uni01alpha-ironic.yml a leftover? We shouldn't have generic hooks so tied to a specific use case.

hooks/playbooks/pre_test_ironic_flavor.yml
```diff
@@ -0,0 +1,26 @@
+---
+- name: Create resources for running ironic scenario tests
```
tosky, line 2: could this message be less generic, to differentiate its counterpart in the other hooks?

hooks/playbooks/pre_test_ironic_enroll_nodes.yml
```diff
@@ -0,0 +1,65 @@
+---
+- name: Create resources for running ironic scenario tests
```
tosky, line 2: could this message be less generic, to differentiate its counterpart in the other hooks?

hooks/playbooks/pre_test_ironic_network.yml
```diff
@@ -0,0 +1,31 @@
+---
+- name: Create resources for running ironic scenario tests
```
tosky, line 2: could this message be less generic, to differentiate its counterpart in the other hooks?

tosky (changes requested):
A few other comments in addition to my previous comment.

hooks/playbooks/pre_test_ironic_network.yml
hjensas: Done.

hooks/playbooks/pre_test_ironic_enroll_nodes.yml
hjensas: Done

hooks/playbooks/pre_test_ironic_flavor.yml
hjensas: Done

hjensas:
> We have other hooks but they don't usually hardcode the stage name (in fact they could be used in pre_admin_setup or post_admin_setup too). My suggestion is to call them simply ironic_enroll_nodes ironic_create_network ironic_create_flavor
> 

done.

> Also, is uni01alpha-ironic.yml a leftover? We shouldn't have generic hooks so tied to a specific use case.

yes, it was a leftover. removed now.

hooks/playbooks/ironic_network.yml
```diff
@@ -0,0 +1,31 @@
+---
+- name: Create network for ironic provisioning
+  hosts: "{{ cifmw_target_hook_host | default('localhost') }}"
+  gather_facts: false
+  vars:
+    _namespace: openstack
+    _subnet_range: '172.20.1.0/24'
+    _subnet_gateway: '172.20.1.1'
```
tosky, line 8: just another question: are those values always valid? I mean, are they independent from any underlying networking in OCP?

hjensas: Won't I be able to override those with `extra_vars`?

```
pre_tests:
  - name: 02 Create ironic provisioning network
    type: playbook
    source: ironic_network.yml
    extra_vars:
      _subnet_range: 172.20.254.0/24'
      _subnet_gateway: '172.20.244.1'
```

tosky: It seems so, but the question was more among the line "couldn't they be dynamically identified by some other configuration", but it seems it's not the case

PR #1864, opened 2024-06-12
test_operator/tempest: auth.tempest_roles empty by default

tosky:
auth.tempest_roles is set to an empty value because otherwise python-tempestconf sets its value as 'member' and this causes failures to SRBAC reader tests, because it assigns both roles to the user.

Beyond that, the roles are defined and assigned in the tests themselves, therefore this parameter isn't needed and should have an empty value by default.

  - [x] README in the role


kopecmartin (approved):
lgtm

konan-abhi (commented):
looks good!

PR #1867, opened 2024-06-12
Converge namespace for hooks

cjeanner:
When running architecture driven deployment, install_yamls isn't
initialized anymore since it's not used.

This patch allows to converge how we pass the namespace down to the
hooks, by checking if we have the install_yamls data and defaulting to a
"not too bad" default value.

In case a user wants to override that namespace, they can use the
extra_vars parameter - since that content comes after our default
setting, it will take over.

  - [X] README in the role

### DIFF
--- a/hooks/playbooks/cinder_multiattach_volume_type.yml
+++ b/hooks/playbooks/cinder_multiattach_volume_type.yml
@@ -11,17 +11,12 @@
           }}
 
     - name: Set a multiattach volume type and create it if needed
-      vars:
-        _namespace: >-
-          {{
-            cifmw_install_yamls_defaults['NAMESPACE'] | default('openstack')
-          }}
       environment:
         KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
         PATH: "{{ cifmw_path }}"
       ansible.builtin.shell: |
         set -xe -o pipefail
-        oc project {{ _namespace }}
+        oc project {{ namespace }}
         oc rsh openstackclient \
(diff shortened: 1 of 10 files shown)


fmount:
Uni* jobs are failing due to:

```
PLAY [Create Manila resources needed for tempest run] **************************

TASK [Creat share type default for manila tempest plugin tests _raw_params=oc -n {{ cifmw_install_yamls_defaults['NAMESPACE'] }} exec -it pod/openstackclient  -- openstack share type create default false
] ***
Wednesday 12 June 2024  19:33:48 -0400 (0:00:00.076)       0:00:00.076 ******** 
  msg: |-
    The task includes an option with an undefined variable. The error was: 'cifmw_install_yamls_defaults' is undefined. 'cifmw_install_yamls_defaults' is undefined
  
    The error appears to be in '/home/zuul/src/github.com/openstack-k8s-operators/ci-framework/hooks/playbooks/manila_create_default_resources.yml': line 6, column 7, but may
  
      tasks:
        - name: Creat share type default for manila tempest plugin tests
          ^ here
```

This patch is supposed to fix that, +1 from me.

fmount (approved):
/lgtm

fmount (approved):
/lgtm

cescgina (commented):
/lgtm

rebtoor (approved):

hjensas:
/lgtm

I think we want to do a touch up on the `hooks/playbooks/ironic_*.yml`  files to use `"{{ namespace }}"` as well. (Replacing the `_namespace` var defined there. But let's do it in a follow up! :+1:

PR #1869, opened 2024-06-13
Add delorean repo filename var to dlrn_promote

viroel:
Adds a new variable to dlrn_promote role which allows us to choose another delorean repo filename inside the same directory, without the need to provide the full path to the file. repo_setup role that is called inside dlrn_promote role can generate multiple delerean repo files, and we we should be able to choose the correct file in order to promote a single component.

  - [X] README in the role


lewisdenny (approved):
/lgtm

No change in default behavior unless you override `cifmw_dlrn_promote_dlrnrepo_filename`

raukadah (approved):
/approve

PR #1870, opened 2024-06-13
Use host_ip in _sushy_url

hjensas:
The `ansible_host` contains a name only available in the ssh client configuration. This results in connection URI with a name that cannot be resolved:
 connection: redfish-virtualmedia+http://controller-0.hypervisor-1:8000/[...]

Instead of using ansible host, use the host_ip which contains the IP address.

This is an intermidiate fix, once we have the changes to use DNS more extensively we should change this to use the FQDN of the controller - or possibly add a CNAME in DNS with `sushy-emulator.$DOMAIN`.

  - [x] README in the role


lewisdenny (dismissed):
Hi Harald, thank you for the patch. This lgtm, could you just add a todo comment for reverting back to hostname once we can use DNS?

hjensas:
> Hi Harald, thank you for the patch. This lgtm, could you just add a todo comment for reverting back to hostname once we can use DNS?

Sure, TODO comment added.

cjeanner:
/approve

makes sense, and we'll be able to clean things once we get #1777 - switching to DNS or something nicer.

PR #1873, opened 2024-06-13
Add default to  cifmw_install_yamls_defaults if undefined

fmount:
The manila_create_default_resources hook is basically a standalone playbook that assumes  cifmw_install_yamls_defaults have been defined or passed to this stage. This patch adds an opinionated default to not fail when those vars are not defined.
This will fix uni* jobs until something like [1] lands.

[1] https://github.com/openstack-k8s-operators/ci-framework/pull/1867

  - [x] README in the role


cjeanner:
Please check https://github.com/openstack-k8s-operators/ci-framework/pull/1867 - it does the same, more or less, but globally.

PR #1880, opened 2024-06-14
Add a role to configure masquerading on hypervisor

hjensas:
Using *nat* type networks in libvirt causes firewall rules to be inserted by libvirt - these rules block traffic between libvirt networks. In some scenarios this is not desired, we want traffic between internal networks and still enable NAT (masquerading) of the traffic going to the external (internet/intranet).

This role's task is to insert these firewall rules to masquerade traffic where the `src` is a libvirt network and the `dst`is external (internet/intranet).

Also enable it in the reproducer. For example:
```yaml
cifmw_masquerade_networks: true
cifmw_masquerade_external_source_ranges:
  - "{{ cifmw_networking_definition.networks.ironic.network }}"
  - "{{ cifmw_networking_definition.networks.ctlplane.network }}"
```

  - [x] README in the role

### DIFF
--- a/roles/masquerade_external/tasks/main.yml
+++ b/roles/masquerade_external/tasks/main.yml
@@ -1,0 +1,37 @@
+---
+# Copyright Red Hat, Inc.
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+- name: Ensure masquerading rules in cifmw post routing chain
+  become: true
+  ansible.builtin.iptables:
+    state: present
+    action: insert
+    table: nat
+    chain_management: true
+    chain: "{{ cifmw_masquerade_external_post_routing_chain_name }}"
+    source: "{{ item }}"
+    out_interface: "{{ cifmw_masquerade_external_out_interface }}"
+    jump: "MASQUERADE"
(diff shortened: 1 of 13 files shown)


hjensas:
recheck

roles/masquerade_external/tasks/main.yml
```diff
@@ -0,0 +1,42 @@
...
+# under the License.
+
+# - name: Ensure forwarding rules
+#   become: true
+#   ansible.builtin.iptables:
+#   loop:
+
+- name: Ensure masquerading rules in cifmw post routing chain
+  become: true
+  ansible.builtin.iptables:
```
cjeanner, line 24: Would it be possible to do the same using firewalld module instead? I'd rather try to converge to ONE tool to manage the firewall.

I see about 2 mentions of actual iptables, vs "many more" for firewalld. You might need some rich_rule though... Care to check?

cjeanner (changes requested):

roles/masquerade_external/tasks/main.yml
hjensas: libvirt itself uses iptables and inserts chains LIBVIRT_FWI, LIBVIRT_FWO, LIBVIRT_FWX, LIBVIRT_PRT etc.
It is important that these rules end up before any of those - I don't see any way to add chains in https://firewalld.org/documentation/man-pages/firewalld.richlanguage.html - we'd need https://firewalld.org/documentation/man-pages/firewalld.direct.html for that - which essentially just makes firewalld run iptables commands afict?

since firewalld ansible role does not support direct, I don't think we can switch.

cjeanner: erf... 2024, still relying on iptables - not even nftables.... great.

cjeanner:
/approve

though I really don't like having two ways to manage the firewall :(.

hjensas:
Moving this to draft - after reading https://discussion.fedoraproject.org/t/f41-change-proposal-libvirt-virtual-network-nftables-self-contained/120329,  I think there may be a way to do this with firewalld.

hjensas:
Closing this, superseeded by https://github.com/openstack-k8s-operators/ci-framework/pull/1928 which is using firewalld.

PR #1886, opened 2024-06-17
hook ironic_network - neutron availability zone

hjensas:
Add variable _availability_zone_hints. Defaults to `null`. When a list of one or more availability zone hints are defined the network create command includes the --availability-zone-hint option so that the networks availability zone hints are set.

  - [x] README in the role


raukadah (approved):
/approve

PR #1888, opened 2024-06-17
Update workflow - remove the set_openstack_container.

sathlan:
This triggers an update of all the pods which triggers an update of
all the containers outside of the update procedure.

We will have a another way to pass the new pods definition.

Jira: [OSPRH-7721](https://issues.redhat.com//browse/OSPRH-7721)

-------

  - [x] README in the role

### DIFF
--- a/playbooks/update.yml
+++ b/playbooks/update.yml
@@ -11,14 +11,6 @@
       ansible.builtin.include_role:
         name: repo_setup
 
-    - name: Update all openstack services containers env vars in meta operator with tag from delorean
-      vars:
-        cifmw_set_openstack_containers_dlrn_md5_path: "{{ cifmw_basedir }}/artifacts/repositories/delorean.repo.md5"
-        cifmw_set_openstack_containers_tag_from_md5: true
-        cifmw_set_openstack_containers_extra_vars: "{{ cifmw_edpm_prepare_extra_vars }}"
-      ansible.builtin.include_role:
-        name: set_openstack_containers
-
 - name: Sync repos for controller to compute
   hosts: computes
   gather_facts: true
--- a/playbooks/update.yml
+++ b/playbooks/update.yml
@@ -10,6 +10,10 @@
     - name: Run repo_setup
       ansible.builtin.include_role:
         name: repo_setup
+    - name: Run update-containers
+      ansible.builtin.include_role:
+        name: update_containers
+      when: cifmw_update_containers | bool
 
 - name: Sync repos for controller to compute
   hosts: computes


ciecierski:
What is the other way you have in mind? Do you plan to replace it with update_containers role or something else?

sathlan:
@ciecierski so for upstream, not sure what will work best.  Maybe the same role, but after the ping test has been set up, ie this should happen during the update.  For downstream we have to deal with OLM, again during update.

Basically we shouldn't update any definition before the update in any cases as far as I understand it.

ciecierski:
@raukadah Will we trigger an update with update_containers role. If not maybe we can replace  set_openstack_containers with update_containers role

ciecierski:
related: https://github.com/openstack-k8s-operators/ci-framework/pull/1884

sathlan:
I've seen the role, not sure if it uses OLM as described, something to check. Eventually, the update container, wether it uses OLM or not should certainly be inside the update procedure instead of before, unless I miss something.

ciecierski:
> I've seen the role, not sure if it uses OLM , something to check. Eventually, the update container, wether it uses OLM or not should certainly be inside the update procedure instead of before, unless I miss something.

It is not using OLM and I don't think there is role in ci-framework that does it. If update_containers applies new openstackversion cr, but doesn't start update it should be good for now until we have workflow/role for OLM

sathlan:
Ok, so as I understand this is that as soon as we apply that CR -
which is the last thing `update_containers` does, we have a updated
pod definition and that will trigger a update of all containers.

Or am I missing something ?

raukadah:
> Ok, so as I understand this is that as soon as we apply that CR - which is the last thing `update_containers` does, we have a updated pod definition and that will trigger a update of all containers.
> 

I think so. Once we do oc apply via update_containers role, This will do a container update. Let me try that on a already deployed environment tomorrow and confirm back to you.

> Or am I missing something ?

raukadah:
> > Ok, so as I understand this is that as soon as we apply that CR - which is the last thing `update_containers` does, we have a updated pod definition and that will trigger a update of all containers.
> 
> I think so. Once we do oc apply via update_containers role, This will do a container update. Let me try that on a already deployed environment tomorrow and confirm back to you.
> 
> > Or am I missing something ?

Hello @sathlan , I tried to update the openstack services using update_containers role on VA1 deployment. It updates the openstack services containers by deleting the pods and recreating it.

Below is the playbook for the same.
```
---
- hosts: localhost
  tasks:
    - name: Update containers to podified-ci-testing
      vars:
        cifmw_update_containers: true
        cifmw_update_containers_registry: <downstream registry name>
        cifmw_update_containers_org: podified-rhos18-rhel9
        cifmw_update_containers_tag: 32724932f6e103dd2a972e1df31dd679 # podified-ci-testing hash, repo-setup will provide it via cifmw_repo_setup_full_hash var
        cifmw_update_containers_openstack: true
        cifmw_update_containers_metadata: controlplane
      include_role:
        name: update_containers

```

ciecierski (changes requested):
update_containers is for hotfix process and running this role in update playbook is not starting minor update procedure.
set_openstack_containers with a slight modification https://github.com/openstack-k8s-operators/ci-framework/pull/1963 just patches openstack-operator csv and enables to start a controlled minor update.

sathlan:
Ok, so our problem here was that pod update was triggered outside the update step.  `update_containers` role is not fit for replacing `set_openstack_container` for our use case as it also triggers a pod update, similar to what a hotfix would do.

#1963 is currently the best fix for our issue as the pod are not updated anymore when we use that role using that patch.

So I'm closing this in favor of the aforementioned patch.

PR #1890, opened 2024-06-17
Enable "backup" for files created by networking_mapper

cjeanner:
This will make debugging easier, since we'll be able to follow the
evolution of the generated content.

Of course, if the generated content isn't modified, it will not replace
the target, meaning it won't create a backup.


hjensas (commented):
/lgtm

PR #1895, opened 2024-06-18
Add missing var in the README

bshewale:
  - [x] README in the role

### DIFF
--- a/roles/update_containers/README.md
+++ b/roles/update_containers/README.md
@@ -7,6 +7,7 @@ openstack services containers, Ansibleee and EDPM Baremetal Image.
 If apply, please explain the privilege escalation done in this role.
 
 ## Parameters
+* `cifmw_update_containers`: The boolean value which will decide to run the 'Update the containers' role. Default to `false`.
 * `cifmw_update_containers_metadata`: The metadata name of podified control plane custom resources. Default to `controlplane`.
 * `cifmw_update_containers_namespace`: The namespace of the podified control plane deployment. Default to `openstack`.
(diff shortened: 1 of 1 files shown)


roles/update_containers/README.md
```diff
@@ -7,6 +7,7 @@ openstack services containers, Ansibleee and EDPM Baremetal Image.
 If apply, please explain the privilege escalation done in this role.
 
 ## Parameters
+* `cifmw_update_containers`: The boolean value which will decide to run the 'Update the containers' task. Default to `false`.
```
arxcruz, line 10: Role, not task. Although technically is the 'main' task, while the other two tasks in the role just create the file, that this task will use... the logic is also a little bit messed, I would put all the tasks in one block with a when cifmw_update_containers is set to true, but 🤷

viroel:
I agree with the changes, but the PR title and description needs update.
molecule job is also failing and will require some fixes too.

roles/update_containers/tasks/main.yml
```diff
@@ -14,21 +14,23 @@
...
-  ansible.builtin.file:
-    state: directory
-    recurse: true
-    path: "{{ cifmw_update_containers_dest_path | dirname }}"
-
-- name: Generate update containers yaml file
-  ansible.builtin.template:
-    src: "update_containers.j2"
-    dest: "{{ cifmw_update_containers_dest_path }}"
-
```
raukadah, line 27: @bshewale Thank you for the patch. The main objective of these two tasks is to allow to generate the container updates cr file only.(It will be useful for anyone using this role standalone.)

Apply only when users want to apply that.

raukadah (commented):
Please check the comment in the role content change.

PR #1897, opened 2024-06-18
Ensure we install tmux

cjeanner:
tmux is a widely used software, and usually we install it manually as
soon as we jump on a host. It was included only on controller-0, but we
also use it from the hypervisor directly.

Let's make that software available anywhere.


raukadah (approved):
/approve

hjensas (approved):

PR #1900, opened 2024-06-18
increase cpu/ram specs for ocp nodes in va-hci scenario

rebtoor:
Trying to deploy the hci scenario, during the control-plane deployment, some pods were in "pending" status due to lack of resources so this PR should address that issue. 


cjeanner:
/approve

fun - the devscripts_override had 16G, pretty sure of that.

lewisdenny:
@rebtoor should we also proactively bump the BGP scenario as well?
```
❯ grep -lr 'cifmw_libvirt_manager_configuration' ./scenarios | xargs yq '.cifmw_libvirt_manager_configuration.vms.ocp.memory'
32
---
32
---
16 <----
---
32
---
null
---
32
```

PR #1903, opened 2024-06-18
Add a new extra variable for holding the update's parameters

sathlan:
Make sure that, when we use downstream repo setup we don't hardcode
cifmw_repo_setup_promotion.

To that end, and to add more flexibility to the variables that need to
be overwritten during update, take the key/value of the
`cifmw_update_extras` hash. It should be defined in a job and would
look like this:

```
cifmw_update_extras:
 cifmw_repo_setup_promotion: promotion
 other_var: value
 ...
```

That enables the job definition to pass all variables needed for the
update job with having new variable name for each override.

It also solves the problem of the hardcoded
`cifmw_repo_setup_promotion` as now the job has total control over the
variables passed.

Jira: https://issues.redhat.com/browse/OSPRH-7811
-------

  - [x] README in the role

### DIFF
--- a/ci/playbooks/edpm/update.yml
+++ b/ci/playbooks/edpm/update.yml
@@ -33,4 +33,8 @@
           {%- endif %}
           -e update_playbook_run=true
           -e "@{{ ansible_user_dir }}/ci-framework-data/artifacts/parameters/zuul-params.yml"
-          -e "cifmw_repo_setup_promotion={{ cifmw_repo_setup_promotion_target }}"
+          {%- if cifmw_update_extras is defined %}
+          {%-   for key, value in cifmw_update_extras.items() %}
+          -e "{{key}}={{value}}"
+          {%-   endfor %}
+          {%- endif %}
--- a/roles/update/README.md
+++ b/roles/update/README.md
@@ -2,6 +2,6 @@
 Role to run update
 
 ## Parameters
-* `param_1`: this is an example
+* ``cifmw_update_extras`: (hash) Hold job variable that get set when running the update playbook.
 
 ## Examples


ciecierski:
I like the change and idea for new cifmw_update_extras. Just can you rephrase description, please? The aim is to add cifmw_update_extras dict defined at zuul job definition level, instead of using pre-defined inline variables.

ci/playbooks/edpm/update.yml
```diff
@@ -33,4 +33,8 @@
           {%- endif %}
           -e update_playbook_run=true
           -e "@{{ ansible_user_dir }}/ci-framework-data/artifacts/parameters/zuul-params.yml"
-          -e "cifmw_repo_setup_promotion={{ cifmw_repo_setup_promotion_target }}"
+          {%- if cifmw_update_extras is defined %}
```
marios, line 36: @sathlan o/ hello. maybe you can join tomorrow's ci reviews call if you have time @ 1100 utc (will add you to the invite momentarily).

For example even this cifmw_repo_setup_promotion is a very specific var which was added (I assume?) for the minor update, so most of us don't even know about that. 

It would be good to get some more context/5 minutes explanation for the team, not just for this review but so we can at least start to hear about the update process and what it will look like in the framework and ci generally. 

sincerely, 

marios

marios (commented):
o/

ci/playbooks/edpm/update.yml
sathlan: Could not join, sorry about that.  I'm a bit confused by that comment though. `cifmw_repo_setup_promotion` is a standard `repo_setup` variable (https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/repo_setup/defaults/main.yml#L23) .  Maybe you meant `cifmw_repo_setup_promotion_target` ?

If so that was the original mechanism used to be able to pass variable to the update playbook run so that we could differentiate them from deployment variables.

Here the review want to get rid of those explicit extra variables and use `cifmw_update_extras` as a list of hashes that will be expanded for the update playbook run.

Then in the job we can define things like this:

```yaml
cifmw_update_extras:
  - cifmw_repo_setup_promotion: latest
```

Does it clarify the matter ?

ciecierski: In example
`cifmw_update_extras:
  - cifmw_repo_setup_promotion: latest
` should be:
`cifmw_update_extras:
  cifmw_repo_setup_promotion: latest
` as cifmw_update_extras is dict, not a list

ciecierski (approved):
Please fix example in description and you have LGTM from me.

ci/playbooks/edpm/update.yml
marios: thanks for explanation folks marking resolved

primarily my concerns were who is going to be affected by removal of 

```
-e "cifmw_repo_setup_promotion={{ cifmw_repo_setup_promotion_target }}"

```

and confirmed by Sofer just now in slack, this update.yml is only used for the minor update job which you are working on

marios (approved):
lgtm

marios:
@sathlan @ciecierski you need to harrass someone from this list https://github.com/openstack-k8s-operators/ci-framework/blob/main/OWNERS to add the /approve here for merge fyi

PR #1904, opened 2024-06-19
Ensure virtqemud-sock is present before cleaning

cjeanner:
While the usual usage is "deploy, then clean", jobs in CI might start
with a "clean, then deploy".

In such case, it might happen all of the dependencies are installed
during the node bootstrap, but libvirt services aren't switched to the
modular daemons layout, leading to a crash with the current
implementation.

This patch adds a sanity check on the virtqemud-sock file availability,
to ensure everything's fine.

Molecule job has also been modified to "clean" before "deploy" so that
we should hopefully catch issues earlier.

### DIFF
--- a/roles/libvirt_manager/molecule/deploy_layout/prepare.yml
+++ b/roles/libvirt_manager/molecule/deploy_layout/prepare.yml
@@ -14,7 +14,6 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
-
 - name: Prepare
   hosts: instance
   vars:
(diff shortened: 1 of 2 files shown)


rebtoor (approved):

roles/libvirt_manager/molecule/deploy_layout/prepare.yml
```diff
@@ -31,8 +31,16 @@
   roles:
     - role: "test_deps"
     - role: "ci_setup"
-    - role: "libvirt_manager"
   tasks:
+    - name: Cleanup layout before anything
+      ansible.builtin.import_role:
+        name: libvirt_manager
+        tasks_from: cleanup_layout.yml
```
rebtoor, line 38:
```suggestion
      ansible.builtin.import_role:
        name: libvirt_manager
        tasks_from: clean_layout.yml
```

rebtoor (changes requested):

rebtoor (approved):

cjeanner:
recheck

rebtoor (approved):

cjeanner:
recheck

PR #1908, opened 2024-06-19
Ensure CRC is using our new dnsmasq

cjeanner:
Since #1777, a special dnsmasq instance is created and managed by the
ci-framework.
While most of the testing was done with OCP cluster in mind, we have to
ensure CRC is able to properly consume that new service.

This allows to remove all of the hacks we had in CRC node related to
DNS, making everything really easier for everyone.

### DIFF
--- a/roles/reproducer/tasks/configure_crc.yml
+++ b/roles/reproducer/tasks/configure_crc.yml
@@ -5,108 +5,79 @@
   ansible.builtin.slurp:
     path: /etc/ci/env/networking-environment-definition.yml
 
-- name: Configure CRC services
-  delegate_to: crc-0
-  vars:
-    _parsed: "{{ _net_conf.content | b64decode | from_yaml}}"
-    _crc: "{{ _parsed.instances['crc-0'].networks.ctlplane }}"
-  block:
-    - name: Ensure crc-0 knows about its second NIC
-      become: true
-      vars:
-        _ctlplane: "{{ _parsed.networks.ctlplane }}"
-      community.general.nmcli:
-        autoconnect: true
-        conn_name: private_net
-        ifname: "{{ _crc.interface_name }}"
-        type: ethernet
-        ip4: "{{ _crc.ip_v4 }}/24"
-        never_default4: true
-        state: present
+- name: Remove CRC managed zone delegation
+  become: true
+  notify: Restart NetworkManager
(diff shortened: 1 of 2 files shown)


roles/reproducer/tasks/configure_crc.yml
```diff
@@ -5,108 +5,62 @@
...
+              - canary-openshift-ingress-canary.apps-crc.testing
+              - console-openshift-console.apps-crc.testing
+              - default-route-openshift-image-registry.apps-crc.testing
+              - downloads-openshift-console.apps-crc.testing
+              - oauth-openshift.apps-crc.testing
+            ipaddr: "{{ ip }}"
+            state: present
+      ansible.builtin.include_role:
+        name: "dnsmasq"
+        tasks_from: "manage_address.yml"
```
hjensas, line 50: I don't this all of these should be `address`
--address is `Specify an IP address to return for any host in the given domains.`

We should use --host-record for api.crc.testing and api-int.crc.testing ?

And then one address entry with `apps-crc.testing` should be enough?

cjeanner: hm, maybe `address` for `crc.testing` then? Not 100% sure about the used domains/others. We indeed have "some" hosts under `apps-crc.testing` - so that's `address` for sure.

~> maybe using address to cover the whole `crc.testing` would be a thing?

cjeanner: updated - we have "just" 2 addresses now (and yeah, I totally mixed things in the previous iteration). Lemme know if that's OK now.

hjensas: We should be careful with the wildcard - I think only apps-crc.testing should be wildcard.

See: https://github.com/openstack-k8s-operators/install_yamls/commit/cd1c8e79828b49eb85234490e754cea1cd1e8590

I used a wildcard - `address=/sno.lab.example.com/${SNO_HOST_IP}` - this cause issues with resolving internally in OCP. Hence I had to change to to only wildcard `apps.$DOMAIN`.

cjeanner: ok, will check that out, but then we'll want to get your dual-stack patch in first I think.

hjensas (commented):
/lgtm

PR #1910, opened 2024-06-19
Add consistency check related to networks

cjeanner:
The new checks will ensure we're not trying to attach a non-existing
network, nor that we're trying to create network that aren't attached to
any VM.

The dedicated molecule scenario should properly cover the various
scenarios this new validation covers.


viroel:
/approve

that's a great validation to add, thanks!

roles/reproducer/tasks/prepare_networking.yml
```diff
@@ -1,4 +1,7 @@
 ---
+- name: Ensure networks are consistent
+  ansible.builtin.import_tasks: "network_assertions.yml"
```
hjensas, line 3: Is there a way to disable the validations?
The ci-framework is a developer / QE tool - I don't think we should block people that need to wire in custom stuff like a pre-existing network etc.

cjeanner:
/hold

need to take a few to add new top-level parameter to disable validations - but that will be something like "-e yes-i-really-know-what-i-am-doing=true"...

roles/reproducer/tasks/prepare_networking.yml
cjeanner: done.

hjensas (commented):
/lgtm

rebtoor (approved):

hjensas (approved):

PR #1912, opened 2024-06-19
Add extra-disks to va1 scenario file

viroel:
Since downstream VA1 jobs uses extra-disks for lvm by default, we are adding this configuration to va-hci scenario file.

  - [X] README in the role


cjeanner:
/approve

no real influence to upstream - adding the approval already. Size/amount would require validation via a DS testproject since it's consumed there.

PR #1914, opened 2024-06-20
Add support to net_map plugin for configuring Multus routes

lewisdenny:
Jira: https://issues.redhat.com/browse/OSPRH-7259

  - [x] Content of the docs/source is reflecting the change

Todo:
- [x] Add support for IPV6
- [x] Test in VA deployment with Octavia enabled

### DIFF
--- a/plugins/module_utils/net_map/networking_definition.py
+++ b/plugins/module_utils/net_map/networking_definition.py
@@ -471,6 +471,26 @@ def check_host_network_ranges_collisions(
     return None, None
 
 
+def _get_field_ip_net_version(raw_field: str) -> typing.Union[int, None]:
+    if not raw_field:
+        return None
+
+    try:
+        # Avoid that start/end integers are takes as IPv4
+        # str allows passing IPs directly without converting to int
+        # and making it fail, so they will be parsed after the except
+        value = int(str(raw_field))
+        if value is not None:
+            return None
+    except ValueError:
+        pass
+
+    try:
+        return ipaddress.ip_network(raw_field).version
+    except ValueError:
+        return None
+
+
 class HostNetworkRange(ansible_encoding.RawConvertibleObject):
     """Parser and validator of network ranges
 
@@ -494,7 +514,7 @@ class HostNetworkRange(ansible_encoding.RawConvertibleObject):
         ] = None,
         length: typing.Union[int, str, None] = None,
     ):
-        """Creates a network range instance form it's net and limits.
+        """Creates a network range instance from it's net and limits.
 
         Creates a range instance for the provided network based on the
         given limits start and end (or length instead of end if provided).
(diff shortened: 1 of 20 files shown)


lewisdenny:
/hold
Currently being tested D/S

tests/unit/module_utils/net_map/test_networking_definitions_network.py
```diff
@@ -794,3 +794,41 @@ def test_network_definition_parse_valid_router_ok():
...
+                    "ranges": [
+                        {
+                            "start": str(network_ip[100]),
+                            "length": 20,
+                        },
+                    ],
+                    "routes": [
+                        {
+                            "destination": "192.168.121.0/24",
+                            "gateway": "192.168.122.1/32",
```
beagles, line 821: The gateway should be just the IP address and not include the mask. This causes an ipam parser error.

beagles (changes requested):
The gateway is just an IP and doesn't need a mask so in th

tests/unit/module_utils/net_map/test_networking_definitions_network.py
lewisdenny: Ah mate, I checked this one when I was implementing. It seemed like it wanted the mask: https://github.com/containernetworking/plugins/blob/acf8ddc8e1128e6f68a34f7fe91122afeb1fa93d/plugins/ipam/static/static_test.go#L59

I will check with you on the error, sorry about that one!

lewisdenny: Oh `s/destination/gateway/` got it!

lewisdenny: This should be resolved now in the latest revision so resolving

beagles (approved):

beagles:
> This is now passing downstream in a UNI* job with Octavia enabled. This patch is functional and ready to merge from my side, @beagles could you please confirm when ready if your okay?

Tested and seems good to go!

> 
> @pablintino from my team will review the code :)

plugins/module_utils/net_map/networking_definition.py
```diff
@@ -471,6 +471,26 @@ def check_host_network_ranges_collisions(
     return None, None
 
 
+def _get_field_ip_net_version(raw_field: str) -> typing.Union[int, None]:
```
pablintino, line 474: I need to update other usages of Union, but I started using another cleaner approach, `typing.Optional[int]`.

lewisdenny: Ooh `Optional[X] is equivalent to X | None (or Union[X, None]).` very nice man.  [1]
[1] https://docs.python.org/3/library/typing.html#typing.Optional

PR #1916, opened 2024-06-20
Revert "Use update_containers role to update openstack services container image"

sdatko:
Reverts openstack-k8s-operators/ci-framework#1884

I am just testing if this is the guilty patch...


sdatko:
Tests done, we know our new workarounds now :grin:

PR #1918, opened 2024-06-20
[polarion] Merge xmls with junitparser 

kopecmartin:
In case there are multiple xml files that need to be reported
to polarion, we need to merge them before passing them to jump.
Using junitparser even in cases where there is only one xml file
will have these benefits:
* we don't have to write when conditions
* the output xml will have the same structure - the content will
be wrapped in <testsuites> tags whether we have more xml results
or not - testsuite in junitparser context equals to one xml input.

  - [x] README in the role

### DIFF
--- a/roles/polarion/tasks/main.yml
+++ b/roles/polarion/tasks/main.yml
@@ -63,6 +63,12 @@
         state: latest # noqa: package-latest
         extra_args: --upgrade
 
+    - name: Install junitparser
+      ansible.builtin.pip:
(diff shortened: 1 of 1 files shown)


kopecmartin:
Testing explained in this comment: https://issues.redhat.com/browse/OSPRH-7552?focusedId=25088182&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-25088182

PR #1921, opened 2024-06-20
Modify libvirt_manager to run linear instead of batches

cjeanner:
The initial approach was batching VMs, meaning we were having multiple
times the same kind of `wait for SSH` and related "pause".

This patch changes how we define, start and configure the various VM, by
using `cifmw_libvirt_manager_all_vms` available since #1777.
This list is showing all of the VM created by the libvirt_manager role,
and with "simple" string transformation, we can extract the VM family
(compute, ocp, networker, etc) so that we can still access the various
configurations for each of them.

This linear approach allows to leverage `async` in various places, as
well as some other helpers to make the whole run smoother and faster.

### not in commit message: testing
- [x] OCP deployment from scratch (via NFV TP)
- [x] OCP deployment, reusing existing base images (via spine/leaf run)
- [x] OCP deployment with volumes, from scratch
- [x] OCP deployment with volumes, reusing existing base images
- [x] OCP deployment with worker, from scratch
- [x] OCP deployment with worker, reusing existing base images
- [x] Spine/Leaf
- [x] 3-nodes (CRC) layout from scratch
- [x] 3-nodes (CRC) layout reusing existing base CRC
- [x] molecule reproducer run (cifmw_job_uri pointing to a molecule run)
- [x] NFV

### DIFF
--- a/roles/libvirt_manager/tasks/clean_layout.yml
+++ b/roles/libvirt_manager/tasks/clean_layout.yml
@@ -56,7 +56,7 @@
       community.libvirt.virt:
         command: undefine
         flags:
-          - keep_nvram
+          - nvram
           - snapshots_metadata
         name: "{{ item }}"
         uri: "qemu:///system"
@@ -187,7 +187,7 @@
            default([])
         }}
 
-    - name: Remove cifmw storage pool
+    - name: Remove OCP volumes storage pool
       tags:
         - never
         - deepscrub
--- a/roles/libvirt_manager/tasks/create_networks.yml
+++ b/roles/libvirt_manager/tasks/create_networks.yml
@@ -81,7 +81,6 @@
   loop_control:
     label: "{{ item.name }}"
 
-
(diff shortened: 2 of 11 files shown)


roles/libvirt_manager/tasks/clean_layout.yml
```diff
@@ -56,7 +56,7 @@
       community.libvirt.virt:
         command: undefine
         flags:
-          - keep_nvram
+          - nvram
```
cjeanner, line 59: @hjensas thoughts on that? AFAIK we can get rid of nvram all the time without any issue - correct? We're using overlays anyway, so the VM itself is undefined, UUID and other things change on each creation... ?

hjensas: yes, I think we should clean nvram. It will be re-created when defining new VMs.

hjensas (approved):

viroel (approved):

PR #1923, opened 2024-06-20
[devscripts] Allow passing a custom user ignition file for OCP nodes

pablintino:
  - [ ] README in the role


roles/devscripts/tasks/139_custom_ignition.yml
```diff
@@ -0,0 +1,35 @@
...
+        _config_data:
+          ignition_extra: "cifmw-ignition.ign"
+      ansible.builtin.set_fact:
+        cifmw_devscripts_config: >-
+          {{
+            cifmw_devscripts_config |
+            combine(_config_data, recursive=true)
+          }}
+
+    - name: Copy the configuration file
```
cjeanner, line 31: wondering if we shouldn't allow a list or, even, a directory?

pablintino:
This functionality is not needed as it was used for troubleshooting an issue and given that we may use other tool to deploy OCP rather than dev-scripts I'm not going to work on this.

PR #1926, opened 2024-06-20
Switch to dhcp-hostsfile

cjeanner:
We hit a race condition with ansible creating the DHCP file with strict
rights, ionitofy kicking in, dnsmasq trying to read the file and
failing.
This prevented the service to load a DHCP host configuration, leading to
issues.

With this change, we ensure we keep a directory, but we have to trigger
the reload ourself - like before. This ensures the rights are all
applied.

### DIFF
--- a/roles/dnsmasq/templates/cifmw-dnsmasq.conf.j2
+++ b/roles/dnsmasq/templates/cifmw-dnsmasq.conf.j2
@@ -9,4 +9,4 @@ dhcp-leasefile=/var/lib/dnsmasq/cifmw-dnsmasq.leases
 {% endif %}
 
 conf-dir={{ cifmw_dnsmasq_basedir }},*.conf
-dhcp-hostsdir="{{ cifmw_dnsmasq_basedir }}/dhcp-hosts.d"
+dhcp-hostsfile="{{ cifmw_dnsmasq_basedir }}/dhcp-hosts.d"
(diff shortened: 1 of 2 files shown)


roles/dnsmasq/templates/cifmw-dnsmasq.conf.j2
```diff
@@ -9,4 +9,4 @@ dhcp-leasefile=/var/lib/dnsmasq/cifmw-dnsmasq.leases
 {% endif %}
 
 conf-dir={{ cifmw_dnsmasq_basedir }},*.conf
-dhcp-hostsdir="{{ cifmw_dnsmasq_basedir }}/dhcp-hosts.d"
+dhcp-hostsfile="{{ cifmw_dnsmasq_basedir }}/dhcp-hosts.d"
```
cjeanner, line 12: From the doc:

> Read DHCP host information from the specified file. If a directory is given,
> then read all the files contained in that directory in alphabetical order. The file
> contains information about one host per line.  The  format of  a line is the same
> as text to the right of '=' in --dhcp-host. The advantage of storing DHCP host
> information in this file is that it can be changed without re-starting dnsmasq: the
> file will be re-read when dnsmasq receives SIGHUP.

viroel:
/approve

with this PR as depends-on, 3 of 3 machines were able to deploy OCP.

hjensas (commented):
/lgtm

PR #1927, opened 2024-06-21
nit: Ensure dnsmasq configuration is really cleaned

cjeanner:
Somewhere in the history of the role, the base configuration directory
wasn't removed anymore, leading to potential conflicts due to dangling
files. This was especially the case when a user had an "old" deployment,
then updates the repository, runs a cleanup, and tries to re-deploy.


hjensas (approved):
/lgtm

rebtoor (approved):

PR #1928, opened 2024-06-21
Enable forward in libvirt zone

hjensas:
For ironic usecase we need to allow traffic forwarding between networks in libvirt - without nat.
    
 Add option: `cifmw_libvirt_manager_firewalld_zone_libvirt_forward`
 Defaults to: `true`, this is the behaviour we tried(?) to have with the code in devscripts that this PR removes.
    
Also, by default set masquerade on the `default` zone i.e `public`.
This allows us to remove the iptables rules inserted by the devscripts role.

  - [x] README in the role


roles/libvirt_manager/tasks/clean_layout.yml
```diff
@@ -168,6 +168,22 @@
...
       loop: "{{ cleanup_nets }}"
 
+    - name: Disable masquerade in firewalld default zone
+      become: true
+      ansible.posix.firewalld:
+        masquerade: false
+        zone: "{{ cifmw_libvirt_manager_firewalld_default_zone }}"
+        permanent: true
+        immediate: true
+        state: enabled
```
cjeanner, line 178: "state enabled" with "disable masquerade" - that module will drive us crazy.

roles/libvirt_manager/tasks/deploy_layout.yml
```diff
@@ -4,6 +4,25 @@
     - _cifmw_libvirt_manager_layout is not defined
   ansible.builtin.include_tasks: generate_layout.yml
 
+# TODO(hjensas): We should have a dedicated role for firewall
+- name: Enable forwarding in the libvirt zone
+  when: cifmw_libvirt_manager_firewalld_zone_libvirt_forward | default(false) | bool
+  become: true
+  notify: Restart firewalld
```
cjeanner, line 11: so you're counting on the existing `flush_handlers` ? Are there running at the right time for you?

hjensas: right, it worked - but better be safe. I added a flush handler close to these tasks.
Good catch!

raukadah:
recheck

raukadah:
recheck

raukadah:
/lgtm

tested with uni01alpha testproject.

PR #1931, opened 2024-06-21
Only call configure_object tasks when rgw_enabled is true

fultonj:
RGW related tasks are called by playbooks/ceph.yml only if `cifmw_ceph_daemons_layout.rgw_enabled` is true but the post tasks in the cifmw_cephadm role need the same rule. This patch adds that rule.

We didn't discover this until PR 1821 merged. The "Create swift service, user and roles" task in post, only ran when `cifmw_ceph_daemons_layout.rgw_enabled` was true. When PR 1821 merged it added tasks to do the same but they didn't have this condition. Those conditions are removed by this patch as the entire configure_boject tasks file is no longer included by post (and only post includes that file) unless `cifmw_ceph_daemons_layout.rgw_enabled` is true.

Jira: https://issues.redhat.com/browse/OSPRH-7889

  - [x] README in the role

### DIFF
--- a/roles/cifmw_cephadm/tasks/configure_object.yml
+++ b/roles/cifmw_cephadm/tasks/configure_object.yml
@@ -35,7 +35,6 @@
     msg: "WARNING: Swift is deployed and the endpoint exists already, ceph RGW cannot be configured as object store service"
   when:
     - cifmw_openshift_kubeconfig is defined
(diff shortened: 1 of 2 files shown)


fmount (approved):
/lgtm thank you!

katarimanojk (commented):
Thanks for the PR.
/lgtm

PR #1934, opened 2024-06-24
Remove nmcli call on controller-0

cjeanner:
Since we now have proper DHCP with fixed IPs on all managed networks, we
can now rely on it instead of manually fixing the IP on the host.

### DIFF
--- a/roles/reproducer/tasks/configure_controller.yml
+++ b/roles/reproducer/tasks/configure_controller.yml
@@ -438,22 +438,6 @@
       ansible.builtin.import_role:
         name: networking_mapper
 
-    - name: Configure ctlplane interface on controller-0
-      become: true
-      vars:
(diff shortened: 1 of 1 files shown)


rebtoor (approved):

hjensas (commented):
/lgtm

roles/reproducer/tasks/configure_controller.yml
```diff
@@ -469,20 +453,6 @@
             owner: zuul
             group: zuul
 
-        - name: Inject crc related host entry
```
hjensas, line 472: We may want to split this out to a separate change? Looks like it is causing issues in the reproducer molecure job?

cjeanner: yup, I'm testing another PR to fix that - I'll update this one shortly.

rebtoor (approved):

hjensas (approved):

PR #1935, opened 2024-06-24
Add support for running tests locally in git worktree

lewisdenny:
This patch adds a warning for users using git worktrees to use non container based pre-commit make target and adds a test to the container file to skip running the git commit command which fails in a worktree.


cescgina (approved):

raukadah (commented):
/approve

PR #1939, opened 2024-06-24
Added var to update IPA agent image

raukadah:
  - [x] README in the role


roles/update_containers/templates/update_containers.j2
```diff
@@ -84,6 +83,9 @@ spec:
 {% if cifmw_update_containers_edpm_image_url is defined %}
     osContainerImage: {{ cifmw_update_containers_edpm_image_url }}
 {% endif %}
+{% if cifmw_update_containers_ipa_image_url is defined %}
```
marios, line 86: so if not set, we just get the default? But I am wondering what is special about this one i.e. why do we need to do this, instead of what we currently have (and where you can already override the cifmw_update_containers_registry _org and _tag ?

raukadah: EDPM build image job builds EDPM and IPA image. EDPM image is used in baremetal job. ipa image is used when we deploy ironic conductor in the ironic podified job. We have only one job which validates the deployment in the ironic component pipeline. component-baremetal-podified-rhel9-rhoso18-crc . That's why we need a separate var to test the same.

raukadah: @marios https://github.com/openstack-k8s-operators/edpm-image-builder/pull/34 will test this flag.

raukadah: and here is the result: https://github.com/openstack-k8s-operators/edpm-image-builder/pull/34#issuecomment-2188588625

marios: thanks for links

marios (approved):

raukadah:
recheck

viroel (approved):

PR #1940, opened 2024-06-24
Correct OCP validations

cjeanner:
Cinder volumes list has a non-empty default, leading to some issues with
the assertions.

This patch corrects the molecule scenarios to match the actual defaults
set in devscripts role, and fixes the validations by considering the
booleans related to volume usage (cifmw_use_lvms and
cifmw_devscripts_create_logical_volume).

This should fix a series of faulty validations hit by end-users.


rebtoor (approved):

cjeanner:
recheck

lewisdenny (approved):
/approve

PR #1942, opened 2024-06-24
[os_net_setup] Disable the use of `openstack-ansible-collection`

rebtoor:
To avoid the usage of the `openstack-ansible-collection`, the modules have been replaced with openstack cli commands.

Signed-off-by: Roberto Alfieri <ralfieri@redhat.com>

  - [x] README in the role


roles/os_net_setup/tasks/main.yml
```diff
@@ -1,53 +1,90 @@
...
+  vars:
+    _template_file: 'subnet_command.j2'
+  block:
+    - name: Delete existing subnets
+      ansible.builtin.shell: |
+        set -euxo pipefail
+        if [
+            $(oc exec -n openstack openstackclient --
+              openstack subnet list -c Name -f value |
+              grep -c {{ item.1.name }}) != 0
```
slawqo, line 45: shouldn't you maybe filter only subnets which belongs to the specific network?

roles/os_net_setup/tasks/main.yml
```diff
@@ -1,53 +1,90 @@
...
-  loop_control:
-    loop_var: net_item
+- name: Create required subnet pools
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  vars:
+    _template_file: 'subnet_pool_command.j2'
+  block:
+    - name: Delete existing subnet pools
```
slawqo, line 71: I'm not sure if that will work as expected. If you will have existing resources created already, and You will first delete old subnets and create new ones, it may use existing subnet pool which will cause failure in the deletion of the subnet pool in this step.
To make it working correctly I think You should have 2 blocks:
1. to delete all existing resources,
2. to create new resources as requested

roles/os_net_setup/tasks/main.yml
rebtoor: done!

roles/os_net_setup/tasks/main.yml
rebtoor: makes sense. Done :)

roles/os_net_setup/tasks/main.yml
```diff
@@ -1,53 +1,87 @@
...
-          combine ({'auth': { 'password' : (osp_secret.resources[0].data['secure.yaml']|b64decode|from_yaml)['clouds']['default']['auth']['password']} }, recursive=true))['auth'] }}
-        region_name: "{{ (clouds_yaml.resources[0].data['clouds.yaml']| from_yaml)['clouds']['default']['region_name'] }}"
-  rescue:
-    - name: Fail if we do not have auth information
-      ansible.builtin.fail:
-        msg: |
-          Failed to Retrive auth Infomation
-          clouds.yaml: {{ clouds_yaml }}
-          osp_secret: {{ osp_secret }}
+    - name: Delete existing subnet pools
```
slawqo, line 9: still I think that this may fail cleaning as if there are subnets created from the existing pool, such pool will not be possible to be deleted. I think You should first delete subnets, then subnet pools and networks (order of networks and subnet pools don't really matters, important is that there will be no subnets using subnetpools before deletion of the subnet pools).

rebtoor: ack! Fixed!

slawqo (commented):
/lgtm for me, thx a lot

roles/os_net_setup/README.md
```diff
@@ -1,17 +1,22 @@
...
+
+* `cifmw_os_net_setup_config`: (list) It contains the definitions for networks and their associated subnets.
+    See an example in roles/os_net_setup/defaults/main.yml
+* `cifmw_os_net_subnetpool_config`: (list) It contains the definitions for subnet pools.
+    See an example in roles/os_net_setup/defaults/main.yml
+* `cifmw_os_net_setup_dry_run`: (bool) Disable the generation of the commands.
 
 ## Molecule
+
 There is no molecule test for this role, since it gets tested via any end-to-end job.
```
lewisdenny, line 22: This is no longer correct and should be removed but this can be done in a follow up

lewisdenny (approved):
/approve

Thank you @rebtoor I appreciate the debug message printing the generated commands as the j2 is hard to parse as always. I also appreciate adding a molecule test.

Just one nit for the README

PR #1944, opened 2024-06-25
Patch nested CRC dnsmasq 

lewisdenny:
This patch moves the dns configuration logic into a central location and
calls it for both multinode and nested CRC jobs.

### DIFF
--- a/roles/openshift_setup/tasks/patch_dnsmasq.yml
+++ b/roles/openshift_setup/tasks/patch_dnsmasq.yml
@@ -1,0 +1,82 @@
+---
+- name: Ensure file is present
+  register: _stat_networking_info
+  ansible.builtin.stat:
+    path: /etc/ci/env/networking-info.yml
+
+- name: Load network parameters
+  when: _stat_networking_info.stat.exists
+  block:
+    - name: Slurp network-info.yml
+      register: _patch_dns_network_info_slurp
+      ansible.builtin.slurp:
+        src: "/etc/ci/env/networking-info.yml"
+
+    - name: Decode network-info.yml
+      ansible.builtin.set_fact:
+        _decoded_net_env: >-
+          {{
+            _patch_dns_network_info_slurp['content'] |
+            b64decode | from_yaml
+          }}
+
+- name: Check which dnsmasq config we must edit
+  register: _dnsmasq
+  ansible.builtin.stat:
+    path: '/srv/dnsmasq.conf'
+    get_attributes: false
+    get_checksum: false
+    get_mime: false
+
+
+- name: Shared parameter across blocks
+  vars:
+    _crc_default_net_ip: >-
+      {{
+        _decoded_net_env.crc_ci_bootstrap_networks_out.crc.default.ip | default('192.168.130.11')
+      }}
+  block:
+    - name: Configure dnsmasq
+      vars:
+        _dnsmasq_config: >-
+          {{
+            _dnsmasq.stat.exists |
+            ternary(_dnsmasq.stat.path, '/etc/dnsmasq.d/crc-dnsmasq.conf')
+          }}
+        _crc_default_net_dns: >-
+          {{
+              _decoded_net_env.crc_ci_bootstrap_provider_dns |
+            default([])
+          }}
+      block:
+        - name: Configure dns forwarders
+          become: true
+          ansible.builtin.blockinfile:
+            path: "{{ _dnsmasq_config }}"
+            block: |-
+              {% for dns_server in _crc_default_net_dns %}
+              server={{ dns_server }}
+              {% endfor %}
+
+        - name: Configure local DNS for CRC pod
+          become: true
+          register: last_modification
+          ansible.builtin.replace:
+            path: "{{ _dnsmasq_config }}"
+            regexp: "192.168.130.11"
+            replace: >-
+              {{ _crc_default_net_ip | ansible.utils.ipaddr('address') }}
+
+        # Note(Lewis): Only needed for CRC => 2.32.0-4.14.8
+        - name: Configure dnsmasq listen-address to listen on both br-ex and ci-private-network
+          when:
+            - not _dnsmasq.stat.exists
+          become: true
+          ansible.builtin.lineinfile:
+            path: "{{ _dnsmasq_config }}"
+            insertafter: '^listen-address='
+            line: "listen-address={{ _crc_default_net_ip | ansible.utils.ipaddr('address') }}"
+
+- name: Restart dnsmasq service
+  ansible.builtin.include_tasks:
+    file: restart_dns.yml
--- a/roles/reproducer/tasks/configure_crc.yml
+++ b/roles/reproducer/tasks/configure_crc.yml
@@ -75,5 +75,13 @@
       local=/testing/
       domain=crc.testing
 
+- name: Configure DNS in CRC VM
+  delegate_to: crc-0
+  block:
+    - name: Configure DNS in CRC VM
+      ansible.builtin.include_role:
+        name: openshift_setup
+        tasks_from: patch_dnsmasq.yml
+
 - name: Flush handlers
   ansible.builtin.meta: flush_handlers
--- a/roles/rhol_crc/molecule/default/converge.yml
+++ b/roles/rhol_crc/molecule/default/converge.yml
@@ -22,5 +22,19 @@
     cifmw_path: "{{ ansible_user_dir }}/.crc/bin:{{ ansible_user_dir }}/.crc/bin/oc:{{ ansible_user_dir }}/bin:{{ ansible_env.PATH }}"
     cifmw_manage_secrets_pullsecret_content: >
       {"auths":{"fake":{"auth": "Zm9vOmJhcgo="}}}
-  roles:
-    - role: "rhol_crc"
+  tasks:
+    - name: Add crc hostname with it's IP to /etc/hosts
+      become: true
+      ansible.builtin.lineinfile:
+        path: /etc/hosts
+        line: "192.168.130.11 crc"
+
+    - name: Add the crc host dynamically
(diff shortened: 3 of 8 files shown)


lewisdenny:
recheck

lewisdenny:
recheck

lewisdenny:
recheck

lewisdenny:
recheck

lewisdenny:
This buildset passed on 4.14 and 4.15: https://review.rdoproject.org/zuul/buildset/ff22bae9fc0e45fb9e49bd6411c86f62

Removing the 4.14 job now to merge this.

roles/rhol_crc/tasks/main.yml
```diff
@@ -137,3 +137,17 @@
 - name: Add crc kubeconfig
   when: cifmw_rhol_crc_creds | bool
   ansible.builtin.import_tasks: add_crc_creds.yml
+
+- name: Configure DNS in CRC VM
+  delegate_to: crc
```
cescgina, line 142: this should use the `ocps` group, otherwise it will fail with the reproducer when the crc vm is called `crc-0`

roles/rhol_crc/molecule/default/converge.yml
```diff
@@ -22,5 +22,19 @@
...
     cifmw_manage_secrets_pullsecret_content: >
       {"auths":{"fake":{"auth": "Zm9vOmJhcgo="}}}
-  roles:
-    - role: "rhol_crc"
+  tasks:
+    - name: Add crc hostname with it's IP to /etc/hosts
+      become: true
+      ansible.builtin.lineinfile:
+        path: /etc/hosts
+        line: "192.168.130.11 crc"
```
cjeanner, line 30: not sure the hard-coded IP is right here? Do we really need that record actually?

cjeanner: scratch that  it's in molecule

roles/openshift_setup/tasks/patch_dnsmasq.yml
```diff
@@ -0,0 +1,81 @@
...
+        src: "/etc/ci/env/networking-info.yml"
+
+    - name: Decode network-info.yml
+      ansible.builtin.set_fact:
+        _decoded_net_env: >-
+          {{
+            _patch_dns_network_info_slurp['content'] |
+            b64decode | from_yaml
+          }}
+
```
cjeanner, line 21: I'd use the networking_mapper/load_env_definition.yml:
https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/networking_mapper/tasks/load_env_definition.yml

roles/rhol_crc/tasks/main.yml
cjeanner: We can keep `crc` since it's the default name, and leverage `hosts: crc, ocps` to cover both use-cases.

roles/rhol_crc/tasks/main.yml
```diff
@@ -137,3 +137,17 @@
...
+  become: true
```
cjeanner, line 143: not sure `become` is wise here?

roles/rhol_crc/tasks/main.yml
```diff
@@ -137,3 +137,17 @@
...
+  block:
+    - name: Patch dnsmasq service
+      ansible.builtin.include_role:
+        name: openshift_setup
+        tasks_from: patch_dnsmasq.yml
+
+    - name: Restart dnsmasq service
```
cjeanner, line 150: ditto - that task is already called from within the patch_dnsmasq.yml

roles/rhol_crc/tasks/main.yml
```diff
@@ -137,3 +137,17 @@
 - name: Add crc kubeconfig
   when: cifmw_rhol_crc_creds | bool
   ansible.builtin.import_tasks: add_crc_creds.yml
+
+- name: Configure DNS in CRC VM
```
cjeanner, line 141: I see some issues calling this from rhols_crc/main.yml:
- there's a good chance the networks would still be CRC default ones in there, so the whole update would be for nothing.
- we won't get the updated dnsmasq config from within the reproducer: it deploys CRC using the defaults, then creates an overlay and then only connects to the networks - meaning we would end with a faulty CRC (yes, it's currently broken in there :( ).

So I propose to NOT call those tasks from within rhol_crc, but from the pelace where CRC is really configured/consumed:
- reproducer/configure_crc.yml
- (the playbook responsible to bootstrap nested CRC)

Any thoughts?

roles/openshift_setup/tasks/patch_dnsmasq.yml
lewisdenny: Thanks, that's much nicer.

roles/rhol_crc/tasks/main.yml
lewisdenny: To be honest I thought rhol_crc was the place crc is really configured/consumed as per the readme `Role to deploy get, configure and start RHOL/CRC instance`

We have 3 situations I'm trying to cover:
- Nested CRC covered by rhol_crc role as I don't see a playbook where we configure nested CRC, I just see us calling the role in `02-infra`
- reproducer CRC now covered by `configure_crc.yml`
- multinode CRC covered by `ci/playbooks/multinode-customizations.yml`

roles/openshift_setup/tasks/patch_dnsmasq.yml
lewisdenny: Actually, we don't use the mapper expect in reproducer/VA jobs right?

lewisdenny: Any downside to using `networking-info.yml`? I believe that is available in all situations.

roles/rhol_crc/tasks/main.yml
cescgina: right now I think we should be ok for the nested and multinode crc, but would fail for the reproducer crc case, since there is no `crc` node and the delegate below would fail

roles/openshift_setup/tasks/patch_dnsmasq.yml
cjeanner: the mapper is widely used in the reproducer. networking-info.yml is CI (deprecated) only - should really rely on the the mapper.

roles/rhol_crc/tasks/main.yml
lewisdenny: For the reproducer case I added[1] which I don't think is defined when we call the rhol_crc role from reproducer so it will be skipped.
[1] 
```
  when:
    - cifmw_use_crc is defined
    - cifmw_use_crc | bool
```
Then in `roles/reproducer/tasks/configure_crc.yml` I use `delegate_to: crc-0`

How does this sound? It's difficult covering all the  scenarios, one more reason I wish we could drop nested CRC already (excluding reproducer :) )

cescgina: I think that should work, I'll duplicate the PR minus the zuul.d changes and test it downstream

roles/openshift_setup/tasks/patch_dnsmasq.yml
```diff
@@ -0,0 +1,82 @@
+---
+- name: Ensure file is present
```
cescgina, line 2: this task doesn't work well with the reproducer, the networking-info is not present in the crc node. I'm testing now delegating this task to the controller with https://github.com/openstack-k8s-operators/ci-framework/pull/1983/files#diff-12658087bbb21e2faf20276318fd200ebb2aa8a6b62ab8b1aab384b32cc1a6d3R3

lewisdenny: Thank you @cescgina, as Cédric has mentioned networking-info deprecated. `networking-environment-definition.yml` annoyingly isn't available everywhere though and I'm not sure what's involved with getting it into all our deployment scenarios.

roles/openshift_setup/tasks/patch_dnsmasq.yml
cescgina: @cjeanner shouldn't we use both then here? for nested and extracted crc I don't think we'll have the mapper setup when calling this, while for the reproducer we can use the mapper. IMO we could rely on networking-info for now and introduce the mapper as a follow-up

roles/openshift_setup/tasks/patch_dnsmasq.yml
```diff
@@ -0,0 +1,82 @@
...
+    get_attributes: false
+    get_checksum: false
+    get_mime: false
+
+
+- name: Shared parameter across blocks
+  vars:
+    _crc_default_net_ip: >-
+      {{
+        _decoded_net_env.crc_ci_bootstrap_networks_out.crc.default.ip | default('192.168.130.11')
```
cescgina, line 36: here we need to add a `default(_decoded_net_env.crc_ci_bootstrap_networks_out['crc-0'].default.ip)` since the crc vm in the reproducer case is called `crc-0`

roles/reproducer/tasks/configure_crc.yml
```diff
@@ -75,5 +75,13 @@
       local=/testing/
       domain=crc.testing
 
+- name: Configure DNS in CRC VM
```
cescgina, line 78: after testing more extensively, this does not seem the right place, the networking info is not pushed to the controller after the `pre-ci-play.yml` file https://github.com/openstack-k8s-operators/ci-framework/blob/415def67377a6e816117046a9b95766aab2467e8/roles/reproducer/files/pre-ci-play.yml#L74 which runs after this

lewisdenny:
I'm going to close this patch and fix the issue in a more simple way, our deployment methods are quite diverged so standardizing the DNS configuration is not correct direction right now as this issue is blocking a CIX.

PR #1951, opened 2024-06-25
Correct OCP related DNS configuration

cjeanner:
This patch correct how and what we inject in the cifmw-dnsmasq instance
for OCP:
- api, api-int and pxe must be actual A/AAAA records, not wildcard
- apps has to be a wildcard

We therefore leverage the dnsmasq role tasks to provision what we can.
The PTR record as well as some configurations are still in a new file,
not directly managed by dnsmasq. We have to add support for creating
PTR (maybe directly in the host-record? but that might lead to some
issues), and injecting "late configurations".

With this PR, the name resolutions are still properly working:
```
[zuul@hostname ~]$ dig +short api.ocp.openstack.lab
192.168.111.2
[zuul@hostname ~]$ dig +short x.apps.ocp.openstack.lab
192.168.111.3
[zuul@hostname ~]$ dig +short pxe.ocp.openstack.lab
172.22.0.2
```

### DIFF
--- a/roles/reproducer/tasks/ocp_layout.yml
+++ b/roles/reproducer/tasks/ocp_layout.yml
@@ -1,4 +1,28 @@
 ---
+# Let's patch the OCP with the needed bits such as amounts.
+- name: Generate patched overrides for devscripts
+  vars:
+    _num_workers: >-
+      {{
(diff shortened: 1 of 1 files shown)


cjeanner:
recheck

raukadah:
/approve

Looks good, Works with VA1.

PR #1952, opened 2024-06-25
Put ssh_jumper to use in libvirt_manager role

cjeanner:
This patch changes some behavior in ssh_jumper to make its integration
smoother.

It also integrate the role in libvirt_manager in order to manage the
various ssh configuration we usually inject on the hypervisor and
ansible controller (laptop, etc).

  - [x] README in the role

### DIFF
--- a/roles/libvirt_manager/tasks/manage_vms.yml
+++ b/roles/libvirt_manager/tasks/manage_vms.yml
@@ -1,44 +1,37 @@
 ---
-- name: "(localhost) Inject ssh jumpers for {{ vm }}"
-  when:
-    - inventory_hostname != 'localhost'
-    - inventory_hostname != 'instance'  # needed for molecule
-  delegate_to: localhost
+- name: "Push ssh jumper/configuration for {{ vm }}"
   vars:
-    proxy_hostname: "{{ ansible_host | default(inventory_hostname) }}"
-  ansible.builtin.blockinfile:
-    create: true
-    path: "{{ lookup('env', 'HOME') }}/.ssh/config"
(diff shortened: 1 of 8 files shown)


roles/libvirt_manager/tasks/manage_vms.yml
```diff
@@ -1,44 +1,37 @@
...
-        UserKnownHostsFile /dev/null
+    _user: "{{ 'core' if vm is match('^(crc|ocp).*') else 'zuul' }}"
+    dataset:
+      ssh_dir: "{{ ansible_user_dir }}/.ssh"
+      user: "{{ _user }}"
+      hostname: "{{ extracted_ip }}"
+      patterns:
+        - "{{ vm }}"
+        - "{{ vm }}.{{ inventory_hostname }}"
+        - "cifmw-{{ vm }}"
```
hjensas, line 12: We are removing the ip address from patterns, I guess that is ok?

hjensas (approved):

roles/libvirt_manager/tasks/manage_vms.yml
cjeanner: I checked the role, it appends the "hostname" value:
https://github.com/openstack-k8s-operators/ci-framework/blob/c6806319c0620e34df502fd2e78ae8235edd6cdf/roles/ssh_jumper/templates/ssh_host.conf.j2#L1

hjensas: ah, right - I did that. :rofl: All good. :+1:

PR #1954, opened 2024-06-26
Hook: ironic_inject_tls - IPA images with certs

hjensas:
Add a hook playbook to inject internal-ca-bundle.pem sourced from the combined-ca-bundle secret.

Also uploads the images to glance, and configures the ironic nodes to use the images.

### DIFF
--- a/hooks/playbooks/ironic_inject_tls.yml
+++ b/hooks/playbooks/ironic_inject_tls.yml
@@ -1,0 +1,121 @@
+---
+- name: Inject CA bundle in ironic-python-agent and create images
+  hosts: "{{ cifmw_target_hook_host | default('localhost') }}"
+  gather_facts: false
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  tasks:
+    - name: Collect facts
+      ansible.builtin.setup:
+        gather_subset:
+          - '!all'
+          - '!min'
+          - user_gid
+          - user_id
+
+    - name: Create directory
+      ansible.builtin.file:
+        path: "{{ cifmw_basedir }}/artifacts/ironic_ipa_tls/initramfs"
+        state: directory
+        mode: "0755"
+
+    - name: Get internal-ca-bundle and IPA image from k8s resources
+      ansible.builtin.set_fact:
+        internal_ca_bundle: >-
+          {{
+            query(
+              'kubernetes.core.k8s',
+              api_version='v1',
+              kind='Secret',
+              resource_name='combined-ca-bundle',
+              namespace=namespace
+            ) | first
+              | community.general.json_query('data."internal-ca-bundle.pem"')
+          }}
+        ironic_python_image: >-
+          {{
+            query(
+              'kubernetes.core.k8s',
+              api_version='v1beta1',
+              kind='Ironic',
+              resource_name='ironic',
+              namespace=namespace
+            ) | map(attribute='spec.images.ironicPythonAgent')
+              | first
+          }}
+
+    - name: Write internal-ca-bundle to file
+      ansible.builtin.copy:
+        dest: "{{ cifmw_basedir }}/artifacts/ironic_ipa_tls/internal-ca-bundle.pem"
+        content: "{{ internal_ca_bundle | b64decode }}"
+        mode: "0644"
+
+    - name: Pull and extract ironic-python-ironic
+      containers.podman.podman_container:
+        image: "{{ ironic_python_image }}"
+        name: ironic-python-agent
+        rm: true
+        volumes:
+          - "{{ cifmw_basedir }}/artifacts/ironic_ipa_tls/:/target:Z"
+
+    - name: Install CA bundle in initramfs
+      become: true
+      cifmw.general.ci_script:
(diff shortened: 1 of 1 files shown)


hooks/playbooks/ironic_inject_tls.yml
```diff
@@ -0,0 +1,117 @@
...
+        internal_ca_bundle: >-
+          {{
+            query(
+              'kubernetes.core.k8s',
+              api_version='v1',
+              kind='Secret',
+              resource_name='combined-ca-bundle',
+              namespace=_namespace
+            ) | first
+              | community.general.json_query('data."internal-ca-bundle.pem"')
```
cjeanner, line 36: we try to avoid `json_query` whenever possible. Would it be possible to use `select`/`selectattr` and `map` instead?

hooks/playbooks/ironic_inject_tls.yml
```diff
@@ -0,0 +1,117 @@
...
+        ironic_python_image: >-
+          {{
+            query(
+              'kubernetes.core.k8s',
+              api_version='v1beta1',
+              kind='Ironic',
+              resource_name='ironic',
+              namespace=_namespace
+            ) | first
+              | community.general.json_query('spec.images.ironicPythonAgent')
```
cjeanner, line 47: ditto - if we can avoid `json_query`, it would be better

hooks/playbooks/ironic_inject_tls.yml
```diff
@@ -0,0 +1,117 @@
...
+      containers.podman.podman_container:
+        image: "{{ ironic_python_image }}"
+        name: ironic-python-agent
+        rm: true
+        volumes:
+          - "{{ cifmw_basedir }}/artifacts/ironic_ipa_tls/:/target:Z"
+
+    - name: Install CA bundle in initramfs
+      become: true
+      ansible.builtin.shell: |
```
cjeanner, line 66: maybe `cifmw.general.ci_script` ?

hooks/playbooks/ironic_inject_tls.yml
```diff
@@ -0,0 +1,117 @@
...
+        mode: "0644"
+
+    - name: Recursively delete the content in artifacts/ironic_ipa_tls/initramfs
+      become: true
+      ansible.builtin.file:
+        path: "{{ cifmw_basedir }}/artifacts/ironic_ipa_tls/initramfs/"
+        state: absent
+
+    - name: Create images in openstack and configure ironic node driver info
+      ansible.builtin.shell: |
```
cjeanner, line 92: ditto - `cifmw.general.ci_script` might be nice

hooks/playbooks/ironic_inject_tls.yml
```diff
@@ -0,0 +1,117 @@
+---
+- name: Inject CA bundle in ironic-python-agent and create images
+  hosts: "{{ cifmw_target_hook_host | default('localhost') }}"
+  gather_facts: false
+  vars:
+    _namespace: openstack
```
cjeanner, line 6: this one is provided by the run_hook/playbook.yml:
https://github.com/openstack-k8s-operators/ci-framework/blob/43072d16e25c0ad62993e027ed20eb5b0b5024ea/roles/run_hook/tasks/playbook.yml#L34-L35

you want `namespace`.

cjeanner (changes requested):

hooks/playbooks/ironic_inject_tls.yml
hjensas: Why do we avoid `json_query`?
I spent a bunch of time trying to figure out how to do it with select/selectattr and map. That stuff is like french to me ...

hjensas: @cjeanner - so how do we escape the `.` in the field?
```
    - name: Get internal-ca-bundle and IPA image from k8s resources
      ansible.builtin.set_fact:
        internal_ca_bundle: >-
          {{
            query(
              'kubernetes.core.k8s',
              api_version='v1',
              kind='Secret',
              resource_name='combined-ca-bundle',
              namespace=namespace
            ) | map(attribute='data') | map(attribute=['internal-ca-bundle.pem']) | first
          }}
```

this code works on other secrets to get for example `data.host` from the `rabbitmq-cell1-default-user`
```
"The task includes an option with an undefined variable. The error was: dict object has no element ['internal-ca-bundle.pem'].
```

hooks/playbooks/ironic_inject_tls.yml
hjensas: For this case I changed it to use map instead. It works because there is no `.` in the key.

hooks/playbooks/ironic_inject_tls.yml
hjensas: Done.

hooks/playbooks/ironic_inject_tls.yml
hjensas: Done.

hooks/playbooks/ironic_inject_tls.yml
rebtoor: we can address this later :) there are already other occurrencies of json_query in the repo.

PR #1956, opened 2024-06-26
Add tolerations to lvms cluster resource

lmiccini:
When using specially tainted workers we usually specify tolerations for pods that should land on these nodes like:

  tolerations:
  - key: "testOperator" value: "true" effect: "NoSchedule
  - key: "testOperator" value: "true" effect: "NoExecute

Similar tolerations must be applied to the lvms pods so that the lvms storageclass can be used across the cluster regardless of how workers are tagged/tainted.

### DIFF
--- a/roles/ci_lvms_storage/templates/lvms-cluster.yaml.j2
+++ b/roles/ci_lvms_storage/templates/lvms-cluster.yaml.j2
@@ -5,6 +5,14 @@ metadata:
   name: {{ cifmw_lvms_cluster_name }}
   namespace: {{ cifmw_lvms_namespace }}
 spec:
+  {% if ci_lvms_storage_tolerations -%}
+  tolerations:
+  {% for item in ci_lvms_storage_tolerations -%}
+  - key: "{{ item['key'] }}"
+    value: "{{ item['value'] }}"
+    effect: "{{ item['effect'] }}"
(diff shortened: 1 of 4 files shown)


fultonj:
Would you please add `cifmw_test_operator_tolerations` to `defaults/main.yml` and document an example of it in the README.md?

Otherwise, this looks fine. Thanks

fultonj:
> Would you please add `cifmw_test_operator_tolerations` to `defaults/main.yml` 

Perhaps default it to `{}`

roles/ci_lvms_storage/templates/lvms-cluster.yaml.j2
```diff
@@ -5,6 +5,14 @@ metadata:
   name: {{ cifmw_lvms_cluster_name }}
   namespace: {{ cifmw_lvms_namespace }}
 spec:
+  {% if cifmw_test_operator_tolerations -%}
```
fultonj, line 8: can we call it `cifmw_lvms_operator_tolerations`?

fultonj: Actually it should be called `ci_lvms_storage_tolerations`

fultonj (approved):
/approve
/lgtm

PR #1957, opened 2024-06-26
[fetch_compute hook]filter controller hostname from groups

raukadah:
In crc reproducer, we have controller-0 as a host for ansible controller but in Zuul CI, we use controller as a hostname.
    
When fetch_compute_facts runs on controller-0 in CI, it does not resolve the hostname in CI but should work in reproducer.
     
In order to fix that, we are filter hostname from groups to set proper controller hostname in order to delegate the task.

- [x] Tested with crc job reproducer downstream


cescgina (commented):
/approve

PR #1961, opened 2024-06-26
Consolidate, correct and test devscripts checks

cjeanner:
With the big changes introduced by #1777, some of the checks done in
devscripts to ensure the OCP cluster was present were broken or not
properly scaling.

This patch corrects this situation, and ensures the tests are properly
working with a dedicated molecule scenario.

It also makes a small logical correction in the reproducer role, to
ensure we don't do same tasks multiple times for nothing, and also to
ensure we're testing the right, up-to-date content.


cescgina (approved):
code wise lgtm, but have little context for the changes. What jobs could this affect, the VA ones?

roles/devscripts/tasks/get_ocp_vips.yml
hjensas, line 1: \o/ we got rid of this.

cjeanner:
Anything using devscripts - so mostly VA, yes.

hjensas (approved):

PR #1962, opened 2024-06-26
Drop set_openstack_containers from edpm_prepare role

raukadah:
All the reference of set_openstack_containers role is replace with update_containers role.

We are keeping set_openstack_containers role as update team wants to use it to perform minor update with installed operators.


raukadah:
Closing this pr, set_openstack_containers may be needed for update testing.

PR #1964, opened 2024-06-26
Create proper DNS records for each host/network pair

cjeanner:
Since we now have a nice DNS service, let's use it.

This patch enables a series of features:

From now on, we build a domain based on the network name + deploy domain
(either crc.testing, or the one set in devscripts). A translation is
done for the "control plane" network (usually, osp_trunk is ctlplane).

While we reserve IPs for the nodes on each networks, we also generate
DNS entries. Here again, we translate the "control plane", taking
advantage of the networking_mapper translation table.

These features should make a lot of things far easier to handle. Note
that those records exist only for the networks where dnsmasq is
configured to provide services.

It also corrects a small nit in dnsmasq role where the built host-record
might have "empty field", leading to potential issues, for example:
`host-record=master-2.openstack.lab,192.168.111.13,`

This fix corrects the line to match the expected format:
`host-record=crc-0.crc.testing,192.168.100.30`

### DIFF
--- a/roles/dnsmasq/templates/network.conf.j2
+++ b/roles/dnsmasq/templates/network.conf.j2
@@ -1,7 +1,8 @@
 # Managed by ci-framework/dnsmasq
-{% if cifmw_dnsmasq_network_definition['ranges'] | selectattr('start_v6', 'defined')               -%}
+{% if cifmw_dnsmasq_network_definition.ranges | selectattr('start_v6', 'defined') |
+    rejectattr('start_v6', 'match', '^$')                                                          -%}
 enable-ra
-{% endif                                                                                            %}
+{% endif                                                                                           -%}
 
 {% for range in cifmw_dnsmasq_network_definition['ranges']                                         -%}
 {%   if range.start_v4 is defined and range.start_v4 | length > 0                                  -%}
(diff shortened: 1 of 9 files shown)


roles/dnsmasq/templates/network.conf.j2
```diff
@@ -1,7 +1,8 @@
 # Managed by ci-framework/dnsmasq
-{% if cifmw_dnsmasq_network_definition['ranges'] | selectattr('start_v6', 'defined')               -%}
+{% if cifmw_dnsmasq_network_definition.ranges.start_v6 is defined and
+        cifmw_dnsmasq_network_definition.ranges.start_v6 | length > 0                              -%}
```
hjensas, line 3: This breaks IPv6 - enabla-ra does not get enabled.

```
[zuul@host26 ~]$ grep enable-ra -R /etc/cifmw-dnsmasq.* || echo "NOT FOUND"
NOT FOUND
```
ranges is a list? so you need to iterate on ranges and check if any of the ranges defined has `start_v6`.

hjensas (changes requested):

roles/dnsmasq/templates/network.conf.j2
cjeanner: thing is, `start_v6` is always defined, it can be empty. Lemme rework that test.

cjeanner:
/hold cancel

Some more tests have been added to cover the `enable-ra` case.

hjensas (approved):
Looks good, thanks for adding the test coverage for enable-ra.

PR #1968, opened 2024-06-27
Create a new .utility DNS zone

cjeanner:
This zone will allow to get fixed names to access utility services, such
as the IPMI.
Using names instead of IPs ensures we can easily inject IPv6 in the
game, without having to chase down IPs in the code: just update the DNS
record to get either IPv4, or IPv6, or even dual-stack.

This patch also updates sushy_emulator role to remove some hacks
injected as "quick workaround". The cifmw_sushy_emulator_listen_ip
parameter is now passed as-is and, if needed, any IPv6 we really pass
there would be properly wrapped.
In the reproducer context, it will get a hostname instead, ensuring we
pass via the .utility zone and appropriate records.

- [x] 3-nodes with CRC (from scratch)
- [x] 3-nodes with CRC (re-using image)
- [x] va-hci with OCP (from scratch)
- [x] va-hci with OCP (re-using images)
- [x] NFV


roles/reproducer/tasks/configure_controller.yml
```diff
@@ -21,7 +21,12 @@
   delegate_to: controller-0
   delegate_facts: false
   vars:
+    cifmw_sushy_emulator_hypervisor_target: "{{ inventory_hostname }}"
+    cifmw_sushy_emulator_install_type: podman
+    cifmw_sushy_emulator_hypervisor_target_connection_ip: "{{ inventory_hostname }}.utility"
+    # cifmw_sushy_emulator_hypervisor_target_connection_ip: "{{ network_bridge_info['cifmw-public'] }}"
     cifmw_sushy_emulator_basedir: "{{ _ctl_reproducer_basedir }}"
+    cifmw_sushy_emulator_listen_ip: "controller-0.utility"
```
hjensas, line 29: By default it listens on all IPs, I don't think we need to set the listen_ip here.

roles/reproducer/tasks/prepare_networking.yml
```diff
@@ -143,6 +143,44 @@
...
+    # TODO: feed IPv6 data in the record
+    - name: Build utility records dataset
+      when:
+        - hostvars[host].ansible_default_ipv4.address | default('') | length > 0
+      vars:
+        _host_record:
+          - names:
+              - "{{ host }}.utility"
+            ips:
+              - "{{ hostvars[host].ansible_default_ipv4.address | default('') }}"
```
hjensas, line 159: Don't we have a map of what IPs is on each network for hosts? If it is not including the hypervisor bridge ip - we could add it? And then instead of hostvars default route interface. We go `our_map[host][pub_net][ip_address]` ?

I think it would make our life simpler if we used one of the bridges, for example the "pub_net" bridge ip address for the "hypervisor". Stay inside the lab - the public interface on the hypervisor is not where we want traffic to terminate?

roles/sushy_emulator/tasks/create_container.yml
```diff
@@ -42,14 +42,8 @@
...
-    _ip_address: >-
-      {%- if cifmw_sushy_emulator_listen_ip in ['::', '0.0.0.0'] -%}
-      {{ hostvars['controller-0'].host_ip | default(hostvars['controller-0']['ansible_host']) }}
-      {%- else -%}
-      {{ cifmw_sushy_emulator_listen_ip }}
-      {%- endif -%}
+    _listen: "{{ cifmw_sushy_emulator_listen_ip | ansible.utils.ipwrap }}"
   ansible.builtin.set_fact:
-    _sushy_url: "http://{{ _ip_address | ansible.utils.ipwrap }}:8000"
+    _sushy_url: "http://{{ _listen }}:8000"
```
hjensas, line 49: I would love to see a CNAME for this, `sushy-emaulator.utility` - then we can drop setting this fact.
... but that should be follow up. :smile:

roles/reproducer/tasks/prepare_networking.yml
hjensas: Also, do we need to put entries for controller-0, compute, ocp, etc on the utlity domain? It will point to same address that we already have in `compute-0.ocp.openstack.lab` no? Will it get confusing to get search domains good in resolv.conf?

roles/reproducer/tasks/configure_controller.yml
```diff
@@ -21,7 +21,12 @@
   delegate_to: controller-0
   delegate_facts: false
   vars:
+    cifmw_sushy_emulator_hypervisor_target: "{{ inventory_hostname }}"
+    cifmw_sushy_emulator_install_type: podman
+    cifmw_sushy_emulator_hypervisor_target_connection_ip: "{{ inventory_hostname }}.utility"
```
lewisdenny, line 26: Now we are adding dns in to the mix this should be changed `connection_ip` to something like `connection_endpoint` or `connection_address`

roles/sushy_emulator/tasks/create_container.yml
lewisdenny: Yes that would be a great improvment on this already great improvment. Nice and clean!

lewisdenny (changes requested):

roles/reproducer/tasks/prepare_networking.yml
cjeanner: Lemme check what we have in the networking_mapper. Using the "gateway" might be something, though we'll have to ensure it's properly mapped on an NFV deployment.

As for "controller, compute, ocp" in the utility "zone": no issue afaik, because `.utility` isn't added to the `domain-search` - meaning it's mostly "unknown". At least that's my understanding, but I'm slightly rusted with DNS, didn't play too much for the past 5-6 years... Lemme deploy, do some testing, and we can also sync live on that matter.

roles/sushy_emulator/tasks/create_container.yml
cjeanner: hmm like a `cifmw_sushy_emulator_address` ? To keep it working outside of the reproducer role. I can check that, once the utility zone is accepted (see other comment).

roles/reproducer/tasks/prepare_networking.yml
cjeanner: I think I solved this by consuming the content of the networking_mapper instead. Lemme know if this matches your ideas?
I used "ctlplane" since this network is always present and "local" to the lab. "public" wouldn't really match the need here, nor "ocppr" - or, at least, they aren't listed in the mapper, meaning not that easy to get.

roles/sushy_emulator/tasks/create_container.yml
cjeanner: I introduced a couple of parameters there to make things "nice" Lemme know your thoughts.

roles/reproducer/tasks/configure_controller.yml
cjeanner: Renamed - might need to update the doc, but I'd rather ensure things are OK as-is.

roles/reproducer/tasks/configure_controller.yml
```diff
@@ -314,26 +323,20 @@
...
-            cifmw_sushy_emulator_install_type: podman
-            cifmw_sushy_emulator_hypervisor_target_connection_ip: "{{ network_bridge_info['cifmw-public'] }}"
-          ansible.builtin.include_role:
+          ansible.builtin.import_role:
             name: sushy_emulator
 
-        - name: Let the project know we have Sushy Emulator available
-          ansible.builtin.set_fact:
-            _sushy_emulator_available: true
+        - name: Generate baremetal-info fact
```
cjeanner, line 333: note that everything is now nested under the same `when` - though it might be an issue for vbmc use-case? @hjensas @lewisdenny lemme know.

roles/reproducer/tasks/generate_bm_info.yml
```diff
@@ -72,6 +72,11 @@
 - name: Generate libvirt_manager_bm_info_data fact
   when: item.key is match('cifmw-')
   vars:
+    _sushy_url: >-
```
cjeanner, line 75: this one is built to select between ocp/non-ocp URI version

roles/reproducer/tasks/ocp_layout.yml
```diff
@@ -90,30 +90,25 @@
     - name: Inject A/AAAA records for api/api-int/pxe
       vars:
         cifmw_dnsmasq_host_record:
-          - names: "{{ record.names }}"
-            ips: "{{ record.ips }}"
+          - names:
```
cjeanner, line 93: let's not loop, since the tasks accepts a list already.

roles/reproducer/tasks/prepare_networking.yml
```diff
@@ -152,6 +152,45 @@
...
+        cifmw_dnsmasq_host_record:
+          - names:
+              - "ipmi.utility"
+              - "{{ inventory_hostname }}.utility"
+            ips:
+              - "{{ _target_net.gw_v4 | default('') }}"
+              - "{{ _target_net.gw_v6 | default('') }}"
+            state: present
+          - names:
+              - "sushy.utility"
```
cjeanner, line 186: this should work - though how can we actually test it? @hjensas maybe?

roles/sushy_emulator/README.md
```diff
@@ -20,10 +20,10 @@ Required to installed required packages.
 * `cifmw_sushy_emulator_driver_openstack_client_config_file`: (String) Path to OpenStack config file, used by OpenStack Sushi Emulator driver. Defaults to `/etc/openstack/clouds.yaml`
 * `cifmw_sushy_emulator_driver_openstack_cloud`: (String) Cloud key reference in OpenStack config file. Defaults to `None`
 * `cifmw_sushy_emulator_hypervisor_target` (String) Hostname of the Libvirt hypervisor to connect to.
-* `cifmw_sushy_emulator_hypervisor_target_connection_ip` (String) IP Address used to connect to hypervisor, optional override for `cifmw_sushy_emulator_hypervisor_target` when specific address is required.
+* `cifmw_sushy_emulator_hypervisor_address` (String) IP Address used to connect to hypervisor, optional override for `cifmw_sushy_emulator_hypervisor_target` when specific address is required.
```
cjeanner, line 23: TODO: update docstring - not IP anymore (though it can still be an IP...)

roles/sushy_emulator/tasks/collect_details.yml
```diff
@@ -25,7 +25,7 @@
 
 - name: Set sushy url for ocp installation
   ansible.builtin.set_fact:
-    _sushy_url: "http://sushy-emulator.{{ _ingress_domain.stdout }}"
+    ocp_sushy_uri: "http://sushy-emulator.{{ _ingress_domain.stdout }}"
```
cjeanner, line 28: creating a new internal thing - needed because we can't override parameters via set_fact...

roles/reproducer/tasks/prepare_networking.yml
```diff
@@ -152,6 +152,45 @@
...
+    - name: Build utility records dataset
+      vars:
+        _target_net: "{{ cifmw_networking_env_definition.networks.ctlplane }}"
+        _controller_net: >-
+          {{
+            cifmw_networking_env_definition.instances['controller-0'].networks
+          }}
+        cifmw_dnsmasq_host_record:
+          - names:
+              - "ipmi.utility"
```
cjeanner, line 179: ipmi.utility can't be on the ctlplane - no access from the ocp-bootstrap, it's plugged on ocpbm and ocppr only... dang.

roles/reproducer/tasks/prepare_networking.yml
hjensas: A bit blocked by that unbound issue in the nat64 right now.
This looks good, if it works for IPv4 it should work for IPv6 - I think we can merge this as is and refine IPv6 support if it does not just work.

roles/reproducer/tasks/prepare_networking.yml
hjensas: That is not a problem, we have routing enabled.

cjeanner: nah, I'll just use the "public" net, I see we have it, so all good.

roles/sushy_emulator/README.md
hjensas, line 23: I like this new parameter, I think we should remove cifmw_sushy_emulator_hypervisor_target. Because the hypervisor might not be managed by ansible. Now that we can set the address - pointing sushy at any hypervisor managed by ansible or not does not matter. When the hypervisor is managed, we can easily set `cifmw_sushy_emulator_hypervisor_address` from inventory/facts in the playbook calling this role.

roles/sushy_emulator/tasks/verify.yml
```diff
@@ -36,28 +36,36 @@
...
-  ansible.builtin.uri:
-    url: "{{ _sushy_url}}/redfish/v1/Systems/{{item}}"
-    return_content: true
-    user: "{{ cifmw_sushy_emulator_redfish_username }}"
-    password: "{{ cifmw_sushy_emulator_redfish_password }}"
-  retries: 5
-  delay: 5
-  register: _sushy_emulator_content_system
-  loop: "{{ _cifmw_sushy_emulator_instances }}"
+    - name: Verify connection to baremetal VMs via Sushy Emulator
```
hjensas, line 57: spaces, instead of tabs?

hjensas (commented):
Looks good. a few nits and opinions. Can be follow ups.

roles/sushy_emulator/tasks/verify.yml
cjeanner: hmm `retab` doesn't change anything. It's now nested under a block, thence the right-shift compared to previous content.

cjeanner:
recheck

cjeanner:
recheck

roles/sushy_emulator/tasks/verify.yml
hjensas: I guess I got thrown off by the highligting github is doing.

hjensas (commented):
/lgtm

I have tested this with IPv6, and it allows me to remove dirty workarounds. :+1:

lewisdenny:
/hold
Adding hold so this can merge when all hands are on deck. Patch LGTM, thanks Cédric.

lewisdenny (approved):
/approve

cjeanner:
/hold cancel

let's get it in, so that we can move forward :)

PR #1973, opened 2024-06-27
Get just the nmstate operator when waiting for install

slagle:
When waiting for the nmstate operator to be installed, only that
operator needs to be returned in the results. As CSV's are visible from
all namespaces, restricting by namespace does not have the effect to
only show the CSV's from that namespace.

This task returns all CSV's in the cluster, regardless of namespace, and
fails if more than one CSV is returned as the expected output does not
match.

This commit adds a filter on the expected label for the nmstate operator
so that it will be the only operator returned.

Signed-off-by: James Slagle <jslagle@redhat.com>

  - [x] README in the role


rebtoor:
/approve

lgtm

rebtoor (approved):

PR #1975, opened 2024-06-28
Create dummy interfaces with ci_nmstate

hjensas:
To ensure bridges are UP, create dummy interfaces on all networks.

For IPv6, if the bridge is not UP the system will not assign a link-local IPv6 address to the interface.

### DIFF
--- a/roles/libvirt_manager/tasks/create_networks.yml
+++ b/roles/libvirt_manager/tasks/create_networks.yml
@@ -81,6 +81,36 @@
   loop_control:
     label: "{{ item.name }}"
 
+# NOTE(hjensas): Using a random string for dummy interface names because
+#                otherwise we hit the 15 char limitation on ifname.
+- name: Create dummy interfaces to ensure bridges are UP
+  vars:
+    cifmw_ci_nmstate_unmanaged_node_config: >-
+      {%- set interfaces = []                                            -%}
+      {%- for network in networks | map(attribute="name") | flatten
+            if cifmw_libvirt_manager_network_interface_types[network | regex_replace('cifmw[-_]','')]
+            | default('bridge') == 'bridge'                              -%}
+      {%- set suffix = lookup('password',
+                              '/dev/null',
+                              chars=['ascii_lowercase', 'digits'],
+                              length=8,
+                              seed=network)                              -%}
+      {%- set _ = interfaces.append(
+        {
+          'name': 'dummy-' + suffix,
+          'type': 'dummy',
+          'ipv4': {'enabled': false},
+          'ipv6': {'enabled': false},
+          'controller': network
+        }
+      )                                                                  -%}
+      {%- endfor                                                         -%}
+      {{ {'interfaces': interfaces } }}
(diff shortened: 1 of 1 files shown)


roles/ci_nmstate/tasks/nmstate_unmanaged_provision_node.yml
```diff
@@ -21,6 +21,17 @@
     state: latest  # noqa: package-latest
     use: "{{ hostvars[cifmw_ci_nmstate_unmanaged_host]['ansible_facts']['pkg_mgr'] }}"
 
+#TODO(hjensas): I don't think the cifmw_ci_nmstate_instance_config "direct pass trough"
```
hjensas, line 24: @pablintino - I initially tried to use `cifmw_ci_nmstate_instance_config` - but it failed.

```
TASK [ci_nmstate : Build NMstates from the Networking Environment Definition] ***
fatal: [instance]: FAILED! =>
  msg: 'An unhandled exception occurred while templating ''{{ cifmw_networking_env_definition | ci_nmstate_map_instance_nets(item) }}''. Error was a <class ''KeyError''>, original message: ''parent_interface'''
```

I think that never worked, because we don't load net-env def when `cifmw_ci_nmstate_instance_config` is set?

cjeanner:
/lgtm

I'll let @pablintino have last say since he created the ci_nmstate role/modules.

roles/libvirt_manager/tasks/create_networks.yml
```diff
@@ -81,6 +81,36 @@
...
+            if cifmw_libvirt_manager_network_interface_types[network | regex_replace('cifmw[-_]','')]
+            | default('bridge') == 'bridge'                              -%}
+      {%- set suffix = lookup('password',
+                              '/dev/null',
+                              chars=['ascii_lowercase', 'digits'],
+                              length=8,
+                              seed=network)                              -%}
+      {%- set _ = interfaces.append(
+        {
+          'name': 'dummy-' + suffix,
```
lewisdenny, line 100: [Suggestion]
I think we should add a description[1] to explain why we are creating this dummy interface. Just because this is unexpected, I can imagine people wondering why this dummy interface exists when troubleshooting some other networking issue.

[1] https://nmstate.io/devel/yaml_api.html#description

lewisdenny (approved):
Thank you Harald, this patch looks good I just have a small suggestion. 

I will approve and hold for you to decide to implement the description suggestion or just remove the hold and merge :)

roles/ci_nmstate/tasks/nmstate_unmanaged_provision_node.yml
hjensas: Marking as resolved, ref: https://github.com/openstack-k8s-operators/ci-framework/pull/1995

cjeanner:
/lgtm

There were already LGTM on that PR - last one from @lewisdenny - with a request for change that @hjensas fixed - so we can go with this release. Mostly re-adding the label.

PR #1976, opened 2024-06-28
nat64 - do-daemonize: no

hjensas:
There is a warning in the logs:

unbound[4558:0] warning:
  use-systemd and do-daemonize should not be enabled at the same time

Manual page:

  do-daemonize: <yes or no>
      Enable or disable whether the Unbound server forks into the
      background as a daemon.  Set the value to no when Unbound
      runs as systemd service.  Default is yes.


hjensas:
Included in https://github.com/openstack-k8s-operators/ci-framework/pull/1991

PR #1982, opened 2024-06-28
Fix variable name so it reaches reproducer-variables

lmiccini:


lewisdenny (approved):
/approve
Thanks, looks good to me, I don't see any `ci_lvms_storage_tolerations` not being renamed.

PR #1983, opened 2024-06-28
Add again DNS configuration for crc reproducer path 

cescgina:
Add again some tasks to properly configure dnsmasq for crc in the
reproducer. This is needed to make the zuul reproducer job green again.
Now the crc node should have the hypervisor in the resolv.conf but
dnsmasq listening on the right addresses.

Started as a duplicate
https://github.com/openstack-k8s-operators/ci-framework/pull/1944/
without zuul changes to test with downstream zuul, but now proposing this fix for the crc reproducer path to close CIX.

### DIFF
--- a/roles/openshift_setup/tasks/restart_dns.yml
+++ b/roles/openshift_setup/tasks/restart_dns.yml
@@ -1,0 +1,40 @@
+- name: Check which dnsmasq config we must edit
+  register: _dnsmasq
+  ansible.builtin.stat:
+    path: '/srv/dnsmasq.conf'
+    get_attributes: false
+    get_checksum: false
+    get_mime: false
+
+- name: Restart dnsmasq service if used
+  become: true
+  when:
+    - not _dnsmasq.stat.exists
+  ansible.builtin.service:
+    name: dnsmasq
+    state: restarted
+
+- name: Manage old dnsmasq container
+  when:
+    - _dnsmasq.stat.exists
+  block:
+    # Avoid 'state: restarted' due to issues with IP not
+    # available when crc-dnsmasq starts
+    - name: Stop dnsmasq
+      become: true
+      ansible.builtin.systemd:
+        state: stopped
+        name: crc-dnsmasq
+
+    - name: Start dnsmasq
+      become: true
+      ansible.builtin.systemd:
+        state: started
+        name: crc-dnsmasq
+      register: _dnsmasq_start_reg
+      retries: 15
+      delay: 20
+      until:
+        - _dnsmasq_start_reg.failed is false
+        - _dnsmasq_start_reg.status is defined
+        - _dnsmasq_start_reg.status.ActiveState == "active"
--- a/roles/reproducer/tasks/configure_crc.yml
+++ b/roles/reproducer/tasks/configure_crc.yml
@@ -5,6 +5,47 @@
   ansible.builtin.slurp:
     path: /etc/ci/env/networking-environment-definition.yml
 
+- name: Configure CRC services
+  delegate_to: crc-0
+  vars:
+    _parsed: "{{ _net_conf.content | b64decode | from_yaml}}"
+    _crc: "{{ _parsed.instances['crc-0'].networks.ctlplane }}"
+  block:
+    - name: Check which dnsmasq config we must edit
+      register: _dnsmasq
+      ansible.builtin.stat:
+        path: '/srv/dnsmasq.conf'
+        get_attributes: false
+        get_checksum: false
+        get_mime: false
+
+    - name: Configure local DNS for CRC pod
+      become: true
+      vars:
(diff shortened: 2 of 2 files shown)


roles/reproducer/tasks/configure_crc.yml
```diff
@@ -5,6 +5,60 @@
...
+      vars:
+        _ctlplane: "{{ _parsed.networks.ctlplane }}"
+      community.general.nmcli:
+        autoconnect: true
+        conn_name: private_net
+        ifname: "{{ _crc.interface_name }}"
+        type: ethernet
+        ip4: "{{ _crc.ip_v4 }}/24"
+        never_default4: true
+        state: present
```
cjeanner, line 25: useless, we have dhcp there with fixed IP

roles/reproducer/tasks/configure_crc.yml
```diff
@@ -5,6 +5,60 @@
...
+
+    - name: Check which dnsmasq config we must edit
+      register: _dnsmasq
+      ansible.builtin.stat:
+        path: '/srv/dnsmasq.conf'
+        get_attributes: false
+        get_checksum: false
+        get_mime: false
```
cjeanner, line 33: not sure if still needed? What's the "oldest" CRC version we have?

roles/reproducer/tasks/configure_crc.yml
```diff
@@ -5,6 +5,60 @@
...
+
+    - name: Configure local DNS for CRC pod
```
cjeanner, line 35: iirc that's more or less covered by @lewisdenny PR (though I just saw he closed #1944)?

cjeanner (changes requested):

roles/reproducer/tasks/configure_crc.yml
```diff
@@ -5,6 +5,66 @@
...
+            - apps-crc.testing
+            - api.crc.testing
+            - api-int.crc.testing
+
+
+        - name: Ensure dnsmasq listens on correct interfaces
+          vars:
+            _listen_adresses: >-
+              {%- if _crc.ip_v6 is defined and _crc.ip_v6 | length > 0 -%}
+              _crc.ip_v6
```
cjeanner, line 55: you're not building a list I think here, but a string with a space between both elements. So no `join(',')` needed

cescgina: yes, I kept the comma because i thought the `listen-address` field in the dnsmasq config requires a comma-separeted string

roles/reproducer/tasks/configure_crc.yml
```diff
@@ -5,6 +5,66 @@
...
+            regexp: "192.168.130.11"
+            replace: "{{ _crc.ip_v4 }}"
+
+        - name: Configure local DNS for CRC pod with ip_v6
+          when:
+            - _crc.ip_v6 is defined
+            - _crc.ip_v6 | length > 0
+          ansible.builtin.lineinfile:
+            path: "{{ _config_file }}"
+            line: "address=/{{ item  }}/ {{ _crc.ip_v6 }}"
```
cjeanner, line 44: @hjensas heh, so here this is how CRC is configuring its DNS - all "address". We can either do like them, or "fix" the v6 version - meaning a list of key/var:
```yaml
loop:
  - key: apps-crc.testing
    type: address
  - key: api.crc.testing
    type: host-record
  - key: api-int.crc.testing
    type: host-record
```
but then we will face the syntax differences - `host-record=foo,IP` vs `address=/foo/IP`

dnsmasq is so easy -.-....

@cescgina FTR, "address" creates a wildcard (*.foo), while apparently we should get simple A/AAAA for api/api-int records...

IMHO we can stick with the current thing - but in any cases, I'd rather go with a `blockinfile` instead of a loop. Faster, and... well, there we can use the right type :).

roles/reproducer/tasks/configure_crc.yml
cescgina: indeed, that did not do what i expected, thanks @cjeanner I'll update

cescgina:
recheck

cjeanner:
/approve

CRC has no IPv6 support, nor immediate plane to get anything there...

roles/openshift_setup/tasks/restart_dns.yml
```diff
@@ -0,0 +1,40 @@
...
+    path: '/srv/dnsmasq.conf'
+    get_attributes: false
+    get_checksum: false
+    get_mime: false
+
+- name: Restart dnsmasq service if used
+  become: true
+  when:
+    - not _dnsmasq.stat.exists
+  ansible.builtin.service:
```
rebtoor, line 13: use `systemd_service` instead of `service`

raukadah (approved):

PR #1986, opened 2024-06-28
Add a hook to increase quotas for project service

Akrog:
When running Glance using Cinder as a backend the `service` project is being used and tests can go over the volume quota.

This patch adds a playbook hook to increase the volume quota for the `service` project to 50.

Uses the region set in the `.config/openstack/clouds.yaml` of the `openstackclient` pod, and assumes the project name is `service`.

  - [x] README in the role

### DIFF
--- a/hooks/playbooks/service-project-more-quota.yml
+++ b/hooks/playbooks/service-project-more-quota.yml
@@ -1,0 +1,13 @@
+---
+- name: Increase quotas for the service project
+  hosts: "{{ cifmw_target_hook_host | default('localhost') }}"
+  gather_facts: false
+  tasks:
+    - name: Set the volumes to 50
+      environment:
+        KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+        PATH: "{{ cifmw_path }}"
+      ansible.builtin.shell: |
+        oc project {{ namespace }}
+        oc rsh openstackclient \
+           openstack quota set --volumes 50 service


hooks/playbooks/service-project-more-quota.yml
```diff
@@ -0,0 +1,13 @@
...
+  gather_facts: false
+  tasks:
+    - name: Set the volumes to 50
+      environment:
+        KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+        PATH: "{{ cifmw_path }}"
+      ansible.builtin.shell: |
+        oc project {{ namespace }}
+        oc rsh openstackclient \
+           openstack quota set --volumes 50 service
```
tosky, line 13: Thank you, this is something I wanted to implement. I didn't move forward because I wasn't sure whether I can assume the name 'service' is kind of hardcoded in operator-based OSP deployments, or it should be a parameter, or it should be gathered from some source dynamically.

Akrog: In today's code this is [hardcoded in the cinder-operator](https://github.com/openstack-k8s-operators/cinder-operator/blob/8f627d558478e8e5b92177bf18db5eb2f32cbfba/templates/cinder/config/00-global-defaults.conf#L94):

```ini
[service_user]
send_service_user_token = True
auth_url = {{ .KeystoneInternalURL }}
auth_type = password
project_domain_name = Default
user_domain_name = Default
project_name = service
```

tosky:
Internal testing shows that the hook does what it is meant to do and the quota is increased. I think this review can be moved out of draft.

fmount (approved):
/lgtm

PR #1991, opened 2024-06-29
nat64 - Fix molecule, and touch ups to stabilize

hjensas:
The molecule tests was failing because the interfaces did not end up in the 'libvirt' firewall zone.
    
Added the zone attribute to the libvirt bridge element. This should work according to libvirt documentation, but for some reason not in this test environment. So, had to do that manually with firewalld.
    
Also:

- Pin dnsmasq to the v2.89 tag.
  Should be more stable than running of the main branch with frequent changes.
- Set do-daemonize: no in unbound config.
  This removes a warning logged by the unbound service, do-daemonize and use-systemd should not be used togheter.
    
There is a lot of "crude" debug info gathering added in the molecule tests, I think we should keep that around as is to make troubleshooting any potentital future issues in this job.

  - [ ] README in the role


roles/nat64_appliance/files/elements/nat64-router/install.d/70-dnsmasq
```diff
@@ -21,7 +21,7 @@ set -o pipefail
 
 pushd /root
 
-git clone git://thekelleys.org.uk/dnsmasq.git
+git clone --depth 1 --branch v2.89 git://thekelleys.org.uk/dnsmasq.git
```
cjeanner, line 24: Guess we can't get a parameter there? static files and all?

roles/nat64_appliance/molecule/default/converge.yml
```diff
@@ -108,6 +129,27 @@
         name: br-mol
         uri: 'qemu:///system'
 
+    # TODO(hjensas): With the zone set in libvirt XML these bridges "should"
```
cjeanner, line 132: don't we need to also get that task in the actual role tasks? If we can't trust libvirt to push the interfaces in the right zone for molecule, we might face the same from standard usage?

roles/nat64_appliance/files/elements/nat64-router/install.d/70-dnsmasq
hjensas: I think we can, but it would require some work. Pass a variable to diskimage-builder. I'd prefer to do it in follow-up.

roles/nat64_appliance/molecule/default/converge.yml
hjensas: hmm, yeah I think you are right. Looking at my lab it is not adding the br-64v6 interface to the zone.

```
libvirt (active)
  target: ACCEPT                    
  icmp-block-inversion: no                                          
  interfaces: br-64v4 ironicendpoint ocpbm ocppr osp_trunk
```

I'll push an update ...

roles/nat64_appliance/files/elements/nat64-router/install.d/70-dnsmasq
hjensas: I added a TODO comment here.

roles/nat64_appliance/molecule/default/converge.yml
```diff
@@ -108,6 +129,23 @@
...
+    # TODO(hjensas): With the zone set in libvirt XML this bridge "should"
+    # already be in this zone. But logs indicate that it is not.
+    # See: https://libvirt.org/formatnetwork.html
+    - name: Make sure br-mol bridge is in the libvirt firewalld zone
+      become: true
+      ansible.posix.firewalld:
+        zone: libvirt
+        interface: br-mol
+        state: enabled
+        permanent: true
```
cescgina, line 141: would setting `immediate: true` here (https://docs.ansible.com/ansible/latest/collections/ansible/posix/firewalld_module.html#parameter-immediate) save as the restart of the service? No idea if it would save much time, or if it's actually a good idea, just asking

cescgina (approved):

PR #1992, opened 2024-06-30
dnsmasq: Use network/prefix in --domain option

hjensas:
We add the `,local` option to --domain configuration statements. According to dnsmasq manual page this only works if using network/prefix for the range.

```
--domain=<domain>[[,<address range>[,local]]|<interface>]

If the address range is given as ip-address/network-size, then a additional
flag "local" may be supplied which has the effect of adding --local declarations
for forward and reverse DNS  queries.
Eg.   --domain=thekelleys.org.uk,192.168.0.0/24,local is identical to
      --domain=thekelleys.org.uk,192.168.0.0/24 --local=/thekelleys.org.uk/ --local=/0.168.192.in-addr.arpa/
```

### DIFF
--- a/roles/dnsmasq/templates/network.conf.j2
+++ b/roles/dnsmasq/templates/network.conf.j2
@@ -9,7 +9,7 @@ enable-ra
 dhcp-range=set:{{ range.label }},{{ range.start_v4 }},static,{{ (range.start_v4 + "/" + range.prefix_length_v4 | default(24) | string) | ansible.utils.ipaddr('netmask') }},{{ range.ttl | default('1h') }}
 {%      if range.domain is defined and range.domain | length > 0                                   -%}
 {%          set range_v4_allowed = (range.start_v4 ~ "/" ~ range.prefix_length_v4 | default('24')) |
-                                    ansible.utils.ipaddr('range_usable') | replace("-",",")         %}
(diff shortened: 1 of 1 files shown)


hjensas:
This is failing because dnsmasq in RHEL is old...
https://thekelleys.org.uk/dnsmasq/CHANGELOG
```
version 2.87
        Allow arbitrary prefix lengths in --rev-server and
	--domain=....,local
```

cjeanner:
Wondering if this wouldn't have some unwanted side effects, such as linking that domain/range to the DHCP and re-injecting the domain in the client "search domain" ? I'd rather get something allowing to inject "local" in a dedicated file rather than using those shortcuts... Thoughts?

hjensas:
> Wondering if this wouldn't have some unwanted side effects, such as linking that domain/range to the DHCP and re-injecting the domain in the client "search domain" ? I'd rather get something allowing to inject "local" in a dedicated file rather than using those shortcuts... Thoughts?

Yes, I think we have to do that considering dnsmasq version in RHEL does not support prefix-length other than /24. /16, /8.

I also agree on the unwanted side effects. It is probably better to just have a rich configuration which is very explicit, rather than using these "helpful" magical things in dnsmasq ...

cjeanner:
@hjensas so I guess we can close that PR?

PR #1993, opened 2024-07-01
Add variables to check if the pod is running or not

arxcruz:
Usually, the default values to check if the pod is running or not is enough for normal jobs, but we notice an increase in time waiting when we are running unijobs, maybe because the amount of services running. Adding variables to set this values, will ensure we can defined a bigger timeout while checking for this without disrupt other jobs.

  - [x] README in the role

### DIFF
--- a/roles/test_operator/README.md
+++ b/roles/test_operator/README.md
@@ -20,6 +20,9 @@ Execute tests via the [test-operator](https://openstack-k8s-operators.github.io/
 * `cifmw_test_operator_node_selector`: (Dict) `nodeSelector` value that is applied to all pods spawned by the test-operator and to the test-operator-controller-manager and test-operator-logs pods. Default value: `{}`
 * `cifmw_test_operator_storage_class_prefix`: (String) Prefix for `storageClass` in generated Tempest CRD file. Defaults to `"lvms-"` only if `cifmw_use_lvms` is True, otherwise it defaults to `""`. The prefix is prepended to the `cifmw_test_operator_storage_class`. It is not recommended to override this value, instead set `cifmw_use_lvms` True or False.
 * `cifmw_test_operator_storage_class`: (String) Value for `storageClass` in generated Tempest or Tobiko CRD file. Defaults to `"lvms-local-storage"` only if `cifmw_use_lvms` is True, otherwise it defaults to `"local-storage"`.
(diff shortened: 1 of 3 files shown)


roles/test_operator/tasks/run-test-operator-job.yml
```diff
@@ -110,8 +110,8 @@
         wait: true
       register: logs_pod
       until: logs_pod.resources[0].status.phase == "Running"
-      delay: 10
-      retries: 20
+      delay: "{{ cifmw_test_operator_check_pod_delay }}"
```
viroel, line 113: Isn't a good idea to set role's default values for both?

roles/test_operator/README.md
```diff
@@ -20,6 +20,8 @@ Execute tests via the [test-operator](https://openstack-k8s-operators.github.io/
 * `cifmw_test_operator_node_selector`: (Dict) `nodeSelector` value that is applied to all pods spawned by the test-operator and to the test-operator-controller-manager and test-operator-logs pods. Default value: `{}`
 * `cifmw_test_operator_storage_class_prefix`: (String) Prefix for `storageClass` in generated Tempest CRD file. Defaults to `"lvms-"` only if `cifmw_use_lvms` is True, otherwise it defaults to `""`. The prefix is prepended to the `cifmw_test_operator_storage_class`. It is not recommended to override this value, instead set `cifmw_use_lvms` True or False.
 * `cifmw_test_operator_storage_class`: (String) Value for `storageClass` in generated Tempest CRD file. Defaults to `"lvms-local-storage"` only if `cifmw_use_lvms` is True, otherwise it defaults to `"local-storage"`.
+* `cifmw_test_operator_check_pod_retries`: (Integer) How many times to check if the test-operator-logs-pod is running
```
cjeanner, line 23: Care to specify the default values for both params in here as well?

roles/test_operator/tasks/run-test-operator-job.yml
```diff
@@ -110,8 +110,8 @@
         wait: true
```
hjensas, line 110: Can we not instead use a `wait_condition` in the k8s_info task? Like we do in other places, for example: https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/kustomize_deploy/tasks/install_operators.yml#L118-L129

pablintino: +1 To this thing, that's the proper way to do so and how we usually wait for a pod.

lewisdenny:
@arxcruz @sdatko Is this one still needed guys? If so please sort out the conflicts and we can finish the review.

lewisdenny:
Closing as stale, please reopen if needed.

PR #1995, opened 2024-07-01
[ci_nmstate] Allow users to directly use the unprovisioned tasks

pablintino:
At the same time, fix a format issue in the Net Env Def that assumes the parent_interface is always present. That's no longer true since we added the new mapper that works in best effort mode.

  - [x] README in the role


hjensas (approved):
Looks good.

Should we update the README with example on how to use the two tasks files in "pass-thru" mode?

hjensas (approved):

PR #1997, opened 2024-07-01
Inject command aliases for faster debugging

cjeanner:
We will probably grow a bit that list while we see repetitive commands.

With the first `oc-job-log` we can directly follow logs from specified
job.


hjensas (commented):
/lgtm

raukadah (approved):
+1 , it will make debugging easier.

PR #1998, opened 2024-07-01
ci_gen_kustomize_values - IPv6

hjensas:
Adds a new role variable: `cifmw_ci_gen_kustomize_values_primary_ip_version` Defaults to `4` 
Set variable `cifmw_ci_gen_kustomize_values_ip_version_var_mapping` with conditionally set `vX` values based on IP version.

In the jinja2 templates, set up variable `_ipv` with the mapping from `cifmw_ci_gen_kustomize_values_ip_version_var_mapping`.

By replacing the version specific variables. i.e `network_v4` with `_ipv.network_vX` etc in the template so that it is easy to switch the primary ip-version when looking up data from facts and loaded data structures.

- [x] uni01alpha
- [x] uni06zeta
- [x] validate-architecture (via https://github.com/openstack-k8s-operators/architecture/pull/319)
- [x] nfv (ci HW)

### DIFF
--- a/roles/ci_gen_kustomize_values/templates/common/edpm-nodeset-values/values.yaml.j2
+++ b/roles/ci_gen_kustomize_values/templates/common/edpm-nodeset-values/values.yaml.j2
@@ -1,5 +1,6 @@
 ---
 # source: common/edpm-nodeset-values/values.yaml.j2
+{% set _ipv = cifmw_ci_gen_kustomize_values_ip_version_var_mapping %}
 {% set instances_names = []                                                            %}
 {% set _original_nodeset = (original_content.data | default({})).nodeset | default({}) %}
 {% set _original_nodes = _original_nodeset.nodes | default({})                         %}
@@ -26,7 +27,7 @@ data:
 {% for instance in instances_names                                                     %}
           edpm-{{ instance }}:
 {%   if hostvars[instance] is defined                                                  %}
-            nic1: "{{ hostvars[instance].ansible_default_ipv4.macaddress }}"
+            nic1: "{{ hostvars[instance][_ipv.ansible_default_ipvX].macaddress }}"
 {%   endif                                                                             %}
             nic2: "{{ cifmw_networking_env_definition.instances[instance].networks.ctlplane.mac_addr }}"
 {% endfor                                                                              %}
@@ -40,7 +41,7 @@ data:
 {% for instance in instances_names                                                     %}
       edpm-{{ instance }}:
         ansible:
-          host: {{ cifmw_networking_env_definition.instances[instance].networks.ctlplane.ip_v4 }}
+          host: {{ cifmw_networking_env_definition.instances[instance].networks.ctlplane[_ipv.ip_vX] }}
         hostName: {{ instance }}
         networks:
 {%      for net in cifmw_networking_env_definition.instances[instance].networks.keys() %}
@@ -48,7 +49,7 @@ data:
             subnetName: subnet1
 {%        if net is match('ctlplane')                                                  %}
             defaultRoute: true
-            fixedIP: {{ cifmw_networking_env_definition.instances[instance].networks.ctlplane.ip_v4 }}
+            fixedIP: {{ cifmw_networking_env_definition.instances[instance].networks.ctlplane[_ipv.ip_vX] }}
 {%        endif                                                                        %}
 {%      endfor                                                                         %}
 {% endfor                                                                              %}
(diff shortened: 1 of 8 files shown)


roles/ci_gen_kustomize_values/templates/common/edpm-nodeset-values/values.yaml.j2
```diff
@@ -1,5 +1,22 @@
...
+                'ipvX_ranges': 'ipv4_ranges'} -%}
+{%- else -%}
+{%- set _ipv = {'ipvX': 'ipv6',
+                'ip_vX': 'ip_v6',
+                'ansible_default_ipvX': 'ansible_default_ipv6',
+                'network_vX': 'network_v6',
+                'vX': 'v6',
+                'gw_vX': 'gw_v6',
+                'ipvX_ranges': 'ipv6_ranges'} -%}
+{%- endif -%}
```
cjeanner, line 19: That sounds exactly like a thing we want to nudge in a "snippet", using the `include` jinja2 instruction, like there for instance:
https://github.com/openstack-k8s-operators/edpm-ansible/blob/f6f28a6d4f8b770624d6ea33f4f67bb8c4bb2fb3/roles/edpm_nftables/templates/ruleset.j2#L13

We can get something like `template/snippets/ip-version.j2` and include it?

hjensas: indeed, I had forgotten about include in jinja2. I'll iron out some other issues and change to use an inlcude.

Thanks for the suggestion!

roles/ci_gen_kustomize_values/templates/include/ip-version.j2
```diff
@@ -0,0 +1,17 @@
+{%- if cifmw_ci_gen_kustomize_values_primary_ip_version == 4 -%}
```
cjeanner, line 1: you might need to get a header comment such as:
`# source: include/ip-verison.j2`
or something like that in order to get a specific linter to pass:
https://github.com/openstack-k8s-operators/ci-framework/blob/main/scripts/check_k8s_snippets_comment.sh

roles/ci_gen_kustomize_values/templates/include/ip-version.j2
```diff
@@ -0,0 +1,17 @@
...
+{%- set _ipv = {'ipvX': 'ipv4',
```
cjeanner, line 2: might need a `namespace` in fact? seeing the issues raised in molecule

roles/ci_gen_kustomize_values/templates/common/edpm-nodeset-values/values.yaml.j2
hjensas: ok, so include did not work. But jinja2 also support imports as well, so allows us o get the one file with the mappings.

roles/ci_gen_kustomize_values/templates/include/ip-version.j2
hjensas: ack, let's see if it is more happy now.

hjensas: Marking as resolved - something went sideways with the import / include approach.
I moved the map building to the vars instead.

roles/ci_gen_kustomize_values/templates/include/ip-version.j2
hjensas: Marking as resolved - something went sideways with the import / include approach.
I moved the map building to the vars instead.

roles/ci_gen_kustomize_values/templates/common/edpm-nodeset-values/values.yaml.j2
hjensas: Marking as resolved - something went sideways with the import / include approach.
I moved the map building to the vars instead.

roles/ci_gen_kustomize_values/vars/main.yml
```diff
@@ -22,3 +22,40 @@
 # All variables within this role should have a prefix of "cifmw_ci_gen_kustomize_values"
 
 cifmw_ci_gen_kustomize_values_original_cm_content_file_name: "01_original.yaml"
+
+cifmw_ci_gen_kustomize_values_ip_version_var_mapping:
```
cjeanner, line 26: That's one way indeed. We might extend it if needed.

cjeanner:
@hjensas did you run any of the ipv4-based uni jobs against it? Maybe running NFV would also be good (I'll let you sync with @elfiesmelfie or @rdiazcam or @pablintino for NFV testing). This change is affecting all of the jobs, we'd rather be 200% sure it's not breaking. It shouldn't, but...

cjeanner:
/approve

The general idea is fine - just waiting for the two last tests to end, and we will be able to tick the last checkboxes and send this PR in.

cjeanner:
I could get up to tempest (tempest failed in both case, for unrelated reasons). So we're good I think.

cjeanner:
/hold

We're about to get a run against NFV by @pablintino

PR #2007, opened 2024-07-02
Fix condition in tcib playbook when change list is empty

cescgina:
If there is no item in zuul items list with the field change_url'
defined, e.g. a periodic jobs, skip the block, as 'zuul_change_list'
will be undefined.

### DIFF
--- a/ci/playbooks/tcib/tcib.yml
+++ b/ci/playbooks/tcib/tcib.yml
@@ -44,7 +44,9 @@
         - "'change_url' in item"
 
     - name: Build containers when needed
-      when: "'os-net-config' not in zuul_change_list"
+      when:
+        - zuul_change_list is defined
+        - "'os-net-config' not in zuul_change_list"
       block:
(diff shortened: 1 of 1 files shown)


ci/playbooks/tcib/tcib.yml
```diff
@@ -44,7 +44,7 @@
         - "'change_url' in item"
 
     - name: Build containers when needed
-      when: "'os-net-config' not in zuul_change_list"
+      when: "zuul_change_list is defined and 'os-net-config' not in zuul_change_list"
```
cjeanner, line 47:
```suggestion
      when:
        -  zuul_change_list is defined
        - "'os-net-config' not in zuul_change_list"
```
slightly more readable - and doing the same

cescgina: thanks, updated the PR

cescgina:
recheck

PR #2008, opened 2024-07-02
Wait until OpenStack operators ready in kustomize_deploy role

abays:
In the `kustomize_deploy` role, we wait until NMState and MetalLB operators are ready, but we do not currently do so for our OpenStack operators.  It is speculated that this is leading to a race condition where, if the OpenStack operators happen to be slow in reaching the ready state, later tasks can fail when trying to apply OpenStack CRs (because the webhooks in the OpenStack operators are not available yet).  Let's try waiting for our OpenStack operators' availability before moving forward.

  - [x] README in the role

### DIFF
--- a/roles/kustomize_deploy/tasks/install_operators.yml
+++ b/roles/kustomize_deploy/tasks/install_operators.yml
@@ -260,3 +260,27 @@
           type: Available
           status: "True"
         wait_timeout: 300
+
+- name: Wait until OpenStack operators are deployed and ready
+  when: not cifmw_kustomize_deploy_generate_crs_only
+  kubernetes.core.k8s_info:
+    kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
+    kind: Deployment
+    namespace: openstack-operators
+    label_selectors:
+      - "{{ item }}"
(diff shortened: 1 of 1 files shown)


cjeanner:
recheck

fultonj:
+2

cjeanner:
recheck

roles/kustomize_deploy/tasks/install_operators.yml
```diff
@@ -260,3 +260,27 @@
           type: Available
           status: "True"
         wait_timeout: 300
+
+- name: Wait until OpenStack operators are deployed and ready
+  when: not cifmw_kustomize_deploy_generate_crs_only
+  kubernetes.core.k8s_info:
+    kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
```
fultonj, line 267: In the ci-framework the calls to `kubernetes.core.k8s)_*` that I have seen also include these lines:

```yaml
    api_key: "{{ cifmw_openshift_token | default(omit) }}"
    context: "{{ cifmw_openshift_context | default(omit) }}"
```

PR #2011, opened 2024-07-02
validate-architecture - sshd_ranges - ipv6

hjensas:
When setting the cifmw_ci_gen_kustomize_values_sshd_ranges variable, try to load both network_v4 and network_v6 values. Default to `none` and pass the list trough `select()` filter to remove any `none` vlaues in the final result.


cjeanner:
/approve

makes sense - we might have to go down that rabbit hole as well, tracking mentions of ipv4 only...

cescgina (approved):

PR #2013, opened 2024-07-02
Disable GSSAPI authentication in vbmc container

cjeanner:
It's mostly useless and might lead to timeout in vbmcd trying to connect
via SSH, and SSH taking too long to get an acceptable result.

Since we know we're using keypair authentication, let's just disable
GSSAPI.

This also fixes broken molecule run due to the major changes in
libvirt_manager.


viroel:
/approve

tested in downstream with depends-on and it fix the issue

PR #2014, opened 2024-07-02
Changes to virtualbmc shouldn't trigger a deployment

cjeanner:
Nothing uses VBMC in upstream CI, we don't need to trigger deployments
upon changes to this role.


viroel:
/approve

thanks, lgtm

PR #2015, opened 2024-07-03
Chasing down last import vs include in reproducer role

cjeanner:
When we have conditional import, it's better to switch to include, since
the condition will be applied to the "do I include this content" rather
than "I'll apply the condition to all of the imported element and
evaluate it again and again".

Note that some stay as "import" to keep loading the role parameters
nevertheless. We've seen issues switching some of them to "include", see
comments in the affected parts.


raukadah (approved):
/approve

PR #2017, opened 2024-07-03
[test-operator] Introduce AnsibleTest CR

lpiwowar:
In this PR [1] we introduce AnsibleTest CR. This patch extends the test-operator so that it can be used to execute tests using the newly introduced CR.

[1] https://github.com/openstack-k8s-operators/test-operator/pull/97

  - [x] README in the role


lpiwowar:
The change was tested the change very here -> https://github.com/openstack-k8s-operators/test-operator/pull/120. It was expected that the job will fail. I just wanted to check whether the ansibletest CR gets created correctly with all the values which seems it is [here|https://logserver.rdoproject.org/20/120/6917ec2f6d63d3224b8bdaea179b6f61e8f8e55e/github-check/podified-multinode-edpm-deployment-crc-test-operator/06abe64/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/crs/ansibletests.test.openstack.org/] (the horizontest name was fixed and the CR is now correctly named ansibletest). 

We need this to be merged to unblock @eshulman2 with here testing. The PR does not touch the already existing logic so it should have no impact on Tempest / Tobiko.

raukadah (approved):
/approve

PR #2018, opened 2024-07-03
Add tripleo_controllers tripleo_computes for DPA ci zuul nodeset

marios:
As part of [1] this add the tripleo_controllers tripleo_computes
groups to the zuul nodeset, to be used by the data plane adoption
ci playbooks in Related below. It also removes a redundant nodeset
no longer used by any job.

[1] https://issues.redhat.com/browse/OSPRH-8291

Related: https://review.rdoproject.org/r/c/rdo-jobs/+/53688

  - [x] README in the role

### DIFF
--- a/zuul.d/nodeset.yaml
+++ b/zuul.d/nodeset.yaml
@@ -63,37 +63,6 @@
           - standalone
 
 - nodeset:
-    name: centos-9-multinode-rhel-9-2-crc-extracted-2.30-3xl
-    nodes:
-      - name: controller
-        label: cloud-centos-9-stream-tripleo
-      - name: crc
-        label: coreos-crc-extracted-2-30-0-3xl
-      - name: undercloud
-        label: cloud-rhel-9-2-tripleo
-      - name: overcloud-controller-0
-        label: cloud-rhel-9-2-tripleo
-      - name: overcloud-controller-1
-        label: cloud-rhel-9-2-tripleo
-      - name: overcloud-controller-2
-        label: cloud-rhel-9-2-tripleo
-      - name: overcloud-novacompute-0
-        label: cloud-rhel-9-2-tripleo
-    groups:
-      - name: computes
-        nodes: []
-      - name: ocps
-        nodes:
-          - crc
-      - name: rh-subscription
-        nodes:
-          - undercloud
-          - overcloud-controller-0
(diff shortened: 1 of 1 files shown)


cescgina (changes requested):
looks good, but we should add the groups for https://github.com/openstack-k8s-operators/ci-framework/pull/2018/files#diff-f3199dc676bf2b8d1ae52a4513c184798c9e9e9981404d6fe6d189097d35604cR66 as well, otherwise when changing the rdo-jobs playbook we'll break the WIP using that nodeset

marios:
> looks good, but we should add the groups for https://github.com/openstack-k8s-operators/ci-framework/pull/2018/files#diff-f3199dc676bf2b8d1ae52a4513c184798c9e9e9981404d6fe6d189097d35604cR66 as well, otherwise when changing the rdo-jobs playbook we'll break the WIP using that nodeset

ack good call can do thanks for reviewing

zuul.d/nodeset.yaml
```diff
@@ -279,6 +248,16 @@
...
+      - name: tripleo_controllers
+        nodes:
+          - overcloud-controller-0
+          - overcloud-controller-1
+          - overcloud-controller-2
+      - name: tripleo_computes
+        nodes:
+          - overcloud-novacompute-0
+          - overcloud-novacompute-1
+          - overcloud-novacompute-2
```
marios, line 260: ack @cescgina on adding these groups for the nova cells but that work is still in progress. I have to check how Bogdan is dealing with those 'fetch compute node hostname' tasks in his patches

so for now i am not going to update

cescgina: oh ok, if there is duplication there it should be fine, I'm approving this change then

cescgina (approved):
/approve

zuul.d/nodeset.yaml
marios: This should not change anything from the cells work, in the sense that you can continue to use delegate_to and target specific nodes (as Bogdan does in https://review.rdoproject.org/r/c/rdo-jobs/+/53192/39/playbooks/data_plane_adoption/setup_tripleo_hostnamemap.yaml for example)

marios:
recheck

raukadah:
recheck

PR #2021, opened 2024-07-03
Enable some debugging if no DNS records

cjeanner:
It seems some scenarios might lead to no DNS records. This isn't normal,
and we need to know why we're hitting that issue.

This patch should help tracking down the issue.


cescgina (approved):

cjeanner:
/override toggle-label

it seems to check in the wrong way..

PR #2022, opened 2024-07-03
Use translated network name

cjeanner:
Instead of hard-coding a name translation, let's use the "translator"
provided by the networking_mapper directly.


hjensas (commented):
Nice!

/lgtm

PR #2029, opened 2024-07-04
Add retries on task: Configure VMs type {{ vm }}

hjensas:
In my IPv6 environment I often see connection related errors in this task. Error is "Connection timed out during banner exchange"

I think this may have something to do with MaxStartups (ref: sshd_config(5)). We just added the host to the ansible inventory -> added ssh keys -> connect again as _init_admin_user etc - so there is potentially a lot of connections going on.

Adding some retires worksaround the issue.

### DIFF
--- a/roles/libvirt_manager/tasks/manage_vms.yml
+++ b/roles/libvirt_manager/tasks/manage_vms.yml
@@ -47,7 +47,7 @@
         default(_init_admin_user)
       }}
 
-- name: "Configure ssh access on type {{ vm }}"
+- name: "Configure ssh access on VM {{ vm }}"
   vars:
     _user: >-
       {{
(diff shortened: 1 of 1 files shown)


roles/libvirt_manager/tasks/manage_vms.yml
```diff
@@ -81,6 +81,10 @@
       sudo -u zuul mkdir -p /home/zuul/.ssh /home/zuul/src/github.com/openstack-k8s-operators;
       sudo cp ${HOME}/.ssh/authorized_keys /home/zuul/.ssh/;
       chown -R zuul: /home/zuul/.ssh;
+  retries: 5
```
cjeanner, line 84: fun. OK - care to also update the task "name" from `Configure VMs type {{ vm }}` to something like `Configure VM {{ vm }}` ? Thanks :)

We're being too fast now, aren't we? ;)

hjensas: ack, I'll push and update an get the task name aligned.

Yeah, I think this might be related to the linear. :)
I guess these connections go via the JumpHost? So probably we hit the MaxStartups there ...

cjeanner: hmm, `delegate_to` so it should come from the hypervisor, without jumper, as I understand that parameter.

cjeanner (commented):
/approve

PR #2030, opened 2024-07-04
Support DCN layout

cjeanner:
DCN has multiple controplane networks, making it different from any
other layout we had until now.

A compute is usually connected to "a public network", and to a specific,
"local" controlplane, while another compute would be connected to a
different controlplane.

In addition to this, with Linux bridge name length limit, we must also
check in the "translation map" in case a network in libvirt isn't named
like the "interesting" ones used in the networking_mapper output.

This patch makes use of that translation when we generate the DHCP
entries - that was apparently the missing part, since we generate the
DNS entries already based on the networking_mapper output, with a
"match" for the `ctlplane` pattern, allowing to capture any network
starting with that specific pattern.

- [x] BGP
- [x] NFV
- [x] DCN

### DIFF
--- a/roles/libvirt_manager/molecule/deploy_layout/prepare.yml
+++ b/roles/libvirt_manager/molecule/deploy_layout/prepare.yml
@@ -62,7 +62,9 @@
         dest: "/etc/NetworkManager/dnsmasq.d/molecule-delegate.conf"
         mode: "0644"
         content: |
+          log-queries
           server=/utility/127.0.0.2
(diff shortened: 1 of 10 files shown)


eduolivares:
Verified on a d/s BGP job.

cjeanner:
/hold

Wait for Monday to merge.

cjeanner:
recheck

rebtoor (approved):
/lgtm

PR #2031, opened 2024-07-04
nat64: Make VM uuid in mac-address idempotent

hjensas:
Use a seed for the random generator to ensure the uuid and mac-addresses are idempotent on re-runs.

### DIFF
--- a/roles/nat64_appliance/tasks/deploy.yml
+++ b/roles/nat64_appliance/tasks/deploy.yml
@@ -87,7 +87,14 @@
     name: firewalld
     state: restarted
 
+- name: List VMs
+  register: _list_running_vms
+  community.libvirt.virt:
+    command: list_vms
+    state: running
+
 - name: "Create the config-drive ISO for the nat64-appliance"
+  when: 'not cifmw_nat64_appliance_name in _list_running_vms.list_vms'
   vars:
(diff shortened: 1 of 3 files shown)


roles/nat64_appliance/tasks/deploy.yml
```diff
@@ -1,8 +1,21 @@
 ---
-- name: Set MAC address facts
+- name: Set nat64-appliance VM uuid and mac-address facts
```
cjeanner, line 2: Wondering if we shouldn't integrate that VM in the libvirt_manager_configuration?
Advantages:
- we pre-generate all the MAC
- we store them in a file, allowing to avoid any issue with re-run

AFAIK, to get that, we're "just" missing a notion of "starting order" to ensure nat64 is built/started first, right? Might be something to consider, maybe post-GA.
Another solution might be to hard-code the MAC, like we do for the ocp-bootstrap node?

hjensas: the way it works now in the reproducer is that nat64 is not cleaned up unless we do deepscrub. Just keeping it around, i.e VM still running is the faster than having it in libvirt manager.

With this change we end up with the same MAC address and UUID generated on every run - on the same hypervisor (inventory_host). For the reproducer, we really don't need it - can be totaly random. I just cooked up this to improve the role in general.

PR #2035, opened 2024-07-05
Allow spine&leaf to not have ctlplane

cjeanner:
Until now, the `.utility` record was only created for nodes connected to
the ctlplane network (or any of its equivalent).
In Spine&Leaf though, some VMs aren't connected to any controlplane,
only to the "Public" network.

This patch:
- allows to fallback on "public network" to create the .utility record
- updated spine_leaf molecule scenario to check that new feature

### DIFF
--- a/roles/libvirt_manager/molecule/spine_leaf/converge.yml
+++ b/roles/libvirt_manager/molecule/spine_leaf/converge.yml
@@ -24,6 +24,8 @@
     cifmw_libvirt_manager_network_interface_types:
       intnet-0: network
       intnet-1: network
+      s0-rtr: network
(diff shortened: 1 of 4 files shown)


cjeanner:
recheck

cjeanner:
/hold

needs #2038

cjeanner:
/hold
we want to double-check with some NFV run, just to be on the safe side.

eduolivares:
/lgtm
PR verified on downstream BGP job.

cjeanner:
/hold cancel
We could get a pass with NFV, we're good to go!

PR #2038, opened 2024-07-05
Ensure libvirt_molecule starts on clean env, each time

cjeanner:
It appears some data were leaking between the runs, due to dangling
files mostly (wanted feature - at least for now).
This lead to some weird situations where the node list was different
from the expeceted one, even though it didn't lead to actualy failure.

This patch also ensures we gather needed bits from molecule run,
especially in the various directories.

### DIFF
--- a/roles/libvirt_manager/molecule/deploy_layout/cleanup.yml
+++ b/roles/libvirt_manager/molecule/deploy_layout/cleanup.yml
@@ -19,13 +19,45 @@
   gather_facts: true
   vars:
     cifmw_basedir: "/opt/basedir"
+    molecule_scenario: deploy_layout
   tasks:
+    - name: Ensure we have the destination directory
+      failed_when: false
+      ansible.builtin.file:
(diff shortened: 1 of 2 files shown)


cjeanner:
/override toggle-label

Reason: this PR is needed by 2 others (#2030 and #2035) - both are rebased onto this one to benefit of that feature. We have to rework the logic of that toggle-label I think.

cescgina (approved):

marios (commented):
will you keep this one or are you going to abandon and merge this squased into https://github.com/openstack-k8s-operators/ci-framework/pull/2035/commits

cjeanner:
> will you keep this one or are you going to abandon and merge this squased into https://github.com/openstack-k8s-operators/ci-framework/pull/2035/commits

I want to get that one in as-is, there are 2 (now 3) PR rebased onto this one, needing this change. I'd rather go step-by-step.

PR #2039, opened 2024-07-05
Change dhcp-host filename pattern

cjeanner:
With the initial name, we just had MAC addresses, and had to open the
file and find which network was associated to the selected IP.

This change allows to know directly which network (and host, if set) is
configured with that file, making debugging and searching faster.


hjensas (commented):
/lgtm

covered in CI - looks good:
https://logserver.rdoproject.org/39/2039/38957c62533174831dbdf32a8526d2fe16bdf865/github-check/cifmw-molecule-dnsmasq/90428c7/ci-framework-data/artifacts/dhcp-hosts.d/

PR #2042, opened 2024-07-08
Correct and add DNS records for *all* networks

cjeanner:
This patch ensures we have proper DNS records for all nodes, on all
networks.
It also ensures "ocp" and "ocp_worker" are properly changed to "master"
and "worker" for all zones ("utility" name was wrong).

This PR also prepare the field to remove some nested loops, leading to
the run trying to create over and over the exact same thing. Not really
an issue for small deploy, but once we get ocp_workers and many
computes, we'll notice a big difference.

We also ensure we inject hosts in the dynamic inventory only once,
allowing to get all the needed data in one place only. This improves
speed, since we don't edit the inventory twice "for nothing".
We have all the needed data since the start, let's just use them
directly.

This PR also adds a new scenario in molecule to ensure DNS are properly
generated for all known networks associated to a compute and an OCP
node.

This should prevent any regression, or changes in the used networks,
especially for the `.utility` zone.

- [x] uni zeta
- [x] NFV
- [x] va-hci
- [x] BGP
- [x] uni alpha
- [x] uni gamma


cjeanner:
/hold

Merging on Monday!

cjeanner:
/hold cancel

We're Monday. Let's get it in.

raukadah (commented):
/approve

PR #2043, opened 2024-07-08
Move final OCP wait closed to end of run

cjeanner:
We're now too fast, meaning the OCP might not have enough time to get
stable.
In parallel, we still have "slow" tasks that can kick in before we
actually need OCP, meaning we can improve overal speed by just moving a
block


hjensas (commented):
/lgtm

cescgina (commented):
/approve

PR #2044, opened 2024-07-08
[os_must_gather] Improve the way how o-m-g directory name is gathered

rebtoor:
We are going to use a regular expression based on the `cifmw_os_must_gather_image` variable to obtain the correct pattern to retrieve the directory name instead of relying on the hardcoded `"quay-io-openstack-k8s-operators-openstack-must-gather"` pattern, so when we are using a different image for `openstack-must-gather` (taken from another registry) it won't fail anymore.

### DIFF
--- a/roles/os_must_gather/tasks/main.yml
+++ b/roles/os_must_gather/tasks/main.yml
@@ -69,10 +69,25 @@
           SOS_DECOMPRESS=$SOS_DECOMPRESS
           gather &> {{ cifmw_os_must_gather_output_dir }}/logs/os_must_gather.log
 
+    # directory name will be generated starting from cifmw_os_must_gather_image
+    # variable e.g.:
+    # EXAMPLE 1
+    # original value: "quay.io/openstack-k8s-operators/openstack-must-gather:latest"
+    # pattern value: "quay-io-openstack-k8s-operators-openstack-must-gather*"
+    # EXAMPLE 2
+    # original value: "foo.bar.example.com/repofoo/openstack-must-gather-rhel9:1.0.0"
+    # patterns value: "foo-bar-example-com-repofoo-openstack-must-gather-rhel9*"
+    # TODO: add molecule testing
     - name: Get exact must-gather output folder name
       ansible.builtin.find:
         paths: "{{ cifmw_os_must_gather_output_dir }}/logs"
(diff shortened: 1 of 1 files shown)


rebtoor:
recheck

roles/os_must_gather/tasks/main.yml
```diff
@@ -72,7 +72,13 @@
     - name: Get exact must-gather output folder name
       ansible.builtin.find:
         paths: "{{ cifmw_os_must_gather_output_dir }}/logs"
-        patterns: "quay-io-openstack-k8s-operators-openstack-must-gather*"
+        patterns: >-
```
cjeanner, line 75: You'll hate me, buuut...

can we at least get a TODO to get some molecule coverage there, like we have for the cifmw_job_uri parsing in the reproducer role?

Also, a couple of example input/output as a leading comment here would be really, really appreciated.

rebtoor: done!

roles/os_must_gather/tasks/main.yml
```diff
@@ -69,10 +69,25 @@
...
           gather &> {{ cifmw_os_must_gather_output_dir }}/logs/os_must_gather.log
 
+    # directory name will be generated starting from cifmw_os_must_gather_image
+    # variable e.g.:
+    # EXAMPLE 1
+    # original value: "quay.io/openstack-k8s-operators/openstack-must-gather:latest"
+    # pattern value: "quay-io-openstack-k8s-operators-openstack-must-gather*"
+    # EXAMPLE 2
+    # original value: "foo.bar.example.com/repofoo/openstack-must-gather-rhel9:1.0.0"
+    # patterns value: "foo-bar-example-com-repofoo-openstack-must-gather-rhel9*"
```
cjeanner, line 79: I'm slightly worried by the potential "path too long" issue if for any reason the leading directory (that one, here) has a extensively long name. A follow-up truncating name might be something to consider?

rebtoor: @cjeanner we are not creating that folder, in fact we move away that long name to a "sane" name. We're only searching for it in that task. I guess that the folder is generated by must-gather itself.

PR #2045, opened 2024-07-08
Pull ansible-utils from github

cjeanner:
Seems galaxy.ansible.com has some issues serving ansible.utils
collection.

Swithing it to github in order to unlock the situation.


cjeanner:

Got green for both ansible-test and pre-commit - they go through the collection install, meaning: we merge.

PR #2009, opened 2024-07-02
Installing go straight from the source

jpodivin:
As a pull request owner and reviewers, I checked that:
- [x] Appropriate testing is done and actually running
- [x] Appropriate documentation exists and/or is up-to-date:
  - [x] README in the role
  - [x] Content of the docs/source is reflecting the changes

Tests in repos using [cifmw-doc](https://review.rdoproject.org/zuul/job/cifmw-doc) or derivatives:

- https://github.com/openstack-k8s-operators/data-plane-adoption/pull/516
- https://github.com/openstack-k8s-operators/openstack-operator/pull/902
- https://github.com/openstack-k8s-operators/openstack-ansibleee-operator/pull/387
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - 

All are passing and appear to produce appropriate results.

PR #2047, opened 2024-07-09
Install golang 1.20.x in docs job

rabi:


bshephar (approved):
Having an unconfined Go version seems to cause problems, so I think this makes sense to keep the Go version consistent across our project.

jpodivin:
We already have a PR on that, tested across repos https://github.com/openstack-k8s-operators/ci-framework/pull/2009

rabi:
Looks like there is already a PR.

PR #2048, opened 2024-07-09
Fix download tool step of the reproducer role.

sathlan:
Given the tags in that task, choose kubectl instead of operator-sdk as
a anchor to check if the task has been done.

-------


cjeanner:
/lgtm

thanks @sathlan - I forgot to update the `creates` back then -.-'

PR #2050, opened 2024-07-09
Add additional listen address to nested CRC

lewisdenny:
With CRC => 2.32.0-4.14.8 the dnsmasq service was changed to a systemd
service and with that change the listen-address was locked down.

We need to be able to query the crc hosted DNS on "192.168.122.10"
(default used by install_yamls)

JIRA: [OSPCIX-336](https://issues.redhat.com//browse/OSPCIX-336)

Test Project Patch: https://review.rdoproject.org/r/c/testproject/+/53754
- [x] [Passing nested baremetal tempest run](https://review.rdoproject.org/zuul/build/0dcab9ca06914d4e8e81f27face3d3d4)

### DIFF
--- a/roles/rhol_crc/tasks/main.yml
+++ b/roles/rhol_crc/tasks/main.yml
@@ -83,14 +83,14 @@
           ansible.builtin.import_tasks: configuration.yml
 
         - name: Setup RHOL/CRC
-          ansible.builtin.shell:  # noqa risky-shell-pipe
+          ansible.builtin.shell: # noqa risky-shell-pipe
             cmd: >-
               {{ cifmw_rhol_crc_binary }} setup 2>&1 |
               tee {{ cifmw_rhol_crc_basedir }}/logs/crc-setup.log
           register: cifmw_rhol_crc_cmd_setup
 
         - name: Start RHOL/CRC
-          ansible.builtin.shell:  # noqa risky-shell-pipe
+          ansible.builtin.shell: # noqa risky-shell-pipe
             cmd: >-
               {{ cifmw_rhol_crc_binary }} start 2>&1 |
               tee {{ cifmw_rhol_crc_basedir }}/logs/crc-start.log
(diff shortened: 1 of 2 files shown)


lewisdenny:
recheck

cescgina (commented):
/approve

roles/rhol_crc/tasks/main.yml
```diff
@@ -83,14 +83,14 @@
           ansible.builtin.import_tasks: configuration.yml
 
         - name: Setup RHOL/CRC
-          ansible.builtin.shell:  # noqa risky-shell-pipe
+          ansible.builtin.shell: # noqa risky-shell-pipe
```
cjeanner, line 86: note that we can remove that `noqa` by adding `set -o pipefail; ` as a first command in that shell.

cjeanner: (or we can use ci_script...)

roles/rhol_crc/tasks/main.yml
```diff
@@ -137,3 +137,37 @@
 - name: Add crc kubeconfig
   when: cifmw_rhol_crc_creds | bool
   ansible.builtin.import_tasks: add_crc_creds.yml
+
+# Note(Lewis): Only needed for CRC => 2.32.0-4.14.8
+- name: Configure DNS in CRC VM
+  delegate_to: crc
+  vars:
+    _crc_ip: "192.168.122.10"
```
pablintino, line 145: Can we avoid hard-coding this?
WDYT? https://github.com/openstack-k8s-operators/ci-framework/blob/main/ci/playbooks/multinode-customizations.yml#L29-L30

roles/rhol_crc/tasks/main.yml
lewisdenny: True, I did not add these lines, my formatter just fixed the double space(disabled now) 

Regardless I'll fix this in a follow up :)

roles/rhol_crc/tasks/main.yml
lewisdenny: `/etc/ci/env/networking-info.yml` is not copied to the nested CRC VM AFAIU. I assume it's on the hypervisor though and I should be able to slurp it from there.

I will test and submit a follow up patch and ping you. I normally wouldn't hardcode such a thing but it's already hardcoded everywhere. I wanted to get the CIX fixed asap but will do it correctly now.

PR #2052, opened 2024-07-09
Allow to not deploy OpenShift provider

cjeanner:
This PR mostly allows to deploy only the controller-0. It might also
allow deploying computes.

The goal is to allow a quick VM bootstrap by setting the `amount` to 0
for both `crc` and `ocp` VM types. Doing so, you can end with a single
controller-0 with all the toolings needed to run your tests (molecule,
for instance).

It introduces a new `_has_openshift` global parameter that's then used
in order to toggle some specific parts of the controller-0 bootstrap.

We can later revisit it in order to extend that support, so that we can
eventually consume an existing OpenShift provider (Hive, externally
managed OCP cluster, etc).

This PR also consolidate the usage of `_use_crc` and `_use_ocp` boolean
that are built based on the VM layout passed by the user.

### DIFF
--- a/roles/reproducer/tasks/configure_controller.yml
+++ b/roles/reproducer/tasks/configure_controller.yml
@@ -244,7 +244,7 @@
         dest: "/home/zuul/.kube/config"
         content: >-
           {{
-            (_cifmw_libvirt_manager_layout.vms.ocp is defined) |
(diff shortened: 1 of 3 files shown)


hjensas (commented):
/lgtm

PR #2054, opened 2024-07-09
Change chap algorithms

vakwetu:
The default CHAP algorithms include MD5 will does not work under FIPS.  Set this parameter to exclude MD5 and SHA-1

  - [x] README in the role

### DIFF
--- a/roles/devscripts/templates/iscsi.j2
+++ b/roles/devscripts/templates/iscsi.j2
@@ -20,7 +20,7 @@ spec:
           group:
             name: root
           contents:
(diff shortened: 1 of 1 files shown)


arxcruz:
I tested, I see tempest running, unfortunately, due another bug (in the wall) I don't have the logs, but from what I could see, there are failures, but I believe we can merge it and tackle what is failing later.

arxcruz (approved):
/lgtm
because github forces me to approve via review

d34dh0r53 (approved):
/lgtm

PR #2055, opened 2024-07-09
Set default value to DLRNAPI_PASSWORD

arxcruz:
When using kerberos for dlrn authentication, it is required to set this variable to empty string otherwise, it fails. Instead of set it in the job definition itself, it is better to set the default value directly in the role that is required to report dlrn.

  - [x] README in the role


raukadah (approved):
/approve

PR #2060, opened 2024-07-10
Change iscsid login timeout in computes

Akrog:
In a previous patch we changed the iSCSI login timeout for the control plane (OpenShift nodes).

In this patch we add a new hook to do the same thing for the edpm nodes, because the default timeout of 2 minutes is too high for some test scenarios.

It is necessary to do it this way because edpm-ansible doesn't currently have a mechanism to change the iscsid.conf file. Once that feature is added this patch can be reverted.

This patch changes the edpm iscsid timeout default to 3 retries and 5 seconds each (15 seconds in total), which is convenient for testing, as any healthy deployment and backend should be able to login to the backend in that amount of time, and if there is a broken path it will not take 2 minutes to give up, just around 15 seconds.

Related-Jira: https://issues.redhat.com/browse/OSPRH-7614

  - [x] README in the role

### DIFF
--- a/hooks/playbooks/compute-iscsi-config.yml
+++ b/hooks/playbooks/compute-iscsi-config.yml
@@ -1,0 +1,19 @@
+---
+- name: Change iscsi login timeout on compute nodes
+  hosts: computes
+  gather_facts: false
+  tasks:
+    - name: Change timeout in iscsid.conf
+      become: true
+      ansible.builtin.lineinfile:
+        path: /etc/iscsi/iscsid.conf
+        line: "{{ item }}"
+      loop:
+        - 'node.session.initial_login_retry_max = 3'
+        - 'node.conn[0].timeo.login_timeout = 5'
+
+    - name: Restart iscsid container to refresh /etcd/iscsid.conf
+      become: true
+      ansible.builtin.systemd:
+        name: edpm_iscsid
+        state: restarted


hooks/playbooks/compute-iscsi-config.yml
```diff
@@ -0,0 +1,19 @@
+---
+- name: Increase quotas for the service project
```
tosky, line 2: I think the task name should be changed (but let's finish the testing of the functionality first)

Akrog: LOL I failed at copy/pasting

hooks/playbooks/compute-iscsi-config.yml
```diff
@@ -0,0 +1,19 @@
...
+  hosts: computes
+  gather_facts: false
+  tasks:
+    - name: Change timeout in iscsid.conf
+      become: true
+      ansible.builtin.lineinfile:
+        path: /etc/iscsi/iscsid.conf
+        line: "{{ item }}"
+      loop:
+        - 'node.session.initial_login_retry_max = 3'
```
Akrog, line 12: I think we can change this to 1, because it either connects or it doesn't, and these are for tests, not production environments...

hooks/playbooks/compute-iscsi-config.yml
Akrog: Done.

tosky:
/lgtm

Internally tested

hooks/playbooks/compute-iscsi-config.yml
```diff
@@ -0,0 +1,19 @@
...
```
pablintino, line 12: Couldn't we add a couple of Ansible variables with these values as defaults?
Allowing adding extra lines passing a list to the playbook sounds good too :)

Akrog: I'm OK with that, I just didn't want to over engineer things.
I have already agreed with @tosky to make a follow up PR to make some improvements here, but we wanted to merge this which we have already tested first to unblock CI jobs.
What do you say I make those changes as a follow up?

pablintino: I'm fine with that approach if someone is blocked.

PR #2061, opened 2024-07-10
[test_operator][wntp] Add controller IP config to each tempest step

eduolivares:
The test_operator role already adds a value for the tempest.conf param `whitebox_neutron_plugin_options.proxy_host_address`, but it only does this at the generic `tempestconfRun` parameter.
That `tempestconfRun` can be overriden per tempest step, using the `workflow` list from the tempest CR.
This patch adds the value for
`whitebox_neutron_plugin_options.proxy_host_address` to each workflow step.


lpiwowar (approved):
Looks good to me! :+1: Thank you for the fix!

eduolivares:
recheck cifmw-crc-podified-edpm-baremetal

PR #2064, opened 2024-07-10
Call rhos-release command only with rhos_release_args

viroel:
The reproducer role is configuring repositories twice, and one of them has 'rhel' args hardcoded. This is a problem when we need to configure a different release version since it end up with duplicated repositories.


hjensas (commented):
/lgtm

PR #2066, opened 2024-07-11
Do not gather facts from networking_mapper on controller-0

cjeanner:
Since we provide the interfaces info, we don't really need to connect to
the hosts to gather the networking facts: we already have everything we
need.

This also means faster infra deployment, since the couple of tasks
associated to the remote facts takes over 2 minutes.

It also converges some paths to use an existing parameter instead of
re-writing over and over `/home/zuul/ci-framework-data`.

- [x] CRC based infrastructure
- [x] va-hci
- [ ] "some" uni job
- [ ] NFV
- [ ] BGP

### DIFF
--- a/roles/reproducer/tasks/configure_controller.yml
+++ b/roles/reproducer/tasks/configure_controller.yml
@@ -287,7 +287,7 @@
 
     - name: Manage secrets on controller-0
       vars:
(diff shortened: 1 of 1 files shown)


cjeanner:
We actually need some data such as MTU and NIC name.
Another approach will be done, leveraging `ip link` and its JSON output.

PR #2067, opened 2024-07-11
Hook: Set properties on default volume type

Akrog:
Cinder always creates a default volume type called __DEFAULT__, and this type is unrestricted, so new volumes can go to any volume back-end running on the deployment if an unrestricted volume type is not specified.

This can be problematic for some tempest tests when running on a multi back-end deployment.

This hook allows a job to define the properties (extra specs) of the default volume type. The most common way of using it is to set the `volume_backend_name` so volumes are created in a specific back-end, but other properties can be added as well.

  - [x] README in the role


tosky:
/lgtm

Tested internally, the patch applies the property as expected. You need to set something like:

cifmw_cinder_default_type_properties:
  volume_backend_name: foobar

PR #2068, opened 2024-07-11
Improve network assertions in reproducer

cjeanner:
We now ensure all VMs have attached networks, and we also output the
built layout in case of failure. This should help debugging issues
without the need to dig into the whole scenario tree.


cescgina (commented):
/approve looks really useful to catch this kind of issues early

cjeanner:
recheck

PR #2069, opened 2024-07-11
Setup minimum environment to run update on architecture.

sathlan:
When the deployment is not using `install_yaml` we need to set up some
variables to be able to trigger the update.

The update is using install_yamls in two places. First, it uses the
`set_openstack_containers` to modify the `openstack-operator` and that
role leverage the `install_yamls` commands.

Second, it uses it for the update itself as it's a target of
`install_yamls`

-------

  - [x] README in the role

### DIFF
--- a/roles/set_openstack_containers/tasks/main.yml
+++ b/roles/set_openstack_containers/tasks/main.yml
@@ -91,7 +91,29 @@
   ansible.builtin.include_role:
     name: 'install_yamls_makes'
     tasks_from: 'make_openstack_wait'
-  when: operator_name == 'openstack'
+  when:
+    - operator_name == 'openstack'
+    - not cifmw_architecture_scenario is defined
+
+# Needed when the deployment wasn't installed using install_yaml.
+- name: Wait for reinstallation of meta operator with arch scenario
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  ansible.builtin.shell:
+    cmd: |
+      set -o pipefail
+      oc get csv -l operators.coreos.com/openstack-operator.openstack-operators \
+      --namespace={{ cifmw_install_yamls_defaults['OPERATOR_NAMESPACE'] }} \
+      --no-headers=true | grep -i "succeeded"
+  register: operator_status
+  until: operator_status.rc == 0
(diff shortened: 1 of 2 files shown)


roles/set_openstack_containers/tasks/main.yml
```diff
@@ -91,7 +91,9 @@
   ansible.builtin.include_role:
     name: 'install_yamls_makes'
     tasks_from: 'make_openstack_wait'
-  when: operator_name == 'openstack'
+  when:
```
ciecierski, line 94: I think we can replace this wait with:
```
- name: Wait for reinstallation of meta operator
  environment:
    PATH: "{{ cifmw_path }}"
  ansible.builtin.shell:
    cmd: >-
      oc get csv -l operators.coreos.com/openstack-operator.openstack-operators
      --namespace={{ cifmw_install_yamls_defaults['OPERATOR_NAMESPACE'] }}
      --no-headers=true | grep -i "succeeded"
  register: operator_status
  until: operator_status.rc == 0
  changed_when: false
  retries: 30
  delay: 30 
 when: operator_name == 'openstack'
```
This is basically the same what openstack_wait does.

sathlan: Hum... the operator.coreos.com is retrieved from https://github.com/openstack-k8s-operators/install_yamls/blob/main/scripts/get-operator-status.sh#L12 :

```
oc get csv -n ${OPERATOR_NAMESPACE} -o jsonpath='{range .items[*]}{@.metadata.name}{"\n"}{end}' | grep -E "^${OPERATOR_NAME}-operator\.v"
```
so maybe we should use that also ... but it's unwrapping the script ... if the makefile target wasn't dependent on openstack target ...

Oki, I'll test with both of them, included.

ciecierski: openstack-operator is the name of operator retrieved by get-operator-status , it might or not change in future
You can implement two tasks, one for revival of exact name and second for waiting for reinstallation

ciecierski, line 94: remove trailing space

sathlan: Done.

marios (approved):
/lgtm

ciecierski (approved):

PR #1333, opened 2024-03-25
[Test-operator] Introduce Tempest workflow

lpiwowar:
Introduce the cifmw_test_operator_tempest_workflow variable which is equivalent to the cifmw_test_operator_tobiko_workflow variable.

This option enables the specification of a value for the workflow field in the Tempest CR. Using the workflow, the user can specify multiple executions of Tempest in a specific order.

Depends-On: https://github.com/openstack-k8s-operators/test-operator/pull/67

As a pull request owner and reviewers, I checked that:
- [x] Appropriate testing is done and actually running
- [x] Appropriate documentation exists and/or is up-to-date:
  - [x] README in the role
  - [x] Content of the docs/source is reflecting the changes

PR #2072, opened 2024-07-11
Allow using parallel option in tempest CR

eshulman2:
NFV test cases cannot run in parallel added a
parallel option to the CR to allow disabaling it
if required.
https://github.com/openstack-k8s-operators/test-operator/blob/44b0b132e34039dbcb9089ad01d23c873e70e7e4/api/bases/test.openstack.org_tempests.yaml#L98

Print CR before applying

  - [x] README in the role


tosky:
This is not needed, there is already cifmw_test_operator_concurrency  - please use that and the workflows (https://github.com/openstack-k8s-operators/ci-framework/pull/1333) if you need to run various set of tests with different concurrency

PR #2073, opened 2024-07-11
Ensure we don't copy useless venv/tmp

cjeanner:
In the "crc_layout" scenario, we deploy an actual layout with
controller-0 and everything done on it, and then we gather everything in
order to make things easier to debug in case of issue.
While it's useful, it also contains useless data, such a virtual env
(venv) and a tmp directory nested in the "path of interest".

This patch excludes those various data to ensure we're not cloaging the
log server.


hjensas (commented):
/lgtm

PR #2076, opened 2024-07-12
Remove async from rhos-release command

viroel:
Some jobs are failing to install packages and the async task may be the culprit. Reverting to the previous behavior here, but keeping the possiblity of providing args, instead of always install 'rhel' repos.


viroel:
recheck

viroel:
/hold 

need to delete the async flag check

PR #2081, opened 2024-07-14
Allow changing CR name and print CR before applying

eshulman2:
Added a var to allow changing tempest job name
Added a debug statment to print the CR that is going to be applied for testing

  - [x] README in the role


raukadah (approved):
/approve

PR #2082, opened 2024-07-15
Enable trash purge scheduler

katarimanojk:
'cifmw_cephadm_enable_trash_scheduler' is set to true by default, it can be disabled if required.

  - [x] README in the role


fultonj:
Can we enable this on the job level instead of enabling it by default?

ISTR this can be fragile and we opted to disable it by default during the tripleo days

https://github.com/openstack-archive/tripleo-ansible/commit/a449df417b36186ce4a97e8670226fedcf574a62

katarimanojk:
Agree with john on enabling it at job level, so closing it.

PR #2090, opened 2024-07-15
Exclude some files from /etc while getting the logs

cjeanner:
While there's little to no security impact, we were leaking host private
keys while fetching remote edpm nodes /etc without any filter.

An attacker could spoof a CI host, and might end grabbing credentials
from a user - though that specific attack is close to 0 impact nor real
usage.

This patch filters out some of the "non"-secrets in the sync - and will
also reduce the amount of logs (by a tiny number).

### DIFF
--- a/roles/artifacts/tasks/edpm.yml
+++ b/roles/artifacts/tasks/edpm.yml
@@ -66,12 +66,14 @@
           set -x
           sudo dnf install -y rsync
           mkdir -p /tmp/{{ host_ip }}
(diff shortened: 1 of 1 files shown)


cjeanner:
We don't have the mentioned/excluded directories in the final output:
https://logserver.rdoproject.org/90/2090/2cdb4a004fb648af2597de9ccf64d88ade73d3d1/github-check/podified-multinode-edpm-deployment-crc/35e7168/controller/ci-framework-data/logs/38.102.83.23/etc/

but we have everything else!

cescgina (approved):

PR #2091, opened 2024-07-15
Reparent adoption github-check job for new job naming scheme

marios:
This reparents the cifmw-data-plane-adoption-osp-17-to-extracted-crc
to the re-named parent job in rdo-jobs (done with the depends-on).

It keeps the existing jobs for now until we can complete the reparenting
across the other repos that reference these [1][2]

[1] https://github.com/openstack-k8s-operators/install_yamls/pull/872
[2] https://github.com/openstack-k8s-operators/edpm-ansible/pull/698

Depends-On: https://review.rdoproject.org/r/c/rdo-jobs/+/53788

  - [x] README in the role

### DIFF
--- a/zuul.d/adoption.yaml
+++ b/zuul.d/adoption.yaml
@@ -95,9 +95,17 @@
                 ip: 172.20.0.100
                 config_nm: false
 
+# TODO(marios): remove this once we finish rename OSPRH-8452
 - job:
     name: cifmw-data-plane-adoption-osp-17-to-extracted-crc
-    parent: data-plane-adoption-osp-17-to-extracted-crc
+    parent: adoption-standalone-to-crc-ceph-provider
+
+- job:
+    name: adoption-standalone-to-crc-ceph-provider
+    description: |
+        Standalone source OSP 17.1 node adopted to crc RHOSO 18.
+        Has ceph, not TLS. Uses a content provider.
+    parent: adoption-standalone-to-crc-ceph
     files:
       - ^playbooks/01-bootstrap.yml
       - ^playbooks/02-infra.yml
@@ -125,9 +133,17 @@
       - OWNERS
       - .*/*.md
 
+# TODO(marios): remove this once we finish rename OSPRH-8452
 - job:
     name: cifmw-data-plane-adoption-osp-17-to-extracted-crc-minimal-no-ceph
-    parent: cifmw-data-plane-adoption-osp-17-to-extracted-crc
+    parent: adoption-standalone-to-crc-no-ceph-provider
+
+- job:
+    name: adoption-standalone-to-crc-no-ceph-provider
+    description: |
+        Standalone source OSP 17.1 node adopted to crc RHOSO 18.
+        Has TLS, not ceph. Uses a content provider.
+    parent: adoption-standalone-to-crc-ceph-provider
     vars:
       enable_tls: "true"
       cloud_domain: "ooo.test"
--- a/zuul.d/project-templates.yaml
+++ b/zuul.d/project-templates.yaml
@@ -24,17 +24,6 @@
         - podified-multinode-hci-deployment-crc: *content_provider
 
 - project-template:
-    name: data-plane-adoption-pipeline
-    description: |
-      Project template to run content provider with data-plane adoption job.
-    github-check:
-      jobs:
-        - openstack-k8s-operators-content-provider
-        - cifmw-data-plane-adoption-osp-17-to-extracted-crc:
-            dependencies:
-              - openstack-k8s-operators-content-provider
-
-- project-template:
     name: podified-ironic-operator-pipeline
     description: |
       Project template to run content provider with ironic podified job.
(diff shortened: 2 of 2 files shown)


zuul.d/adoption.yaml
```diff
@@ -125,9 +130,14 @@
...
       - .*/*.md
 
+# TODO(marios): remove this once we finish rename OSPRH-8452
 - job:
     name: cifmw-data-plane-adoption-osp-17-to-extracted-crc-minimal-no-ceph
-    parent: cifmw-data-plane-adoption-osp-17-to-extracted-crc
+    parent: adoption-rhel-osp17-standalone-to-crc-no-ceph-provider
+
+- job:
+    name: adoption-rhel-osp17-standalone-to-crc-no-ceph-provider
```
marios, line 139: so i'm thinking we keep the names we agree on with in https://review.rdoproject.org/r/c/rdo-jobs/+/53788/2/zuul.d/_data_plane_adoption.yaml#4 and then just have a common suffix 

```
-provider
```

to signal that these jobs are in fact content consumers. really 'consumer' is more appropriate but I think some folks would be more familiar with the concept of 'content provider' which is why i used 'provider' here.

WDYT?

@jistr @cescgina @frenzyfriday

jistr: I'm wondering whether we should explicitly naming the 'provider' ones or the 'no-provider' ones, since i think 'provider' is the usual setup in other repos?

marios: so indeed, only data-plane-adoption repo runs this without a content provider.

```
adoption-standalone-to-crc-no-ceph-no-provider
adoption-standalone-to-crc-no-ceph
```

interesting idea and I don't hate it but if we do it then you'd have to "know" that the one without 'no-provider' in the name is the one that has a content provider. 

```
adoption-standalone-to-crc-no-ceph-provider
adoption-standalone-to-crc-no-ceph
```

I am also having second thoughts about not using 'consumer' here... maybe

```
adoption-standalone-to-crc-no-ceph-consumer
adoption-standalone-to-crc-no-ceph
```

I'll update https://review.rdoproject.org/r/c/rdo-jobs/+/53788 today after yesterday's discussion since that one has to go first anyway. then i'll update this and the others that depends-on it and we can finalise this discussion

zuul.d/adoption.yaml
```diff
@@ -95,9 +95,14 @@
                 ip: 172.20.0.100
                 config_nm: false
 
+# TODO(marios): remove this once we finish rename OSPRH-8452
 - job:
     name: cifmw-data-plane-adoption-osp-17-to-extracted-crc
-    parent: data-plane-adoption-osp-17-to-extracted-crc
+    parent: adoption-standalone-to-crc-ceph-provider
```
marios, line 101: TODO: add description (via Lewis/reviews call)

ciecierski (changes requested):
Can you add description of jobs, the same you did for downstream jobs?

zuul.d/adoption.yaml
```diff
@@ -125,9 +130,14 @@
       - OWNERS
       - .*/*.md
 
+# TODO(marios): remove this once we finish rename OSPRH-8452
```
lewisdenny, line 133: +1 I like this workaround :)

zuul.d/adoption.yaml
lewisdenny: Ack, I'm happy with this once otherwise

zuul.d/adoption.yaml
```diff
@@ -95,9 +95,15 @@
...
 - job:
     name: cifmw-data-plane-adoption-osp-17-to-extracted-crc
-    parent: data-plane-adoption-osp-17-to-extracted-crc
+    parent: adoption-standalone-to-crc-ceph-provider
+
+- job:
+    name: adoption-standalone-to-crc-ceph-provider
+    description: Standalone source OSP 17.1 node adopted to crc RHOSO 18. Has ceph, not TLS. Uses a content provider.
```
raukadah, line 105: We can use pipe here and make the description multi line like this. What do you say?
```suggestion
    description: |
         Standalone source OSP 17.1 node adopted to crc RHOSO 18. 
         Has ceph, not TLS. Uses a content provider.
```

zuul.d/adoption.yaml
```diff
@@ -125,9 +131,15 @@
...
 
+# TODO(marios): remove this once we finish rename OSPRH-8452
 - job:
     name: cifmw-data-plane-adoption-osp-17-to-extracted-crc-minimal-no-ceph
-    parent: cifmw-data-plane-adoption-osp-17-to-extracted-crc
+    parent: adoption-standalone-to-crc-no-ceph-provider
+
+- job:
+    name: adoption-standalone-to-crc-no-ceph-provider
+    description: Standalone source OSP 17.1 node adopted to crc RHOSO 18. Has TLS, not ceph. Uses a content provider.
```
raukadah, line 141: same comment regarding description.

zuul.d/adoption.yaml
marios: sure

zuul.d/adoption.yaml
marios: ack

just wanted it to stay one line to be honest and it was short enough i felt, but ok updating

ciecierski (approved):

PR #2094, opened 2024-07-15
Add nodesets for CRC 2.39 (OCP 4.16)

lewisdenny:


lewisdenny:
recheck

PR #2102, opened 2024-07-16
Update cifmw-molecule-rhol_crc zuul job to CRC 2.39 - OCP 4.16

lewisdenny:
Jira: https://issues.redhat.com/browse/OSPRH-8665
Jira: https://issues.redhat.com/browse/OSPCIX-376

  - [ ] README in the role

### DIFF
--- a/ci/config/molecule.yaml
+++ b/ci/config/molecule.yaml
@@ -15,7 +15,7 @@
     nodeset: centos-9-crc-2-39-0-xl
 - job:
     name: cifmw-molecule-rhol_crc
-    nodeset: centos-9-crc-2-36-0-xxl #TODO(Lewis): Bump to 2-39 once OSPCIX-376 is resolved
+    nodeset: centos-9-crc-2-39-0-xxl
     timeout: 5400
 - job:
     name: cifmw-molecule-operator_deploy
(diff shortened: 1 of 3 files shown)


pablintino:
recheck

pablintino:
recheck

pablintino:
recheck

cjeanner:
recheck

lewisdenny:
recheck

cjeanner:
recheck

roles/rhol_crc/tasks/cleanup.yml
cjeanner, line 1: I'd challenge the need there.

But there's indeed a need to edit the role: bumping the version in defaults/main.yml is mandatory, else the current way of doing things in the role will make it fetch the old CRC - and crash because no real pull-secret available.

cjeanner (changes requested):

roles/rhol_crc/tasks/cleanup.yml
cjeanner: To be precise:
- removing that directory is a bad idea, since it means we have to re-download the bundle (~4G, maybe more, sometimes less)
- there's a better way to clean things up (I'm working on it) while keeping the cache
- in any cases, such a destructive thing should get the `never, deepscrub` tags, so that we hit it only when running a `--tags deepscrub`

We can of course discuss that precise change here, but I'm reworking the role, and that file is being changed to implement a better way to clean (calling delete -f + cleanup to ensure everything's removed but cached bundle)

lewisdenny:
Closing in favor of https://github.com/openstack-k8s-operators/ci-framework/pull/2123

PR #2107, opened 2024-07-17
Add some more debugging to the network_assertion failures

cjeanner:
Having only the final built layout isn't enough whenever we want to
debug some weird situations.

This patch allows to see everything used to build that final layout:
- original configuration
- any applied "patches"
- even the CI reproducer generated layout

This should really help debugging any issues related to the networking
assertions.

### DIFF
--- a/roles/reproducer/tasks/network_assertions.yml
+++ b/roles/reproducer/tasks/network_assertions.yml
@@ -47,6 +47,39 @@
               defined networks. Those nets don't seem to be attached:
               {{ _defined | difference(_all_nets) }}
   rescue:
+    # We don't want it when cifmw_libvirt_manager_configuration_gen is
+    # defined: it could confuse people.
+    - name: Dump initial layout
+      when:
+        - cifmw_libvirt_manager_configuration_gen is undefined
+      ansible.builtin.debug:
+        var: cifmw_libvirt_manager_configuration
+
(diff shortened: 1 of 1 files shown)


roles/reproducer/tasks/network_assertions.yml
```diff
@@ -47,6 +47,33 @@
...
               {{ _defined | difference(_all_nets) }}
   rescue:
+    - name: Dump initial layout
+      ansible.builtin.debug:
+        var: cifmw_libvirt_manager_configuration
+
+    # This param exists only in case of CI reproducer run
+    # so it's expected to usually see it as "undefined" in
+    # the debugging output.
+    - name: Dump CI reproducer generated layout
```
rebtoor, line 57: how about adding a `when: cifmw_libvirt_manager_configuration_gen is defined` ?

cjeanner: it would just output "UNDEFINED" or something like that. No failure, so imho we can go as-is

rebtoor: but why print a value if it's undefined? :)

cjeanner: filtered both (default + gen) according to the inline comment.

rebtoor (approved):
/lgtm

PR #2114, opened 2024-07-18
Use dedicated edpm_kernel_hugepages var

bshephar:
This change switches from using the generic edpm_kernel_args variable to set hugepages, and instead uses the dedicated edpm_kernel_hugepages variable

  - [x] README in the role


bshephar:
recheck

rebtoor (approved):

PR #2117, opened 2024-07-18
Improve DHCP configuration in dnsmasq

cjeanner:
Until now, we were doing far too many things far too many times for
nothing: check if the networks exists, check dataset consistency, etc,
etc.

All of that takes time, far too much time for little added value.

This patch switches to "batch": we can pass a list with all the
requested DHCP entries, and the tasks will loop on that data set.

This allows to check networks only once per set, and leverage faster
treatment, making a bigger difference as we add more nodes to the
layout.

We also bump the libvirt_manager molecule timeout to 1h, the
initial/default 30 minutes seems to be a little bit too tight with the
amount of new checks added lately.

  - [x] README in the role

### DIFF
--- a/roles/dnsmasq/README.md
+++ b/roles/dnsmasq/README.md
@@ -159,24 +159,29 @@ supported in libvirt).
 
 ### New host parameters
 
-* `cifmw_dnsmasq_host_network`: (String) Existing network name.
-* `cifmw_dnsmasq_host_state`: (String) Host status. Must be either `present` or `absent`.
-* `cifmw_dnsmasq_host_mac`: (String) Host MAC address.
-* `cifmw_dnsmasq_host_ips`: (List) Host IP addressees.
-* `cifmw_dnsmasq_host_name`: (String) Host name. Optional.
+* `cifmw_dnsmasq_dhcp_entries`: (List[mapping]) List of DHCP entries.
+
+### DHCP entry mapping
+
+* `state`: (String) Entry status. Must be either `present` or `absent`. Defaults to `present`
+* `network`: (String) Entry network. Must already exist in dnsmasq. Mandatory.
+* `mac`: (String) Entry MAC address. Mandatory.
+* `ips`: (List[string]) List of IP addresses associated to the MAC (v4, v6). Mandatory.
+* `name`: (String) Host name. Optional.
 
 #### Examples
 
 ```yaml
     - name: Inject some node in starwars network
       vars:
-        cifmw_dnsmasq_host_network: starwars
-        cifmw_dnsmasq_host_state: present
-        cifmw_dnsmasq_host_mac: "0a:19:02:f8:4c:a7"
-        cifmw_dnsmasq_host_ips:
-          - "2345:0425:2CA1::0567:5673:cafe"
-          - "192.168.254.11"
-        cifmw_dnsmasq_host_name: r2d2
+        cifmw_dnsmasq_dhcp_entries:
+          - network: starwars
+            state: present
+            mac: "0a:19:02:f8:4c:a7"
+            ips:
+              - "2345:0425:2CA1::0567:5673:cafe"
+              - "192.168.254.11"
+            name: r2d2
       ansible.builtin.include_role:
         name: dnsmasq
         tasks_from: manage_host.yml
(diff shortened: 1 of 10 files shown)


cjeanner:
recheck

roles/dnsmasq/README.md
```diff
@@ -159,24 +159,29 @@ supported in libvirt).
 
 ### New host parameters
 
-* `cifmw_dnsmasq_host_network`: (String) Existing network name.
-* `cifmw_dnsmasq_host_state`: (String) Host status. Must be either `present` or `absent`.
-* `cifmw_dnsmasq_host_mac`: (String) Host MAC address.
-* `cifmw_dnsmasq_host_ips`: (List) Host IP addressees.
-* `cifmw_dnsmasq_host_name`: (String) Host name. Optional.
+* `cifmw_dnsmasq_dhcp_entries`: (List[mapping]) List of DHCP entries.
+
```
marios, line 163: could there be folks out there using the 'old' vars ... can/should we do something about them?

roles/dnsmasq/tasks/manage_host.yml
```diff
@@ -14,72 +14,100 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+- name: Ensure we have the right data and type
```
marios, line 17: do we also need to check if they are using the 'old' vars and let them know? or this one will catch it?

marios (commented):
seems reasonable just wondering about existing users

roles/dnsmasq/README.md
cjeanner: Little to no chance anyone is using this role outside of the reproducer/libvirt_manager.
If really, they'll get an error with the updated  `ansible.builtin.assert` and will have to update their things.

No way we maintain the old interface, it would make things overcomplicated for no actual value imho

roles/dnsmasq/tasks/manage_host.yml
cjeanner: if they use the old one, they won't provide the new one - so crash

roles/dnsmasq/tasks/manage_host.yml
```diff
@@ -14,72 +14,100 @@
...
+  ansible.builtin.assert:
+    that:
+      - cifmw_dnsmasq_dhcp_entries is defined
+      - (cifmw_dnsmasq_dhcp_entries | type_debug) == 'list'
```
rebtoor, line 21:
```suggestion
      - cifmw_dnsmasq_dhcp_entries is iterable
```

works better for this kind of tests See [here](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_tests.html#type-tests).

roles/dnsmasq/tasks/manage_host.yml
```diff
@@ -14,72 +14,100 @@
...
 - name: Assert we have needed host data
   ansible.builtin.assert:
     quiet: true
     that:
-      - cifmw_dnsmasq_host_network is defined
-      - cifmw_dnsmasq_host_state is defined
-      - cifmw_dnsmasq_host_state in ['present', 'absent']
-      - cifmw_dnsmasq_host_mac is defined
-      - cifmw_dnsmasq_host_ips is defined
+      - (entry | type_debug) == 'dict'
```
rebtoor, line 31:
```suggestion
      - entry is mapping
```

as above

rebtoor (changes requested):

roles/dnsmasq/tasks/manage_host.yml
cjeanner: would love to know why they advise against using their own filters... comparing string might be "ugly", but `is iterable` alone wouldn't fit. To check if it's really a list, we'll have to use
`cifmw_dhcp_entries is not string and cifmw_dnsmasq_dhcp_entries is not mapping and cifmw_dnsmasq_dhcp_entries is iterable`

which is... well...

roles/dnsmasq/tasks/manage_host.yml
cjeanner: so here, the replacement isn't that straightforward:

`entry is not string and entry is iterable and entry is mapping` according to their doc
which, here again, is pretty not nice -.-

rebtoor (approved):
/approve

/lgtm

PR #2122, opened 2024-07-19
Do not edit sensitive data set in machineconfig

danpawlik:
By editing the sensitive data for the OpenShift, later the machineconfig daemon will set the node as degraded, so later the operations that we would like to be done via machineconfig manifests, will be not triggered.
Some projects are using the machineconfig manifests to configure the host, not directly by using the Ansible.

### DIFF
--- a/ci/playbooks/collect-logs.yml
+++ b/ci/playbooks/collect-logs.yml
@@ -117,6 +117,8 @@
               chown -R "{{ ansible_user }}" libvirt
               chown "{{ ansible_user }}" *
               cp /etc/containers/registries.conf {{ ansible_user_dir }}/zuul-output/logs/
+              cp -r /etc/containers/registries.conf.d {{ ansible_user_dir }}/zuul-output/logs/
+              chown -R "{{ ansible_user }}" {{ ansible_user_dir }}/zuul-output/logs/registries.conf.d
 
         - name: Copy generated documentation if available
           when:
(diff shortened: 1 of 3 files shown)


danpawlik:
Based on https://docs.redhat.com/en/documentation/openshift_container_platform/4.6/html/post-installation_configuration/post-install-machine-configuration-tasks#machineconfig-modify-journald_post-install-machine-configuration-tasks

danpawlik:
recheck

danpawlik:
recheck

raukadah:
recheck

ci/playbooks/collect-logs.yml
```diff
@@ -118,6 +118,15 @@
               chown "{{ ansible_user }}" *
               cp /etc/containers/registries.conf {{ ansible_user_dir }}/zuul-output/logs/
 
+        - name: Copy registries.conf.d directory
+          become: true
+          ansible.builtin.copy:
+            src: /etc/containers/registries.conf.d
+            dest: "{{ ansible_user_dir }}/zuul-output/logs/registries.conf.d"
```
raukadah, line 125: Task defined at Line 110 is already copying content from `/etc/containers/registries.conf`. Instead of this task, Can we replace this task with `cp /etc/containers/registries.conf.d {{ ansible_user_dir }}/zuul-output/logs/registries.conf.d` at line 120?

It will keep the existing pattern.

raukadah (commented):
Rest looks good, left one minute comment.

PR #2132, opened 2024-07-22
Add the `when` case for import vs include doc

cjeanner:
Also re-generate the dictionnary that was obviously manually edited

- [X] Appropriate documentation is up to date.


rebtoor (approved):
/lgtm

PR #2133, opened 2024-07-22
[devscripts] Added when clause in the `async_status` task

rebtoor:
Since we weren't check for the dry_run execution of the devscripts
(via the `cifmw_devscripts_dry_run` boolean variable), the task "Check
async deploy status" failed in the `default` molecule scenario where
the above mentioned variable was set to `true`.

### DIFF
--- a/roles/devscripts/tasks/main.yml
+++ b/roles/devscripts/tasks/main.yml
@@ -49,6 +49,8 @@
   poll: 0
 
 - name: Check async deploy status
+  when:
+    - not cifmw_devscripts_dry_run | bool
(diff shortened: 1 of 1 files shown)


roles/devscripts/tasks/main.yml
```diff
@@ -49,6 +49,8 @@
   poll: 0
 
 - name: Check async deploy status
+  when:
+    - not cifmw_devscripts_dry_run | bool
```
cjeanner, line 53: we could also `when: _devscripts_deploy.ansible_job_id is defined` - but this one should be good as well.

PR #2134, opened 2024-07-22
Correct ansible-lint issues

cjeanner:
It seems ansible-lint didn't bother warning when the new
community.general.dig lookup was introduced in
415c1439185f23f5f90003706e2f904870b439e7 while it does now complain with
a more recent patch.

Let's fix it in a dedicated PR, making things easier to check.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -33,7 +33,7 @@ repos:
       - id: black
 
   - repo: https://github.com/ansible/ansible-lint
-    rev: v6.22.1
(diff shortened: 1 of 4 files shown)


rebtoor (approved):
/approve 
/lgtm

cjeanner:

Reason: molecule and pre-commit did pass; now Zuul seems to have some issues scheduling deployments, but those deployments won't test anything related to the changes, since they are in molecule, and pre-commit related configuration.

PR #2136, opened 2024-07-23
Add skip subscription manager patch to devscripts role

raukadah:
https://github.com/openshift-metal3/dev-scripts/pull/1674 enables CRB repos on RHEL using subscription-manager to devscript. It breaks downstream baremetal job.

In downstream, we add custom crb repos. The systems are not subscribed to subscription manager. It is blocking the pipelines.

https://github.com/openshift-metal3/dev-scripts/pull/1681 runs subscription-manager command when system is subscribed. Since it is not merged yet. Let's copy the patch in devscripts role to unblock it.


cjeanner:
/approve

makes sense. Enforcing subscription isn't a good :/. PR has also been submitted against dev-scripts repository via https://github.com/openshift-metal3/dev-scripts/pull/1681

PR #2137, opened 2024-07-23
Allow to toggle no_log for debugging purpose

cjeanner:

### DIFF
--- a/roles/cifmw_external_dns/tasks/cert.yml
+++ b/roles/cifmw_external_dns/tasks/cert.yml
@@ -43,7 +43,7 @@
     - name: Ask k8s for the new Certficate Secret
       delegate_to: localhost
       run_once: true
-      no_log: true
+      no_log: "{{ cifmw_nolog | default(true) | bool }}"
       kubernetes.core.k8s_info:
         kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
         api_key: "{{ cifmw_openshift_token | default(omit) }}"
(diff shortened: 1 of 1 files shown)


fultonj:
I am +2 to merge this

roles/cifmw_external_dns/tasks/cert.yml
```diff
@@ -43,7 +43,7 @@
     - name: Ask k8s for the new Certficate Secret
       delegate_to: localhost
       run_once: true
-      no_log: true
+      no_log: "{{ cifmw_nolog | default(true) | bool }}"
```
marios, line 46: i just wonder/hope that whoever sets cifmw_nolog would be aware of the implications. Not as part of this change but it would be better to have a dedicated 'nolog' parameter for this stuff. Like 'cifmw_secrets_nolog' just an example.

marios (approved):
inline concern but otherwise lgtm/can see how it is needed if you are debugging something here so setting approve

roles/cifmw_external_dns/tasks/cert.yml
cjeanner: There are two kinds of `no_log: true` reasons in the framework:
- mask endless output, such as nested playbook called via command
- mask secrets

The first one usually has its own dedicated file. Toggling the no_log for them is, usually, because it fails to find the script, or changedir, or something related that prevents even the logging in the standard location

The second one is, usually, the one we really want to see, because the no_logs also masks the reason of the failure (right permission, missing parameter, wrong path, whatever).

Also, keep in mind most (if not all) of those secrets are temporary to the CI run anyway. Good thing to mask them, but, really... not THAT important since temporary.

lewisdenny (approved):
/lgtm

PR #2140, opened 2024-07-23
Add a tag to the "fix marketplace" imported tasks

cjeanner:
Adding this tag will allow people to get a better re-run experience:
when running multiple time the role, it may happen the marketplace is
broken "for good".

This tag allows now to get a first run up to whatever stage you want,
the re-run with `--skip-tags fix_marketplace` to avoid "unfixing" it.

### DIFF
--- a/docs/source/usage/01_usage.md
+++ b/docs/source/usage/01_usage.md
@@ -167,6 +167,7 @@ or `--skip-tags`:
 - `bootstrap_libvirt`: Run the [reproducer](../reproducers/01-considerations.md) libvirt bootstrap only.
 - `bootstrap_repositories`: Run the [reproducer](../reproducers/01-considerations.md) repositories bootstrap steps only.
 - `devscripts_layout`: Run the [reproducer](../reproducers/01-considerations.md) devscripts bootstrap only.
+- `fix_marketplace`: When skipped, allows to not re-run the `openshift_setup/fix_openshift_marketplace.yml` tasks.
 - `infra`: Denotes tasks to prepare host virtualization and Openshift Container Platform when deploy-edpm.yml playbook is run.
 - `build-packages`: Denotes tasks to call the role [pkg_build](../roles/pkg_build.md) when deploy-edpm.yml playbook is run.
 - `build-containers`: Denotes tasks to call the role [build_containers](../roles/build_containers.md) when deploy-edpm.yml playbook is run.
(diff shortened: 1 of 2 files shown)


raukadah:
/approve
thank you for adding more tags, it means more powers!

marios (approved):

pablintino:
I have some doubts about this one, not only about the PR itself, but on the topic it touches and the way it evolved to the point of requiring a tag.
So, a couple of points:

- I'm fine doing this hack of killing the pods, but why should we kill them if the environment is fine? I'd prefer to avoid doing so if all the pods in that namespace are in the `Running` state.
- This hack was relevant in OCP 4.15, in 4.16.z it's solved. I hope the issues we are still facing are because we are testing with 4.15 or an early release of 4.16. https://issues.redhat.com/browse/OCPBUGS-31645

cjeanner:
> I have some doubts about this one, not only about the PR itself, but on the topic it touches and the way it evolved to the point of requiring a tag. So, a couple of points:
> 
>     * I'm fine doing this hack of killing the pods, but why should we kill them if the environment is fine? I'd prefer to avoid doing so if all the pods in that namespace are in the `Running` state.
> 
>     * This hack was relevant in OCP 4.15, in 4.16.z it's solved. I hope the issues we are still facing are because we are testing with 4.15 or an early release of 4.16. https://issues.redhat.com/browse/OCPBUGS-31645

Indeed, if we can get something more automated, I'm all for it. My use-case covers mostly "dev" - QE usually don't need to re-run the whole thing, unless they want to test scaling in/out the EDPM nodes (and, even with that, they can leverage other tags to speed up the second deploy).

So, what shall we do with this PR? close it in favor of something else, or get it in, and iterate in a follow-up, @pablintino ? Either way are "ok" for me. I just won't have time to work on the "better" solution right now -.-'

lewisdenny:
@cjeanner I agree with Pablo on this one, as we are coming to an inflection point with the project I think we should shifting the needle over to doing things the right way rather than the fast way. 

Does this patch fix the issue? Yes and thank you. Could it be done better? Yes for sure but you know that but just don't have time tight now. 

I would opt to add it to the backlog and if it's an issue bugging someone they will fix it when they have time. 

The reason I don't just approve this one is it adds complexity, one more skip tag that we could avoid.

PR #2143, opened 2024-07-24
Prepare for FQDN unique names: controller-0 match

cjeanner:
We'll have to use FQDN, unique names for the deployment at some point.

This PR slightly modifies how we extract controller-0 related data from
the hostvars, in order to prepare for the FQDN integration (there's a
good chance the inventory will get updated).

It also introduces a try/catch schema to provide hopefully useful output
in case of failure.


cescgina (commented):
/lgtm

PR #2146, opened 2024-07-24
Ensure we're talking with boolean

cjeanner:
Usually we want to cast user provided booleans using the `| bool`
filter, just to be sure we're doing the right thing in the play.


lewisdenny (approved):
/lgtm

PR #2148, opened 2024-07-24
Ensures we can use FQDN in ceph.yml playbook

cjeanner:
Until now, there was an inconsistency between the 06-deploy-edpm.yml and
06-deploy-architecture.yml playbooks when it comes to ceph
configuration:
in the first one, ceph.yml play was instructed to use FQDN (via
ansible_fqdn fact), while the latter was using shortnames (via
ansible_hostname).

This leads to issues whenever we want to set FQDN on the various nodes
and in the inventory, since some stages of the ceph configuration is
checking for name consistency.

### DIFF
--- a/playbooks/06-deploy-architecture.yml
+++ b/playbooks/06-deploy-architecture.yml
@@ -168,6 +168,22 @@
         - storage
         - edpm_bootstrap
 
+    # The ceph_spec_fqdn parameter is set to true in the
+    # 06-deploy-edpm.yml play already. This will instruct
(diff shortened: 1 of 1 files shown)


fultonj:
This might be sufficient. cephadm's hostname handling is complex. Let's make sure delta and delta-ipv6 can pass with this before merging.

cjeanner:
Seems we can remove all of those `*_spec_fqdn` and leverage `inventory_hostname` and its variations (`item` in loop).

This is implemented in #2164 . Closing this one.

PR #2153, opened 2024-07-24
Use latest version of workload_launch for update.

sathlan:
This improve the comptatibility with different public network
setup (vlan segmentation for instance) and pave the way to get the
continuous control plane testing.

-------

  - [x] README in the role

### DIFF
--- a/roles/update/templates/workload_launch.sh.j2
+++ b/roles/update/templates/workload_launch.sh.j2
@@ -1,20 +1,133 @@
 #!/bin/bash
 #
 # Script that spawns an instance
-# Has 3 operational modes:
+# Has 5 operational modes:
 #   - workload: spawn VM with floating ip (default) or external ip in case of
 #               SRIOV
 #   - sanity:   spawn VM with floating/external ip assert it's ssh'able,
 #               delete VM and associated resources
+#   - sanityfast: same than sanity but fails faster
+#
+#   - workload_traffic: spawn a vm and generate traffic to it before
+#     anything else so that the interface counters are near 0.
+#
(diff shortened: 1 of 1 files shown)


raukadah:
recheck Ensure that user is logged crc failure on bm job

roles/update/templates/workload_launch.sh.j2
```diff
@@ -394,9 +499,18 @@ if [[ "${MODE}" == "cleanup" ]]; then
 fi
 
 if [[ "${MODE}" == "sanity" ]]; then
+    trap cleanup_on_exit EXIT
```
ciecierski, line 502: Why trap command here and next mode too?

ciecierski (changes requested):
Is `trap `required?

roles/update/templates/workload_launch.sh.j2
sathlan: Sanity and sanifyfast mode, are special modes where vm are created and then destroyed.  sanityfast just has lower timeout values.

trap EXIT is a nice bash signal extension to make sure the cleanup code run whatever happen during deployment, either an error or a normal exit.

The code then doesn't have to be cluttered with conditional on every command to check if it succeed or not.  We are just certain that cleanup will try to run no matter what.

ciecierski: thank you for answer, makes sense to me

ciecierski (approved):

cjeanner:
/approve

won't really break "us" afaik, so if it kills update, that's on them :).

PR #2154, opened 2024-07-24
Update adoption-standalone-*-provider irrelevant-files list

viroel:
"adoption-standalone-to-crc-ceph-provider" job depends on a content-provider job to run but since it has a different list of irrelevant-files, Zuul can throw an error like: " adoption-standalone-to-crc-ceph-provider depends on openstack-k8s-operators-content-provider which was not run". This PR copies the missing irrelevant-files from content-provider job to this adoption job.


marios (approved):
lgtm 

thanks @viroel

PR #2156, opened 2024-07-25
Skip provider/edpm jobs on validations role

raukadah:
Validations role is used downstream. It does not make sense to run provider/edpm jobs on this role.


lewisdenny (approved):
/approve

bshephar (approved):

PR #2159, opened 2024-07-25
Add missing connection option to a virsh command

cjeanner:


rebtoor (commented):
/approve

/lgtm

rebtoor:
Successfully tested locally.

cjeanner:

First run got a timeout, that same job is green, but now we're short in zuul resources. All jobs are green here, so let's save some time/resources.

PR #2161, opened 2024-07-25
Allow image pull from gcr.io during EDPM deployment

raukadah:
Adding gcr.io to secure registry will allow to fix following error:
```
Normal   BackOff         102s (x44 over 11m)  kubelet
Back-off pulling image "gcr.io/kubebuilder/kube-rbac-proxy@sha256:d99a8d144816b951a67648c12c0b988936ccd25cf3754f3cd85ab8c01592248f"
```
during edpm deployment and fix operator image ImagePullBackOff error.


ci/playbooks/tasks/set_crc_insecure_registry.yml
```diff
@@ -12,14 +12,15 @@
     }
     }' image.config.openshift.io/cluster
 
-- name: Patch the image.config.openshift.io resource to allow insecure registry
+- name: Patch the image.config.openshift.io resource to allow secure/insecure registry
```
rebtoor, line 15: This task define only secure registries :)

cjeanner: `Patch the image.config.openshift.io resource to allow registries` would be good imho, if we really want to edit that task name :).

cjeanner:
/approve

won't obsess on a task name at that point :).

ci/playbooks/tasks/set_crc_insecure_registry.yml
rebtoor: ++

raukadah: Done

rebtoor:
/hold

debugging result on node

PR #2163, opened 2024-07-25
Remove duplicated job

cjeanner:
the cifmw-baremetal-nested-crc job was mostly useless, since it was a
duplicate of cifmw-crc-podified-edpm-baremetal.

While the nested-crc one had a `files` entry to limit its triggering,
the other one doesn't have any limitation, probably because it's using
the content-provider.

A potential follow-up would be to limit the triggers in the
project-templates.yml.


raukadah (commented):
/approve

PR #2164, opened 2024-07-25
Use inventory_hostname in ceph related roles and play

cjeanner:
Until now, the ceph playbook and associated roles were trying to be
smart, but it created many other issues, and continues to create weird
situations.

With this patch, we reverse the burden, and base everything on the
inventory.
Since the inventory is generated by the Framework, and the Framework
also sets the virtual machines hostname, we should be able to rely on
the `inventory_hostname` only.

This means the inventory must be consistent, if provided by external
means.

We therefore update/correct the hostname set on the various nodes to
match the inventory.

  - [x] README in the role

- [x] va-hci (locally tested)
- [x] delta-v6 (downstream TP)

### DIFF
--- a/ci/playbooks/multinode-customizations.yml
+++ b/ci/playbooks/multinode-customizations.yml
@@ -346,5 +346,5 @@
     - name: Inject hostname in configuration file
       become: true
       ansible.builtin.hostname:
-        name: "{{ inventory_hostname }}.ci-rdo.local"
+        name: "{{ inventory_hostname }}"
         use: "systemd"
--- a/playbooks/06-deploy-edpm.yml
+++ b/playbooks/06-deploy-edpm.yml
@@ -122,7 +122,6 @@
         (cifmw_edpm_deploy_hci | default('false') | bool) and
         cifmw_architecture_scenario is undefined
       }}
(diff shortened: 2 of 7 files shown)


fultonj:
Thank you Cedric. This is a nice clean up. Let's make sure it doesn't break anything (e.g. IPv6) before merging.

cjeanner:
It's failing in multinode-hci-deployment job:
`ERROR: hostname "compute-0.ci-rdo.local" does not match expected hostname "compute-0"`

When checking the source of cephadm, we can see the inventory doesn't reflect the actual hostname:
https://github.com/ceph/ceph/blob/7fe91d5d5842e04be3b4f514d6dd990c54b29c76/src/cephadm/cephadm.py#L8128-L8133

To correct that, we have 2 ways:
- either ensure the inventory passed to ceph.yml is consistent (need to see how it's built, I suppose it's passing the zuul inventory as-is)
- or modify the compute node hostname as a pre-task during the job preparation.

Having consistency between hostnames and inventory names sounds like the right thing imho. "just" need to make that consistency a realty.

fultonj (approved):
/lgtm
/approve

PR #2124, opened 2024-07-19
[test-operator] Enable creation of multiple PVCs

kstrenkova:
This patch enables the creation of multiple PVCs when there are multiple steps in workflow and parallel option is set to true. This change is needed, because the LVM backed storage does not support the RWX, which causes issues in some cases.

As a pull request owner and reviewers, I checked that:
- [x] Appropriate testing is done and actually running
- [x] Appropriate documentation exists and/or is up-to-date:
  - [x] README in the role
  - [x] Content of the docs/source is reflecting the changes

PR #2168, opened 2024-07-26
[test-operator] Fix volume mounting

lpiwowar:
In this patch we introduced a small bug that appears when we run both tempest and tobiko tests in a single job [1]. The issue is that when tobiko is executed both variables volume and volume_mounts contain values from the tempest run.

This PR ensures that both volume and volume_mounts are initialized to an empty list when logs are being collected.

[1] https://github.com/openstack-k8s-operators/ci-framework/pull/2124

  - [x] README in the role

### DIFF
--- a/roles/test_operator/tasks/run-test-operator-job.yml
+++ b/roles/test_operator/tasks/run-test-operator-job.yml
@@ -67,6 +67,11 @@
     - not cifmw_test_operator_dry_run | bool
     - not testjob_timed_out
   block:
+    - name: Reset volumes and volume_mounts to an empty list
+      ansible.builtin.set_fact:
+        volumes: []
+        volume_mounts: []
+
     - name: Get information about PVCs that store the logs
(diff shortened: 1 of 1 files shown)


lpiwowar:
Testing here -> https://github.com/openstack-k8s-operators/test-operator/pull/145

roles/test_operator/tasks/run-test-operator-job.yml
```diff
@@ -67,6 +67,11 @@
     - not cifmw_test_operator_dry_run | bool
     - not testjob_timed_out
   block:
+    - name: Initialize volumes and volume_mounts
```
viroel, line 70: You are actually reseting the values right? Maybe the task name could say that. Not a blocking to merge

lpiwowar:
recheck

frenzyfriday:
/lgtm 
the TP result is green

PR #2170, opened 2024-07-29
Use cifmw_dlrn_report_promote_source to set proper dlrn reporting target

raukadah:
In update job, we deploy the system using cifmw_repo_setup_promotion: current-podified var and update it to podified-ci-testing via different update vars.

dlrn_report role works on repo-setup vars. In this case, dlrn will report the job status against current-podified. But we want to report the status against podified-ci-testing. In this case, current dlrn-report does not work.

In order to fix that we are adding a new var `cifmw_dlrn_report_promote_source` to set the proper dlrn target against which job should report. It will not affect existing job. As it's value depends on cifmw_repo_setup_promotion case.

  - [x] README in the role


roles/dlrn_report/tasks/dlrn_report_results.yml
```diff
@@ -5,6 +5,12 @@
     name: set-zuul-log-path-fact
 
 - name: Get hash related data from repo_setup role
+  vars:
+    cifmw_repo_setup_promotion: >-
+      {{
+        (cifmw_dlrn_report_promote_source | length > 0) |
+        ternary(cifmw_dlrn_report_promote_source, cifmw_repo_setup_promotion)
+      }}
```
raukadah, line 13: @pablintino Hello, I am not sure why I am hitting [following error](https://logserver.rdoproject.org/70/2170/3e33d800f341ce9c10a996876730075edbf2b1b4/github-check/cifmw-pod-pre-commit/c845bd8/pre_commit.log):
```
Error was a <class 'ansible.errors.AnsibleError'>, original message: recursive loop detected in template string: {{
  (cifmw_dlrn_report_promote_source | length > 0) |
  ternary(cifmw_dlrn_report_promote_source, cifmw_repo_setup_promotion)
}}. maximum recursion depth exceeded while calling a Python object
```

raukadah: Thank you @ciecierski , I have changed to simpler set_fact. The task will also look much simpler now.

PR #2173, opened 2024-07-29
manila share type: do not fail if share exists

queria:
When cifmw deployment/reproducer is rerun (e.g. for scale-out), it always fails in manila-create-default-resources hook since the default share already exists.

[OSPRH-8983](https://issues.redhat.com//browse/OSPRH-8983)

  - [x] README in the role

### DIFF
--- a/hooks/playbooks/manila_create_default_resources.yml
+++ b/hooks/playbooks/manila_create_default_resources.yml
@@ -2,12 +2,33 @@
 - name: Create Manila resources needed for tempest run
   hosts: "{{ cifmw_target_hook_host | default('localhost') }}"
   gather_facts: false
+  vars:
+    share_type_name: "default"
+    driver_handles_share_servers: "False"
+    extra_specs:
+      snapshot_support: "True"
+      create_share_from_snapshot_support: "True"
   tasks:
-    - name: Creat share type default for manila tempest plugin tests
+    - name: Create share type default for manila tempest plugin tests
       environment:
         KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
         PATH: "{{ cifmw_path }}"
-      ansible.builtin.shell: |
+      ansible.builtin.command: |
         oc -n {{ namespace }} exec -it pod/openstackclient \
-          -- openstack share type create default false \
-          --snapshot-support True --create-share-from-snapshot-support True
+          -- openstack share type create {{ share_type_name }} {{ driver_handles_share_servers }}
+      register: _manila_share_creation
+      failed_when: >-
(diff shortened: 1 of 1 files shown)


hooks/playbooks/manila_create_default_resources.yml
```diff
@@ -11,3 +11,8 @@
         oc -n {{ namespace }} exec -it pod/openstackclient \
           -- openstack share type create default false \
           --snapshot-support True --create-share-from-snapshot-support True
```
tosky, line 13: Maybe it would make it sense to split it into the create (and not fail if it exists)) and one to assign the additional spec

gouthampacha: yes; wouldn't it?


```
---
- name: Create Manila resources needed for tempest run
  hosts: "{{ cifmw_target_hook_host | default('localhost') }}"
  gather_facts: false
  vars:
    share_type_name: "default"
    driver_handles_share_servers: "False"
    extra_specs:
      snapshot_support: "True"
      create_share_from_snapshot_support: "True"
  tasks:
    - name: Check if the Manila share type already exists
       environment:
        PATH: "{{ cifmw_path }}"
      ansible.builtin.shell: |
        oc -n {{ namespace }} exec -it pod/openstackclient \
        -- openstack share type list -f value -c Name | grep -w "{{ share_type_name }}"
      register: share_type_exists
      ignore_errors: true
    - name: Create the Manila share type if it does not exist
      environment:
        PATH: "{{ cifmw_path }}"
      ansible.builtin.shell: |
        oc -n {{ namespace }} exec -it pod/openstackclient \
        openstack share type create "{{ share_type_name }}" \
           "{{ driver_handles_share_servers }}" \
           {{ ' '.join('--extra-specs ' + k + '=' + v for k, v in extra_specs.items()) }}
      when: share_type_exists.rc != 0
```

queria: hm this seems as just more verbose 'exists - or create',
it does not cover the case when it may exist but without expected properties

what tosky suggests (and i agree its at least safer) is more like:
- share type create (failed_when: failed && not already-exists)
- share type update to set properties // assuming this does not fail if properties already match (will check)

queria: I've updated this to align with both - to do create|ignore-existing + update properties ... and use the syntax/approach based on vars for name and extra_specs dictionary.

tosky:
Looks good now - @gouthampacha what do you think?

gouthampacha (approved):
Great; thanks @queria

PR #2175, opened 2024-07-29
[reproducer] Call switches config if available

pablintino:
  - [x] README in the role

### DIFF
--- a/docs/source/usage/01_usage.md
+++ b/docs/source/usage/01_usage.md
@@ -67,6 +67,7 @@ are shared among multiple roles:
 - `cifmw_openshift_ingress_ip_address` (String) contains the OpenShift Ingress IP address. Note: it is computed internally and should not be user defined.
 - `cifmw_nolog`: (Bool) Toggle `no_log` value for selected tasks. Defaults to `true` (hiding those logs by default).
 - `cifmw_parent_scenario`: (String or List(String)) path to existing scenario/parameter file to inherit from.
+- `cifmw_configure_switches`: (Bool) Specifies whether switches should be configured. Computes in `reproducer.yml` playbook. Defaults to `false`.
 
 ```{admonition} Words of caution
 :class: danger
--- a/playbooks/switches_config.yml
+++ b/playbooks/switches_config.yml
@@ -1,0 +1,30 @@
+---
+- name: Switches configuration requirements
+  hosts: "{{ cifmw_target_host | default('localhost') }}"
+  gather_facts: false
+  pre_tasks:
+    # end_play will end this current playbook and go the the next
+    # imported play.
+    - name: Early playbook stop
+      when:
+        - not cifmw_configure_switches | default(true)
+      ansible.builtin.meta: end_play
+  tasks:
+    - name: Install required packages
+      delegate_to: localhost
+      ansible.builtin.pip:
+        name:
+          - paramiko
+          - ncclient
+        extra_args: --user
+
+- name: Switches configuration
+  hosts: switches
+  gather_facts: false
+  pre_tasks:
+    - name: Early playbook stop
+      when:
+        - not cifmw_configure_switches | default(true)
+      ansible.builtin.meta: end_play
+  roles:
+    - switch_config
--- a/reproducer.yml
+++ b/reproducer.yml
@@ -1,6 +1,5 @@
 ---
-
-- name: Reproducer Play
+- name: Reproducer prepare play
   hosts: "{{ cifmw_target_host | default('localhost') }}"
   gather_facts: true
   pre_tasks:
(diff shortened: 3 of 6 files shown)


reproducer.yml
```diff
@@ -57,11 +56,18 @@
       ansible.builtin.import_role:
         name: repo_setup
         tasks_from: rhos_release.yml
-
   roles:
     - role: ci_setup
-    - role: reproducer
 
+- name: Prepare switches if present
+  ansible.builtin.import_playbook: playbooks/switches_config.yml
```
cjeanner, line 63:
```suggestion
  ansible.builtin.import_playbook: playbooks/switch_config.yml
```

reproducer.yml
```diff
@@ -57,11 +56,18 @@
...
+  when: "'switches' in groups"
```
cjeanner, line 64: that will lead to "skipped" tasks. You want a pre_tasks in the `switch_config.yml` play imho, like we did for ceph.yml for instance.

playbooks/switches_config.yml
```diff
@@ -1,14 +1,16 @@
+---
 - name: Switches configuration requirements
-  hosts: 'localhost'
+  hosts: "{{ cifmw_target_host | default('localhost') }}"
   gather_facts: false
   tasks:
```
cjeanner, line 5: Adding a pre_tasks here, with a `when` condition allowing an early end_play, would be really good.

reproducer.yml
cjeanner: (scratch that, I didn't see you renamed it)

reproducer.yml
```diff
@@ -57,11 +56,18 @@
       ansible.builtin.import_role:
         name: repo_setup
         tasks_from: rhos_release.yml
-
   roles:
     - role: ci_setup
-    - role: reproducer
 
+- name: Prepare switches if present
```
cjeanner, line 62:
```yaml
vars:
  has_switches: "{{ groups['switches'] is defined }}"
```

then you can check on that `has_switches` as a condition to early end_play in the playbook itself. Doing so would allow to still run the play as a hook by passing the boolean.

playbooks/switches_config.yml
```diff
@@ -0,0 +1,25 @@
+---
+- name: Switches configuration requirements
+  hosts: "{{ cifmw_target_host | default('localhost') }}"
+  gather_facts: false
+  tasks:
+    # end_play will end this current playbook and go the the next
+    # imported play.
+    - name: Early stop ceph related work
```
cjeanner, line 8: I'd use pre_tasks for that one

reproducer.yml
pablintino: I've switched to this suggested approach.

playbooks/switches_config.yml
```diff
@@ -0,0 +1,25 @@
+---
+- name: Switches configuration requirements
+  hosts: "{{ cifmw_target_host | default('localhost') }}"
+  gather_facts: false
+  tasks:
+    # end_play will end this current playbook and go the the next
+    # imported play.
+    - name: Early playbook stop
+      when:
+        - not cifmw_configure_switches | default(true)
```
lewisdenny, line 10: I don't see this top level var documented in `docs/source/usage/01_usage.md`

lewisdenny (changes requested):

playbooks/switches_config.yml
```diff
@@ -0,0 +1,25 @@
...
+
+    - name: Install required packages
+      delegate_to: localhost
+      ansible.builtin.pip:
+        name:
+          - paramiko
+          - ncclient
+        extra_args: --user
+
+- name: Switches configuration
```
cjeanner, line 21: ah, you need to end_play *all* of the plays listed in that file. Check ceph.yml, we're facing the same.

Also, using `pre_tasks` ensures a really early end.

docs/source/usage/01_usage.md
```diff
@@ -67,6 +67,7 @@ are shared among multiple roles:
 - `cifmw_openshift_ingress_ip_address` (String) contains the OpenShift Ingress IP address. Note: it is computed internally and should not be user defined.
 - `cifmw_nolog`: (Bool) Toggle `no_log` value for selected tasks. Defaults to `true` (hiding those logs by default).
 - `cifmw_parent_scenario`: (String or List(String)) path to existing scenario/parameter file to inherit from.
+- `cifmw_configure_switches`: (Bool) Specifies whether switches should be configured. Defaults to `false`.
```
cjeanner, line 70: if we document that one, we must also mention the way it's computed in the reproducer.yml play imho.

rebtoor (commented):
/lgtm

PR #2179, opened 2024-07-30
[DNM] Remove nodes from SR-IOV Template

jamepark4:
Currently evaluating removal of nodes since I am explicitly adding them downstream via values.yaml.


jamepark4:
Based on feedback from @pablintino, this was an issue with how job details were defined downstream when defining `cifmw_baremetal_hosts` .  This approach is not needed to progress deployments with fixedIPs manually configured, closing.

PR #1907, opened 2024-06-19
Dropping dataplane-operator references

fao89:
As a pull request owner and reviewers, I checked that:
- [x] Appropriate testing is done and actually running
- [x] Appropriate documentation exists and/or is up-to-date:
  - [x] README in the role
  - [x] Content of the docs/source is reflecting the changes

PR #2180, opened 2024-07-30
Remove dataplane-operator references

rlandy:
Now that the dataplane-operator is combined and
not installed  through a separate repo.

  - [x] README in the role

### DIFF
--- a/roles/edpm_deploy/tasks/main.yml
+++ b/roles/edpm_deploy/tasks/main.yml
@@ -18,7 +18,6 @@
   when:
     - zuul is not defined
     - cifmw_operator_build_output is defined
-    - "'dataplane-operator' in operators_build_output.operators"
   tags:
     - always
   vars:
@@ -27,7 +26,6 @@
     _install_yamls_repos:
       DATAPLANE_BRANCH: ""
       GIT_CLONE_OPTS': "-l"
-      DATAPLANE_REPO': "{{ operators_build_output['dataplane-operator'].git_src_dir }}"
 
 - name: Set EDPM related vars
   when:
--- a/roles/edpm_deploy_baremetal/tasks/main.yml
+++ b/roles/edpm_deploy_baremetal/tasks/main.yml
@@ -32,7 +32,6 @@
         ) |
         combine(
           {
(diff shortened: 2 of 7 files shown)


rlandy:
Note alternative review: https://github.com/openstack-k8s-operators/ci-framework/pull/1907

roles/edpm_deploy/tasks/main.yml
```diff
@@ -18,7 +18,6 @@
   when:
     - zuul is not defined
     - cifmw_operator_build_output is defined
-    - "'dataplane-operator' in operators_build_output.operators"
```
chkumar246, line 21: I think we can drop this whole task "Define minimal set of repo variables when not running on Zuul" as it is specific to dataplane operator.

raukadah, line 21: I think we can drop this whole task "Define minimal set of repo variables when not running on Zuul" as it is specific to dataplane-operator.

roles/edpm_deploy_baremetal/tasks/main.yml
```diff
@@ -32,7 +32,6 @@
         ) |
         combine(
           {
-            'DATAPLANE_REPO': operators_build_output['dataplane-operator'].git_src_dir,
```
raukadah, line 35: This dict from line 33 to 39, needs to be removed.

tests/integration/targets/make/files/get_makefiles_env/makefiles/Makefile
pablintino, line 1: fyi: These files are examples used for integration test of the `get_makefiles_env` module. The values here are never used against install_yamls, but I'm ok removing the dataplane vars if you want to be consistent.

roles/edpm_deploy/tasks/main.yml
pablintino: +1

roles/edpm_deploy_baremetal/tasks/main.yml
pablintino: +1

PR #2185, opened 2024-08-01
nit: Update project guidelines to define approval policy

lewisdenny:
This patch documents the projects policy on PR approval, mainly that PRs should have two different reviewers (excluding the PR owner) approval rather than one reviewer adding both votes.


lewisdenny:
/hold
Holding as I would like more reviewers to see this.

PR #2196, opened 2024-08-05
Use separate test_operator artifacts basedir for update test run

ciecierski:
Save test_operator test runs to separate test_operator_update dir to prevent overriding test_operators deployment run results with update run.

  - [ ] README in the role


lpiwowar (approved):
LGTM! :+1: 

This makes sense. We want to store the logs into a separate folder when running the test-operator after an update.

PR #2199, opened 2024-08-05
Ensure vbmc is stable

cjeanner:
It seems libvirt_manager goes so fast that, on some system, vBMC doesn't
have time to properly stabilize.

This patch moves the vBMC deployment to a more logical place (reproducer
role), and introduce a way to ensure the service is properly running.


raukadah (commented):
/lgtm

PR #2210, opened 2024-08-07
[switch_config] Fix pip install

pablintino:
If the reproducer runs in a venv the pip install will fail as we are passing the --user option to pip.
As we are not usign become: true in the install task enforcing --user is not needed, since pip will install the dependencies inside the user site-packages as the root one is not writable.

### DIFF
--- a/playbooks/switches_config.yml
+++ b/playbooks/switches_config.yml
@@ -16,7 +16,6 @@
         name:
           - paramiko
           - ncclient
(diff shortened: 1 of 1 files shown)


rebtoor (approved):
/approve

pablintino:
Holding till testing is done.

pablintino:
Test in the NFV environment.

PR #2213, opened 2024-08-08
Ensure we get a timeout for node listing

cjeanner:
It seems getting the nodes might hang, and it never ever returns.

This patch ensures we set a wait_condition, meaning we can toggle the
`wait` parameter to `true` and benefit from the default `wait_timeout`
of 120s[1] - which should be more than enough to get the OCP cluster
nodes list.

This PR also extract the node listing into its own tasks file, so that
we can import it instead of copy/pasting the same code.

This PR also adds a debugging capability when we're waiting for the
cluster to stabilize after removing the certificates. Using our
ci_script action, we're able to get logs, meaning we can easily review
the command output. Usually it lists the failing service(s) and provides
data about stability duration.

[1] https://docs.ansible.com/ansible/latest/collections/kubernetes/core/k8s_info_module.html#parameter-wait_timeout

### DIFF
--- a/roles/openshift_adm/tasks/api_cert.yml
+++ b/roles/openshift_adm/tasks/api_cert.yml
@@ -48,8 +48,11 @@
   environment:
     KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
     PATH: "{{ cifmw_path }}"
-  ansible.builtin.command:
-    cmd: "oc adm wait-for-stable-cluster --minimum-stable-period=3m"
+  cifmw.general.ci_script:
+    output_dir: "{{ cifmw_openshift_adm_basedir }}/artifacts"
+    script: >-
+      oc adm wait-for-stable-cluster
+      --minimum-stable-period={{ cifmw_openshift_adm_stable_period }}
 
 - name: Gather the kubelet signer secret.
   kubernetes.core.k8s_info:
--- a/roles/openshift_adm/README.md
+++ b/roles/openshift_adm/README.md
@@ -14,11 +14,13 @@ requires administrator level privileges for the deployed OpenShift.
 
 This role requires the following parameters to be configured.
 
+* `cifmw_openshift_adm_basedir` (str) Framework base directory, defaults to `cifmw_basedir` or
+  `~/ci-framework-data`.
 * `cifmw_openshift_api` (str) Cluster endpoint to be used for communication.
 * `cifmw_openshift_user` (str) Name of the user to be used for authentication.
 * `cifmw_openshift_password` (str) Password of the provided user.
 * `cifmw_openshift_kubeconfig` (str) Absolute path to the kubeconfig file.
-* `cifmw_base_dir` (str) Absolute path to the base directory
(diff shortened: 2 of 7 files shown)


roles/openshift_adm/tasks/api_cert.yml
```diff
@@ -44,13 +44,6 @@
   loop_control:
     label: "{{ item }}"
 
-- name: Wait until the OpenShift cluster is stable after certificate update.
-  environment:
-    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
-    PATH: "{{ cifmw_path }}"
-  ansible.builtin.command:
-    cmd: "oc adm wait-for-stable-cluster --minimum-stable-period=3m"
-
```
rebtoor, line 53: I'm a bit unsure about this.

This step is required by the [procedure](https://access.redhat.com/articles/regenerating_cluster_certificates) of the certificates regeneration.

The command itself doesn't simply check for the cluster operators readiness, but it ensure their stability in a defined timeframe, as you can guess by the command line argument and from the [source code](https://github.com/openshift/oc/blob/release-4.16/pkg/cli/admin/waitforstable/command.go) of the command itself.

It's true that the cluster won't be used as-is, but what happens if we introduce an issue given by the new certificates?

cjeanner: that's why I'm testing it right now. Thing is, the certificate is regenerated once we boot the cluster from the overlays. I really don't think this command should be placed here.

If really, it should be done once the certificates are really re-generated. Not before shutting the cluster down. Note that we have the same command, but with a shorter minimum-stable-period:
https://github.com/openstack-k8s-operators/ci-framework/blob/6148ebc68a63451f9e7701b04a03d79649596122/roles/openshift_adm/tasks/wait_for_cluster.yml#L62-L68

And that one is called at the "right" place.

cjeanner:
/hold

Roberto has some power issues, I want to give him enough time to validate. But for the rest, it's mostly "ready for review".

roles/openshift_adm/tasks/api_cert.yml
cjeanner: after some discussions : we seem to need to keep it - but I'll check the doc to see the default value for the minimum-stable-period - the doc doesn't set any parameter for that one

cjeanner:
With the latest iteration, we now have a nice log to review about the wait-for-stable, such as:
```
[...]
clusteroperators/console stabilized at 2024-08-09T15:21:51+02:00 after 20s
clusteroperators/kube-apiserver is still progressing after 3m50s

clusteroperators/kube-apiserver is still progressing after 4m0s

clusteroperators/kube-apiserver is still progressing after 4m10s

clusteroperators/kube-apiserver is still progressing after 4m20s

clusteroperators/kube-apiserver is still progressing after 4m30s

clusteroperators/kube-apiserver is still progressing after 4m40s
[...]
```

rebtoor (approved):
/approve
/lgtm

PR #2218, opened 2024-08-11
Set required vars for update_containers role in reproducer tcib job

raukadah:
Currently reproducer tcib playbook pushes the containers with build_containers role default's namespace and tag.

Ci reproducer job have no idea about the above information leading to ImagePullerror failure during podified deployment.

This pr make sure we use update_containers namespace and tag to push the containers in reproducer tcib playbook run and also pass the correct the update_containers registry ip. If an user overrides the update_containers registry ip, it needs to be ignored to above ImagePullerror issue.

Note: run_hook role loads the parameters files from artifacts/parameters. Content provider yaml file is needed in fetch
 compute fact hook to set proper edpm registry url otherwise it will not work. Copying the content provider registry url at two places, fixes theissue.

### DIFF
--- a/roles/reproducer/templates/content-provider.yml.j2
+++ b/roles/reproducer/templates/content-provider.yml.j2
@@ -121,6 +121,8 @@
           -e @./scenarios/centos-9/base.yml
           -e @./scenarios/centos-9/tcib.yml
           -e cifmw_zuul_target_host=controller-0
+          -e "cifmw_build_containers_registry_namespace="{{ cifmw_update_containers_org | default('podified-antelope-centos9') }}""
+          -e "cifmw_build_containers_image_tag="{{ cifmw_update_containers_tag | default('current-podified') }}""
           -e "cifmw_rp_registry_ip="{{ hostvars['localhost']['cifmw_rp_registry_ip'] }}""
           -e "@{{ ansible_user_dir }}/ci-framework-data/parameters/reproducer-variables.yml"
           -e "@~/{{ job_id }}-params/zuul-params.yml"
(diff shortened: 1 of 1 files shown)


roles/reproducer/templates/content-provider.yml.j2
```diff
@@ -121,18 +121,24 @@
           -e @./scenarios/centos-9/base.yml
           -e @./scenarios/centos-9/tcib.yml
           -e cifmw_zuul_target_host=controller-0
+          -e "cifmw_build_containers_registry_namespace="{{ cifmw_update_containers_org | default('podified-antelope-centos9') }}""
```
cjeanner, line 124: you're in a `raw` block, it won't generate the expected content.

roles/reproducer/templates/content-provider.yml.j2
```diff
@@ -121,18 +121,24 @@
...
+          -e "cifmw_build_containers_image_tag="{{ cifmw_update_containers_tag | default('current-podified') }}""
           -e "cifmw_rp_registry_ip="{{ hostvars['localhost']['cifmw_rp_registry_ip'] }}""
           -e "@{{ ansible_user_dir }}/ci-framework-data/parameters/reproducer-variables.yml"
           -e "@~/{{ job_id }}-params/zuul-params.yml"
           -e "cifmw_rp_registry_port=5001"
           --skip-tags build_openstack_packages
 
     - name: Output needed content into consumable environment
+      loop:
```
cjeanner, line 133: instead of generating twice, we might generate once, and symlink?

cjeanner (changes requested):

roles/reproducer/templates/content-provider.yml.j2
cjeanner: after discussion: this doesn't need to be interpreted during the file generation.

cjeanner (approved):
/approve

PR #2219, opened 2024-08-12
Don't use ansible_domain if empty string

rabi:


cjeanner:
/lgtm

Looks OK, but would love to get some TP result to ensure it does the needed filtering.

cjeanner:
/lgtm

We got a downstream job passing with this code, correcting an issue raised with the removal of `ci-rdo.local` domain in #2164

PR #2220, opened 2024-08-12
[libvirt_manager] Adopt cloud-init usage while provisioning EL based vms

rebtoor:
In an effort to increase deployment speed and maintainability, a lot of tasks performed using `virt-sysprep` on the vm disk image and using ansible `shell`/`command` modules have been replaced by a `user-data` file provided to `cloud-init` using crafted ISO files attached to the vms before their first boot.

Also, since a parameter to configure chrony at boot time (only for controller nodes) has been added, the reproducer scenario files have been updated accordingly.


rebtoor:
> Any reason for not using the existing role? https://github.com/openstack-k8s-operators/ci-framework/tree/main/roles/config_drive

🤦🏻‍♂️ i didn't check for it, let me fix my pr. Thanks @hjensas !

roles/libvirt_manager/tasks/create_cloud_init_iso.yml
```diff
@@ -0,0 +1,36 @@
...
+    _images_dir: "{{ [ cifmw_libvirt_manager_basedir, 'workload', ] | ansible.builtin.path_join }}"
+  block:
+    - name: "Acquire authorized_keys from hypervisor"
+      ansible.builtin.slurp:
+        src: "{{ ansible_user_dir ~ '/.ssh/authorized_keys' }}"
+      register: _authorized_keys_b64
+
+    - name: "Generate instance random UUID"
+      ansible.builtin.set_fact:
+        _instance_uuid: "{{ 99999999 | random | to_uuid | lower }}"
```
hjensas, line 14: suggest we use a seed, VM name or something to make the UUID idempotent?
```
{{ 99999999 | ansible.builtin.random(seed=$VM_NAME)| to_uuid | lower }}
```

roles/libvirt_manager/tasks/create_cloud_init_iso.yml
```diff
@@ -0,0 +1,36 @@
...
+
+    - name: "Render vm type based user-data"
+      vars:
+        _authorized_keys: "{{ _authorized_keys_b64.content | b64decode }}"
+        _static_ip_script: >-
```
hjensas, line 19: I think we should use network_config that cloud-init understands instead of this script? But that might be a follow up?

roles/libvirt_manager/tasks/create_cloud_init_iso.yml
```diff
@@ -0,0 +1,36 @@
...
+          {{
+            (
+              cifmw_libvirt_manager_ip_script_basedir,
+              'static-ip-' + item.key + '.sh'
+            ) | ansible.builtin.path_join
+          }}
+      ansible.builtin.set_fact:
+        _userdata: "{{ lookup('ansible.builtin.template', 'user-data.yml.j2') }}"
```
hjensas, line 27: Do we need this template? Can we not just build up cifmw_config_drive_userdata? similar to - https://github.com/openstack-k8s-operators/ci-framework/pull/1813/files#diff-0b33aeeecf16501c026ef2b08cfd8eb03d767fb837e927e630ff6c8dd5ebb356R96-R130

roles/libvirt_manager/tasks/create_cloud_init_iso.yml
rebtoor: I like the idea!

roles/libvirt_manager/tasks/create_cloud_init_iso.yml
rebtoor: I have to admit that I barely translated instructions from "old" shell instructions into cloud-init params, ignoring `network_config`. I'll have a look and try to integrate it in this PR or in a follow-up.

rebtoor:
/hold cancel

...and fixed :)

rebtoor:
recheck

roles/libvirt_manager/tasks/create_cloud_init_iso.yml
```diff
@@ -0,0 +1,72 @@
...
+  vars:
+    _images_dir: "{{ [ cifmw_libvirt_manager_basedir, 'workload', ] | ansible.builtin.path_join }}"
+  block:
+    - name: "Acquire authorized_keys from hypervisor"
+      ansible.builtin.slurp:
+        src: "{{ ansible_user_dir ~ '/.ssh/authorized_keys' }}"
+      register: _authorized_keys_b64
+
+    - name: Create a list of ssh pub keys from authorized_keys
+      ansible.builtin.set_fact:
```
pablintino, line 13: Setting this as a fact makes me think it's used multiple times but it's used only in the man set_fact below. Why aren't we merging this with the next other set_fact? Clarity?

roles/libvirt_manager/tasks/create_cloud_init_iso.yml
```diff
@@ -0,0 +1,72 @@
...
+        _ssh_authorized_keys: |-
+          [
+          {% for ssh_key in (_authorized_keys_b64.content | b64decode).split('\n') | select() %}
+            "{{ ssh_key | trim }}",
+          {% endfor %}
+          ]
+
+    - name: "Generate instance random UUID"
+      ansible.builtin.set_fact:
+        _instance_uuid: "{{ 99999999 | random(seed=vm) | to_uuid | lower }}"
```
pablintino, line 23: Do we really need a fact for this?

roles/libvirt_manager/tasks/create_cloud_init_iso.yml
```diff
@@ -0,0 +1,72 @@
...
+    - name: "Define the network config if needed"
+      when: item.value.ip_address is defined
+      ansible.builtin.set_fact:
+        _network_config:
+          network:
+            version: 2
+            renderer: NetworkManager
+            ethernets:
+              match:
+                name: eth0
```
pablintino, line 34: I'm not sure if hard-coding eth0 will escalate properly...

roles/libvirt_manager/tasks/create_cloud_init_iso.yml
rebtoor: I created it to have the vars while calling the role as concise as possible. 

Yesterday I learned that defining a var using `set_fact` with its default, is just _yet another way_ to define a variable (at host level but still). A fact becomes an `ansible_fact` (with its pro and cons) only defining `cacheable` property as `true`. Interesting, isn't it? https://docs.ansible.com/ansible/latest/collections/ansible/builtin/set_fact_module.html#parameter-cacheable

roles/libvirt_manager/tasks/create_cloud_init_iso.yml
rebtoor: I agree but i have just translated the [old](https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/libvirt_manager/templates/static_ip.j2) bash script. I never seen a real life example where we need to define this (and by this I mean where we define "ip_address" in the vm layout, so we probably can get rid of it.

roles/libvirt_manager/tasks/create_cloud_init_iso.yml
rebtoor: Clarity mainly, yep, but we can merge it since I don't have a strong opinion about it. I changed idea (at least) 3 times.

roles/libvirt_manager/tasks/create_cloud_init_iso.yml
rebtoor: @cjeanner do you know something more about it?

hjensas: We should have the MAC address available, it would be better to match on that?

Also, each of the "eternets" connection need to be defined under a "connaction id" key - see: https://cloudinit.readthedocs.io/en/latest/reference/network-config-format-v2.html#device-configuration-ids

```
  ethernets:
    <connection_id>:
      match:
        macaddress: 52:54:00:bd:8f:cb
      set-name: eth0
      addresses:
        - ...
```

I think it would be a good idea to use set-name to define an interface name matching the name of network. like `osptrunk`, `ctlplane` etc - matching the name of network as defined in libvirt manager config?

roles/libvirt_manager/tasks/create_cloud_init_iso.yml
```diff
@@ -0,0 +1,72 @@
...
+                - "{{ item.value.ip_address.address }}"
+              gateway4: "{{ item.value.ip_address.gw }}"
+              nameservers:
+                addresses:
+                  - "{{ item.value.ip_address.dns }}"
+
+    - name: "Define the user data"
+      ansible.builtin.set_fact:
+        _user_data:
+          users:
```
hjensas, line 46: It would be nice to implement this so that it can be overriden? In case we ever need a different set of users? To make it "easy" we could require the entire list when overriding.

rebtoor: ack, sounds good!

roles/libvirt_manager/tasks/create_cloud_init_iso.yml
rebtoor: As i thought, this is not required anymore. I chose the KISS approach here, leaving to the user the ability to pass a dict based on `network-config` v2.

roles/libvirt_manager/tasks/create_cloud_init_iso.yml
rebtoor: @hjensas so... now we have 3 choices:
- accept the defaults with `root` and `zuul` users
- define a `user` field in the libvirt layout to change the "sudo" enabled user from `zuul` to something else
- define a `cifmw_libvirt_manager_users` to completely override the list of users.

roles/libvirt_manager/tasks/create_cloud_init_iso.yml
```diff
@@ -0,0 +1,61 @@
...
+            ignore_growroot_disabled: true
+            mode: growpart
+          resize_rootfs: noblock
+
+    - name: "Call the config_drive role"
+      vars:
+        cifmw_config_drive_iso_image: "{{ [ _images_dir, vm ~ '.iso' ] | ansible.builtin.path_join }}"
+        cifmw_config_drive_uuid: "{{ _instance_uuid }}"
+        cifmw_config_drive_hostname: "{{ vm }}"
+        cifmw_config_drive_networkconfig: "{{ item.value.networkconfig | default(omit) }}"
```
pablintino, line 58: omit doesn't work in vars, it's redered as the __omit__<uuid> string. It only works in module args

cjeanner (changes requested):

roles/libvirt_manager/tasks/create_cloud_init_iso.yml
```diff
@@ -0,0 +1,61 @@
+---
+- name: Generate cloud-init iso
+  become: false
```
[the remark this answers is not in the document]
cjeanner: (it is, from the create_vms.yml file)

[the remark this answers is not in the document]
rebtoor: exactly :)

roles/libvirt_manager/tasks/create_cloud_init_iso.yml
rebtoor: ack, makes sense

roles/libvirt_manager/tasks/create_cloud_init_iso.yml
rebtoor: ack

roles/libvirt_manager/tasks/create_cloud_init_iso.yml
```diff
@@ -0,0 +1,61 @@
...
+
+    - name: "Generate instance random UUID"
+      ansible.builtin.set_fact:
+        _instance_uuid: "{{ 99999999 | random(seed=vm) | to_uuid | lower }}"
+
+    - name: "Define the default users for the vm"
+      when:
+        - cifmw_libvirt_manager_vm_users is not defined or
+          cifmw_libvirt_manager_vm_users | length == 0
+      ansible.builtin.set_fact:
```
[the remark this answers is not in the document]
rebtoor: ack

roles/libvirt_manager/tasks/create_cloud_init_iso.yml
rebtoor: *sigh* (thanks for the hint)

roles/libvirt_manager/tasks/create_vms.yml
```diff
@@ -51,6 +50,18 @@
         owner: "{{ ansible_user_id }}"
         state: file
 
+    - name: Create cloud-init ISO
+      when:
+        - item.value.disk_file_name != 'blank'
+        - item.key is not match('^(crc|ocp).*$')
+        - vm is not match('^(crc|ocp).*$')
+        - vm is match('^.*' ~ item.key ~ '.*$')
```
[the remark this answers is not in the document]
rebtoor: i removed the loop for now and adapted the variable names, let's see how it goes.

roles/libvirt_manager/tasks/create_vms.yml
```diff
@@ -51,6 +50,14 @@
         owner: "{{ ansible_user_id }}"
         state: file
 
+    - name: Create cloud-init ISO
+      when:
+        - vm_data.disk_file_name != 'blank'
```
cjeanner, line 55: we also probably don't want to create the ISO for VM that:
- aren't set to start
- aren't set to be managed

and, therefore, we don't want to attach that ISO with the same conditions. Maybe you can move this task in a common block with the one attaching the ISO, so that you have one `when` block?

roles/libvirt_manager/molecule/check_dns/converge.yml
```diff
@@ -19,7 +19,7 @@
   gather_facts: true
   vars:
     ansible_user_dir: "{{ lookup('env', 'HOME') }}"
-    cifmw_basedir: "/opt/basedir"
```
cjeanner, line 22: that shouldn't be part of the current change.

cjeanner: (this leads to the removal of all the logs - you really don't want that, do you?)

cjeanner (changes requested):

rebtoor:
recheck

rebtoor:
recheck

roles/libvirt_manager/tasks/create_vms.yml
rebtoor: makes sense, let me try.

roles/libvirt_manager/tasks/manage_vms.yml
```diff
@@ -65,28 +65,6 @@
...
-  ansible.builtin.shell:
-    executable: /bin/bash
-    cmd: |-
-      test -d /home/zuul && exit 0;
-      set -xe -o pipefail;
-      echo "{{ vm }}" | sudo tee /etc/hostname;
-      sudo hostname -F /etc/hostname;
-      sudo useradd -m -d /home/zuul zuul;
-      echo "zuul ALL=(ALL)  NOPASSWD: ALL" | sudo tee /etc/sudoers.d/zuul;
-      sudo -u zuul mkdir -p /home/zuul/.ssh /home/zuul/src/github.com/openstack-k8s-operators;
```
cjeanner, line 82: wondering if we shouldn't ensure the `src/...` tree exists via the cloud-init? Or we'll need to make it from some other place, such as the reproducer role. IIRC the `git` module creates the parent tree, but in case we're using something else such as `synchronize` or even plain `rsync` command, we might face an issue if the parent tree doesn't exist.

rebtoor: oh frack, i lost the `runcmd` in one of the commits...

cjeanner: Note for other reviewers: this directory tree is created in the reproducer role, via the push_code.yml - so we're all good.

cjeanner (approved):
/approve

roles/libvirt_manager/tasks/create_cloud_init_iso.yml
```diff
@@ -0,0 +1,51 @@
...
+        ignore_growroot_disabled: true
+        mode: growpart
+      resize_rootfs: noblock
+
+- name: "Call the config_drive role"
+  vars:
+    cifmw_config_drive_iso_image: "{{ _iso_path }}"
+    cifmw_config_drive_uuid: "{{ 99999999 | random(seed=vm) | to_uuid | lower }}"
+    cifmw_config_drive_hostname: "{{ vm }}"
+    cifmw_config_drive_networkconfig: "{{ vm_data.networkconfig | default(None) }}"
```
hjensas, line 48: If we are allowing to pass networkconfig, it needs to be possible to define separate config for each VM in the group? i.e if I have compute with amount 3 and want to pass static IPs to each of them? It looks like that won't work, i.e we'd end up setting the same IPs on each VM?

Can we change `networkconfig` to be a list? (Or optional - if it is dict apply same to all VMs - if it is list pick the config from the list using the loop index?

rebtoor: i should have fixed this @hjensas , PTAL when you can :)

hjensas: Thanks @rebtoor - looks good now!

hjensas (approved):

cjeanner:
/hold cancel

Per discussion within the Team last Friday, removing the hold that was placed ~2 weeks ago. There are enough positive reviews on it, and we need that feature.

PR #2221, opened 2024-08-13
Create a role for uploading results to Report portal

jirimacku:
The role is using Data router for uploading the test results into an Report portal instance.

### DIFF
--- a/roles/reportportal/README.md
+++ b/roles/reportportal/README.md
@@ -1,0 +1,33 @@
+# Report portal
+The `reportportal` role uses Data Router tool for uploading XML test results to Report portal. Access to the specific Report portal instance including URL and credentials has to be specified via Data router web interface.
+
+## Parameters
+* `cifmw_reportportal_basedir`: (String) Base directory. Defaults to `cifmw_basedir` which defaults to `~/ci-framework-data`.
+* `cifmw_reportportal_datarouter_url`: (String) URL with running Data router service (mandatory).
+* `cifmw_reportportal_droute_client_url`: (String) URL of Data router client repository (mandatory).
+* `cifmw_reportportal_datarouter_username`: (String) username for Data router client (mandatory).
+* `cifmw_reportportal_datarouter_password`: (String) password for Data router client (mandatory).
+* `cifmw_reportportal_datarouter_result_dir`: (String) Test results directory. Based on `cifmw_run_test_role` defaults to `~/ci-framework-data/tests/tempest/` or `~/ci-framework-data/tests/test_operator/`. One or more properly formatted xml results files are expected to be found in this directory.
+* `cifmw_reportportal_project`: (String) Report portal project for uploading results (mandatory).
+* `cifmw_reportportal_launch_name`: (String) Name of the Report portal launch defaults to `Dummy launch`.
+* `cifmw_reportportal_launch_description`: (String) Description of the Report portal launch defaults to `Test results sent via Data router`.
+* `cifmw_reportportal_droute_version`: (String) Data router client version defaults to `1.2.1`.
+* `cifmw_reportportal_droute_binary`: (String) Data router binary name defaults to `droute-linux-amd64`.
+
+## Examples
+```YAML
+- name: Play
+  hosts: localhost
+  vars:
+    cifmw_reportportal_droute_client_url: "https://example.com/data_router_client"
+    cifmw_reportportal_datarouter_url: "https://data-router.example.service.com"
+    cifmw_reportportal_datarouter_username: "<username>"
+    cifmw_reportportal_datarouter_password: "<password>"
+    cifmw_reportportal_project: "Example"
+
+  roles:
+    - reportportal
+```
+
+## Note
+This role exclusively relies on the internal Report portal and Data router instances and other internal repositories.
--- a/roles/reportportal/tasks/main.yaml
+++ b/roles/reportportal/tasks/main.yaml
@@ -1,0 +1,71 @@
+---
+# Copyright Red Hat, Inc.
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+- name: Check if mandatory variables are defined and not empty
+  ansible.builtin.assert:
+    that:
+      - cifmw_reportportal_datarouter_url is defined and cifmw_reportportal_datarouter_url | length > 0
+      - cifmw_reportportal_datarouter_username is defined and cifmw_reportportal_datarouter_username | length > 0
+      - cifmw_reportportal_datarouter_password is defined and cifmw_reportportal_datarouter_password | length > 0
+      - cifmw_reportportal_project is defined and cifmw_reportportal_project | length > 0
+      - cifmw_reportportal_droute_client_url is defined and cifmw_reportportal_droute_client_url | length > 0
+
+- name: Make sure the droute directory exists
+  ansible.builtin.file:
+    path: "{{ cifmw_reportportal_droute_dir }}"
+    state: directory
+    mode: "0755"
+
+- name: Get the Data router client
+  vars:
+    _droute_url: >-
+      {{
+        (cifmw_reportportal_droute_client_url,
+         cifmw_reportportal_droute_version,
+         cifmw_reportportal_droute_binary) |
+        path_join
+      }}
+  ansible.builtin.get_url:
+    url: "{{ _droute_url }}"
+    dest: "{{ cifmw_reportportal_droute_dir }}"
+    backup: true
+    mode: "0755"
+  when: not cifmw_reportportal_dry_run | bool
+
+- name: Create the metadata file
+  ansible.builtin.template:
+    src: metadata.json.j2
+    dest: "{{ cifmw_reportportal_metadata_file }}"
+    mode: "0755"
+
+- name: Send the results via Data router client call
+  ansible.builtin.command:
+    chdir: "{{ cifmw_reportportal_droute_dir }}"
+    cmd: >-
+      {{ cifmw_reportportal_droute_dir }}/{{ cifmw_reportportal_droute_binary }} send
+      --wait
+      --metadata={{ cifmw_reportportal_metadata_file }}
+      --results={{ cifmw_reportportal_results_search }}
+      --username={{ cifmw_reportportal_datarouter_username }}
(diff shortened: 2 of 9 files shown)


roles/reportportal/tasks/main.yaml
```diff
@@ -0,0 +1,79 @@
...
+# under the License.
+
+- name: Check if mandatory variables are defined and not empty
+  ansible.builtin.assert:
+    that:
+      - cifmw_reportportal_datarouter_url is defined and cifmw_reportportal_datarouter_url != ""
+      - cifmw_reportportal_datarouter_username is defined and cifmw_reportportal_datarouter_username != ""
+      - cifmw_reportportal_datarouter_password is defined and cifmw_reportportal_datarouter_password != ""
+      - cifmw_reportportal_project is defined and cifmw_reportportal_project != ""
+      - cifmw_reportportal_droute_client_url is defined and cifmw_reportportal_droute_client_url != ""
```
rebtoor, line 24: It's considered a best practice (especially for code clarity) to avoid comparing a string to "empty" and rather check for its length.

```suggestion
      - cifmw_reportportal_datarouter_url is defined and cifmw_reportportal_datarouter_url | length > 0
      - cifmw_reportportal_datarouter_username is defined and cifmw_reportportal_datarouter_username | length > 0
      - cifmw_reportportal_datarouter_password is defined and cifmw_reportportal_datarouter_password | length > 0
      - cifmw_reportportal_project is defined and cifmw_reportportal_project | length > 0
      - cifmw_reportportal_droute_client_url is defined and cifmw_reportportal_droute_client_url | length > 0
```

roles/reportportal/tasks/main.yaml
```diff
@@ -0,0 +1,79 @@
...
+        (cifmw_reportportal_droute_client_url,
+         cifmw_reportportal_droute_version,
+         cifmw_reportportal_droute_binary) |
+        path_join
+      }}
+
+- name: Make sure the droute directory exists
+  ansible.builtin.file:
+    path: "{{ cifmw_reportportal_droute_dir }}"
+    state: directory
```
rebtoor, line 39: You should define a default `mode` while you're creating files and/or directories.

roles/reportportal/tasks/main.yaml
```diff
@@ -0,0 +1,79 @@
...
+
+- name: Get the Data router client
+  ansible.builtin.get_url:
+    url: "{{ droute_url }}"
+    dest: "{{ cifmw_reportportal_droute_dir }}"
+    backup: true
+    mode: 0755
```
rebtoor, line 46: Mode should be declared as string (since it's an octal number)

```suggestion
    mode: "0755"
```

roles/reportportal/tasks/main.yaml
```diff
@@ -0,0 +1,79 @@
...
+      {{
+        (cifmw_reportportal_droute_dir,
+         'metadata.json') |
+        path_join
+      }}
+
+- name: Create the metadata file
+  ansible.builtin.template:
+    src: metadata.json.j2
+    dest: "{{ cifmw_reportportal_metadata_file }}"
```
rebtoor, line 60: You should define a default `mode` while you're creating file from a template.

roles/reportportal/tasks/main.yaml
```diff
@@ -0,0 +1,79 @@
...
+
+- name: Send the results via Data router client call
+  ansible.builtin.shell:
```
rebtoor, line 63: You aren't using any shell capability here, so you can safely switch to `command` module.

roles/reportportal/tasks/main.yaml
```diff
@@ -0,0 +1,79 @@
...
+      --results={{ cifmw_reportportal_results_search }}
+      --username={{ cifmw_reportportal_datarouter_username }}
+      --password={{ cifmw_reportportal_datarouter_password }}
+      --url={{ cifmw_reportportal_datarouter_url }}
+      --verbose
+  register: droute_result
+
+- name: Log droute result
+  ansible.builtin.debug:
+    var: droute_result
```
rebtoor, line 78: To avoid printing all fields from the register var, i'd filter for its stdout.

```suggestion
    var: droute_result.stdout
```

roles/reportportal/tasks/main.yaml
jirimacku: Done

roles/reportportal/tasks/main.yaml
jirimacku: Done

roles/reportportal/tasks/main.yaml
jirimacku: Done

roles/reportportal/tasks/main.yaml
jirimacku: Done

roles/reportportal/tasks/main.yaml
jirimacku: Done

roles/reportportal/tasks/main.yaml
jirimacku: Done

jirimacku:
Thank you for your comments @rebtoor !

rebtoor:
Role as-is can work and I will give my +2 to its inclusion in the framework once you solve a couple of things first:
- fix the errors raised by pre-commit
- include a (basic?) molecule test. You can test if you can download the client, pass a basic xml and maybe launch a command in dry-run

roles/reportportal/tasks/main.yaml
```diff
@@ -0,0 +1,81 @@
...
+        path_join
+      }}
+
+- name: Make sure the droute directory exists
+  ansible.builtin.file:
+    path: "{{ cifmw_reportportal_droute_dir }}"
+    state: directory
+    mode: "0755"
+
+- name: Get the Data router client
```
lewisdenny, line 42: We try avoid setting facts when we can. I would suggest this:
```suggestion
- name: Get the Data router client
  vars:
    _droute_url: >-
      {{
        (cifmw_reportportal_droute_client_url,
        cifmw_reportportal_droute_version,
        cifmw_reportportal_droute_binary) |
        path_join
      }}
```

This allows you to remove the `- name: Url for downloading data router client` task and the `_` prefix indicates `_droute_url` is internal and not for user consumption (and not documented in the readme etc.)

roles/reportportal/tasks/main.yaml
```diff
@@ -0,0 +1,81 @@
...
+  ansible.builtin.get_url:
+    url: "{{ droute_url }}"
+    dest: "{{ cifmw_reportportal_droute_dir }}"
+    backup: true
+    mode: "0755"
+
+- name: Path to metadata file
```
lewisdenny, line 49: As we are just setting a fact here this could just be defined in the defaults and this task removed

roles/reportportal/tasks/main.yaml
```diff
@@ -0,0 +1,81 @@
...
+      --wait
+      --metadata={{ cifmw_reportportal_metadata_file }}
+      --results={{ cifmw_reportportal_results_search }}
+      --username={{ cifmw_reportportal_datarouter_username }}
+      --password={{ cifmw_reportportal_datarouter_password }}
+      --url={{ cifmw_reportportal_datarouter_url }}
+      --verbose
+  register: droute_result
+
+- name: Log droute result
```
lewisdenny, line 78: [Suggestion]
How large is the output for this? if it's just a few lines this is fine, if it's 10+ I would log it to a file.

lewisdenny (changes requested):
Really nice role @jirimacku 

Just a few requests, let us know if you need a hand with the Molecule test.

roles/reportportal/README.md
```diff
@@ -0,0 +1,33 @@
+# Report portal
+The `reportportal` role uses Data Router tool for uploading XML test results to Report portal. Access to the specific Report portal instance including URL and credentials has to be specified via Data router web interface.
+
+## Parameters
+* `cifmw_reportportal_basedir`: (String) Base directory. Defaults to `cifmw_basedir` which defaults to `~/ci-framework-data`.
+* `cifmw_reportportal_datarouter_url`: (String) URL with running Data router service (mandatory).
+* `cifmw_reportportal_droute_client_url`: (String) URL of Data router client repository (mandatory).
+* `cifmw_reportportal_username`: (String) username for Data router client (mandatory).
```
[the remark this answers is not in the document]
jirimacku: Good catch @lewisdenny! Done.

roles/reportportal/tasks/main.yaml
jirimacku: Done.

roles/reportportal/tasks/main.yaml
jirimacku: Done.

roles/reportportal/tasks/main.yaml
lewisdenny: Great thank you @jirimacku I will review this once more tomorrow morning and resolve my threads.

roles/reportportal/tasks/main.yaml
pablintino: In fact, there's a linter rule for that, that I assume is not enabled if it's not blaming

rebtoor: correct, it's an opt-in rule

lewisdenny:
/hold

Adding hold until until Molecule test is added

jirimacku:
@rebtoor @lewisdenny I've just realized all the systems this role is talking to are in downstream (behind the FW). I'm not sure I'm able to prepare any reasonable molecule tests.

lewisdenny:
> @rebtoor @lewisdenny I've just realized all the systems this role is talking to are in downstream (behind the FW). I'm not sure I'm able to prepare any reasonable molecule tests.

What I suggest is:
- A simple Molecule test that calls your role providing any required vars for the role to run. (This will act as a working example)
- Add a new var called `cifmw_reportportal_dry_run` and set it to `false` by default and set it `true` in your molecule test before you call your role.
- Add a when clause to both `Get the Data router client` and `Send the results via Data router client` tasks to allow skipping them when `cifmw_reportportal_dry_run` is `true`. Example: `when: not cifmw_reportportal_dry_run | bool`. This allows us to skip any task that can't be tested due to unreachable 3rd party requirements as you explained. 

You can check the `mirror_registry` role for how simple the Molecule test can be.

Once you have your role running in Molecule, you can add some tests you see fit, for example; providing a rendered copy of the `metadata.json` file and comparing that to the copy generated by the molecule run of your role.

jirimacku:
recheck

jirimacku:
@rebtoor @lewisdenny I pushed some basic testing. Please check it out. Thanks

lewisdenny:
> @rebtoor @lewisdenny I pushed some basic testing. Please check it out. Thanks

Just need to run `make role_molecule`, you created your molecule job but this command renders the config for zuul to run the job. As you can see on this PR, your molecule job isn't running :)

jirimacku:
@rebtoor @lewisdenny the molecule test are running.

jirimacku:
> /hold
> 
> Adding hold until until Molecule test is added

Molecule test added and passed.

lewisdenny (approved):
/lgtm

Thank you @jirimacku, great work on this role and thank you for the contribution!

PR #2222, opened 2024-08-13
Ensure we don't have shortnames in the DNS

cjeanner:
Without the dhcp-fqdn option, dnsmasq might push shortnames in the DNS,
leading to some weird situation with the resolution of such names.

Before that change, "dig compute-0" might resolve to the public/ocpbm
network, or, later, on the osp_trunk/ctlplane network.

This is because, in the DHCP configuration, we use only shortnames,
meaning we get "compute-0" twice, with different IP/networks.
Then, it depends on the order of the resolvers in /etc/resolv.conf -
order that might change upon DHCP lease updates.

With this patch, we ensure no shortnames are injected, leading to a more
stable DNS resolution, while it can also lead to slower resolution,
since the request might have to go through multiple resolvers to get an
answer.

### DIFF
--- a/roles/reproducer/tasks/configure_crc.yml
+++ b/roles/reproducer/tasks/configure_crc.yml
@@ -115,6 +115,7 @@
       expand-hosts
       local=/testing/
       domain=crc.testing
+      dhcp-fqdn
 
 - name: Flush handlers
   ansible.builtin.meta: flush_handlers
--- a/roles/reproducer/tasks/ocp_layout.yml
+++ b/roles/reproducer/tasks/ocp_layout.yml
@@ -138,6 +138,7 @@
           {% for ip in _api_addr %}
           ptr-record={{ ip | ansible.utils.ipaddr('revdns') }},api-int.{{ _cluster }}.{{ _domain }}
           {% endfor %}
+          dhcp-fqdn
         validate: "/usr/sbin/dnsmasq -C %s --test"
 
 - name: Ensure services are reloaded


cjeanner:
/override toggle-label

This check is still wrong ;_; cc @lewisdenny

cjeanner:
/hold

there are some planed tests for that one - and since I'll be off a couple of days, I'd rather avoid seeing it merged until I'm back.

hjensas (approved):

raukadah:
Test results from VA HCI trigger job: https://review.rdoproject.org/r/c/tripleo-downstream-trigger-nested-virt/+/54154/5#message-3d8682beb0303c295d035f24e6df13b2494eaaad
```
ci-framework-reserve-static-node-va-hci https://sf.hosted.upshift.rdu2.redhat.com/zuul/t/tripleo-ci-internal/build/5aab3c969490457a91e490fc48fe6545 : SUCCESS in 2h 49m 29s

ci-framework-baremetal-static-node-rhel9-va-hci https://sf.hosted.upshift.rdu2.redhat.com/zuul/t/tripleo-ci-internal/build/a774960113e64a6ea4134ef23de2e00b : SUCCESS in 2h 49m 09s
```

lewisdenny (commented):
Hi @cjeanner, this change looks good and past d/s testing a while ago. 

Is there a reason this didn't merge?

lewisdenny:
Closing as stale, please reopen if needed.

PR #2223, opened 2024-08-13
Inject FQDN in the DHCP IP reservation

cjeanner:
Until now, we could face a clash in the way dnsmasq re-injects its DHCP
nodes into its DNS:
with the previous state of the code, we ended with, for instance,
"compute-0" name associated to two IP, one in the ocpbm/public network,
and one on the osp_trunk (ctlplane) network. DNS resolution was slightly
off and, at least, unreliable.

This change, in addition to the addition of the `dhcp-fqdn` option in


hjensas (commented):
/lgtm

PR #2224, opened 2024-08-13
Enable unique names for certain types of VMs

cjeanner:
Unique names are needed for some specific usecases, but not for all VMs.

This patch ensures anything that's not OCP/CRC nor controller VM gets a
unique, random string in their hostname.

We also converge names, ensuring no underscore (`_`) are present in the VM
names - we replace them with dashes (`-`) to ensure it's a valid hostname.
This leads to a modification in the way we build the VM listing, it's
now a dict associating the name to the type, in order to make things
"slightly" easier to handle later.

This patch introduces a new top-level parameter, `cifmw_run_id`, that
can be either passed down by the user, or generated using the
random_string lookup from the community.general collection.

Allowing to set the custom ID would allow, for instance, to use the Zuul
job ID as the identifier, making things slightly easier to follow
across the various runs.

- [X] va-hci (local testing)
- [x] uni01alpha (via ds tp 609)
- [x] uni02beta (via ds tp 609)
- [ ] ~~uni04delta (via ds tp 609)~~ (failed due to unrelated issues, need a fix at dev-scripts level)
- [ ] BGP/spine-leaf (via ds tp 608) ("should" work - but TP was using midstream in the controller-0, leading to issues)
- [x] NFV (via ds tp 611)

### DIFF
--- a/docs/source/usage/01_usage.md
+++ b/docs/source/usage/01_usage.md
@@ -71,6 +71,8 @@ are shared among multiple roles:
 - `cifmw_crc_default_network`: (String) name of the untagged network used to address DNS on the crc node. Default is `default`.
 - `cifmw_run_operators_compliance_scans`: (Bool) Specifies whether to run operator compliance scans.  Defaults to `false`.
 - `cifmw_run_compute_compliance_scans`: (Bool) Specifies whether to run compliance scans on the first compute.  Defaults to `false`.
+- `cifmw_run_id`: (String) CI Framework run identifier. This is used in libvirt_manager, to add some uniqueness to some types of virtual machines (anything that's not OCP, CRC nor controller).
+  If not set, the Framework will generate a random string for you, and store it on the target host, in `{{ cifmw_basedir }}/artifacts/run-id`
 
 ```{admonition} Words of caution
 :class: danger
--- a/reproducer-clean.yml
+++ b/reproducer-clean.yml
@@ -33,7 +33,7 @@
         name: nat64_appliance
         tasks_from: cleanup.yml
 
-    - name: Remove remote selected data directories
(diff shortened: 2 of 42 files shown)


docs/source/usage/01_usage.md
```diff
@@ -68,6 +68,7 @@ are shared among multiple roles:
 - `cifmw_nolog`: (Bool) Toggle `no_log` value for selected tasks. Defaults to `true` (hiding those logs by default).
 - `cifmw_parent_scenario`: (String or List(String)) path to existing scenario/parameter file to inherit from.
 - `cifmw_configure_switches`: (Bool) Specifies whether switches should be configured. Computes in `reproducer.yml` playbook. Defaults to `false`.
+- `cifmw_run_id`: (String) CI Framework run identifier. This is used in libvirt_manager, to add some uniqueness to some times of virtual machines (anything that's not OCP, CRC nor controller). If not set, the Framework will generate a random string for you, and store it on the target host, in `{{ cifmw_basedir }}/artifacts/run-id`
```
eduolivares, line 71: `to some times of virtual machines`
you mean `to some types of virtual machines`, right?

cjeanner: oh, right. Damn. Do I need more time to type in correctly? :rofl:

lewisdenny:
`build-push-container-cifmw-client` failed due to unrelated issue to this patch. 

Not rechecking as jobs are still running but next run `build-push-container-cifmw-client` should not fail.

lewisdenny:
/approve

I'm happy with this patch and have reviewed it best I can for the size.

Once we have another +, downstream testing done and molecule passing, I think best we can do is merge this.

PR #2226, opened 2024-08-14
Allow to chose which network is the default gateway

cjeanner:
The new parameter allows to provide a single network, or a list of nets,
to use as "default gateway".

By default, the Framework uses the "public network" (usually "public" or
"ocpbm"), but it doesn't properly work with DCN layout, where computes
might need some other routing table content.


sbekkerm:
/lgtm

Tested. It allows setting the default gateway network

PR #2230, opened 2024-08-15
Add cifmw client container

lewisdenny:
This patch adds a new containerfile defining the container we will use with the upcoming CLI, maybe in CI jobs soon as well.

With every merge into the ci-framework repo the container will be build using the ref content and tagged with the Git SHA. If the PR is merged into the main branch the "latest" tag will be updated.


.github/workflows/build-and-push-client-container.yml
```diff
@@ -0,0 +1,49 @@
+name: Build and Push Image
+on: [ push ]
```
lewisdenny, line 2: So a new container will only be pushed when code is merged, but do we want/need a container PR revision? I don't so, not initially at least

lewisdenny:
/hold
Holding for now, need to check if GitHub action is the right direction, if not maybe OpenShift CI, if not create tiny Zuul job to build and push container.

lewisdenny:
recheck

lewisdenny:
recheck

lewisdenny:
recheck

lewisdenny:
recheck

lewisdenny:
recheck

lewisdenny:
recheck

lewisdenny:
recheck

lewisdenny:
recheck

lewisdenny:
recheck

lewisdenny:
recheck

lewisdenny:
recheck

containerfiles/Containerfile.client
```diff
@@ -0,0 +1,35 @@
...
+
+RUN /usr/bin/make -C /home/cifmw/ci-framework setup_molecule USE_VENV=yes
+
+# todo(Lewis): Possible enhancements:
+# - Add pre-commit install so more dependencies are bundled
+#     RUN /home/cifmw/test-python/bin/pre-commit install --install-hooks
+# - Add entrypoint script to source python venv
+# - Use ansible-dev-environment
+#     https://github.com/ansible/ansible-dev-environment
+#     RUN ade install --venv ~/test-python/ -r requirements.yml
```
rebtoor, line 32: ❤️

lewisdenny:
recheck

lewisdenny:
recheck

lewisdenny:
recheck

ci/playbooks/build_push_container_runner.yml
```diff
@@ -0,0 +1,13 @@
+---
+- name: Run build_push_container role
+  hosts: "{{ cifmw_zuul_target_host | default('controller') }}"
+  tasks:
+    - name: Run build_push_container role
+      vars:
+        ci_framework_src_dir: "{{ ansible_user_dir }}/src/github.com/openstack-k8s-operators/ci-framework"
+      ansible.builtin.command:
```
cjeanner, line 8: with long commands like that, I'd use `ci_script`. That would also generate the reproducer script, allowing to track potential issues such as "oh, we delegate to a host, and the ansible_user_dir is different" and related potential headaches.

cjeanner: (IF ci_script is available at that level...)

lewisdenny: Nope it's not, even when asking Zuul to pull the role to the executor it doesn't work. 

Maybe there is a way to make it work but now my reasons against have stacked up:

- It makes the stupid runner playbook more complex, I need to create directories etc. 
- It's not how the other playbooks that are used to call nested playbooks are written so goes against the standard. 
- ci_script is not available for the Zuul executor and we want to run as little as possible in that space for the same reasons you mention, reproducing and tracking down issues with node holds etc.

cjeanner: Makes sense - I wasn't sure about the ci_script availability ;).

rebtoor (approved):
/lgtm

PR #2232, opened 2024-08-15
Remove more skipped ansible-lint rules

lewisdenny:
This patch removes the following skipped ansible-lint rules:
- schema[meta]
- schema[vars]

schema[meta] is fixed by removing the galaxy_info.platforms section. This was not maintained and has now been removed from Ansible Galaxy and soon to be removed from ansible-lint.[1]

schema[vars] was fixed by changing `roles/openshift_provisioner_node/molecule/defaults` to `roles/openshift_provisioner_node/molecule/default`, this was confusing ansible-lint

This patch also adds the following new opt-in rules as these are things we would like to avoid:
- empty-string-compare
- risky-shell-pipe

[1] https://forum.ansible.com/t/role-meta-data-platform-key-broken/4651/10


rebtoor (approved):
/approve

PR #2234, opened 2024-08-16
Update precommit hooks and add shellcheck

lewisdenny:
This patch adds a new precommit hook that calls shellcheck [1]

The hook is written in a way that all findings are reported but only
errors cause failure.

In follow up patches the excluded rule will be removed and the severity
lowered, for now this will protect us from new patches with erroneous shell scripts.

[1] https://github.com/koalaman/shellcheck


rebtoor (approved):
/approve

PR #2235, opened 2024-08-16
[Reproducer] Add support for disabling ocp overlays

lewisdenny:
This patch adds a new top level variable `cifmw_use_ocp_overlay` used to control whether the OCP VMs created by dev_scripts via the reproducer role use image overlays or not.

Prior to this patch the use of image overlays was not configurable, while image overlays are useful for local deployments, in CI they are not used. The downside to image overlays is the process to create them takes a very long time.

The OCP cluster needs come up and stabilize, then we shut the VMs down and copy the images to use as base images, we then need to start the OCP VMs again and wait for it to stabilize, this is all wasted time because in CI environments we start fresh every time.

This patch also removes the undocumented internal variable `cifmw_use_reproducer` and it's use replaced with
`cifmw_use_ocp_overlay`

Jira: https://issues.redhat.com/browse/OSPRH-7771

---

- [x] Test deployment with 3 masters
- [x] Test a uni* job - Testing downstream in TP/501
- [x] Update and verify cleanup

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/2320

### DIFF
--- a/docs/source/usage/01_usage.md
+++ b/docs/source/usage/01_usage.md
@@ -66,6 +66,7 @@ are shared among multiple roles:
 - `cifmw_nolog`: (Bool) Toggle `no_log` value for selected tasks. Defaults to `true` (hiding those logs by default).
 - `cifmw_parent_scenario`: (String or List(String)) path to existing scenario/parameter file to inherit from.
 - `cifmw_configure_switches`: (Bool) Specifies whether switches should be configured. Computes in `reproducer.yml` playbook. Defaults to `false`.
+- `cifmw_use_ocp_overlay`: (Boolean) Specifies whether OCP nodes deployed via devscripts should use overlay images. Using overlay images speeds up the redeployment when using the reproducer role locally but in CI each job is cleaned up and redeployed. Creating the overlay image takes time so should be disabled when not used. Defaults to `true`.
 - `cifmw_crc_default_network`: (String) name of the untagged network used to address DNS on the crc node. Default is `default`.
 - `cifmw_run_operators_compliance_scans`: (Bool) Specifies whether to run operator compliance scans.  Defaults to `false`.
 - `cifmw_run_compute_compliance_scans`: (Bool) Specifies whether to run compliance scans on the first compute.  Defaults to `false`.
--- a/roles/devscripts/tasks/300_post.yml
+++ b/roles/devscripts/tasks/300_post.yml
@@ -26,42 +26,10 @@
     - not cifmw_devscripts_ocp_online | bool
   ansible.builtin.import_tasks: set_cluster_fact.yml
 
-- name: Prepare for disk overlay configuration.
-  tags:
-    - devscripts_deploy
+- name: Prepare for disk overlay configuration
   when:
     - not cifmw_devscripts_ocp_comply | bool
+    - cifmw_use_ocp_overlay | default(true) | bool
+  tags:
+    - devscripts_deploy
   ansible.builtin.include_tasks: 310_prepare_overlay.yml
-
-- name: Bringing cluster online.
-  when:
-    - not cifmw_use_reproducer | default(false) | bool
-  block:
-    - name: Deploy layout on target host
-      tags:
-        - libvirt_layout
-      when:
-        - not cifmw_devscripts_ocp_comply | bool
-      ansible.builtin.include_role:
-        name: libvirt_manager
-        tasks_from: deploy_layout
-
-    - name: Apply VLAN configuration for vnet interfaces.
-      tags:
-        - devscripts_post
-      when:
-        - cifmw_libvirt_manager_configuration_gen.networks is defined
-      become: true
-      cifmw.general.bridge_vlan:
-        networks: >-
-          {{
-            cifmw_libvirt_manager_configuration_gen.networks.keys() | list
-          }}
-      failed_when: false
-
-    - name: Ensure the OpenShift cluster is accessible.
-      tags:
-        - devscripts_post
-      when:
-        - not cifmw_devscripts_ocp_online | bool
-      ansible.builtin.include_tasks: 330_wait_ocp.yml
--- a/roles/libvirt_manager/tasks/create_vms.yml
+++ b/roles/libvirt_manager/tasks/create_vms.yml
@@ -29,12 +29,12 @@
     _workload: "{{ cifmw_libvirt_manager_basedir }}/workload"
     _img_dir: "{{ cifmw_libvirt_manager_basedir }}/images"
     _chdir: >-
-      {{ (is_base_img | default(false) | bool) | ternary(_img_dir, _workload) }}
+      {{
+        (_is_base_img | default(false) | bool) |
+        ansible.builtin.ternary(_img_dir, _workload)
+      }}
   block:
     - name: "Create VM image for {{ vm }}"
-      vars:
-        _vm_img: >-
-          {{ vm }}.qcow2
       ansible.builtin.command:
         cmd: >-
           qemu-img create
@@ -48,9 +48,6 @@
         chdir: "{{ _chdir }}"
 
     - name: "Ensure file ownership and rights for {{ vm }}"
-      vars:
-        _vm_img: >-
-          {{ vm }}.qcow2
       ansible.builtin.file:
         path: "{{ (_chdir, _vm_img) | path_join }}"
         group: "qemu"
--- a/roles/libvirt_manager/tasks/deploy_layout.yml
+++ b/roles/libvirt_manager/tasks/deploy_layout.yml
@@ -175,6 +175,8 @@
       }}
     pub_key: "{{ pub_ssh_key.content | b64decode }}"
     priv_key: "{{ priv_ssh_key.content | b64decode }}"
+    _vm_img: >-
+      {{ vm }}.qcow2
   ansible.builtin.include_tasks:
     file: create_vms.yml
   loop: "{{ cifmw_libvirt_manager_all_vms | dict2items }}"
--- a/roles/libvirt_manager/tasks/ocp_layout.yml
+++ b/roles/libvirt_manager/tasks/ocp_layout.yml
@@ -6,7 +6,7 @@
   notify: Restart firewalld
   ansible.builtin.command:
     cmd: >-
-       firewall-cmd --permanent --zone libvirt --add-forward
+      firewall-cmd --permanent --zone libvirt --add-forward
 
 - name: Enable masquerading for public traffic
   when: cifmw_libvirt_manager_firewalld_default_zone_masquerade | default(true) | bool
@@ -64,6 +64,7 @@
         state: directory
       loop:
         - "{{ cifmw_libvirt_manager_basedir }}/images"
+        - "{{ cifmw_libvirt_manager_basedir }}/workload"
         - "{{ cifmw_libvirt_manager_ocp_pool_dir }}"
 
     - name: Create pool in libvirt
@@ -79,7 +80,9 @@
           {{
             _ocp_layout.vms[vm_type]
           }}
-        is_base_img: true
+        _vm_img: >-
+          {{ vm }}.qcow2
+        _is_base_img: "{{ cifmw_use_ocp_overlay | default(true) | bool }}"
       ansible.builtin.include_role:
         name: "libvirt_manager"
         tasks_from: "create_vms.yml"
--- a/roles/reproducer/tasks/ocp_layout.yml
+++ b/roles/reproducer/tasks/ocp_layout.yml
@@ -336,8 +336,6 @@
         permanent: true
 
     - name: Run devscripts role
-      vars:
-        cifmw_use_reproducer: true
       ansible.builtin.include_role:
         name: devscripts
 


pablintino:
@lewisdenny do not forget to test the cleanup, as the one we have may be not prepare to an scenario without the overlays.

lewisdenny:
> @lewisdenny do not forget to test the cleanup, as the one we have may be not prepare to an scenario without the overlays.

Thank you @pablintino , I hadn't considered that. I will add that to my check list.

roles/devscripts/tasks/300_post.yml
```diff
@@ -31,11 +31,12 @@
...
   when:
     - not cifmw_devscripts_ocp_comply | bool
+    - cifmw_use_ocp_overlay | default(false) | bool
   ansible.builtin.include_tasks: 310_prepare_overlay.yml
 
 - name: Bringing cluster online.
   when:
-    - not cifmw_use_reproducer | default(false) | bool
+    - cifmw_use_ocp_overlay | default(false) | bool
   block:
```
cjeanner, line 40: no need for a block if you have only one task nested there

docs/source/usage/01_usage.md
```diff
@@ -68,6 +68,7 @@ are shared among multiple roles:
 - `cifmw_nolog`: (Bool) Toggle `no_log` value for selected tasks. Defaults to `true` (hiding those logs by default).
 - `cifmw_parent_scenario`: (String or List(String)) path to existing scenario/parameter file to inherit from.
 - `cifmw_configure_switches`: (Bool) Specifies whether switches should be configured. Computes in `reproducer.yml` playbook. Defaults to `false`.
+- `cifmw_use_ocp_overlay`: (Boolean) Specifies whether devscripts deployed OCP nodes should use overlay images. Defaults to `false`.
```
cjeanner, line 71: I'd rather NOT change the default behaviour. QE and other ppl not wanting the overlay should really make it clear in their env (i.e. toggling that boolean).

roles/libvirt_manager/tasks/create_vms.yml
```diff
@@ -21,17 +21,20 @@
...
-          {{ vm }}.qcow2
+      # when:
+      #   - (_is_ocp_vm and cifmw_use_ocp_overlay | default(false) | bool) or
+      #     not _is_ocp_vm
       ansible.builtin.command:
         cmd: >-
           qemu-img create
           {% if vm_data.disk_file_name != 'blank' %}
-          -o backing_file={{ _img }},backing_fmt=qcow2
+            {% if (_is_ocp_vm and cifmw_use_ocp_overlay | default(false) | bool) or not _is_ocp_vm -%}
```
cjeanner, line 35: I don't think this is needed. the `creates: {{ _vm_img }}` should do it properly already. What you want is to ensure the "overlay-less" OCP cluster is deployed so that the `_vm_img` matches. To do that, I think you should toggle the value of `_is_base_img` set in the `ocp_layout.yml` (in libvirt_manager role iirc).

roles/libvirt_manager/tasks/deploy_layout.yml
```diff
@@ -226,13 +270,23 @@
...
         _cifmw_libvirt_manager_layout.vms[vm_type]
       }}
     pub_key: "{{ pub_ssh_key.content | b64decode }}"
     priv_key: "{{ priv_ssh_key.content | b64decode }}"
+    # If $vm is an OCP master node, capture it's index
+    # and craft the _vm_img varible, examples:
+    # ocp-0 -> ocp_master-0.qcow (match and modify)
+    # ocp_worker-0 -> ocp_worker-0.qcow (pass)
+    # compute-1 -> compute-1.qcow (pass)
+    _vm_img: "{{ vm | regex_search(_ocp_master_matcher) | ternary(vm | regex_replace(_ocp_master_matcher, 'ocp_master-\\1.qcow2'), vm ~ '.qcow2') }}"
```
cjeanner, line 289: Not sure this is really needed? the disk_image_name is supposed to be `ocp_master` already. So I'd expect it to be "easy" to build onto that... ? Extracting the ID is needed indeed - but for the rest... I might be easier.

roles/libvirt_manager/tasks/ocp_layout.yml
```diff
@@ -30,7 +30,7 @@
       disk_file_name: "blank"
     _ocp_layout:
       vms:
-        ocp_master: >-
+        ocp: >-
```
cjeanner, line 33: ok, so that will create "ocp" instead of "ocp_master" I think. I'm probably missing some brain cells to grasp the whole overview.

roles/libvirt_manager/tasks/ocp_layout.yml
```diff
@@ -61,26 +61,80 @@
...
+        vm_data: >-
+          {{
+            _ocp_layout.vms[vm_type]
+          }}
+        # If $vm is an OCP master node, capture it's index
+        # and craft the _vm_img varible, examples:
+        # ocp-0 -> ocp_master-0.qcow (match)
+        # ocp_worker-0 -> ocp_worker-0.qcow (pass)
+        # compute-1 -> compute-1.qcow (pass)
+        _vm_img: "{{ vm | regex_search(_ocp_master_matcher) | ternary(vm | regex_replace(_ocp_master_matcher, 'ocp_master-\\1.qcow2'), vm ~ '.qcow2') }}"
```
cjeanner, line 102: hm... imho it should try to match the disk_file_name instead, and build the expected value. WDYT?

docs/source/usage/01_usage.md
lewisdenny: Done

roles/devscripts/tasks/300_post.yml
lewisdenny: Done

roles/libvirt_manager/tasks/create_vms.yml
lewisdenny: Done

roles/libvirt_manager/tasks/ocp_layout.yml
lewisdenny: Yep correct, this makes it so the VMs deployed via devscripts are the same name as the ones deployed from the overlay images. 

We only use one or the other at one time so there is no overlap and later code uses the vm name for things so it needs to be the same.

roles/libvirt_manager/tasks/deploy_layout.yml
```diff
@@ -169,12 +169,28 @@
...
+      }}
     vm_data: >-
       {{
         _cifmw_libvirt_manager_layout.vms[vm_type]
       }}
     pub_key: "{{ pub_ssh_key.content | b64decode }}"
     priv_key: "{{ priv_ssh_key.content | b64decode }}"
+    # If $vm is an OCP master node, capture it's index
+    # and craft the _vm_img varible, examples:
+    # ocp-0 -> ocp_master-0.qcow (match and modify)
```
cjeanner, line 189: `ocp-(master|worker)-ID.qcow2` actually, with the last change converging all names to use hyphens instead of underscores.

roles/libvirt_manager/tasks/ocp_layout.yml
```diff
@@ -73,13 +74,30 @@
 
     - name: Create blank images for OCP cluster resources
       vars:
+        _matcher: >-
```
cjeanner, line 77: if we could avoid injecting that again - afaik there's no use for it anymore.

roles/libvirt_manager/tasks/ocp_layout.yml
```diff
@@ -73,13 +74,30 @@
...
+          {{
+            '(cifmw-)?' ~
+            '([a-z0-9]+)-[0-9]+'
+          }}
+        _ocp_master_matcher: >-
```
cjeanner, line 82: not sure this is needed either

roles/libvirt_manager/tasks/ocp_layout.yml
```diff
@@ -73,13 +74,30 @@
...
           {{
             _ocp_layout.vms[vm_type]
           }}
-        is_base_img: true
+        # If $vm is an OCP master node, capture it's index
+        # and craft the _vm_img varible, examples:
+        # ocp-0 -> ocp_master-0.qcow (match)
+        # ocp_worker-0 -> ocp_worker-0.qcow (pass)
+        # compute-1 -> compute-1.qcow (pass)
+        # _vm_img: "{{ vm | regex_search(_ocp_master_matcher) | ternary(vm | regex_replace(_ocp_master_matcher, 'ocp_master-\\1.qcow2'), vm ~ '.qcow2') }}"
```
cjeanner, line 97: image name should be converged to use hyphens, not dashes.

roles/libvirt_manager/templates/domain.xml.j2
```diff
@@ -29,7 +29,7 @@
     <emulator>/usr/libexec/qemu-kvm</emulator>
     <disk type='file' device='disk'>
       <driver name='qemu' type='qcow2'/>
-      <source file='{{ _chdir }}/{{ vm }}.qcow2'/>
+      <source file='{{ _chdir }}/{{ _vm_img }}'/>
```
cjeanner, line 32: I'd keep the `vm` there, really.

cjeanner (changes requested):
There's the mix with underscore vs hyphens that should be converged to hyphens only.

roles/libvirt_manager/tasks/deploy_layout.yml
lewisdenny: Removed

roles/libvirt_manager/tasks/ocp_layout.yml
lewisdenny: Removed

roles/libvirt_manager/tasks/ocp_layout.yml
lewisdenny: Removed

roles/libvirt_manager/tasks/ocp_layout.yml
lewisdenny: Removed

roles/libvirt_manager/templates/domain.xml.j2
lewisdenny: Done

lewisdenny:
recheck

lewisdenny:
recheck

lewisdenny:
recheck

https://softwarefactory-project.io/zuul/t/rdoproject.org/build/b19c21ee086a428cbf8685c49540af76

lewisdenny:
recheck

lewisdenny:
stdout: '2024-09-05 19:16:31,182 - repo-setup-get-hash - ERROR - Invalid response received from the delorean server. Queried URL: https://trunk.rdoproject.org/centos9-master/current-podified/delorean.repo.md5. Response code: -1. Response text: The read operation timed out. Failed to create HashInfo object.'

lewisdenny:
recheck

-_-

lewisdenny:
recheck

roles/libvirt_manager/tasks/create_vms.yml
```diff
@@ -29,12 +29,9 @@
     _workload: "{{ cifmw_libvirt_manager_basedir }}/workload"
     _img_dir: "{{ cifmw_libvirt_manager_basedir }}/images"
     _chdir: >-
-      {{ (is_base_img | default(false) | bool) | ternary(_img_dir, _workload) }}
+      {{ (_is_base_img | default(false) | bool) | ansible.builtin.ternary(_img_dir, _workload) }}
```
cjeanner, line 32:
```suggestion
      {{
           (_is_base_img | default(false) | bool) |
           ansible.builtin.ternary(_img_dir, _workload)
       }}
```

Shorter line :)

roles/libvirt_manager/tasks/ocp_layout.yml
```diff
@@ -79,7 +80,9 @@
           {{
             _ocp_layout.vms[vm_type]
           }}
-        is_base_img: true
+        _vm_img: >-
```
cjeanner, line 83: ah, I was wondering where it was. Makes sense, cleaner!

roles/libvirt_manager/tasks/create_vms.yml
```diff
@@ -48,9 +45,7 @@
         chdir: "{{ _chdir }}"
 
     - name: "Ensure file ownership and rights for {{ vm }}"
-      vars:
-        _vm_img: >-
-          {{ vm }}.qcow2
+      become: true
```
cjeanner, line 48: hmmm isn't that task nested under `name: "Create VM overlays or images for {{ vm_type }}"` block, meaning we have the `become: true` already?

roles/devscripts/tasks/300_post.yml
```diff
@@ -26,42 +26,39 @@
...
-  tags:
-    - devscripts_deploy
+- name: Handle OCP disk overlay
   when:
     - not cifmw_devscripts_ocp_comply | bool
-  ansible.builtin.include_tasks: 310_prepare_overlay.yml
-
-- name: Bringing cluster online.
-  when:
-    - not cifmw_use_reproducer | default(false) | bool
```
cjeanner, line 38: ok, beware, this cifmw_use_reproducer was defaulting to "false". And toggled to "true" in the reproducer/tasks/ocp_layout.yml.

Here you're enabling those tasks bellow "by default", while it shouldn't be the case imho.

Since the condition is removed, you should probably remove those tasks instead.

cjeanner (changes requested):

roles/libvirt_manager/tasks/deploy_layout.yml
```diff
@@ -175,6 +175,8 @@
       }}
     pub_key: "{{ pub_ssh_key.content | b64decode }}"
     priv_key: "{{ priv_ssh_key.content | b64decode }}"
+    _vm_img: >-
+      {{ vm }}.qcow2
```
rebtoor, line 179:
```suggestion
    _vm_img: {{ vm }}.qcow2
```
why don't use this simpler form?

roles/libvirt_manager/tasks/ocp_layout.yml
```diff
@@ -79,7 +80,9 @@
...
+          {{ vm }}.qcow2
```
rebtoor, line 84:
```suggestion
        _vm_img: "{{ vm }}.qcow2"
```

rebtoor (commented):
I'm totally ok with this PR, just left a couple of comments for my curiosity :D

/approve

roles/libvirt_manager/tasks/deploy_layout.yml
lewisdenny: Copy Paste, out of interested I checked the blame and historically the line was longer[1]

Then it was pushed to two lines to not be so long, then it was shortened without going back to a single line and then I Copy/Pasted it without thinking to change to a single line :P

[1] https://github.com/openstack-k8s-operators/ci-framework/pull/1151

PR #2236, opened 2024-08-16
Optimize ceph conf overrides

katarimanojk:
This patch will optimize the code by applying all conf at once using the asimilate conf file instead of many config commands.

JIRA: https://issues.redhat.com/browse/OSPRH-7427

### DIFF
--- a/roles/cifmw_cephadm/tasks/pre.yml
+++ b/roles/cifmw_cephadm/tasks/pre.yml
@@ -144,18 +144,3 @@
         mode: '0644'
       become: true
       when: stat_bootstrap_conf.stat.exists
-
-- name: Generate cifmw_cephadm_assimilate_conf on bootstrap node
-  when:
-    - cifmw_cephadm_bootstrap_conf is not defined or
-      cifmw_cephadm_bootstrap_conf | length == 0
-  become: true
-  action: config_template
-  args:
-    src: "ceph.conf.j2"
-    dest: "{{ cifmw_cephadm_assimilate_conf }}"
-    owner: "{{ cifmw_cephadm_uid }}"
-    group: "{{ cifmw_cephadm_uid }}"
-    mode: "0644"
-    config_overrides: "{{ cifmw_cephadm_conf_overrides }}"
-    config_type: ini
--- a/roles/cifmw_cephadm/tasks/rgw.yml
+++ b/roles/cifmw_cephadm/tasks/rgw.yml
@@ -74,12 +74,20 @@
     mount_spec: true
     cifmw_cephadm_spec: "{{ cifmw_ceph_rgw_spec_path }}"
 
+- name: Generate cifmw_cephadm_assimilate_conf for ceph rgw
+  become: true
+  ansible.builtin.template:
+    src: "ceph_rgw.conf.j2"
+    dest: "{{ cifmw_cephadm_assimilate_conf }}"
+    owner: "{{ cifmw_cephadm_uid }}"
+    group: "{{ cifmw_cephadm_uid }}"
+    mode: "0644"
+
 - name: Apply ceph rgw keystone config
   ansible.builtin.command: |
-    {{ cifmw_cephadm_ceph_cli }} config set global {{ item.key }} {{ item.value }}
+    {{ cifmw_cephadm_ceph_cli }} config assimilate-conf -i {{ cifmw_cephadm_assimilate_conf_container }}
   changed_when: false
   become: true
-  loop: "{{ lookup('ansible.builtin.dict', cifmw_ceph_rgw_config) }}"
 
 - name: Apply ceph config-key set ssl_option
   ansible.builtin.command: |
--- a/roles/cifmw_cephadm/templates/ceph_rgw.conf.j2
+++ b/roles/cifmw_cephadm/templates/ceph_rgw.conf.j2
@@ -14,7 +14,7 @@ fsid = {{ cifmw_cephadm_fsid }}
 mon host = {{ cifmw_cephadm_first_mon_ip }}
 
 {# Build CephConfigOverrides #}
-{% for key, value in (ceph_conf_overrides | default({})).items() %}
+{% for key, value in (cifmw_ceph_rgw_config | default({})).items() %}
 {%   if not (['global', 'mon', 'mgr', 'osd', 'mds', 'client']) | intersect([key]) %}
 {{ key }} = {{ value }}
 {%   elif (['global']) | intersect([key]) %}
@@ -24,7 +24,7 @@ mon host = {{ cifmw_cephadm_first_mon_ip }}
 {% endfor %}
 
 {# Build the remaining sections #}
-{% for key, value in (ceph_conf_overrides | default({})).items() %}
+{% for key, value in (cifmw_ceph_rgw_config | default({})).items() %}
 {%   if (['mon', 'mgr', 'osd', 'mds', 'client']) | intersect([key]) %}
 [{{ key }}]
 {# Render the current section #}


playbooks/ceph.yml
```diff
@@ -357,6 +357,8 @@
       ansible.builtin.import_role:
         name: cifmw_cephadm
         tasks_from: pre
+      vars:
+        cifmw_cephadm_bootstrap_conf: ""
```
katarimanojk, line 361: `cifmw_cephadm_bootstrap_conf: /tmp/initial_ceph.conf ` is done in [1] , so the conf file generation from template at [2] is never executed.
moreover the code in [3] and [4]  relies on cifmw_cephadm_bootstrap_conf, so i had to unset it only for pre.yaml .
We shall discuss and do it in a better way.

[1] https://github.com/openstack-k8s-operators/ci-framework/blob/main/playbooks/ceph.yml#L253
[2] https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/cifmw_cephadm/tasks/pre.yml#L147
[3]  https://github.com/openstack-k8s-operators/ci-framework/blob/main/playbooks/ceph.yml#L324
[4]  https://github.com/openstack-k8s-operators/ci-framework/blob/main/playbooks/ceph.yml#L396

fultonj:
@katarimanojk is this something you noticed just by looking at the code or was a bug or undesired behavior reported which this should fix?

fultonj:
AFAICT `ceph_conf_overrides` has no default and isn't used in ci-framework but could be used. We got it from tripleo in order to support the passing of `CephConfigOverrides` in Heat on stack updates, provided `cifmw_cephadm_apply_ceph_conf_overrides_on_update` was true.

```
[fultonj@runcible ci-framework{net}]$ grep -R ceph_conf_overrides
roles/cifmw_ceph_client/tasks/main.yml:      {% set overrides = ceph_overrides.get('ceph_conf_overrides', {}) %}
roles/cifmw_cephadm/defaults/main.yml:cifmw_cephadm_apply_ceph_conf_overrides_on_update: false
roles/cifmw_cephadm/templates/ceph.conf.j2:{% for key, value in (ceph_conf_overrides | default({})).items() %}
roles/cifmw_cephadm/templates/ceph.conf.j2:{% for key, value in (ceph_conf_overrides | default({})).items() %}
[fultonj@runcible ci-framework{net}]$ 
```

katarimanojk:
> @katarimanojk is this something you noticed just by looking at the code or was a bug or undesired behavior reported which this should fix?

Francesco opened this JIRA: https://issues.redhat.com/browse/OSPRH-7427

fultonj:
> > @katarimanojk is this something you noticed just by looking at the code or was a bug or undesired behavior reported which this should fix?
> 
> @fultonj Francesco opened this JIRA: https://issues.redhat.com/browse/OSPRH-7427

Thanks, I recommend adding `Jira: https://issues.redhat.com/browse/OSPRH-7427` to the commit message and PR request.

playbooks/ceph.yml
fultonj: > `cifmw_cephadm_bootstrap_conf: /tmp/initial_ceph.conf ` is done in [1] , 
> so the conf file generation from template at [2] is never executed 
> and the `/home/ceph-admin/assimilate.conf` in the compute node will always have `initial_ceph.conf`

Yes, you're right. When ci-framework is used we always expect the initial_ceph.conf to be passed directly and created by another role, so we have no concept of supporting additional overrides via `ceph_conf_overrides`.
 
> moreover the code in [3] and [4] relies on cifmw_cephadm_bootstrap_conf, 

We don't have a problem that I know of in the code for [3] and [4]. It works and the items in the initial ceph.conf are looked up successfully in order to get the networks. It works because we're always passing ` cifmw_cephadm_bootstrap_conf`.

 > so i had to unset it only for pre.yaml . We shall discuss and do it in a better way.

What do you mean you had to unset it in pre.yml? Do you mean you had to (literally or effectively) remove this part of (lines 148-150) of TASK `Generate cifmw_cephadm_assimilate_conf on bootstrap node`?

```yaml
  when:
    - cifmw_cephadm_bootstrap_conf is not defined or
      cifmw_cephadm_bootstrap_conf | length == 0
```

The change I see in pre.yml is that it's switched from a `config_template` to a `ansible.builtin.template`.


> [1] https://github.com/openstack-k8s-operators/ci-framework/blob/main/playbooks/ceph.yml#L253
>  [2] https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/cifmw_cephadm/tasks/pre.yml#L147
>  [3] https://github.com/openstack-k8s-operators/ci-framework/blob/main/playbooks/ceph.yml#L324 
> [4] https://github.com/openstack-k8s-operators/ci-framework/blob/main/playbooks/ceph.yml#L396

roles/cifmw_cephadm/templates/ceph.conf.j2
```diff
@@ -14,7 +14,7 @@ fsid = {{ cifmw_cephadm_fsid }}
 mon host = {{ cifmw_cephadm_first_mon_ip }}
 
 {# Build CephConfigOverrides #}
-{% for key, value in (ceph_conf_overrides | default({})).items() %}
+{% for key, value in (cifmw_ceph_rgw_config | default({})).items() %}
```
fultonj, line 17: Since this template will only contain RGW configuration (and we don't need to support `ceph_conf_overrides`) perhaps it should be renamed to `ceph_rgw.conf.j2` so as to avoid confusion.

You could then remove the call in pre.yml to create this template and call it instead from rgw.yml since it's not needed until RGW configuration and not used at bootstrap. That would make the code more clear.

playbooks/ceph.yml
katarimanojk: Thanks for your thoughts.
yeah my only worry is about this task at [1] being skipped due to cifmw_cephadm_bootstrap_conf

[1] https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/cifmw_cephadm/tasks/pre.yml#L149

but like i you mentioned your other comment, if we create another task rendering the overrides only for rgw in rgw.yaml, there is no confustion here.

roles/cifmw_cephadm/templates/ceph.conf.j2
katarimanojk: @fultonj Thanks and agree with you. As overrides are only related to rgw, i will make sure to move/add the code in rgw.yaml.

playbooks/ceph.yml
fmount: The problem is if we want support or not a `cephconfigoverride` interface. If we want to rely on a generic mechanism that apply overrides that are not necessarily part of rgw, then we should improve the existing code to handle them. 
If the plan is to not support it and **only** optimize the RGW config option generation, then we can move that code in rgw.yaml.
If in the future we need a generic mechanism to handle overrides, we need to rework this patch.

roles/cifmw_cephadm/tasks/pre.yml
```diff
@@ -149,12 +149,9 @@
     - cifmw_cephadm_bootstrap_conf is not defined or
       cifmw_cephadm_bootstrap_conf | length == 0
   become: true
-  action: config_template
-  args:
+  ansible.builtin.template:
```
fmount, line 152: +1 to not use `config_template`

roles/cifmw_cephadm/tasks/rgw.yml
```diff
@@ -76,10 +76,9 @@
 
 - name: Apply ceph rgw keystone config
   ansible.builtin.command: |
-    {{ cifmw_cephadm_ceph_cli }} config set global {{ item.key }} {{ item.value }}
+    {{ cifmw_cephadm_ceph_cli }} config assimilate-conf -i {{ cifmw_cephadm_assimilate_conf_container }}
```
fmount, line 79: nit: this variable `cifmw_cephadm_assimilate_conf_container` is confusing to me: I understand this path is something resolved within the cephadm container, but can we rename to something better that ci-framework users understand in case we need to override it?

roles/cifmw_cephadm/templates/ceph.conf.j2
fmount: If we're going to create a template to render RGW overrides, I'm wondering if it makes sense to you to have a generic conf file where we not only render RGW, but we leave space to any potential override we want to inject in the ceph-cluster (it would be useful for job override purposes).
In that case the struct should be extended to have multiple entries:

```yaml
ceph_conf_overrides:
    cifmw_ceph_rgw_config:
       <params>
    other_overrides:
       <params>
```

In both cases (the scope is reduced to RGW or not), I agree to move it to a `ceph_<something>.conf.j2` to avoid confusion and for better maintenance.

- <something> can be `overrides` or `rgw`, I really do not want to enforce a particular scope here; it depends if we want an interface to `assimilate` more than just RGW, or not, and if in the future we plan for this.

roles/cifmw_cephadm/tasks/rgw.yml
katarimanojk: @fmount  `cifmw_cephadm_assimilate_conf_container ` is the path [1] to assimilate conf in the container , and we bind mount it in  [2]
can you suggest on what it should be renamed to ?  

[1] https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/cifmw_cephadm/defaults/main.yml#L22
[2] https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/cifmw_cephadm/tasks/ceph_cli.yml#L29

roles/cifmw_cephadm/tasks/rgw.yml
```diff
@@ -74,12 +74,20 @@
     mount_spec: true
     cifmw_cephadm_spec: "{{ cifmw_ceph_rgw_spec_path }}"
 
+- name: Generate cifmw_cephadm_assimilate_conf for ceph rgw
+  become: true
+  ansible.builtin.template:
+    src: "ceph_rgw.conf.j2"
+    dest: "{{ cifmw_cephadm_assimilate_conf }}"
```
katarimanojk, line 81: @fultonj this will override the generic assimilate conf (intial_ceph.conf) set at [1], is it ok ? do you see a need to use path specific to rgw like `cifmw_cephadm_rgw_assmilate.conf` and `cifmw_cephadm_rgw_assmilate.conf_container`?

[1] https://github.com/openstack-k8s-operators/ci-framework/pull/2236/files#diff-ffe549a4bf8e540dc9d3a37b219441bfaaf6d4e236bb202b6a32d55949e42f56R138

fultonj: @katarimanojk Yes, it's OK.

Note, we do not have a generic assimilate conf as you're implying. Because...

When `playbooks/ceph.yml` runs, it always generates and sets `cifmw_cephadm_bootstrap_conf` with this role.

  https://github.com/openstack-k8s-operators/ci-framework/tree/main/roles/cifmw_ceph_spec

TASK "Push it to cifmw_cephadm_assimilate_conf on bootstrap node" in pre.yml pushes it to the node and it's bootstraped. 

TASK "Generate cifmw_cephadm_assimilate_conf on bootstrap node" in pre.yml is never run because  `cifmw_cephadm_bootstrap_conf`  is never empty.

The "general" way to override the ceph.conf  in ci-framework is to use roles/cifmw_ceph_spec

playbooks/ceph.yml
fultonj: Per our conversation we do NOT support the cephconfigoverride interface.

fultonj: Per our conversation the generic mechanism is via 

https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/cifmw_ceph_spec/templates/initial_ceph.conf.j2

roles/cifmw_cephadm/tasks/rgw.yml
katarimanojk: Got it, thanks.

roles/cifmw_cephadm/templates/ceph.conf.j2
katarimanojk: As discussed , template file is renamed as `ceph_rgw.conf.j2`, and it rendered and applied in rgw.yaml

roles/cifmw_cephadm/tasks/rgw.yml
fmount: no I think we can keep as it is, maybe we can add a comment to indicate that is referring to the path within the container, but we can always follow up if we need.

fmount:
It would be nice to test  this patch when #2248 lands, so we can have evidence of the log dump at deployment time, where we expect that `assimilate <file>` worked as expected (as we run `ceph config dump`).

katarimanojk:
> It would be nice to test this patch when #2248 lands, so we can have evidence of the log dump at deployment time, where we expect that `assimilate <file>` worked as expected (as we run `ceph config dump`).

yeah good idea, after rebase, i will retest it and observe the generated logs.

katarimanojk:
> > It would be nice to test this patch when #2248 lands, so we can have evidence of the log dump at deployment time, where we expect that `assimilate <file>` worked as expected (as we run `ceph config dump`).
> 
> yeah good idea, after rebase, i will retest it and observe the generated logs.

@fmount I have tested it along with #2248 , and i see the log dump with all the config in 'assimilate conf'.

fultonj:
Looks like it worked in `podified-multinode-hci-deployment-crc`

  https://logserver.rdoproject.org/36/2236/b4c9c5461d02b9f6e48f86e56d10a9e9f36de738/github-check/podified-multinode-hci-deployment-crc/1984f3f/controller/ci-framework-data/logs/ceph/ceph_config.log

I see the following in  https://logserver.rdoproject.org/36/2236/b4c9c5461d02b9f6e48f86e56d10a9e9f36de738/github-check/podified-multinode-hci-deployment-crc/1984f3f/controller/ci-framework-data/logs/ansible.log  

```
2024-08-26 04:27:16,601 p=29891 u=zuul n=ansible | TASK [cifmw_cephadm : Create a Ceph RGW spec src=templates/ceph_rgw.yml.j2, dest={{ cifmw_ceph_rgw_spec_path }}, mode=0644, force=True] ***
2024-08-26 04:27:16,601 p=29891 u=zuul n=ansible | Monday 26 August 2024  04:27:16 -0400 (0:00:00.029)       0:31:25.357 ********* 
2024-08-26 04:27:17,547 p=29891 u=zuul n=ansible | changed: [compute-0]
2024-08-26 04:27:17,559 p=29891 u=zuul n=ansible | TASK [cifmw_cephadm : Get ceph_cli _raw_params=ceph_cli.yml] *******************
2024-08-26 04:27:17,559 p=29891 u=zuul n=ansible | Monday 26 August 2024  04:27:17 -0400 (0:00:00.957)       0:31:26.315 ********* 
2024-08-26 04:27:17,603 p=29891 u=zuul n=ansible | included: /home/zuul/src/github.com/openstack-k8s-operators/ci-framework/roles/cifmw_cephadm/tasks/ceph_cli.yml for compute-0
2024-08-26 04:27:17,618 p=29891 u=zuul n=ansible | TASK [cifmw_cephadm : Set ceph CLI cifmw_cephadm_ceph_cli={{ cifmw_cephadm_container_cli }} run --rm {{ cifmw_cephadm_container_options }} {% if mount_certs|default(false) %} --volume {{ cifmw_cephadm_certs }}:/etc/pki/tls:z {% endif %} {% if sensitive_data|default(false) %} --interactive {% endif %} --volume {{ cifmw_cephadm_config_home }}:/etc/ceph:z {% if not (external_cluster|default(false) or crush_rules|default(false)) -%} --volume {{ cifmw_cephadm_assimilate_conf }}:{{ cifmw_cephadm_assimilate_conf_container }}:z {% endif %} {% if mount_spec|default(false) %} --volume {{ cifmw_cephadm_spec }}:{{ cifmw_cephadm_container_spec }}:z {% endif %} {% if admin_daemon|default(false) %} --volume /var/run/ceph/{{ cifmw_cephadm_fsid }}:/var/run/ceph:z {% endif %} --entrypoint {{ ceph_command | default('ceph') }} {{ cifmw_cephadm_container_ns }}/{{ cifmw_cephadm_container_image }}:{{ cifmw_cephadm_container_tag }} {% if ceph_command|default('ceph') == 'ceph' or ceph_command|default('ceph') == 'rados' or ceph_command|default('ceph') == 'rbd' -%}
  {% if not admin_daemon|default(false) -%}
  --fsid {{ cifmw_cephadm_fsid }} -c /etc/ceph/{{ cifmw_cephadm_cluster }}.conf -k /etc/ceph/{{ cifmw_cephadm_cluster }}.client.{{ select_keyring| default('admin') }}.keyring
  {%- endif %}
  {% if external_cluster|default(false) -%}
  -n client.{{ select_keyring }}
  {%- endif %}
{%- endif %}] ***
2024-08-26 04:27:17,618 p=29891 u=zuul n=ansible | Monday 26 August 2024  04:27:17 -0400 (0:00:00.059)       0:31:26.375 ********* 
2024-08-26 04:27:17,671 p=29891 u=zuul n=ansible | ok: [compute-0]
2024-08-26 04:27:17,680 p=29891 u=zuul n=ansible | TASK [cifmw_cephadm : Generate cifmw_cephadm_assimilate_conf for ceph rgw src=ceph_rgw.conf.j2, dest={{ cifmw_cephadm_assimilate_conf }}, owner={{ cifmw_cephadm_uid }}, group={{ cifmw_cephadm_uid }}, mode=0644] ***
2024-08-26 04:27:17,680 p=29891 u=zuul n=ansible | Monday 26 August 2024  04:27:17 -0400 (0:00:00.061)       0:31:26.436 ********* 
2024-08-26 04:27:18,701 p=29891 u=zuul n=ansible | changed: [compute-0]
2024-08-26 04:27:18,719 p=29891 u=zuul n=ansible | TASK [cifmw_cephadm : Apply ceph rgw keystone config _raw_params={{ cifmw_cephadm_ceph_cli }} config assimilate-conf -i {{ cifmw_cephadm_assimilate_conf_container }}
] ***
2024-08-26 04:27:18,719 p=29891 u=zuul n=ansible | Monday 26 August 2024  04:27:18 -0400 (0:00:01.039)       0:31:27.476 ********* 
2024-08-26 04:27:19,820 p=29891 u=zuul n=ansible | ok: [compute-0]
2024-08-26 04:27:19,837 p=29891 u=zuul n=ansible | TASK [cifmw_cephadm : Apply ceph config-key set ssl_option _raw_params={{ cifmw_cephadm_ceph_cli }} config-key set ssl_option {{ cifmw_cephadm_config_key_set_ssl_option }}
] ***
2024-08-26 04:27:19,837 p=29891 u=zuul n=ansible | Monday 26 August 2024  04:27:19 -0400 (0:00:01.117)       0:31:28.593 ********* 
2024-08-26 04:27:21,055 p=29891 u=zuul n=ansible | ok: [compute-0]
```

fultonj:
@katarimanojk would you please do a downstream test-project with this?

fultonj:
A downstream test project has passed with gamma and epsilon so I feel comfortable merging

fultonj (approved):
/approve
/lgtm

PR #2237, opened 2024-08-18
Update switch role modules

eshulman2:
- updated switch module for Junos switches
- allow ipv6 on l3 interfaces

### DIFF
--- a/galaxy.yml
+++ b/galaxy.yml
@@ -81,8 +81,8 @@ dependencies:
   'git+https://github.com/openstack/ansible-collections-openstack': 'master'
   'git+https://github.com/ansible-collections/ansible.netcommon': 'main'
   'git+https://github.com/openstack/ansible-config_template': 'master'
-  'git+https://github.com/ansible-collections/junipernetworks.junos': '2.10.0'
-  'git+https://github.com/ansible-collections/cisco.ios': '3.3.2'
+  'git+https://github.com/ansible-collections/junipernetworks.junos': 'main'
+  'git+https://github.com/ansible-collections/cisco.ios': 'main'
   'git+https://github.com/ansible-collections/mellanox.onyx': 'master'
   'git+https://github.com/openshift/community.okd': 'main'
 
--- a/roles/switch_config/tasks/ios_config.yml
+++ b/roles/switch_config/tasks/ios_config.yml
@@ -14,11 +14,15 @@
   when: "'dhcp_relay' in switch_vars"
 
 - name: Set VLANs
-  cisco.ios.ios_vlan:
-    name: "vlan{{ item }}"
-    vlan_id: "{{ item }}"
-    state: present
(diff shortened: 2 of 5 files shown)


galaxy.yml
```diff
@@ -81,7 +81,7 @@ dependencies:
   'git+https://github.com/openstack/ansible-collections-openstack': 'master'
   'git+https://github.com/ansible-collections/ansible.netcommon': 'main'
   'git+https://github.com/openstack/ansible-config_template': 'master'
-  'git+https://github.com/ansible-collections/junipernetworks.junos': '2.10.0'
+  'git+https://github.com/ansible-collections/junipernetworks.junos': 'main'
```
lewisdenny, line 84: Please also update this in the requirements:

https://github.com/openstack-k8s-operators/ci-framework/blob/00bdc9430c01230d6965015274d72af035991cb3/requirements.yml#L39

lewisdenny (changes requested):

galaxy.yml
eshulman2: done

lewisdenny (approved):
/lgtm

roles/switch_config/tasks/ios_config.yml
```diff
@@ -14,11 +14,15 @@
   when: "'dhcp_relay' in switch_vars"
 
 - name: Set VLANs
-  cisco.ios.ios_vlan:
-    name: "vlan{{ item }}"
-    vlan_id: "{{ item }}"
-    state: present
-  loop: "{{ vlan_list }}"
+  vars:
+    vlan_config: |
```
pablintino, line 18: You can avoid the yaml conversion by simple tweaking the jinja part:

```yaml
  vars:
    vlan_config: |-
      [   
      {% for vlan in vlan_list %}
      - name: "vlan{{ vlan }}"
        vlan_id: "{{ vlan }}"
      {% endfor %}
      ]
```
With that in place you can directly pass the var like this:
```yaml
  cisco.ios.ios_vlans:
    config: "{{ vlan_config }}"
```

Or even simpler:
```yaml
  cisco.ios.ios_vlans:
    config: |-
      [   
      {% for vlan in vlan_list %}
      - name: "vlan{{ vlan }}"
        vlan_id: "{{ vlan }}"
      {% endfor %}
      ]
    state: merged

```

eshulman2: I tried doing semilar things but for some reason the module failed to accept them, not sure if this is a specific issue to the modules but unfortunately I was unable to find a better way to implement it with the time I had to work on it
```
fatal: [nfv_private_sw02]: FAILED! => changed=false 
  invocation:
    module_args:
      config: []
      running_config: null
      state: merged
  msg: dictionary requested, could not parse JSON or key=value
```

PR #2240, opened 2024-08-20
Update switch role modules

eshulman2:
- fix layer3 interfaces configuration


lewisdenny (approved):
/approve

@eshulman2 I suggest adding some testing to these large j2 template vars. 

One idea that might work is storing these big template vars in a vars file rather than in the task file. 

In the Molecule job could render these vars into a file and compare it with a rendered copy you provide as a golden file. 

In the task file you can just call them from the vars file.

PR #2231, opened 2024-08-15
Update the source of containers to pull from cifmw_update_containers_…

frenzyfriday:
…registry

PR #2242, opened 2024-08-21
[VA HCI]Adds cindervolume and manilashare names for update_containers

raukadah:
HCI job, ceph cinder volume and share1 manila share is used. We need to add the cindervolume and manila share name so that proper containers gets updated by update_containers role.

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/2231


arxcruz:
This is not the best solution but I'm intended to approve it in order to fix the problem, and then work on a proper solution.

raukadah:
recheck

PR #2243, opened 2024-08-21
Remove checking if commit in multiple PRs

lewisdenny:
This check was not requested, I added as it was a side product of the
main check but it doesn't work as I expected and can block good patches

### DIFF
--- a/.github/workflows/review-ready-label.yml
+++ b/.github/workflows/review-ready-label.yml
@@ -3,7 +3,6 @@
 #   - There are no failed CI jobs
 #   - There are no Pending CI jobs excluding Tide
 #   - The PR isn't in draft state
-#   - The commit isn't found in more than one PR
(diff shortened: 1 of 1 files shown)


.github/workflows/review-ready-label.yml
```diff
@@ -59,7 +58,7 @@ jobs:
           pr_search_result_length=$(echo "${pr_search_result}" | jq length)
           if [ "$pr_search_result_length" -ne 1 ]; then
               echo "Commit not found or in multiple PRs"
-              exit 1
+              exit 0
```
cjeanner, line 61: hm, so we'll keep the exit message in place, and don't act at all if we hit that case. Is this really what we want?
Guess we can't test it until we merge, right?

raukadah:
/lgtm

merging to test it!

PR #2244, opened 2024-08-21
Update pod-jobs to use cifmw container

lewisdenny:
Just testing


lewisdenny:
Oh cool, seems the pod Zuul spawns can just point to an image ref so that could point to our cifmw container

```
  237     │ , centos-9-stream = Base.Container::{
  238     │ │ , Dockerfile = ""
  239     │ │ , label = Some "pod-centos-9-stream"
  240     │ │ , image-ref = Some "quay.io/software-factory/zuul-worker-c9s:latest"
  241     │ │ }
```

PR #2245, opened 2024-08-21
Switch modules update

eshulman2:
- merge to replace - For most of the usecases merge is good enough but certain edge cases
might be effected due to certain configuration. changed some merge
statments to replaced.
- allow reusing ae members


lewisdenny (approved):
/lgtm

PR #2246, opened 2024-08-21
Add a hook that deploys Ceph on BMO nodes

pablintino:
This patch is to allow discovering of nodeset computes and adding them to a in memory inventory before calling ceph, this is to enable NFV+HCI job to run.

### DIFF
--- a/hooks/playbooks/ceph-bm.yml
+++ b/hooks/playbooks/ceph-bm.yml
@@ -1,0 +1,39 @@
+- name: Deploy ceph on OSP BMO deployed nodes
+  hosts: "{{ cifmw_target_hook_host | default('localhost') }}"
+  gather_facts: false
+  tasks:
+    - name: Fetch OSP BMO nodesets
+      environment:
+        KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+        PATH: "{{ cifmw_path }}"
+      ansible.builtin.command:
+        cmd: >-
+          oc get OpenStackBaremetalSet -n "{{ _namespace}}"  -o yaml
+      register: _osp_bmo_nodsets_oc_out
+
+    - name: Add OSP BMO nodesets to Ansible
+      ansible.builtin.add_host:
+        name: "{{ item.name }}"
+        groups: "{{ item.group }}"
+        ansible_ssh_user: "{{ item.user }}"
+        ansible_host: "{{ item.ip }}"
+        ansible_ssh_private_key_file: "{{ ansible_user_dir }}/.ssh/id_cifw"
+        ansible_ssh_extra_args: '-o StrictHostKeyChecking=no'
+      loop: >-
+        {% set hosts = [] -%}
+        {% set nodesets = (_osp_bmo_nodsets_oc_out.stdout | from_yaml)['items'] | default([]) -%}
+        {% for spec in nodesets | map(attribute='spec') -%}
+        {%   for host_key, host_val in spec.baremetalHosts.items() -%}
+        {%     set _ = hosts.append(
+          {
+          'name': host_key,
+          'ip': host_val['ctlPlaneIP'] | ansible.utils.ipaddr('address'),
+          'user': spec.cloudUserName,
+          'group': host_key | split('-') | first + 's'
+        }) -%}
+        {%   endfor -%}
+        {% endfor -%}
+        {{ hosts }}
+
+- name: Deploy Ceph on target nodes
+  ansible.builtin.import_playbook: ../../playbooks/ceph.yml
--- a/playbooks/ceph.yml
+++ b/playbooks/ceph.yml
@@ -63,7 +63,13 @@
   become: true
   vars:
     cifmw_admin_user: ceph-admin
-    ansible_ssh_private_key_file: "{{ lookup('env', 'ANSIBLE_SSH_PRIVATE_KEY') }}"
+    _target_group: "{{ cifmw_ceph_target | default('computes') }}"
+    _target: "{{ groups[_target_group] | default([]) | first }}"
+    ansible_ssh_private_key_file: >-
+      {{
+        hostvars[_target]['ansible_ssh_private_key_file'] |
+        default(lookup('env', 'ANSIBLE_SSH_PRIVATE_KEY'))
+      }}
   pre_tasks:
     # end_play will end this current playbook and go the the next
     # imported play.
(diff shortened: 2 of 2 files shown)


playbooks/ceph.yml
```diff
@@ -63,7 +63,7 @@
   become: true
   vars:
     cifmw_admin_user: ceph-admin
-    ansible_ssh_private_key_file: "{{ lookup('env', 'ANSIBLE_SSH_PRIVATE_KEY') }}"
+    ansible_ssh_private_key_file: "{{ hostvars[(groups[cifmw_ceph_target | default('computes')] | default([]) | first)]['ansible_ssh_private_key_file'] | default(lookup('env', 'ANSIBLE_SSH_PRIVATE_KEY')) }}"
```
fmount, line 66: I assume this preserves the default behavior as the default value is what we previously had on L66.

hooks/playbooks/ceph-bm.yml
```diff
@@ -0,0 +1,39 @@
...
+  hosts: "{{ cifmw_target_hook_host | default('localhost') }}"
+  gather_facts: false
+  tasks:
+    - name: Fetch OSP BMO nodesets
+      environment:
+        KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+        PATH: "{{ cifmw_path }}"
+      ansible.builtin.command:
+        cmd: >-
+          oc get OpenStackBaremetalSet -n openstack -o yaml
```
fmount, line 11: is the $namespace something we should pass as a parameter? I saw this pattern somewhere and I'm wondering if it makes sense to have a `$namespace | default('openstack')`.

fmount (approved):
/lgtm

fmount:
Just a small nit about the `namespace`, otherwise looks good to me. Not sure we have any job we can look at but I know @eshulman2 had positive results.

hooks/playbooks/ceph-bm.yml
cjeanner: `namespace` is passed down by the run_hook role:
https://github.com/openstack-k8s-operators/ci-framework/blob/0f9117448b23647fa9051064c3e23cad87404651/roles/run_hook/tasks/playbook.yml#L35
You can/should consume it.

playbooks/ceph.yml
eshulman2: yes

cjeanner, line 66: care to split a bit that line? It's barely readable...
Something like:
```yaml
vars:
  _target_group: "{{ cifmw_ceph_target | default('computes') }}"
  _target: "{{ groups[_target] | default([]) | first }}"
 ansible_ssh_private_key_file: >-
  {{
    hostvars[_target]['ansible_ssh_private_key_file'] |
    default(lookup('env', 'ANSIBLE_SSH_PRIVATE_KEY'))
  }}
```

Also, I'm not sure the `| default([]) | first` works, iirc `first` will crash because there's no first element in an empty list... Care to validate that on the side?

playbooks/ceph.yml
```diff
@@ -119,7 +119,7 @@
           {{ num_osds }}
     - name: Create Block Device on EDPM Nodes
       vars:
-        ansible_ssh_private_key_file: "{{ lookup('env', 'ANSIBLE_SSH_PRIVATE_KEY') }}"
+        ansible_ssh_private_key_file: "{{ hostvars[(groups[cifmw_ceph_target | default('computes')] | default([]) | first)]['ansible_ssh_private_key_file'] | default(lookup('env', 'ANSIBLE_SSH_PRIVATE_KEY')) }}"
```
cjeanner, line 122: same, please split that up to keep it readable...

hooks/playbooks/ceph-bm.yml
eshulman2: It is possible but I don't think it is actually required having that I don't think anyone is using a different namespace for openstack other than openstack but lets hear some more opinions

cjeanner (changes requested):

playbooks/ceph.yml
eshulman2: done

playbooks/ceph.yml
```diff
@@ -63,7 +63,13 @@
...
   vars:
     cifmw_admin_user: ceph-admin
-    ansible_ssh_private_key_file: "{{ lookup('env', 'ANSIBLE_SSH_PRIVATE_KEY') }}"
+    _target_group: "{{ cifmw_ceph_target | default('computes') }}"
+    _target: "{{ groups[_target_group] | default([]) | first }}"
+    ansible_ssh_private_key_file: >-
+      {{
+        hostvars[_target]['ansible_ssh_private_key_file'] |
+        default(lookup('env', 'ANSIBLE_SSH_PRIVATE_KEY'))
+      }}
```
fultonj, line 72: these changes look like they will be fine but could we please test project them downstream to make sure they do not affect the unijobs?

hooks/playbooks/ceph-bm.yml
```diff
@@ -0,0 +1,39 @@
...
+  hosts: "{{ cifmw_target_hook_host | default('localhost') }}"
+  gather_facts: false
+  tasks:
+    - name: Fetch OSP BMO nodesets
+      environment:
+        KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+        PATH: "{{ cifmw_path }}"
+      ansible.builtin.command:
+        cmd: >-
+          oc get OpenStackBaremetalSet -n "{{ _namespace}}"  -o yaml
```
fultonj, line 11: I see this is one command to generate a yaml file so it's probably fine.

Did you consider using the `kubernetes.core.k8s` module? For example.

https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/ci_lvms_storage/tasks/main.yml#L42-L52'

In general I think this is nicer than running oc commands but I don't think we have to block the review over it.

playbooks/ceph.yml
```diff
@@ -119,7 +125,13 @@
           {{ num_osds }}
     - name: Create Block Device on EDPM Nodes
       vars:
-        ansible_ssh_private_key_file: "{{ lookup('env', 'ANSIBLE_SSH_PRIVATE_KEY') }}"
+        _target_group: "{{ cifmw_ceph_target | default('computes') }}"
```
fultonj, line 128: ditto

fultonj (commented):
Thank you. In general I'm positive on this. Can we make sure it tested downstream before merging? If you confirm I could approve if you like.

playbooks/ceph.yml
eshulman2: yes I am testing it now

hooks/playbooks/ceph-bm.yml
eshulman2: I'll look into it but I would prefer doing it after getting this patch merge having that those are leftovers we wanted to get done ASAP.

eshulman2:
recheck

fultonj (approved):
/approve
/lgtm

PR #2256, opened 2024-08-23
Added trigger downstream job github action

raukadah:
Since we have limited hardware in downstream. If all developers want to run github-experimental pipeline downstream trigger job, then We may run of resources.

In order to avoid that, This pr adds github-action-trigger-downstream-job to allow approvers to trigger downstream job. Non-approvers will be greeted with message that they are not allowed to run this workflow.

Currently downstream trigger jobs are invoked via 'trigger github-experimental'. Once https://review.rdoproject.org/r/c/config/+/54157 merges, We will trigger it via github-experimental label. There it will be useful.

Testresults:
* [When approver is not in allowed list](https://github.com/raukadah/ci-playground/pull/10#issuecomment-2298751594)
* [When approver is in allowed list](https://github.com/raukadah/ci-playground/pull/12#issuecomment-2298786162)
* [When a new approver is added via pull request and on that pr, approver about to get added run trigger job, they are not allowed to run the trigger job](https://github.com/raukadah/ci-playground/pull/21#issuecomment-2319995489)
  * [Once the PR get merged, they are allowed to run trigger workflow.](https://github.com/raukadah/ci-playground/pull/21#issuecomment-2320004406)

### DIFF
--- a/.github/workflows/github-action-trigger-downstream-job.yaml
+++ b/.github/workflows/github-action-trigger-downstream-job.yaml
@@ -1,0 +1,52 @@
+name: Github Action to trigger downstream job
+on:
+  issue_comment:
+    types: [created, edited]
+
+jobs:
+  jobs01:
+    # Check if comment contains /trigger github-experimental
+    if: contains(github.event.comment.body, '/trigger github-experimental')
+    runs-on: [ubuntu-latest]
+    permissions:
+      pull-requests: write
+    outputs:
+      approvers_check: ${{ steps.approvers_check.outputs.approvers_check }}
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+        with:
+          sparse-checkout: |
+            OWNERS
+          sparse-checkout-cone-mode: false
+          ref: main
+
+      - name: Get approvers list
+        id: approvers_list
+        uses: mikefarah/yq@master
+        with:
+          cmd: yq -o=csv ".approvers" OWNERS
+
+      - name: Print Approvers list
+        run: echo ${{ steps.approvers_list.outputs.result }}
+
+      - name: Check commenter status
+        id: approvers_check
+        run: |
+          APPROVERS_CHECK=$(python -c "print('${{ github.event.comment.user.login }}' in '${{ steps.approvers_list.outputs.result }}')")
+          echo "approvers_check=$APPROVERS_CHECK" >> "$GITHUB_OUTPUT"
+
+      - name: Add proper labels to trigger workflow
+        if: steps.approvers_check.outputs.approvers_check == 'True'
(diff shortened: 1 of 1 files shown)


.github/workflows/github-action-trigger-downstream-job.yaml
```diff
@@ -0,0 +1,49 @@
...
+    runs-on: [ubuntu-latest]
+    permissions:
+      pull-requests: write
+    outputs:
+      approvers_check: ${{ steps.approvers_check.outputs.approvers_check }}
+    steps:
+      - name: Install wget and tar
+        run: sudo apt-get install wget tar
+
+      - name: Download yq binary
```
cjeanner, line 19: isn't there already some pre-built image with yq binary on the marketplace? That would allow to NOT have to manage that dependency over new releases, security and so on...

rebtoor: https://github.com/mikefarah/yq?tab=readme-ov-file#github-action

.github/workflows/github-action-trigger-downstream-job.yaml
```diff
@@ -0,0 +1,49 @@
...
+        run: |
+          wget https://github.com/mikefarah/yq/releases/download/v4.44.3/yq_linux_amd64.tar.gz -O - |  tar xz && sudo mv yq_linux_amd64 /usr/bin/yq
```
rebtoor, line 21:
```suggestion
          curl https://github.com/mikefarah/yq/releases/download/v4.44.3/yq_linux_amd64.tar.gz | tar xz && sudo mv yq_linux_amd64 /usr/bin/yq
```

.github/workflows/github-action-trigger-downstream-job.yaml
```diff
@@ -0,0 +1,49 @@
...
+    # Check if comment contains /trigger github-experimental
+    if: contains(github.event.comment.body, '/trigger github-experimental')
+    runs-on: [ubuntu-latest]
+    permissions:
+      pull-requests: write
+    outputs:
+      approvers_check: ${{ steps.approvers_check.outputs.approvers_check }}
+    steps:
+      - name: Install wget and tar
+        run: sudo apt-get install wget tar
```
rebtoor, line 17:
```suggestion
        run: sudo apt-get install tar
```

.github/workflows/github-action-trigger-downstream-job.yaml
raukadah: Replace it with github action

.github/workflows/github-action-trigger-downstream-job.yaml
raukadah: Removed, no longer needed

PR #2258, opened 2024-08-26
Introduce Ceph Upgrade tasks

fmount:
This patch improves the existing `cephadm` role and introduces a set of tasks that are supposed to perform a `Ceph update/upgrade`. The main idea behind this work is to be able to include it from [1] and rollout new `Ceph` daemons before starting an `OpenStack` minor update.
Including this new set of tasks in [1] is not part of this PR, and it will be done in a follow up.

[1] https://github.com/openstack-k8s-operators/ci-framework/blob/main/update-edpm.yml#L18

Jira: https://issues.redhat.com/browse/OSPRH-9697

### DIFF
--- a/roles/cifmw_cephadm/tasks/ceph_upgrade.yml
+++ b/roles/cifmw_cephadm/tasks/ceph_upgrade.yml
@@ -1,0 +1,78 @@
+- name: Fail if Ceph FSID is not set
+  ansible.builtin.fail:
+    msg: "Ceph FSID must be defined"
+  when: cifmw_cephadm_fsid is undefined
+
+- name: Get ceph_cli
+  ansible.builtin.include_tasks: ceph_cli.yml
+  vars:
+    mount_certs: true
+
+- name: Get Ceph Health
+  become: true
+  ansible.builtin.command:
+    cmd: "{{ cifmw_cephadm_ceph_cli }} -s -f json"
+  register: ceph
+
+- name: Load ceph data
+  ansible.builtin.set_fact:
+    ceph_health: "{{ ceph.stdout | from_json }}"
+
+- name: Fail if health is HEALTH_WARN || HEALTH_ERR
+  ansible.builtin.fail:
+    msg: Ceph is in {{ ceph_health.status }} state.
+  when:
+    - ceph_health.status == 'HEALTH_WARN' or
+      ceph_health.status == 'HEALTH_ERR'
+
+- name: Build container image target
+  ansible.builtin.set_fact:
+    cifmw_cephadm_update_target_image: >-
+      "{{ cifmw_cephadm_update_container_ns }}/{{ cifmw_cephadm_update_container_image }}:{{ cifmw_cephadm_update_container_tag }}"
+
+- name: Check if a Ceph update is required
+  become: true
+  ansible.builtin.command:
+    cmd: "{{ cifmw_cephadm_ceph_cli }} orch upgrade check {{ cifmw_cephadm_update_target_image }} -f json"
+  register: ceph_needs_update
+
+- name: Ceph upgrade
+  when: ceph_needs_update.stdout | from_json | community.general.json_query('needs_update') | default('') | length > 0
+  become: true
+  block:
+    - name: Rollout new container images using cephadm
+      ansible.builtin.command:
+        cmd: "{{ cifmw_cephadm_ceph_cli }} orch upgrade start --image {{ cifmw_cephadm_update_target_image }}"
+    - name: Wait while Ceph update is in progress
+      ansible.builtin.command:
+        cmd: "{{ cifmw_cephadm_ceph_cli }} orch upgrade status -f json"
+      register: ceph_upgrade_progress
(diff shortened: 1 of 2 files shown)


katarimanojk:
@fmount the tasks in ceph_upgrade.yaml looks good to me as it
-  checks the health of cluster
-  performs 'orch upgrade'  
-  wait for the upgrade to succeed 
-  collect logs.

Along with this,  i think we should do additional work (may be i can do in a followup ) for upgrading `monitoring_stack` and `ganesha` daemons similar to tripleo.
for monitoring_stack upgrade, before 'orch upgrade start' we should call [1] by overriding the images in [2]

[1] https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/cifmw_cephadm/tasks/cephadm_config_set.yml#L84
[2] https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/cifmw_cephadm/defaults/main.yml#L101

fmount:
> @fmount the tasks in ceph_upgrade.yaml looks good to me as it
> 
>     * checks the health of cluster
> 
>     * performs 'orch upgrade'
> 
>     * wait for the upgrade to succeed
> 
>     * collect logs.
> 
> 
> Along with this, i think we should do additional work (may be i can do in a followup ) for upgrading `monitoring_stack` and `ganesha` daemons similar to tripleo. for monitoring_stack upgrade, before 'orch upgrade start' we should call [1] by overriding the images in [2]
> 
> [1] https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/cifmw_cephadm/tasks/cephadm_config_set.yml#L84 [2] https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/cifmw_cephadm/defaults/main.yml#L101

Hi @katarimanojk ,
thanks for the review.
-1 on ganesha: in 18 ganesha is only managed by cephadm, the tripleo model doesn't exist anymore, and Ganesha uses the same image of the rest of the Ceph daemons.
WRT the dashboard, I would say is out of scope here, as there's no interaction w/ the rest of the OpenStack client. We might provide an upgrade path as a nice have in the future, but I guess it's not required here. In addition, the container images related to the dashboard change when you do a major `upgrade`, but they do not change during a minor update.

fmount:
recheck

roles/cifmw_cephadm/tasks/ceph_upgrade.yml
```diff
@@ -0,0 +1,84 @@
+- name: Fail if target image is not available
+  ansible.builtin.fail:
+    msg: "Ceph target image is not available"
+  when: cifmw_cephadm_update_container_image is undefined or
```
ciecierski, line 4: this two variables are defined always

roles/cifmw_cephadm/tasks/ceph_upgrade.yml
```diff
@@ -0,0 +1,84 @@
...
+
+- name: Load ceph data
+  ansible.builtin.set_fact:
+    ceph_health: "{{ ceph.stdout | from_json }}"
+
+- name: Fail if health is HEALTH_WARN || HEALTH_ERR
+  ansible.builtin.fail:
+    msg: Ceph is in {{ ceph.health.status }} state.
+  when:
+    - ceph.health.status == 'HEALTH_WARN' or
```
ciecierski, line 31: s/ceph.health/ceph_health

roles/cifmw_cephadm/tasks/ceph_upgrade.yml
```diff
@@ -0,0 +1,84 @@
...
+    cmd: "{{ cifmw_cephadm_ceph_cli }} -s -f json"
+  register: ceph
+
+- name: Load ceph data
+  ansible.builtin.set_fact:
+    ceph_health: "{{ ceph.stdout | from_json }}"
+
+- name: Fail if health is HEALTH_WARN || HEALTH_ERR
+  ansible.builtin.fail:
+    msg: Ceph is in {{ ceph.health.status }} state.
```
ciecierski, line 29: same as in next line

roles/cifmw_cephadm/tasks/ceph_upgrade.yml
fmount: correct, we don't need them: we have defaults or the user is supposed to provide an override. If the image is not valid, cephadm itself will fail checking for `needs_update` using the provided container image.

roles/cifmw_cephadm/tasks/ceph_upgrade.yml
fmount: good catch, Done.

roles/cifmw_cephadm/tasks/ceph_upgrade.yml
fmount: Done

fmount:
CI is green (even though not really testing this change).
@fultonj @ciecierski @katarimanojk can we land this one or do you have additional comments I should address?

ciecierski (approved):
lgtm

PR #2260, opened 2024-08-26
[DNM] Testing extracing network info from controller

jamepark4:
When executing reproducer with cifmw_job_uri, crc_ci_bootstrap_networking is being stored in controller instead of crc.  Updating to point to controller when populating bootstrap network info.

### DIFF
--- a/roles/reproducer/tasks/ci_deploy_data.yml
+++ b/roles/reproducer/tasks/ci_deploy_data.yml
@@ -29,7 +29,7 @@
       }}
   ansible.builtin.set_fact:
     cacheable: true
-    ci_job_networking: "{{ zuul_inventory.all.hosts.crc.crc_ci_bootstrap_networking }}"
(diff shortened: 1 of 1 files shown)


roles/reproducer/tasks/ci_deploy_data.yml
```diff
@@ -29,7 +29,7 @@
       }}
   ansible.builtin.set_fact:
     cacheable: true
-    ci_job_networking: "{{ zuul_inventory.all.hosts.crc.crc_ci_bootstrap_networking }}"
+    ci_job_networking: "{{ zuul_inventory.all.hosts.controller.crc_ci_bootstrap_networking }}"
```
cjeanner, line 32: Note this is a quick workaround - I'll try to find a better suited solution. There are DS jobs running and zuul inventory doesn't seem to have crc node, just "controller", leading to a crash in case of a reproducer run.

I'm waiting to get some feedback about that workaround, in order to implement a better solution.

PR #2262, opened 2024-08-27
Replace content_provider_registry_ip with _insecure_registries

raukadah:


raukadah:
/close

no longer needed, approach changed!

PR #2264, opened 2024-08-27
NFV update workload scripts changes

Jaganathancse:
This change is to add workload automation support for NFV OVS DPDK.

### DIFF
--- a/roles/update/templates/workload_launch.sh.j2
+++ b/roles/update/templates/workload_launch.sh.j2
@@ -132,7 +132,7 @@ function generate_traffic {
 function prepare_env {
 
     export IMAGE_URL='{{ workload_image_url|default("http://download.cirros-cloud.net/0.5.2/cirros-0.5.2-x86_64-disk.img") }}'
-    export IMAGE_NAME="upgrade_workload_${SUFFIX}"
+    export IMAGE_NAME="{{ workload_image_name|default('upgrade_workload_${SUFFIX}')}}"
     export IMAGE_FILE="$(mktemp -d )/upgrade_workload_image.qcow2"
     export INSTANCE_NAME="instance_${SUFFIX}"
     export INSTANCE_USER='{{ workload_user|default("cirros") }}'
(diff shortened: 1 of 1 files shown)


lewisdenny:
recheck
`build-push-container-cifmw-client` failed due to unrelated issue to this patch.

rdiazcam (approved):
lgtm

cjeanner:
/appove
/lgtm

NFV team tested it.

roles/update/templates/workload_launch.sh.j2
```diff
@@ -319,6 +324,9 @@ function workload_launch {
             --ram {{ workload_memory|default(512) }} \
             --disk {{ workload_disk|default(10) }} \
             --swap {{ workload_swap|default(1) }} \
+            --property hw:mem_page_size=large \
```
ciecierski, line 327: For non-nfv workload I don't think we should set this properties.
I suggest to create nfv flavor and non-nfv inside if-else condition

Jaganathancse: @ciecierski ack

Jaganathancse: done

roles/update/templates/workload_launch.sh.j2
```diff
@@ -188,6 +189,8 @@ function sanity_teardown {
     local elapsed_seconds=0
     {% if workload_sriov|default(false) | bool -%}
     openstack port delete "${SRIOV_PORT}"
+    {% elif workload_dpdk|default(false) | bool -%}
+    openstack port delete "${DPDK_PORT}"
     {% else -%}
```
sathlan, line 194: Oki, so I'm not an expert in `dpdk` - or `sriov` for that matter - but just to make sure.  Here (at L193) we don't remove the FIP from the INSTANCE.

But on L50 of that same file (`set_vm_ip` function called on L438), we do create a FIP.  So it means that just deleting the port remove the FIP.  Is that correct ?  Also, also around line 50, for `sriov `the command to get FIP was different, in the sense that creating the port seems to create the FIP.  Here for `dpdk` we don't have any special treatment. 

Overall, that looks strange to me: we only need to delete the port to get rid of the FIP, but we do have to run `openstack floating ip create` to create it.

So is that ok and it just shows my lack of knowdegde in dpdk or is there something wrong there ?

Thanks,

sathlan (commented):
Hi,

the destruction of the FIP only need the destruction of the port, but the creation of the FIP needs the `openstack floating ip create`. Is that correct ?

Thanks,

sathlan:
lgtm, @cjeanner I think there's no more comment from the update team on this one.

Jaganathancse:
> Hi,
> 
> the destruction of the FIP only need the destruction of the port, but the creation of the FIP needs the `openstack floating ip create`. Is that correct ?
> 
> Thanks,

@sathlan  
correct,  FIP deletion part is also already handled in latest code. 
https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/update/templates/workload_launch.sh.j2#L198

so we can merge this PR.

viroel:
/lgtm

code change is protected with `workload_dpdk` var, which defaults to `false`. looks safe to merde

PR #2265, opened 2024-08-27
[pre-commit] Remove `bashate` hook

rebtoor:
Since we're now using `shellcheck` to lint and validate shell scripts, we can remove `bashate` which contains only a subsets of checks implemented by shellcheck.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,20 +12,6 @@ repos:
       - id: check-symlinks
       - id: debug-statements
 
-  - repo: https://github.com/openstack-dev/bashate.git
-    rev: 2.1.1
-    hooks:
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -40,11 +26,11 @@ repos:
         verbose: true
         # entry: bash -c "shellcheck "$@" || true" --
```
lewisdenny, line 27: Could you also please remove this line.

lewisdenny (changes requested):

.pre-commit-config.yaml
rebtoor: done!

lewisdenny (approved):
/approve

PR #2268, opened 2024-08-27
Correct shellcheck errors

cjeanner:
Since we're moving to shellcheck instead of bashate, let's get it green.

This patch fixes, but also disables some checks on specific lines. It
also integrate a "bashate" related ignore since the check is still in
place, we can remove it later.

### DIFF
--- a/docs/source/build-docs.sh
+++ b/docs/source/build-docs.sh
@@ -6,8 +6,9 @@ VENV_DIR="${DOCS_DIR}/_venv"
 BUILD_TYPE="${BUILD_TYPE:=static}"
 
 # Create a virtual environment and activate it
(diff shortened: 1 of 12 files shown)


docs/source/build-docs.sh
```diff
@@ -6,8 +6,9 @@ VENV_DIR="${DOCS_DIR}/_venv"
 BUILD_TYPE="${BUILD_TYPE:=static}"
 
 # Create a virtual environment and activate it
-python -m venv ${VENV_DIR} && source ${VENV_DIR}/bin/activate
-pip3 install -r ${DOCS_DIR}/doc-requirements.txt
+# shellcheck disable=SC1091
```
rebtoor, line 9: `SC1091` is annoying, imho we can safely add to ignore list.

PR #2270, opened 2024-08-27
fix(ci_setup): restore platform compatibility for CentOS 9 and EL 9

son-vyas:
- Re-added platform support for CentOS 9 and EL 9 in meta/main.yml
- Ensured the installation of required packages like bash-completion and tmux
Related-Bug: https://issues.redhat.com/browse/OSPCIX-432

### DIFF
--- a/roles/ci_setup/meta/main.yml
+++ b/roles/ci_setup/meta/main.yml
@@ -21,6 +21,20 @@ galaxy_info:
   license: Apache-2.0
   min_ansible_version: "2.14"
   namespace: cifmw
+   #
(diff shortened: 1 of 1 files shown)


lewisdenny (changes requested):
This was removed in https://github.com/openstack-k8s-operators/ci-framework/pull/2232

This file is used only for Ansible Galaxy metadata, and the platform section is no longer used by Ansible Galaxy which is why I removed it. More details here [1] but it's unrelated to the CIX you linked. 

[1] https://forum.ansible.com/t/role-meta-data-platform-key-broken/4651/10

PR #2274, opened 2024-08-28
Prefer the zuul items from run before the job to reproduce

cescgina:
When running the zuul reproducer, prefer to use the 'items' section from
the job running, to the one from the job it's reproducing. That will
allow to take a depends-on when running a testproject, and use the code
version that was used in the job to reproduce if not.

Tested downstream in TP 599


cjeanner:
/approve

We could successfully put this PR to use in ds tp 599, allowing to test #2263 as a Depends-On.

PR #2276, opened 2024-08-28
[polarion] Each result dir equals polarion test run

kopecmartin:
Each of the dirs found under the cifmw_polarion_jump_result_dir will be understood as a new test run in polarion. The polarion role will merge all .xml result files found inside each of the directories. Then it will iterate over all merged files and will upload them to polarion as a test run.

The reason behind this is that the same test can be executed several times within one job, each time with a different configuration or in a different cloud state (e.g. before/after service reboot). However, if polarion receives multiple results of the same test it will take into account only the last one.

The results from the tests that are executed repeatedly need to be in separate result directories so that they are reported as separate test runs.

### DIFF
--- a/roles/polarion/defaults/main.yml
+++ b/roles/polarion/defaults/main.yml
@@ -22,6 +22,7 @@ cifmw_polarion_basedir: "{{ cifmw_basedir | default(ansible_user_dir ~ '/ci-fram
 cifmw_polarion_jump_repo_dir: "{{ cifmw_polarion_basedir }}/polarion-jump"
 cifmw_polarion_jump_result_dir: "{{ cifmw_polarion_basedir }}/tests/{{ cifmw_run_test_role | default('tempest') }}/"
 cifmw_polarion_jump_repo_branch: "master"
+cifmw_polarion_testrun_title: "testrun-name"
 cifmw_polarion_use_stage: false
 cifmw_zuul_project: "tripleo-ci-internal"
 cifmw_zuul_url: "https://sf.hosted.upshift.rdu2.redhat.com/zuul/t/{{ cifmw_zuul_project }}/build"
--- a/roles/polarion/tasks/main.yml
+++ b/roles/polarion/tasks/main.yml
@@ -94,12 +94,27 @@
       ansible.builtin.debug:
         var: pylero_output.stdout_lines
 
-    - name: Merge result XML files
+    - name: Look for test result directories in artifacts directory
+      ansible.builtin.find:
+        paths: "{{ cifmw_polarion_jump_result_dir }}"
+        recurse: false
+        file_type: directory
+      register: result_dirs
+
+    - name: Merge result XML files of each directory
       ansible.builtin.shell:
         chdir: "{{ cifmw_polarion_jump_result_dir }}"
         cmd: >-
           source "{{ cifmw_polarion_jump_repo_dir }}/jump-venv/bin/activate" &&
-          junitparser merge {{ xml_files.files | json_query('[*].path') | join(' ') }} ./results_merged.xml
+          junitparser merge {{ item.path }}/*.xml {{item.path }}/results_merged.xml
+      loop: "{{ result_dirs.files }}"
+
+    - name: Look for test result XML files in artifacts directory
+      ansible.builtin.find:
+        paths: "{{ cifmw_polarion_jump_result_dir }}"
+        patterns: "results_merged.xml"
+        recurse: true
+      register: merged_xml_files
 
     - name: Update Test Runs
       ansible.builtin.shell:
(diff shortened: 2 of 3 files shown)


roles/polarion/tasks/main.yml
```diff
@@ -108,10 +123,11 @@
           source "{{ cifmw_polarion_jump_repo_dir }}/jump-venv/bin/activate" &&
           {{ cifmw_polarion_jump_repo_dir }}/jump-venv/bin/python jump.py
           --testrun-id={{ cifmw_polarion_testrun_id }}
```
lpiwowar, line 125: The change will **probably** not work as intended when `cifmw_polarion_testrun_id` is specified (no `create-new-test-run`). Update of the README.md might be a good idea if it is really true. It might surprise somebody later.

lpiwowar (approved):
Looks good to me :+1: 

One small comment.

roles/polarion/tasks/main.yml
kopecmartin: I'll update the README to reflect this behavior

kopecmartin:
this is ready, just note, it has to be merged around the same time as the downstream patch (ci-framework-jobs repo, MR 1022), ideally if the downstream patch was merged first and this one would follow within minutes.

roles/polarion/defaults/main.yml
```diff
@@ -22,6 +22,7 @@ cifmw_polarion_basedir: "{{ cifmw_basedir | default(ansible_user_dir ~ '/ci-fram
 cifmw_polarion_jump_repo_dir: "{{ cifmw_polarion_basedir }}/polarion-jump"
 cifmw_polarion_jump_result_dir: "{{ cifmw_polarion_basedir }}/tests/{{ cifmw_run_test_role | default('tempest') }}/"
 cifmw_polarion_jump_repo_branch: "master"
+cifmw_polarion_testrun_title: "testrun-name"
```
lpiwowar, line 25: Nit: `cifmw_polarion_testrun_title` is missing in README.

tosky:
This is a good result in the way it introduces the concept, but there is no direct correspondence between a stage and the results. If we just split by step, we may end up introducing more reports than needed. The way I see (and the way I thought when I raised the issue that led to this) is that we need a new concept of "phase" and aggregate the test results by phase.

kopecmartin:
> This is a good result in the way it introduces the concept, but there is no direct correspondence between a stage and the results. If we just split by step, we may end up introducing more reports than needed. The way I see (and the way I thought when I raised the issue that led to this) is that we need a new concept of "phase" and aggregate the test results by phase.

stage/phase .. all the same to me, test-operator (the role in ci-framework or test-operator itself) and this polarion role don't have all the context to tell what results belong to what stage or phase - test-operator works with stages, a phase as i understand is a concept above that, a test run in polarion is also a concept above that .. now, the mapping between stages, phases and tests runs isn't simple, one test run may include more stages and thus more phases. I don't see any part of the whole automation process that could tell what should be in which test run.

Therefore, I proposed this (honestly, I think it's clever :D), everything that is in one directory (the dir is most likely created by test-operator .. the stage) is gonna be one test run. Therefore it doesn't matter what a phase or a stage is, we don't have to define those terms and find a mapping between them, simply what's in one directory it's gonna be reported as one test run. We may end up with a bit more test runs then we need, so what? Polarion can handle that, if not, it's time to get rid of that. Having potentially more test runs than needed is the price to pay for being able to run the same test multiple times during the same job.

Of course, feel free to propose a different soltution if you have a better idea how to address this.

jparoly:
Can we get this approved and merged?  What is holding it up?

tosky:
> Can we get this approved and merged? What is holding it up?

As I mentioned, I don't think the proposed change address the issue that led to its creation. Even if this is merged, we will need to work to redo this differently because different test-operator stages doesn't mean different test runs. IMHO. Others will disagree.

kopecmartin:
>  different test-operator stages doesn't mean different test runs

I agree, this patch is not saying that. In fact, it was written this way so that we don't have to deal with that. But I already said it, twice, once in the commit message and once in the comment I wrote earlier. So far I haven't heard any different ideas or better solutions. As far as I know, this resolves the issue and can be merged right after we merge the downstream patch (which is btw left in the review queue .. seems like owners of that repo need to revisit their review processes).

jparoly:
Please lets get this resolved (sync up if necessary) so we can support adoption, etc.

roles/polarion/tasks/main.yml
```diff
@@ -108,10 +123,14 @@
...
-          --xml-file={{ cifmw_polarion_jump_result_dir }}/results_merged.xml
+          --testrun-title={{ cifmw_polarion_testrun_title }}-{{ loop_idx }}
+          --xml-file={{ item.path }}
           --update_testcases={{ cifmw_polarion_update_testcases | default(false) }}
-          --custom-fields cijoburl={{ cifmw_zuul_url }}/{{ zuul.build }}
+          --custom-fields cijoburl={{ cifmw_zuul_url }}/{{ zuul.build }},jobteststage={{ result_dirs.files[ loop_idx ].path | basename }}
           {{ cifmw_polarion_jump_extra_vars | default ('') }}
+      loop: "{{ merged_xml_files.files }}"
+      loop_control:
+        index_var: loop_idx
```
marios, line 133: @kopecmartin can you please have a look at my comment there https://issues.redhat.com/browse/OSPCIX-470?focusedId=25560054&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-25560054

I think this patch introduced a new bug.

if needed I can file a dedicated CIX for that (the referenced cix is tracking a different polarion reporting issue)

kopecmartin: thanks for bringing this up, fix has been proposed - https://github.com/openstack-k8s-operators/ci-framework/pull/2345

PR #2278, opened 2024-08-28
Add logging for `setup_tests` and `setup_molecule` targets in Makefile

rebtoor:
In order to debug weird behaviors during CI jobs, it's useful to have logs for those targets.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,11 +53,11 @@ new_role: create_new_role role_molecule ## Create a new Ansible role and related
 ##@ Setup steps
 .PHONY: setup_tests
 setup_tests: ## Setup the environment
-	bash scripts/setup_env
+	bash scripts/setup_env 2>&1 | tee $(LOG_DIR)/setup_env.log
 
 .PHONY: setup_molecule
 setup_molecule: setup_tests ## Setup molecule environment
-	bash scripts/setup_molecule
+	bash scripts/setup_molecule 2>&1 | tee $(LOG_DIR)/setup_molecule.log
 
 ##@ General testing
 .PHONY: tests


Makefile
```diff
@@ -83,18 +83,18 @@ pre_commit_nodeps: ## Run pre-commit tests without installing dependencies
 	pushd $(BASEDIR); \
 	git config --global safe.directory '*'; \
 	if [ "x$(USE_VENV)" ==  'xyes' ]; then \
-		${HOME}/test-python/bin/pre-commit run --all-files 2>&1 | ansi2txt | tee $(LOG_DIR)/pre_commit.log; \
+		${HOME}/test-python/bin/pre-commit run --all-files 2>&1 | ansi2txt | tee $(LOG_DIR)/$@.log; \
```
cjeanner, line 86: iirc those files were saved in zuul output. I think it's related to the pod run, it might be good to ensure they are still around...

cjeanner (commented):
/hold need to ensure we didn't lose the existing logs in the process. I don't 100% remember how I implemented it in the zuul pods though...

Makefile
cjeanner: 'k, so the artifacts link is broken indeed:
https://softwarefactory-project.io/zuul/t/rdoproject.org/build/5dc22b012af340058d46be298cecbc29
click on "Artifacts", then "[pre_commit logfile](https://logserver.rdoproject.org/78/2278/ab80f77ebfbd745f4e61b7a67198774dc433b1c6/github-check/cifmw-pod-pre-commit/5dc22b0/pre_commit.log)"

This comes from:
https://github.com/openstack-k8s-operators/ci-framework/blob/5fa99bbb78d4cab5246d6c711bf9c08e19043a1d/ci/playbooks/pod-jobs.yml#L44-L53

rebtoor:
recheck

PR #2279, opened 2024-08-29
Add metric collection via njmon

cjeanner:
This patch adds a new role to manage "njmon". This tool allows to
collect metrics on the hypervisor so that we can check how the resources
were used on it.

By toggling `cifmw_monitoring` boolean to `true`, you will enable its
installation at the very beginning of the reproducer.yml run, as well as
the graphs generation at the very end of the run.

This can help understanding potential issues, for instance related to
memory shortage (oom-killer), slow I/O (disks) or clogged CPU
(overprovisioning).

One of the advantages of njmon, in addition to being really small and
light, is its capacity to send data to a remote InfluxDB. This would
allow you to display the graphs in realtime in Grafana, for instance.

### DIFF
--- a/docs/dictionary/en-custom.txt
+++ b/docs/dictionary/en-custom.txt
@@ -107,6 +107,7 @@ dataplane
 dataplanedeployments
 dataplanenodeset
 dataplanenodesets
+dataset
 dcn
 dd'd
 ddr
@@ -206,6 +207,7 @@ igmp
 igogicbjyxbzig
 ihbyb
 img
+influxdb
 ingressvips
 ini
 init
@@ -319,6 +321,8 @@ nfs
 nftables
 nic
 nigzpbgugpsavdmfyl
+njmon
+njmonchart
 nlcggvjgnsdxn
 nmcli
 nmstate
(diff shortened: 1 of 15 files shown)


cjeanner:
An example of the metrics is visible in the molecule output, for instance here:
https://logserver.rdoproject.org/79/2279/d468184914fc9944da8cd68873f78812aabe69cd/github-check/cifmw-molecule-njmon/4955177/ci-framework-data/artifacts/njmon_stats/np0004868171_20240829_0631.html

You can then click on the various button at the top to see the generated graphs.
The job is gathering just 30 seconds, to allow building the graph, so it's not that "beautiful", but at least you can get an idea of the data it collects.

rebtoor (changes requested):
I left some comments but the problem with this PR is not its scope, which i strongly support, but rather the tool of choice.

Honestly I don't like this tool: it doesn't provide the bare minimum requirement for a modern supply chain requirement (no gpg signature on downloading and even no checksum for downloaded file), it's clearly written for a pre systemd world and it feels poorly integrated in the system. 

We can find a better solution for the problem, probably something already present in EL repo or, at least, well packaged and, above all, modern. :)

cjeanner:
(brain dropping before I forget things)
Among packages alternatives, I can list:
- collectd (but it's probably too big for the usage, and its configuration isn't that straightforward)
- munin

Those tools can write on disk for later visualization, which is interesting in many cases. In the njmon case, we can get the resource usage directly from the zuul logs. Keeping this functionality seems a good idea imho.

Another solution would be Node Exporter, but it requires some more things than just one single service, and more configuration. Also not sure about the footprint.

Let's discuss that topic once we get more ppl on board. At least, it sparked some discussion and interest on the metric topic :)

PR #2281, opened 2024-08-29
Add cifmw_set_openstack_containers_prefix_filter variable to set_openstack_cotaniners

ciecierski:
Add cifmw_set_openstack_containers_prefix_filter variable to replace it in update_env_vars template with cifmw_set_openstack_containers_prefix variable. This will enable to replace prefix of openstack containers if prefixes are not following the same convention between container registries.


raukadah:
recheck network issue between edpm job and content provider which is no longer seen.

raukadah:
recheck molecule job random issue

raukadah:
/lgtm

Tested it in post adoption upgrade job, works fine!

PR #2283, opened 2024-08-29
Use libvirt_manager extra_disks for Ceph OSDs

fultonj:
Update Ceph playbook to not call `cifmw_block_device` role if `cifmw_ceph_spec_data_devices` is set. This parameter should only be overridden if different disks than the ones created by the `cifmw_block_device` role are going to be used.

Document an example of how to use the `libvirt_manager` role extra_disks_ parameter to create EDPM nodes with extra disks and how to override `cifmw_ceph_spec_data_devices` to use those disks.

Following the example provides an alterative to get better storage performance but this option is only available for jobs which use the reproducer and thus the `cifmw_block_device` role cannot be removed.

Also update scenarios/reproducers/va-hci.yml to add extra_disks.

Jira: https://issues.redhat.com/browse/OSPRH-9489

(cherry-picked from f75fff6dbdb9c46b56857d043aacf5422f2db5ca)

### DIFF
--- a/docs/dictionary/en-custom.txt
+++ b/docs/dictionary/en-custom.txt
@@ -364,6 +364,7 @@ operatorgroup
 opn
 orchestrator
 osd
+osds
 osp
 osprh
 otz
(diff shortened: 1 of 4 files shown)


cjeanner:
/approve
/lgtm

This is a clean backport, needed to make ceph based deployment more reliable in downstream QE.

PR #2285, opened 2024-08-30
Introduce two new playbooks for adoption infra

cjeanner:
create-infra.yml allows to create all the VMs related to adoption
scenarios.
deploy-ocp.yml allows to get the OCP cluster in, independently to the
rest of the Framework. You usually want to run that playbook after the
"create-infra.yml", but it's not mandatory.

The infrastructure and networks changes needed for adoption are to be
pushed in a scenario file, either under scenarios/adoption/SCENARIO.yml,
or in a 3rd party repository (mostly for Downstream).

All of the VMs are shut off after the create-infra.yml is finished, and
only the OCP cluster members are running after the deploy-ocp.yml run.


cjeanner:
recheck

playbooks/unique-id.yml
```diff
@@ -0,0 +1,56 @@
...
+      ansible.builtin.copy:
+        dest: "{{ _unique_id_file }}"
+        content: "{{ cifmw_run_id | default(_unique_id) | lower }}"
+
+    # Since the user might pass their own run ID, we can just consume it.
+    # If, for a subsequent run, the user doesn't pass the run ID, we will
+    # just get it from the file and consume it.
+    - name: Load existing run ID
+      when:
+        - cifmw_run_id is undefined
```
marios, line 47: is this right? this is when user is specifying their own unique ID, so does that imply that cifmw_run_id is undefined?

playbooks/unique-id.yml
```diff
@@ -0,0 +1,56 @@
...
+      block:
+        - name: Slurp unique id file if needed
+          register: _unique_id_content
+          ansible.builtin.slurp:
+            path: "{{ _unique_id_file }}"
```
marios, line 52: and following from above comment, when cifmw_run_id is undefined does it imply that _unique_id_file is there? 

should this be part of the when conditional i.e. when  _unique_id_stat.stat.exists

playbooks/unique-id.yml
cjeanner: User would define `cifmw_run_id` as a parameter themselves.
Also, this isn't really part of the review - the "unique-id.yml" playbook is just "moving a block" from the reproducer.yml playbook, for code re-usability and de-duplication (it's used in the new plays as well).

playbooks/unique-id.yml
cjeanner: it nested (2-level blocks)

cescgina:
I've tested this a few times and it seems to work well if run in conjuntion with https://github.com/openstack-k8s-operators/ci-framework/pull/2327. I think we can merge it for now and iterate if needed in the future

marios (approved):

PR #2286, opened 2024-08-30
name space doesn't available

eshulman2:


eshulman2:
recheck

hooks/playbooks/ceph-bm.yml
```diff
@@ -8,7 +8,7 @@
         PATH: "{{ cifmw_path }}"
       ansible.builtin.command:
         cmd: >-
-          oc get OpenStackBaremetalSet -n "{{ _namespace}}"  -o yaml
+          oc get OpenStackBaremetalSet -n openstack  -o yaml
```
rlandy, line 11: would it be safer to default the value "{{ _namespace|default('openstack') }}"?

viroel: IIUC, the openstack namespace should be available with `{{ namespace }}`  and not `{{ _namespace }}`, according with [this](https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/run_hook/tasks/playbook.yml#L34-L35)
So I think that would be better to replace with:
`"{{ namespace|default('openstack') }}"`

eshulman2: ok sounds good I'll test it like that

tosky: In any case, if the change is still required, please explain the reason for the change in the commit message.
https://chris.beams.io/posts/git-commit/

eshulman2:
recheck

PR #2287, opened 2024-08-30
Introduce cifmw_test_operator_insecure parameter

lpiwowar:
In order to be able to consume certain functionalities in the CI the test pods need to be spawned with:

- allowPrivilegedEscalation: true and
- default Linux capabilities.

These parameters are turned off by default on the side of the test-operator. However, in CI we need to use all test-operator functionalities. If this parameter is set to false then the following operations won't work:

- installation of external rpms and
- certain set of tobiko tests


lpiwowar:
https://github.com/openstack-k8s-operators/ci-framework/pull/2295

PR #2289, opened 2024-09-02
Store update related repo_setup repos in update_repository

raukadah:
Currently we use repo-setup default repositories location to store repo during pre-update and during update process.

There is no way to find out from what repos update happens. This pr changes the cifmw_repo_setup_output to update_repositories directory. It will help to find out what repos used during update.


ciecierski (approved):

viroel (approved):
Going to be used for adoption jobs, safe to merge since doesn't impact current jobs.

PR #2292, opened 2024-09-03
Use cifmw_architecture_scenario to set proper automation file

cjeanner:
https://github.com/openstack-k8s-operators/architecture/pull/375 splits
the default vars into different scenario files. We need to fix the
reproducer to adapt this change.

This pull-request uses cifmw_architecture_scenario to set proper automation file.
It drops cifmw_arch_automation_file params from scenario files.

Depends-On: https://github.com/openstack-k8s-operators/architecture/pull/375

Signed-off-by: Chandan Kumar (raukadah) <raukadah@gmail.com>

### DIFF
--- a/docs/source/usage/01_usage.md
+++ b/docs/source/usage/01_usage.md
@@ -44,8 +44,6 @@ are shared among multiple roles:
 - `cifmw_ssh_keysize`: (Integer) Size of ssh keys that will be injected into the controller in order to connect to the rest of the nodes. Defaults to 521.
 - `cifmw_architecture_repo`: (String) Path of the architecture repository on the controller node.
   Defaults to `~/src/github.com/openstack-k8s-operators/architecture`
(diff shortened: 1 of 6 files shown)


cjeanner:
/approve

self-approving. This is the 1:1 match of @raukadah patch that got merged slightly too fast. Keeping it as draft to ensure it doesn't merge before we've been able to test all the things.

raukadah:
> /approve
> 
> self-approving. This is the 1:1 match of @raukadah patch that got merged slightly too fast. Keeping it as draft to ensure it doesn't merge before we've been able to test all the things.

I can confirm this change has passed on VAHCI, VA2, OSASinfra and BGP, We are good to merge it.

cjeanner:
- Zuul's loaded right now
- molecule failed on reproducer due to totally unrelated issue (repository issue)
- that patch is a 1:1 match with the original one, it was all green
- that patch went through extensive testing DS.

PR #2293, opened 2024-09-03
Listallowed telemetry tests

vyzigold:
tempest.api.telemetry.* actually doesn't exist. The only place with telemetry related tempest api tests is in telemetry_tempest_plugin.aodh.api


lewisdenny (approved):
/approve

PR #2296, opened 2024-09-03
Allow to NOT connect some VMs to the public network

cjeanner:
Until now, we assumed every VM were connected to the "public network".
This seems to not be right, so we have to prevent patching the
networking-definition for all the nodes.

This patch checks if the VM group/type is connected to the public
network, so that we generate the DHCP configuration only for the ones
that need it.
This should prevent some corner cases.


hjensas (commented):
/lgtm

PR #2300, opened 2024-09-03
Ensure we don't hardcode "compute-0" anymore

cjeanner:
Depending on what's providing the infra, the "compute-0" might be
non-existing, especially now that we provide unique hostnames.

This patch should allow to be slightly more dynamic regarding the
inventory names, fixing the issue.


cescgina (commented):
/approve

PR #2301, opened 2024-09-03
Configure Ceph RGW with a object storage network

fultonj:
When Ceph RGW is used, an endpoint for object storage is hosted not in a pod on k8s but on an EDPM node. Thus, a service hosted on an EDPM node will need to be accessed from a separate network.

This patch adds the object storage network (storageobj) with VLAN 25 and range 172.22.0.0/24 in the reproducers networking-definition.yml. It also adds a multus range for this network so that the Tempest pod can access this network for testing. The object storage network is added to the OCP nodes for the same reason.

The Ceph playbook cifmw_cephadm_rgw_network parameter may then be set to the new network so that object storage may be configured on a separate network. If this network is parameter is not set, then the storage network is used.

Jira: https://issues.redhat.com/browse/OSPRH-6675

### DIFF
--- a/ci/playbooks/files/networking-env-definition.yml
+++ b/ci/playbooks/files/networking-env-definition.yml
@@ -151,6 +151,15 @@ instances:
                 parent_interface: eth1
                 skip_nm: false
                 vlan_id: 22
+            swift:
+                interface_name: eth1.25
+                ip_v4: 172.22.0.100
+                mac_addr: '52:54:00:0b:1c:e7'
+                mtu: 1500
+                network_name: swift
+                parent_interface: eth1
+                skip_nm: false
+                vlan_id: 25
     compute-1:
         hostname: compute-1
         name: compute-1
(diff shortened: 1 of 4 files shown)


roles/kustomize_deploy/tasks/execute_step.yml
```diff
@@ -204,6 +204,18 @@
...
+          environment:
+            PATH: "{{ cifmw_path | default(ansible_env.PATH) }}"
+          ansible.builtin.command:
+            chdir: "{{ _chdir }}"
+            cmd: "oc version"
+
+        - name: "show oc version"
+          ansible.builtin.debug:
+            var: _oc_version
+
```
fultonj, line 218: https://github.com/openstack-k8s-operators/architecture/pull/386#issuecomment-2327694656

ci/playbooks/files/networking-env-definition.yml
```diff
@@ -336,6 +363,15 @@ instances:
                 parent_interface: enp6s0
                 skip_nm: false
                 vlan_id: 22
+            swift:
+                interface_name: enp6s0.25
+                ip_v4: 172.22.0.10
+                mac_addr: '52:54:00:18:a0:g6'
```
fultonj, line 369: `g` is not hex.

fultonj:
@abays @fmount @cjeanner @pablintino 

A test project (uni03gamma-rhel9-rhoso18.0-titan105-ceph) which depends on this patch is green so we are safe to merge.

rebtoor (approved):
/approve

fmount:
Thank you @fultonj @rebtoor @pablintino !

PR #2304, opened 2024-09-04
Use if instead of elif to set container overrides

raukadah:
Currently set_openstack_containers either updates the operator environment variables or apply the overrides. It does not perform both the things together.

In downstream for post-adoption job, we want to update the environment variable as well as overrides. By changing the condition to if, we can achieve the same.


roles/set_openstack_containers/templates/update_env_vars.sh.j2
```diff
@@ -32,11 +32,12 @@ export PATH="{{ cifmw_path }}"
 {%     set _container_address = _container_registry + '/' + _container_prefix + _container_name + ':' + cifmw_set_openstack_containers_tag -%}
 {%     set _ = containers_dict.update({_container_env: _container_address}) -%}
 {#   This section overrides the non openstack services containers #}
-{%   elif cifmw_set_openstack_containers_overrides.keys() | length > 0 -%}
```
ciecierski, line 35: You should close previous if before opening new if
`{% endif %}`

ciecierski (approved):

cjeanner (approved):
/approve

PR #2305, opened 2024-09-04
[test-operator] Add the `wait_timeout` parameter to cleanup tasks

lkuchlan:
We recently encountered a timeout issue during the deletion of the test-operator. The default timeout is set to 120 seconds. By adding the `wait_timeout` parameter to the cleanup tasks, we can extend this time frame.

### DIFF
--- a/roles/test_operator/tasks/cleanup.yml
+++ b/roles/test_operator/tasks/cleanup.yml
@@ -25,6 +25,7 @@
     name: test-operator
     namespace: "{{ cifmw_test_operator_namespace }}"
     wait: true
+    wait_timeout: 600
 
 - name: Delete CatalogSource
   kubernetes.core.k8s:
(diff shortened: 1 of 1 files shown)


roles/test_operator/tasks/cleanup.yml
```diff
@@ -25,6 +25,7 @@
     name: test-operator
     namespace: "{{ cifmw_test_operator_namespace }}"
     wait: true
+    wait_timeout: 600
```
lpiwowar, line 28: I was thinking about parametrizing this but I see that in ci-framework they use the 600 / 300 secs timeout without any parameter quite often (for example - [1]). 

[1] https://github.com/openstack-k8s-operators/ci-framework/blob/7f82829499d69b5fbeead0a521530ffa29c1cd85/roles/kustomize_deploy/tasks/cleanup.yml#L35

lpiwowar (approved):
Looks good to me!

rlandy:
\lgtm

PR #2309, opened 2024-09-04
Add some more validations

cjeanner:
Ensure we're consuming projects from the same source.

### DIFF
--- a/roles/reproducer/tasks/validations.yml
+++ b/roles/reproducer/tasks/validations.yml
@@ -83,3 +83,58 @@
       cifmw_network_dnsmasq_config is set and not empty. This feature is NOT
       supported anymore. Please contact the CI Framework team to check your
       configuration, and see if you really need to pass anything to dnsmasq.
+
+- name: Verify cifmw_arch_automation_file is undefined
+  ansible.builtin.assert:
+    quiet: true
+    that:
+      - cifmw_arch_automation_file is undefined
+    msg: >-
+      cifmw_arch_automation_file is defined. This parameter is NOT supported
+      anymore. You can use cifmw_architecture_automation_file instead, providing
+      the absolute path to the automation file. Note that you shouldn't need to
+      set it at all.
+
+# Ensure we don't mix upstream and midstream content
+- name: Ensure we're consuming consistent project sources
+  when:
+    - cifmw_reproducer_validate_project_sources | default(true) | bool
+  vars:
+    _custom_rep_dict: >-
+      {{
+        cifmw_reproducer_repositories |
+        items2dict(key_name='dest', value_name='src')
+      }}
+    # Don't consider local paths, only remotes.
+    # This should prevent failing whenver we sync local repositories
+    # while developping with the Framework.
+    _projects_src: >-
+      {{
+        cifmw_reproducer_default_repositories |
+        items2dict(key_name='dest', value_name='src') |
+        combine(_custom_rep_dict) |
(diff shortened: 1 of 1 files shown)


roles/reproducer/tasks/validations.yml
```diff
@@ -83,3 +83,58 @@
...
+      {{
+        cifmw_reproducer_repositories |
+        items2dict(key_name='dest', value_name='src')
+      }}
+    # Don't consider local paths, only remotes.
+    # This should prevent failing whenver we sync local repositories
+    # while developping with the Framework.
+    _projects_src: >-
+      {{
+        cifmw_reproducer_default_repositories |
```
viroel, line 113: shouldn't we worry about default values for this var here? IIRC it defaults to upstream

roles/reproducer/tasks/validations.yml
```diff
@@ -83,3 +83,58 @@
...
+        var: _projects_src
+
+    - name: Debug _src_name
+      ansible.builtin.debug:
+        var: _src_name
+
+    - name: Assert everything comes from the same source
+      ansible.builtin.assert:
+        that:
+          - _src_name | length == 0
```
viroel, line 136: `length == 1` , no?

roles/reproducer/tasks/validations.yml
cjeanner: we need to extend the whole thing. For now, it doesn't check the value, but only the consistency. It's weak, to be fair, and we probably want/need something better here, to ensure we're in the right context.

I don't know if we could get anything from within zuul env showing "where" we run, and create some "validation map" stating something like this:
```yaml
cifmw_src_base_validation:
  upstream:
    src_base: https://github.com/openstack-k8s-operators
    branch: main
  downstream_trunk:
    src_base: /home/zuul/src/....
    branch: trunk  # whatever it should be
  downstream_18_proposed:
    src_base: https://github.com/openstack-k8s-operators.
    branch: 18.0-proposed  # whatever it should be
```

That would really consolidate the whole thing. Any thoughts? Of course, it must be extensible in an easy way, to be future proof (new branch, new entry in the mapping).

roles/reproducer/tasks/validations.yml
cjeanner: woops, right.

cjeanner:
IMHO it's a good opportunity to extract the validations in their own role. @lewisdenny you created the initial validations.yml in the reproducer role - any thoughts on extracting them all in a "cifmw_validations" or something like that, with dedicated tasks files maybe? 
For the git checks, we could even leverage some deeper inspections with re-usable tasks files, loops and all.

Of course, we'll have to keep some parameters allowing to disable the validations - maybe per "validation groups". We can discuss this and, if you're willing to tackle that, we can do some proper handover :)

lewisdenny:
> IMHO it's a good opportunity to extract the validations in their own role. @lewisdenny you created the initial validations.yml in the reproducer role - any thoughts on extracting them all in a "cifmw_validations" or something like that, with dedicated tasks files maybe? For the git checks, we could even leverage some deeper inspections with re-usable tasks files, loops and all.
> 
> Of course, we'll have to keep some parameters allowing to disable the validations - maybe per "validation groups". We can discuss this and, if you're willing to tackle that, we can do some proper handover :)

Yeah it would be nice to support some different types/groups of validations, a role would be best. Let's get a Jira card created and we can flesh it out :)

PR #2310, opened 2024-09-04
Ensure we can re-run the framework without issue on config_drive

cjeanner:
Until now, if we were to re-run the same deploy without any cleanup, it
would fail on the geniso from config_drive role because it can't
override the existing ISO.

Adding the "creates" parameter to the ci_script corrects this issue.
The assertions also ensures we don't try to edit an existing ISO - this
wasn't supported before, but since the run crashed, the user would know
there was an issue. The assertion makes it clear now.

The next issue is related to the usage of the ISO from within
libvirt_manager: the ISO isn't removed once it's attached, and
re-running fails since it's already attached to the VM.

Checking for the attached volumes, and removing the useless "removes"
from the command (converted to shell) ensures it's working fine.

This patch also attempts to fix the molecule jobs for libvirt_manager:
from time to time, it might crash due to the incomplet cleanup done in
the common "prepare" stage: the ocp_volumes pool is still present, but
the underneath directory is absent, leading to a crash whenever we want
to refresh the pools:
```
msg: 'cannot open directory ''/opt/basedir/ocp_volumes'': No such file or directory'
```

### DIFF
--- a/roles/config_drive/tasks/main.yml
+++ b/roles/config_drive/tasks/main.yml
@@ -41,18 +41,21 @@
     mode: "0755"
 
 - name: Generate meta-data
+  register: _meta_data_change
   ansible.builtin.template:
     src: "meta-data.j2"
     dest: "{{ cifmw_config_drive_instancedir }}/meta-data"
     mode: '0644'
 
 - name: Generate user-data
+  register: _user_data_change
   ansible.builtin.template:
     src: "user-data.j2"
     dest: "{{ cifmw_config_drive_instancedir }}/user-data"
     mode: '0644'
 
 - name: Generate network-data
+  register: _net_data_change
   when:
     - "cifmw_config_drive_networkconfig is defined"
     - "cifmw_config_drive_networkconfig is not none"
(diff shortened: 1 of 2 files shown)


hjensas (commented):
/lgtm

I think this is ok. There is one "gotcha" that might trip users off with this behaviour - if the use changed the data going into the config drive that data is never updated. It was not before either, but it would fail in either case of no-change/change.

Would a better approach be to condition the generate of the ISO based on the previous tasks `Generate meta-data`, `Generate user-data` and ` Generate network-data`? - i.e only try to re-create the ISO if one of those changed? (And if they changed, and the ISO already exist raising the error is not a bad thing?)

cjeanner:
> /lgtm
> 
> I think this is ok. There is one "gotcha" that might trip users off with this behaviour - if the use changed the data going into the config drive that data is never updated. It was not before either, but it would fail in either case of no-change/change.
> 
> Would a better approach be to condition the generate of the ISO based on the previous tasks `Generate meta-data`, `Generate user-data` and ` Generate network-data`? - i.e only try to re-create the ISO if one of those changed? (And if they changed, and the ISO already exist raising the error is not a bad thing?)

I thought about the "what about data change". But since the ISO is associated to the VM (at least in the libvirt_manager usage), it can't be overridden/recreated. We can try add some assertions? Like check if the ISO exists, and if the dataset changed, then fail with a nice message?

I have to check molecule anyway, there seems to be some issues with a specific directory, I thought I corrected it. Lemme know your thoughts about adding the assertion, and I can integrate it.

cjeanner:
recheck

hjensas (commented):
/lgtm

PR #2311, opened 2024-09-04
Add missing cifmw_architecture_automation_file

cescgina:
After a recent change [1] we define the architecture automation to be used in the reproducer, based on the VA used. The HCI VA crc job does not use the reproducer role, so it does not get that parameter defined, thus we need to add it to the job definition.

[1] https://github.com/openstack-k8s-operators/ci-framework/pull/2292

### DIFF
--- a/playbooks/06-deploy-architecture.yml
+++ b/playbooks/06-deploy-architecture.yml
@@ -128,6 +128,18 @@
                 msg: >-
                   Error detected. Check debugging output above.
 
+    - name: Set cifmw_architecture_automation_file if not set before
+      when: cifmw_architecture_automation_file is not defined
+      ansible.builtin.set_fact:
+        cifmw_architecture_automation_file: >-
+          {{
+            (
+              cifmw_architecture_repo | default(ansible_user_dir+'/src/github.com/openstack-k8s-operators/architecture'),
+              'automation/vars',
+              cifmw_architecture_scenario~'.yaml'
+            ) | ansible.builtin.path_join
+          }}
+
     - name: Load architecture automation file
       tags:
         - edpm_deploy


zuul.d/architecture-jobs.yaml
```diff
@@ -53,6 +53,7 @@
       cifmw_openshift_setup_skip_internal_registry_tls_verify: true
       cifmw_networking_env_def_file: "networking-env-definition.yml"
       cifmw_networking_mapper_networking_env_def_path: "/etc/ci/env/networking-environment-definition.yml"
+      cifmw_architecture_automation_file: "{{ cifmw_architecture_repo }}/automation/vars/hci.yaml"
```
cjeanner, line 56: err... nope? the reproducer is supposed to build the filename based on `cifmw_architecture_scenario` ... ?

cescgina: yes, but this is for the crc VA job which doesn't use the architecture role

cjeanner: oh. ok, "reproducer" role you mean. Indeed, that parameter is injected by the `configure_controller.yml` nested in the reproducer.

Maybe use the `cifmw_architecture_scenario` instead of hard-coding the `hci.yaml` ? That's my only concern here. Or move that specific parameter to `cifmw-architecture-validate-hci` job definition ?

cescgina: fair enough, I though I added it to the child job which is hci specific, let me make that more general

cescgina: updated  the PR to use `cifmw_architecture_scenario`

cjeanner:
I don't see the job being triggered. Care to either ensure at least crc-hci is triggered, or get a test project to be sure the parameter properly "leaks" between parent and child?

cescgina:
> I don't see the job being triggered. Care to either ensure at least crc-hci is triggered, or get a test project to be sure the parameter properly "leaks" between parent and child?

good point, I was testing it downstream yesterday, but  will trigger it here as well to verify the new version

playbooks/06-deploy-architecture.yml
```diff
@@ -128,6 +128,18 @@
...
+      ansible.builtin.set_fact:
+        cifmw_architecture_automation_file: >-
+          {{
+            (
+              cifmw_architecture_repo | default(ansible_user_dir+'/src/github.com/openstack-k8s-operators/architecture'),
+              'automation/vars',
+              cifmw_architecture_scenario~'.yaml'
+            ) | ansible.builtin.path_join
+          }}
+      when: cifmw_architecture_automation_file is not defined
```
cjeanner, line 141: usually the `when` is put after the `name` to make reading slightly easier

cescgina: ack, will change once the current CI run is finished

PR #2313, opened 2024-09-04
Don't harcode "compute-0" reference in hci_prepare

cjeanner:
With the new unique hostname feature, we can't hardcode "compute-0"
anymore. Wile it still works in Zuul multinode, it breaks on other runs
driven by the "reproducer.yml" playbook.

### DIFF
--- a/roles/hci_prepare/tasks/load_parameters.yml
+++ b/roles/hci_prepare/tasks/load_parameters.yml
@@ -25,14 +25,24 @@
 
 - name: Try/catch pattern for debugging
   block:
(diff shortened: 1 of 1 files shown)


cescgina (commented):
/approve

raukadah (commented):
/approve 

hci job passed!
https://softwarefactory-project.io/zuul/t/rdoproject.org/buildset/0bfbde6e60fd45428a75a4b5ae813d7b

podified-multinode-hci-deployment-crc SUCCESS

PR #2317, opened 2024-09-05
Introduce cifmw_reproducer_default_repositories_override_checkout var

raukadah:
https://github.com/openstack-k8s-operators/ci-framework/pull/1791 already supports passing override-checkout to
cifmw_reproducer_default_repositories vars.

Let's expand this functionality to pass override_checkout to all the repo. It assumes that all the repos have same default branch name.


raukadah:
no longer needed!

PR #2318, opened 2024-09-05
Do not add the unique ID to VM names in specific case

cjeanner:
The unique hostname is needed for architecture driven deployment. When a
run is using install_yamls, it can't properly kustomize the data since
the names are all hard-coded in that external project.

In order to avoid issues, this patch limits the unique hostname to
scenarios that are 100% handled by the Framework, while keeping the old,
non-unique names for install_yamls usage.


cjeanner:
recheck

cescgina (commented):
/approve

PR #2320, opened 2024-09-05
Ensure we deepscrube between tests

cjeanner:
There are a couple of tasks that aren't triggered over standard
cleaning. Ensuring the "always" tag is set should ensure we have a
really clean environment before each stage.

This allowed to find and fix an issue in the way pools were cleaned. Due
to a wrong parameter name, the `ocp_volumes` was dangling after a
deepscrub. This is now properly fixed.


hjensas (commented):
/lgtm

PR #2321, opened 2024-09-05
Integrate set_openstack_containers in validated architecture playbook

ciecierski:
set_openstack_containers role is used to update openstack container images using openstack version cr in update jobs. Let's use the same to update the containers in va.

It will be used in downstream VA jobs to use RHEL openstack services container.

### DIFF
--- a/playbooks/06-deploy-architecture.yml
+++ b/playbooks/06-deploy-architecture.yml
@@ -227,6 +227,14 @@
         - update_containers
         - edpm_bootstrap
 
+    - name: Update containers in deployed OSP operators using set_openstack_containers role
+      when: cifmw_set_openstack_containers | default(false) | bool
+      ansible.builtin.include_role:
+        name: set_openstack_containers
+      tags:
+        - set_openstack_containers
+        - edpm_bootstrap
+
     - name: Configure LVMS Storage Class
       ansible.builtin.include_role:
         name: ci_lvms_storage
--- a/docs/source/usage/01_usage.md
+++ b/docs/source/usage/01_usage.md
@@ -35,6 +35,7 @@ are shared among multiple roles:
 - `cifmw_deploy_edpm`: (Bool) toggle deploying EDPM. Default to false.
 - `cifmw_use_vbmc`: (Bool) Toggle VirtualBMC usage. Defaults to `false`.
 - `cifmw_use_sushy_emulator`: (Bool) Toggle Sushy Emulator usage. Defaults to `true`.
+- `cifmw_set_openstack_containers`:  (Bool) Run set_openstack_containers role during deployment to update openstack containers. Defaults to `false`.
 - `cifmw_sushy_redfish_bmc_protocol`: (String) The RedFish BMC protocol you would like to use with Sushy Emulator, options are `redfish` or `redfish-virtualmedia`. Defaults to `redfish-virtualmedia`
 - `cifmw_config_nmstate`: (Bool) toggle NMstate networking deployment. Default to false.
 - `cifmw_config_bmh`: (Bool) toggle Metal3 BareMetalHost CRs deployment. Default to false.


roles/reproducer/templates/script.sh.j2
```diff
@@ -12,5 +12,8 @@ ansible-playbook -i ~/ci-framework-data/artifacts/zuul_inventory.yml \
 {% for extravar in extravars %}
   -e {{ extravar }} \
 {% endfor %}
+{% if cifmw_reproducer_play_skip_tags | length > 0 -%}
+  --skip-tags ' + (cifmw_reproducer_play_skip_tags | join(',')) \
+{% endif -%}
```
sdatko, line 17: also please remove `-` from here and line 15 :pray:

playbooks/06-deploy-architecture.yml
```diff
@@ -197,6 +197,15 @@
...
 
+    - name: Update containers in deployed OSP operators using set_openstack_containers role
+      vars:
+        cifmw_set_openstack_containers_extra_vars: "{{ cifmw_edpm_prepare_extra_vars }}"
+      ansible.builtin.include_role:
+        name: set_openstack_containers
+      tags:
+        - set_openstack_containers
+        - edpm_bootstrap
+
```
sathlan, line 208: I see.  In the update job then you just add:

```yaml
      cifmw_reproducer_play_skip_tags:
        - update_containers
```
and you get the right type of deployment, but by default nothing change of other uni job, they still deploy the "hotfix" kind of deployment.

sathlan (approved):
Nice.

playbooks/06-deploy-architecture.yml
```diff
@@ -197,6 +197,15 @@
         - update_containers
         - edpm_bootstrap
 
+    - name: Update containers in deployed OSP operators using set_openstack_containers role
+      vars:
+        cifmw_set_openstack_containers_extra_vars: "{{ cifmw_edpm_prepare_extra_vars }}"
```
pablintino, line 202: edpm_prepare vars are not used in VA/DT deployments

playbooks/06-deploy-architecture.yml
```diff
@@ -197,6 +197,15 @@
...
+      ansible.builtin.include_role:
+        name: set_openstack_containers
+      tags:
+        - set_openstack_containers
```
pablintino, line 206: A few comments:
- You assume we call deploy-edpm only using the reproducer, that with this change, will inject the tag in the skip list
- You assume that we want the role to be executed by default, that's the meaning of adding a tag here without a when condition.
- This skip strategy is not aligned with the rest of "skip flags" we have all over the framework. Please, use a flag like `cifmw_update_containers` with a sane `false` default.

pablintino (changes requested):

playbooks/06-deploy-architecture.yml
pablintino: @ciecierski sorry mate, I may miss-expressed what I wanted. `cifmw_update_containers` was just an example of the kind of flag we usually do, for this new task create a new one called: `cifmw_set_openstack_containers` that defaults to false. Yes, false, as the regular behavior we had (and we still want) is to preserve what's deployed. If someone wants to patch the images all he/she needs is to make the flag true.

pablintino (changes requested):

playbooks/06-deploy-architecture.yml
pablintino: PS: We doc those vars [here](https://github.com/openstack-k8s-operators/ci-framework/blob/main/docs/source/usage/01_usage.md), so add the new one in that file too.

ciecierski: @pablintino I added cifmw_update_containers to not create more vars, but then I realized that there are deployments jobs which have cifmw_update_containers set to false. Those deployment jobs are using containers from compose. I'm going to add cifmw_set_openstack_containers for  set_openstack_containers role include. 
I'm not sure if I should remove cifmw_update_containers from update_containers include. update_containers role default value for  cifmw_update_containers is false, but it used only for applying generated template: https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/update_containers/tasks/main.yml#L30
Do you think we want to generate template always? Should I remove when: cifmw_update_containers or it can stay?

playbooks/06-deploy-architecture.yml
ciecierski: cifmw_reproducer_play_skip_tags mechanism is removed after review from Pablo

playbooks/06-deploy-architecture.yml
pablintino: @ciecierski I thought about telling you to do so, but I'd prefer to keep your change as scoped to your issue as posible, so no worries, we will do it during the cleanup of the framework.
Care to add the new var to this doc? https://github.com/openstack-k8s-operators/ci-framework/blob/main/docs/source/usage/01_usage.md

playbooks/06-deploy-architecture.yml
```diff
@@ -197,6 +197,14 @@
         - update_containers
         - edpm_bootstrap
 
+    - name: Update containers in deployed OSP operators using set_openstack_containers role
+      when: cifmw_set_openstack_containers | default(false)
```
lewisdenny, line 201:
```suggestion
      when: cifmw_set_openstack_containers | default(false) | bool
```

Using bool filters to "cast" public API variables to ensure type safety is best practice.

lewisdenny (changes requested):
lgtm other than a small nit, thanks!

playbooks/06-deploy-architecture.yml
ciecierski: done

lewisdenny (approved):
/lgtm

PR #2322, opened 2024-09-05
Rename extraMounts to extraConfigmapsMounts

kopecmartin:
We introduced extraConfigmapsMounts parameter for the Tempest CR that has the same behaviour as extraMounts parameter in AnsibleTest CR.

Let's keep the naming consistent and rename the extraMounts to extraConfigmapsMounts in the AnsibleTest CR.

This follows changes from the second commit of
https://github.com/openstack-k8s-operators/test-operator/pull/192/


lpiwowar:
Looks good to me. Note, that this change probably requires an update of downstream job.

PR #2324, opened 2024-09-06
Convert trigger action into reusable workflow

raukadah:
this pull request:
- Renames github-action-trigger-downstream-job.yaml to github-action-trigger-downstream-job-reusable-workflow.yaml and adds on_workflow call to make it reusable github action workflow. It will allow us to reuse the trigger github action in multiple repos

- It adds trigger-downstream-job.yaml to call trigger github action on pull request.

It is being triggered at https://github.com/openstack-k8s-operators/ci-playground/pull/21#issuecomment-2333068050 and tested here: https://github.com/openstack-k8s-operators/ci-playground/pull/21#issuecomment-2333068143


lewisdenny (approved):
/approve

PR #2325, opened 2024-09-06
Add a new var to override openstack services container prefix

raukadah:
In downstream for post_adoption upgrade job. We are using upstream openstack operators. Upstream meta operator openstack services env variables have openstack prefix.

In order to create the post adoption pre update environment, we need to use the downstream openstack services containers. They starts with rhoso-openstack. Since it is a new prefix. We need a way to update the containers prefix to new prefix.

By using set_openstack_containers_new_prefix, we can achieve the same.


raukadah:
closing in favor of https://github.com/openstack-k8s-operators/ci-framework/pull/2281

PR #2327, opened 2024-09-09
Ensure we configure any "managed" VM

cjeanner:
UNtile now, we didn't generate any config drive for VM that weren't
started, even though we wanted them managed.

This patch corrects the issue, and also ensure we don't rely on SSH to
configure anything on the VMs.

We just keep plain SSH accesses for OCP/CRC, since they can't rely on
cloud-init.


cescgina:
we need this for the adoption infra

hjensas (commented):
/lgtm

PR #2329, opened 2024-09-09
Introduce DCN network-values template and job scenario parameters

sbekkerm:
Network values templates:
 - Configures multiple spine and leaf control plane networks
 - Incorporates static routing for network communication across nodes

Scenario parameters:
 - Adds parameters to provision three separate sites, each containing three compute nodes


cjeanner:
/approve

This has been tested with downstream TP 250. It doesn't affect any other job than a new DCN one.

PR #2330, opened 2024-09-10
Enable "backup" for config_drive generated files

cjeanner:
While we keep failing in case of file changes, we might want to know
what changed in those files over a re-run.
Enabling the "backup" feature ensures we are able to spot the
differences and adjust our runs if needed.


hjensas (commented):
/lgtm

cescgina (commented):
/approve

PR #2331, opened 2024-09-10
Trigger libvirt_manager molecule on config_drive changes

cjeanner:
Since config_drive is used in the libvirt_manager, we want to ensure
it's still properly working.

Also, moving the config_drove workdir to the standard "artifacts" tree,
ensuring we gather the content from the jobs.

### DIFF
--- a/zuul.d/molecule.yaml
+++ b/zuul.d/molecule.yaml
@@ -471,6 +471,7 @@
     - ^.config/molecule/.*
     - ^roles/dnsmasq/(defaults|files|handlers|library|lookup_plugins|module_utils|tasks|templates|vars).*
     - ^roles/networking_mapper/(defaults|files|handlers|library|lookup_plugins|module_utils|tasks|templates|vars).*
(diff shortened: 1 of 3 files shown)


cjeanner:
recheck

zuul.d/molecule.yaml
```diff
@@ -471,6 +471,7 @@
     - ^.config/molecule/.*
     - ^roles/dnsmasq/(defaults|files|handlers|library|lookup_plugins|module_utils|tasks|templates|vars).*
     - ^roles/networking_mapper/(defaults|files|handlers|library|lookup_plugins|module_utils|tasks|templates|vars).*
+    - ^roles/config_drive/(defaults|files|handlers|library|lookup_plugins|module_utils|molecule|tasks|templates|vars).*
```
hjensas, line 474: I think we should also add `name: cifmw-molecule-nat64_appliance` molecule job to run on config_drive role changes. ~ L510

cjeanner: care to provide another PR to get it covered?

hjensas: yes, we can do it in a separate PR.

PR #2332, opened 2024-09-10
Use the VM UUID instead of some other randomly value

cjeanner:
Until now, config_drive was using a random UUID built using the vm name.
It makes tracking slightly more difficult.

This PR reorganize some tasks so that we get the VM UUID earlier. It
also removes the XML parsing to rely on `virsh domuuid` instead, making
it stronger and safer.

The UUID is then injected in the config_drive role directly.


cjeanner:
recheck

PR #2335, opened 2024-09-10
Add a comment to sushy_driver SSH pub key

cjeanner:
Adding this comment, especially when injecting the pub key in the
hypervisor authorized_keys, allows to then filter out that specific key
when we generate the cloud-init "user-data".

This is needed in order to ensure we can re-run the deployment, for
instance when doing a scale-up or scale-down, or just re-running for
whatever reason:
since the sushy_driver key is injected "late" to the hypervisor
authorized_keys via reproducer role, it would then modify the user-data
generated during the libvirt_manager stage, leading to a file change,
leading eventually to a failure since it would re-generate the connfig
drive ISO (and this is prevented).

This patch ensures indempotency of the runs, by ensuring the
"late" sushy_driver ssh key isn't injected.

We also take the opportunity to remove the sushy_driver pub key via the
reproducer-clean.yml, ensuring we don't get an ever-growing file on the
hypervisor.

### DIFF
--- a/reproducer-clean.yml
+++ b/reproducer-clean.yml
@@ -33,6 +33,13 @@
         name: nat64_appliance
         tasks_from: cleanup.yml
 
+    - name: Cleanup config_drive data
+      tags:
+        - deepscrub
+      ansible.builtin.import_role:
+        name: config_drive
+        tasks_from: cleanup.yml
+
     - name: Remove remote selected data
       tags:
         - deepscrub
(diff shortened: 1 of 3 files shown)


reproducer-clean.yml
```diff
@@ -44,6 +51,14 @@
         - reproducer-network-env
         - artifacts/run-id
 
+    - name: Ensure sushy_driver key is removed
+      tags:
+        - deepscrub
+      ansible.builtin.lineinfile:
+        path: "{{ ansible_user_dir }}/.ssh/authorized_keys"
+        state: absent
+        regexp: ".*cifmw-sushy_driver$"
```
hjensas, line 60: Can we rename the comment to `cifmw-sushy_emulator` to match the role name?
AFICT, `sushy_driver` is not a term used anywhere before this change.

hjensas (changes requested):

PR #2337, opened 2024-09-10
Ensure we can re-run the deploy_layout multiple times

cjeanner:
Calling the tasks twice will ensure we're able to re-run the whole
deployment more than once without cleaning up in-between.

This should help preventing issues when we add new features, by ensuring
tasks are either properly isolated, or idempotent.


queria:
I think it's good idea, although wonder if it would be better to repeat it as whole - importing the libvirt_manager role as also the asserts/verifications.

hjensas (commented):
/lgtm

PR #2349, opened 2024-09-13
Run ceph update if part of the deployment

fmount:
In case we perform a minor update job where `Ceph` is deployed, as part of the minor update procedure we might want to update `Ceph` (according to a given `container image tag`) to verify the combination between the target Ceph version and the target OpenStack version is still healthy. This patch introduces the tasks that are supposed to call the `Ceph` update before OpenStack:
- it results in a noop in case the ceph container image tag is the <= of the current deployed `Ceph` cluster
- it is skipped by default if no `cifmw_ceph_update` variable is passed or it is set to `false` (which is the default)
- it can be skipped by passing `--skip-tags ceph_update` to the playbook

By doing this we have the ability to trigger a Ceph update and make sure we continue to test the combination between the two projects, but at the same time we are not introducing an hard dependency as it can easily be skipped.

Jira: https://issues.redhat.com/browse/OSPRH-9697

### DIFF
--- a/roles/cifmw_ceph_spec/templates/initial_ceph.conf.j2
+++ b/roles/cifmw_ceph_spec/templates/initial_ceph.conf.j2
@@ -7,7 +7,9 @@ ms_mon_client_mode = secure
 ms_mon_cluster_mode = secure
 ms_mon_service_mode = secure
 {% endif %}
+{% if cifmw_cephadm_single_host_defaults %}
 osd pool default size = 1
+{% endif %}
 {% if cifmw_ceph_spec_public_network %}
 public_network = {{ cifmw_ceph_spec_public_network }}
 {% endif %}
--- a/playbooks/update.yml
+++ b/playbooks/update.yml
@@ -49,6 +49,42 @@
         dest: "/etc/yum.repos.d/"
         src: "{{ cifmw_basedir }}/artifacts/repositories/"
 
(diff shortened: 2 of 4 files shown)


fmount:
I'd like to testproject this before doing any change or marking as ready.

fultonj:
+2

This patch looks good to me. Want to rebase it and take it out of draft ?

fmount:
> +2
> 
> This patch looks good to me. Want to rebase it and take it out of draft ?

Thank you @fultonj for looking at it. I'm currently trying this patch as part of unigamma in a testproject. When we have a green CI I'll post the results and we can move forward!

roles/cifmw_ceph_spec/templates/initial_ceph.conf.j2
```diff
@@ -7,7 +7,9 @@ ms_mon_client_mode = secure
 ms_mon_cluster_mode = secure
 ms_mon_service_mode = secure
 {% endif %}
+{% if cifmw_cephadm_single_host_defaults %}
 osd pool default size = 1
+{% endif %}
```
fultonj, line 12: This will affect other jobs so let's invert the logic please (per our discussion)

fmount: Done

fultonj (approved):
/approve
/lgtm

PR #2357, opened 2024-09-17
Fix rbd pools trash purge scheduling

katarimanojk:
This patch ensures to use the right conditions
and parameters while trash purge scheduling on
the rbd pools.

Jira: https://issues.redhat.com/browse/OSPRH-8504

### DIFF
--- a/roles/cifmw_cephadm/tasks/pools.yml
+++ b/roles/cifmw_cephadm/tasks/pools.yml
@@ -47,8 +47,9 @@
 
 - name: Configure the RBD trash purge scheduler
   when:
-    - cifmw_enabled_services | default([]) | intersect(['cinder_volume'])
     - cifmw_cephadm_enable_trash_scheduler | default(false)
+    - cifmw_cephadm_pools is defined
+    - cifmw_cephadm_pools | length > 0
   block:
     - name: Get the RBD ceph_cli
       ansible.builtin.include_tasks: ceph_cli.yml
@@ -56,10 +57,11 @@
         ceph_command: rbd
 
     - name: Set trash interval
+      when: item.trash_purge_enabled | default(false)
       ansible.builtin.command:
-        cmd: |
-          {{ cifmw_cephadm_ceph_cli }} trash purge schedule add \
-          {{ cifmw_cephadm_rbd_trash_interval | default(15) }} --pool {{ item }}
+        cmd: >-
+          {{ cifmw_cephadm_ceph_cli }} trash purge schedule add
+          {{ cifmw_cephadm_rbd_trash_interval | default(15) }} --pool {{ item.name }}
       changed_when: false
       become: true
-      loop: "{{ [ cinder_pool.name | default('volumes') ] + cinder_pool.cinder_extra_pools | default([]) }}"
+      loop: "{{ cifmw_cephadm_pools | default([]) }}"
--- a/playbooks/ceph.yml
+++ b/playbooks/ceph.yml
@@ -279,6 +279,7 @@
         pg_autoscale_mode: true
         target_size_ratio: 0.3
         application: rbd
+        trash_purge_enabled: true
       - name: backups
         pg_autoscale_mode: true
         target_size_ratio: 0.1


roles/cifmw_cephadm/tasks/pools.yml
```diff
@@ -47,19 +47,21 @@
 
 - name: Configure the RBD trash purge scheduler
   when:
-    - cifmw_enabled_services | default([]) | intersect(['cinder_volume'])
     - cifmw_cephadm_enable_trash_scheduler | default(false)
+    - cifmw_cephadm_pools is defined
+    - cifmw_cephadm_pools | length > 0
```
fmount, line 52: do we need a ` | default([])` here ?

roles/cifmw_cephadm/tasks/pools.yml
```diff
@@ -47,19 +47,21 @@
...
-        cmd: |
-          {{ cifmw_cephadm_ceph_cli }} trash purge schedule add \
-          {{ cifmw_cephadm_rbd_trash_interval | default(15) }} --pool {{ item }}
+        cmd: >-
+          {{ cifmw_cephadm_ceph_cli }} trash purge schedule add
+          {{ cifmw_cephadm_rbd_trash_interval | default(15) }} --pool {{ item.name }}
       changed_when: false
       become: true
-      loop: "{{ [ cinder_pool.name | default('volumes') ] + cinder_pool.cinder_extra_pools | default([]) }}"
+      loop: "{{ cifmw_cephadm_pools | default([]) }}"
```
fmount, line 67: Is the goal to set the rbd trash purge scheduler to all the ceph pools w/ an rbd application set? Or the goal was to enabled it only for cinder pools?

roles/cifmw_cephadm/tasks/pools.yml
```diff
@@ -47,19 +47,21 @@
...
+    - cifmw_cephadm_pools is defined
+    - cifmw_cephadm_pools | length > 0
   block:
     - name: Get the RBD ceph_cli
       ansible.builtin.include_tasks: ceph_cli.yml
       vars:
         ceph_command: rbd
 
     - name: Set trash interval
+      when: item.application == 'rbd'
```
fmount, line 60: This seems fragile to me, because if by any chance `application` is not set in the data structure, this task fails w/ `application` not defined.
am I right? @katarimanojk not sure you can double check this.

roles/cifmw_cephadm/tasks/pools.yml
katarimanojk: you are right, i think goal here is set it only for cinder pools

in tripleo, cinder_pool was defined in `ceph_pools ` like this
ceph_pools:
  cinder_pool:
    name: 'volumes'
    enabled: true
    cinder_extra_pools: [altrbd, pool2, pool3]

in cifmw , we have

cifmw_cephadm_pools:
- name: volumes
  pg_autoscale_mode: True
  target_size_ratio: 0.3
  application: rbd

@fultonj 
so we have only one pool `volumes` for cinder, should we set the rbd trash purge scheduler only for that pool and remove the loop ?

fmount: I would only set for volumes **if** is part of the `cifmw_cephadm_pools` (this should be the when condition). Right now this is wrong because you're setting the trash scheduler for all pools, and I don't think this was the original goal.

fultonj: +1 to what fmount said.

roles/cifmw_cephadm/tasks/pools.yml
fultonj: I think these checks make `| default([])` unnecessary. I think we specifically want to avoid this block if there are no `cifmw_cephadm_pools` so I think the update of the `when` is fine. I assume you were just proposing some alternative method for the `when` and not pointing out a logic problem. If there's a logic problem, then please elaborate.

roles/cifmw_cephadm/tasks/pools.yml
fultonj: I think most of the time people will use the hard coded `cifmw_cephadm_pools` and I think @katarimanojk was trying to set it for all RBD and exclude it for cephfs. However, I agree with @fmount that only want to set it for cinder, which will be called volumes. Unless someone overrides it, but in that case we'll have no way knowing which is the cinder volume unless we have some other sort of label. We could introduce a separate variable like `volumes_with_trash_purge` and default it to `['volumes']`. That at least makes a clear interface and allows the user to override which volumes will have it set. I'm open to other methods.

https://github.com/openstack-k8s-operators/ci-framework/blob/main/playbooks/ceph.yml#L273-L297

roles/cifmw_cephadm/tasks/pools.yml
```diff
@@ -47,19 +47,21 @@
 
 - name: Configure the RBD trash purge scheduler
   when:
-    - cifmw_enabled_services | default([]) | intersect(['cinder_volume'])
     - cifmw_cephadm_enable_trash_scheduler | default(false)
```
fultonj, line 50: If we introduce `cifmw_cephadm_pools_with_trash_scheduler` as a list, then we could get rid of the `cifmw_cephadm_enable_trash_scheduler` boolean remove the `when` and then loop on it.

We could add an additional condition in the when to ensure an intersection with `cifmw_cephadm_enable_trash_scheduler` and `cifmw_cephadm_pools` but that might be unnecessary and lead to further unnecessary complication in the future.

roles/cifmw_cephadm/tasks/pools.yml
fmount: Yeah, that solution is feasible but more complicated (I'm thinking for long term maintenance).
I was thinking more about something like "tagging" the pool where we want to enable trashing:

```
cifmw_cephadm_pools:
    name: volumes
    pg_autoscale_mode: True
    target_size_ratio: 0.3
    application: rbd
    trash_enabled: True
```

so we can keep iterating over the data structure, and enable the trash purge scheduler only for pools where we passed that `trash_enabled` boolean (that can be `|default(false)` for a general use case).
This should be simple enough and make that code less error prone.

fultonj: I like your proposal of the tag more. Let's go for that.

roles/cifmw_cephadm/tasks/pools.yml
fultonj: Let's go with the tag in the original `cifmw_cephadm_pools` instead.

roles/cifmw_cephadm/tasks/pools.yml
fmount: +1

roles/cifmw_cephadm/tasks/pools.yml
katarimanojk: @fmount Thanks for your suggestion, i will update the code accordingly.

roles/cifmw_cephadm/tasks/pools.yml
katarimanojk: @fmount agree with you, code is now updated and it sets trash purge scheduler only to volumes pool.

roles/cifmw_cephadm/tasks/pools.yml
katarimanojk: @fmount i have implemented your suggestion and tested it.

fmount (approved):
/lgtm

fultonj (approved):
/lgtm
/approve

PR #2361, opened 2024-09-18
Added workaround to fix OSPCIX-487

raukadah:
Currently tempest run in all jobs are broken due to missing testscenarios dependency[1]. This pr installed python3-testscenarios package using  cifmw_test_operator_tempest_extra_rpms and workaround this issue.

It modifies the project template to include this var to avoid affecting downstream/periodic job.

[1]. https://issues.redhat.com/browse/OSPCIX-487


karelyatin:
i think we don't need this workaround, as i see jobs using openstack-tempest-all image and that's not impacted. also required fixes are in place so we should have a promotion soon

PR #2364, opened 2024-09-19
[test-operator] Add wait_timeout for removing CRD instances in test-o…

lkuchlan:
…perator

Follow-up on PR [1]: Still encountering timeout issues during the deletion of the test-operator.

[1] https://github.com/openstack-k8s-operators/ci-framework/pull/2305

### DIFF
--- a/roles/test_operator/tasks/run-test-operator-job.yml
+++ b/roles/test_operator/tasks/run-test-operator-job.yml
@@ -228,6 +228,7 @@
         name: "{{ test_operator_job_name }}"
         namespace: "{{ cifmw_test_operator_namespace }}"
         wait: true
+        wait_timeout: 600
 
     - name: Delete CRD for {{ run_test_fw }}
       kubernetes.core.k8s:
@@ -240,6 +241,7 @@
         name: "{{ test_operator_crd_name }}"
         namespace: "{{ cifmw_test_operator_namespace }}"
         wait: true
+        wait_timeout: 600
 
 - name: Delete test-operator-logs-pod
   kubernetes.core.k8s:
@@ -252,6 +254,7 @@
     name: "test-operator-logs-pod-{{ run_test_fw }}"
     namespace: "{{ cifmw_test_operator_namespace }}"
     wait: true
+    wait_timeout: 600
   when:
     - cifmw_test_operator_cleanup | bool and not cifmw_test_operator_dry_run | bool or
       cifmw_test_operator_delete_logs_pod | bool


lewisdenny (changes requested):
Please add this to the roles defaults. 

I see your comment here [1] but a default is the right move here.

Also if you can reproduce the issue, please test this patch via depends on so we can be sure it helps.  

[1] https://github.com/openstack-k8s-operators/ci-framework/pull/2305#discussion_r1743647392

lpiwowar (approved):
/lgmt

kopecmartin:
> Please add this to the roles defaults.
> 
> I see your comment here [1] but a default is the right move here.
> 
> Also if you can reproduce the issue, please test this patch via depends on so we can be sure it helps.
> 
> [1] [#2305 (comment)](https://github.com/openstack-k8s-operators/ci-framework/pull/2305#discussion_r1743647392)

should be really timeout value in the defaults?

lewisdenny:
> > Please add this to the roles defaults.
> > I see your comment here [1] but a default is the right move here.
> > Also if you can reproduce the issue, please test this patch via depends on so we can be sure it helps.
> > [1] [#2305 (comment)](https://github.com/openstack-k8s-operators/ci-framework/pull/2305#discussion_r1743647392)
> 
> should be really timeout value in the defaults?

For better or worse the standard in this project that configurable variables be defined in the role defaults and documented the the roles README.

lpiwowar:
> > > Please add this to the roles defaults.
> > > I see your comment here [1] but a default is the right move here.
> > > Also if you can reproduce the issue, please test this patch via depends on so we can be sure it helps.
> > > [1] [#2305 (comment)](https://github.com/openstack-k8s-operators/ci-framework/pull/2305#discussion_r1743647392)
> > 
> > 
> > should be really timeout value in the defaults?
> 
> For better or worse the standard in this project that configurable variables be defined in the role defaults and documented the the roles README.

There is no new configurable parameter introduced into the ci-framework role with this PR. The `wait_timeout: xxx` is sprinkled all over the place in ci-framework without any parameter [1]. 

@lewisdenny do you think that we really can't merge it like this? If no then I guess @lkuchlan will have to modify the PR and introduce a new parameter `cifmw_test_operator_wait_timeout` (default value: 600). This will require update of every wait_timeout call in the test_operator role [2].

```
wait_timeout: {{ cifmw_test_operator_wait_timeout }}
```

[1] [here](https://github.com/search?q=repo%3Aopenstack-k8s-operators%2Fci-framework%20wait_timeout%3A&type=code)
[2] [here](https://github.com/openstack-k8s-operators/ci-framework/blob/8d239cfef4e27788102e3da7d344d82e8e543ae5/roles/test_operator/tasks/cleanup.yml#L67)

PR #2367, opened 2024-09-22
[reproducer] Move the installation of the timesync role in its block

rebtoor:


rebtoor:
I'm closing this PR because this approach will fail. I'll write an ntp role to use within ci_framework.

PR #2376, opened 2024-09-24
Adding toolbelt catalog metadata file

jpodivin:
Including this file in our documentation will enable us to list ci-framework in the catalog of tools for QE.

### DIFF
--- a/docs/toolbelt-catalog.yaml
+++ b/docs/toolbelt-catalog.yaml
@@ -1,0 +1,41 @@
+# Catalog entry for Backstage [backstage.io]
+
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+    name: ci-framework
+    title: ci-framework
+    description: |
+        CI Framework - used for CI, QE and Devs to run OSP 18+ jobs in a converged way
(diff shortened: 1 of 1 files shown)


docs/toolbelt-catalog.yaml
```diff
@@ -0,0 +1,41 @@
...
+        - python
+        - openstack
+        - openshift
+        - cloud
+        - continuous-integration
+    namespace: quality-community
+spec:
+    type: tool
+    owner: group:redhat/openstack-k8s-operators-ci
+    lifecycle: production
```
lewisdenny, line 41: Do you have a link to what this metadata actually means. Everything looks correct to me, I would just like to know what `lifecycle: production` actually means because ci-framework is not a customer product... if that's what it means but as I say idk what it means :)

jpodivin: It doesn't mean that. Catalog is purely for our QE, so they can look up the best tools for the job.

lewisdenny: In that case it's okay for me

PR #2383, opened 2024-09-25
Only gather required ansible facts

rabi:
This would ensure we can test the roles as expected with `gather_facts: false`.


cjeanner:
/approve

makes sense to collect only a selected subset of facts to ensure roles do the right things. All of the existing molecule jobs are green in the ci-framework, and the edpm-ansible PR depending on this one is also all green.

slagle (approved):

lewisdenny (approved):
/lgtm

PR #2385, opened 2024-09-25
Gather ssh private key from the hostvars in ceph playbook

rebtoor:


rebtoor:
This patch was successfully tested by @rdiazcam

PR #2386, opened 2024-09-26
Add support to deploy sushy in extracted jobs

lewisdenny:
- Add support for deploying Sushy Emulator in non reproducer jobs
- Add support for configuring the controller node as a hypervisor
- Add new var `cifmw_enable_virtual_baremetal_support` to toggle feature
- Enable `cifmw_enable_virtual_baremetal_support` in
  `podified-multinode-edpm-deployment-crc-bootstrap-staging-ironic` for
  testing.

### DIFF
--- a/playbooks/02-infra.yml
+++ b/playbooks/02-infra.yml
@@ -86,6 +86,22 @@
       ansible.builtin.include_role:
         name: cert_manager
 
+    - name: Deploy Sushy Emulator and configure controller as hypervisor
+      when:
+        - cifmw_enable_virtual_baremetal_support | default(false) | bool
+      block:
+        - name: Deploy Sushy Emulator service pod
+          vars:
+            cifmw_sushy_emulator_hypervisor_address: "{{ hostvars['controller'].ansible_host }}"
+            cifmw_sushy_emulator_hypervisor_target: controller
+            cifmw_sushy_emulator_install_type: ocp
+          ansible.builtin.include_role:
+            name: sushy_emulator
+
+        - name: Setup Libvirt on controller
+          ansible.builtin.include_role:
+            name: libvirt_manager
+
     - name: Configure hosts networking using nmstate
       when:
         - cifmw_config_nmstate is defined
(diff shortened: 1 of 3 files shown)


lewisdenny:
recheck

lewisdenny:
recheck

lewisdenny:
recheck

lewisdenny:
recheck

playbooks/02-infra.yml
```diff
@@ -86,6 +86,22 @@
...
+      when: cifmw_use_sushy_emulator | default(false) | bool
+      vars:
+        cifmw_sushy_emulator_hypervisor_address: "{{ hostvars['controller'].ansible_host }}"
+        cifmw_sushy_emulator_hypervisor_target: controller
+        cifmw_sushy_emulator_install_type: ocp
+      ansible.builtin.include_role:
+        name: sushy_emulator
+
+    - name: Setup Libvirt on controller
+      when: cifmw_use_libvirt_extracted | default(false) | bool
```
lewisdenny, line 100: Not sure what var to use here, cifmw_use_libvirt is already used in the reproducer scenario so using cifmw_use_libvirt_extracted as a placeholder until discussion.

lewisdenny:
recheck

playbooks/02-infra.yml
evallesp, line 100: (non-blocking) question: Make sense to "Setup libvirt on controller" when cifmw_use_sushy_emulator is missing or false?

On the top of my head, I think if both are required for support deploying Sushy if the first is not applicable, the second should not be applied as well.

frenzyfriday:
lgtm from the commit msg

playbooks/02-infra.yml
lewisdenny: Hi @evallesp, 

Yes, I think this is a case of "YAGNIi"

I was just adding the option for other use cases outside this Ironic/Sushy one to setup the controller node as a hypervisor but I can't even imagine now what that would be so I will lessen the scope.

lewisdenny: Done

zuul.d/edpm_multinode.yaml
```diff
@@ -444,6 +444,7 @@
     name: podified-multinode-edpm-deployment-crc-bootstrap-staging-ironic
     parent: podified-multinode-edpm-deployment-crc-bootstrap-staging
     vars:
+      cifmw_enable_virtual_baremetal_support: true
```
pablintino, line 447: Do we want this in non-BMO jobs? I'd say no.

PR #2388, opened 2024-09-26
Make `^nodes?(_[0-9]+)?$` keys removal configurable

hjensas:
When creating the ConfigMap baseline any keys matching regular expression `^nodes?(_[0-9]+)?$` is removed from architecture source  values.yaml files.
    
This change introduces a new parameter to make this behavior configurable, `cifmw_ci_gen_kustomize_values_remove_keys_expressions` takes a list of regular expressions. 

By default these are:
```
      - "^nodes$"
      - "^node(_[0-9]+)?$"
```    

Disabling the keys removal is useful in the case when the values used in architecture repo example files is a 1-to-1 match with the deployed infrastructure - i.e there is no need to jinja template the values based on the network_mapper data.

### DIFF
--- a/roles/ci_gen_kustomize_values/tasks/generate_snippets.yml
+++ b/roles/ci_gen_kustomize_values/tasks/generate_snippets.yml
@@ -80,8 +80,9 @@
     _cifmw_gen_kustomize_values_base_cm_content: >-
       {{
         _config_map_content |
-        ansible.utils.remove_keys(target=['^nodes?(_[0-9]+)?$'],
-                                  matching_parameter='regex')
+        ansible.utils.remove_keys(
+          target=cifmw_ci_gen_kustomize_values_remove_keys_expressions,
+          matching_parameter='regex')
       }}
     cacheable: false
 
--- a/roles/ci_gen_kustomize_values/README.md
+++ b/roles/ci_gen_kustomize_values/README.md
@@ -35,6 +35,8 @@ with a message.
 * `ci_gen_kustomize_fetch_ocp_state`: (Boolean) If true it enables generating CI templates based on the OCP state. Defaults to `true`.
 * `cifmw_ci_gen_kustomize_values_storage_class_prefix`: (String) Prefix for `storageClass` in generated values.yaml files. Defaults to `"lvms-"` only if `cifmw_use_lvms` is True, otherwise it defaults to `""`. The prefix is prepended to the `cifmw_ci_gen_kustomize_values_storage_class`. It is not recommended to override this value, instead set `cifmw_use_lvms` True or False.
 * `cifmw_ci_gen_kustomize_values_storage_class`: (String) Value for `storageClass` in generated values.yaml files. Defaults to `"lvms-local-storage"` only if `cifmw_use_lvms` is True, otherwise it defaults to `"local-storage"`.
+* `cifmw_ci_gen_kustomize_values_remove_keys_expressions`: (List) Remove keys matching the regular expressions from source ConfigMaps (values.yaml).
+  Defaults to `["^nodes$", "^node(_[0-9]+)?$"]`.
 
 ### Specific parameters for edpm-values
 This configMap needs some more parameters in order to properly override the `architecture` provided one.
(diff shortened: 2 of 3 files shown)


roles/ci_gen_kustomize_values/tasks/generate_snippets.yml
```diff
@@ -78,11 +78,14 @@
...
       }}
     _cifmw_gen_kustomize_values_base_cm_content: >-
-      {{
-        _config_map_content |
-        ansible.utils.remove_keys(target=['^nodes?(_[0-9]+)?$'],
-                                  matching_parameter='regex')
-      }}
+      {%- if cifmw_ci_gen_kustomize_values_remove_nodes_keys | default(true) -%}
+      {%- set _content = _config_map_content |
+         ansible.utils.remove_keys(target=['^nodes?(_[0-9]+)?$'],
```
cjeanner, line 83: beware we're talking about two different keys, with different meaning:
- `node_[0-9]` is for OCP cluster members
- `nodes` is for edpm nodes (computes)

Not 100% sure we should have 1:1 matching data in a public repository, showing internal IPs and so on...

hjensas: I fully agree we should not leak any actual RedHat internal IPs - like DNS servers - repo URL's, BMC addresses of physical servers etc in a public repo. The use-case for this is only when we have virtual labs with lab internal addressing that is not routed in Red Hat.

Doing this takes days of developing a job for a new DT, without this we need to 1. Create the DT and test that values that we provide in examples actually work. 2. Create the Jinja2 templating and test. This is pretty much duplicate work, and iterations during test/dev is slow.

Thanks for pointing out that this is actually two keys. `node_[0-9]` and `nodes` - we may want to make this more granular - or move the actual reg_ex to the parameter?

hjensas: So, I changed this to set the regular expressions for keys to remove a parameter instead. Parameter is a list, so we can remove just the OCP cluster members, edpm nodes or both - also someone might need to add additional keys to remove.

cjeanner: OK, I understand the usecase and agree.

I'd rather split the regex rather than pushing it as a parameter:
```YAML
{%- set _content = _config_map_content -%}
{%- if [first param for OCP nodes cleanup] -%}
{%-   set _content = _content | ansible.utils.remove_keys(target=['^node_[0-9]+$'], ...) -%}
{%- endif -%}
{%- if [second param for EDPM nodes cleanup] -%}
{%-   set _content = _content | ansible.utils.remove_keys(target=['nodes']) -%}
{%- endif -%}
{{ _content }}
```

That should do.

Also, as a safeguard, we might want to check for those parameters and, maybe, add a "pause" with prompt if either of those params are set to disable the key cleanup. That would prevent ppl from running it in the CI automation, since it would "pause" until it gets the expected answer.... Might look lame, but that might prevent misuse of that capability:
```yaml
- name: Block when cleanup is disabled
  when:
    - [first param for OCP nodes cleanup] or
       [second param for EDPM nodes cleanup]
  block:
    - name: Check for user availability
      register: _no_cleanup
      ansible.builtin.pause:
        prompt: "You disabled nodes or node_* cleanup, are you sure?"
    - name: Fail if no right answer is provided
      when:
        - _no_cleanup != 'yes I know'
      ansible.builtin.fail:
        msg: "Don't try me"
```
dumb. simple. and might help ensure ppl are aware.

hjensas: I disagree with custom parameters per key-filter. Right now I just want to get rid of them ... But there are other parameters that I can imagine I'd want to add key removal for, example's `barmetalhosts` and `bmcredentials`.

I don't want to add another patch to ci-framework everytime we find another key we want to filter.

I don't want to add the use check - I want to make it possible to run jobs with these filters disabled.

lewisdenny (approved):
/lgtm

Thanks for the clean PR Harald!

PR #2389, opened 2024-09-26
Make sure set_openstack_container is less depedent on instal_yamls

ciecierski:
Cleanup set_openstack_container role by setting default values and omitting
setting fact if install_yamls was not used during deployment.

This is first atempt of reducing dependency in set_openstack_container role
from install_yamls. The work that is left for later is for updating
ENVs in operators other than openstack-operator. Task that waits
for an operator(other than openstack-operator) update to complete will only work in install_yaml deployed environments and is skipped in architecture_scenario context.

Jira: https://issues.redhat.com/browse/OSPRH-10285

### DIFF
--- a/roles/set_openstack_containers/tasks/main.yml
+++ b/roles/set_openstack_containers/tasks/main.yml
@@ -37,23 +37,27 @@
 - name: Set common facts for the role
   ansible.builtin.set_fact:
     operator_name: "{{ cifmw_set_openstack_containers_operator_name }}"
+    operator_namespace: >-
+      {{
+        cifmw_install_yamls_defaults['OPERATOR_NAMESPACE'] |
+        default(cifmw_set_openstack_containers_operators_namespace)
+      }}
     cifmw_set_openstack_containers_common_env: >-
       {{
-        cifmw_install_yamls_environment |
+        ( cifmw_install_yamls_environment |
         combine({'PATH': cifmw_path}) |
         combine(cifmw_set_openstack_containers_extra_vars | default({}))
+        ) if cifmw_install_yamls_environment is defined else omit
       }}
     cacheable: true
 
 - name: Get meta operator environment variable
-  vars:
-    _operator_ns: "{{ cifmw_install_yamls_defaults['OPERATOR_NAMESPACE'] }}"
   environment:
     KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
     PATH: "{{ cifmw_path }}"
   cifmw.general.ci_script:
     output_dir: "{{ cifmw_set_openstack_containers_basedir }}/artifacts"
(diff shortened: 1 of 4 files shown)


roles/set_openstack_containers/tasks/main.yml
```diff
@@ -105,14 +109,17 @@
   ansible.builtin.include_role:
     name: 'install_yamls_makes'
     tasks_from: 'make_wait'
+  when:
+    - operator_name != 'openstack'
+    - not cifmw_architecture_scenario is defined
```
sathlan, line 114: Oki, if we want to run this all the time where 

```
  NOT  (operator_name == 'openstack' AND cifmw_architecture_scenario is defined)
```

That should be
```
  operator_name != 'openstack' OR not cifmw_architecture_scenario is defined
```
so if cifmw_arch_scenario is defined but operator name is not openstack this can still run ... but we don't want to run install_yaml when cifmw_architecture_scenario is defined.

I'm not sure now if we need the operator_name conditional.

sathlan (approved):
LGTM, just a note about the extra conditional for operator_name, as I'm not sure we really need this.  But not a blocker.

roles/set_openstack_containers/tasks/main.yml
ciecierski: Strictly speaking this code path has never been used, but I'm leaving it just in case there will be use case for modifying envs in operator other than openstack-operator.
We don't want to run it in cifmw_architecture_scenario and operator_name != 'openstack'. We want to run it only in the context of install_yamls jobs and only  operator_name != 'openstack' for instace baremetal-operator.

sathlan (approved):
LGTM

roles/set_openstack_containers/tasks/main.yml
```diff
@@ -37,23 +37,27 @@
 - name: Set common facts for the role
   ansible.builtin.set_fact:
     operator_name: "{{ cifmw_set_openstack_containers_operator_name }}"
+    operator_namespace: >-
+      {{
+        cifmw_install_yamls_defaults['OPERATOR_NAMESPACE'] |
+        default('openstack-operators')
```
pablintino, line 43: Care to use a role default var for this?

roles/set_openstack_containers/tasks/main.yml
pablintino, line 114: The fact we need to check against the presence of `cifmw_architecture_scenario` in this role is a big red flag. But it's out of the scope of your change.

pablintino (changes requested):

pablintino:
recheck

lewisdenny (approved):
/lgtm

+1 for being less dependent on install_yamls.

PR #2390, opened 2024-09-26
Fix molecule - add swift network

hjensas:
The swift network was added to the HCI scenario used in the molecule tests here: https://github.com/openstack-k8s-operators/architecture/pull/404


rebtoor (approved):
/approve

PR #2393, opened 2024-09-26
Only consider ocps networks for network-values template

cescgina:
The common template for the network-values in ci_gen_kustomize_values
iterates over all the hosts in the cifmw_networking_env_definition and
stores in a map the relationship between a network and the interface.

For this to work, the ocp nodes must be the last ones, since if there
are other nodes after them in the cifmw_networking_env_definition,
for example nodes named osp-*, and they use a different network
interface, the resulting nncp will fail when applied.

### DIFF
--- a/roles/ci_gen_kustomize_values/templates/bgp/network-values/values.yaml.j2
+++ b/roles/ci_gen_kustomize_values/templates/bgp/network-values/values.yaml.j2
@@ -5,18 +5,16 @@
                       lb_tools={}) %}
 data:
 {% for host in cifmw_networking_env_definition.instances.keys() -%}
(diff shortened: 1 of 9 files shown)


hjensas:
+1 from me.

Since the network-values is overridden in most (if not all) of the VAs and DTs I think it would be good to make this change accross the board.

cescgina:
> +1 from me.
> 
> Since the network-values is overridden in most (if not all) of the VAs and DTs I think it would be good to make this change accross the board.

thanks @hjensas I've added a new commit copying the change to the rest of VAs and DTs templates, ptal

lewisdenny (approved):
/lgtm

Nice catch @cescgina !

PR #2396, opened 2024-09-27
[sushy_emulator] Make instance whitelist optional

lewisdenny:


lewisdenny:
recheck

lewisdenny:
recheck

marios (approved):
came via slack request, would be nice to have some more info in the commit message but guessing based on the commit title, indeed it moves tasks into a block and executes only when uuid file exists

roles/sushy_emulator/tasks/collect_details.yml
```diff
@@ -69,20 +69,30 @@
...
     - name: "Slurp content of: {{ _uuid_file }}"
-      ansible.builtin.slurp:
-        src: "{{ _uuid_file }}"
-      register: _libvirt_uuids_file
+      when: _uuid_file_status.stat.exists
+      block:
+        - name: "Slurp content of: {{ _uuid_file }}"
+          ansible.builtin.slurp:
+            src: "{{ _uuid_file }}"
+            register: _libvirt_uuids_file
```
pablintino, line 83: This register is wrongly indented.

lewisdenny: Interesting the linter didn't catch this :(

lewisdenny: Done

pablintino:
@frenzyfriday care to add your approval after the small tweak Lewis did?

roles/sushy_emulator/tasks/collect_details.yml
frenzyfriday: haha, I missed that as well 😅

PR #2397, opened 2024-09-27
Pass tcib-extras tcib_package while building containers from source

raukadah:
We need to pass `--tcib-extras tcib_package=` while [using tcib from source](https://github.com/openstack-k8s-operators/tcib?tab=readme-ov-file#building-images-with-local-changes). This pr updates the same.

Tested it here: https://review.rdoproject.org/r/c/testproject/+/54761/11#message-db813ffd30e270c15753f711d65f2e8079bbb741 and  https://logserver.rdoproject.org/61/54761/11/check/periodic-container-tcib-build-push-centos-9-master/ed662ef/ci-framework-data/artifacts/build_containers.sh


lewisdenny (approved):
/lgtm

PR #2398, opened 2024-09-27
Make HCI architecture repo parameterizable

rdiazcam:


tosky:
(maybe specify this is about a specific reproducer in the first line of the commit message?)

scenarios/reproducers/dt-nfv-ovs-dpdk-sriov-hci.yml
```diff
@@ -5,7 +5,7 @@ cifmw_architecture_scenario: "nfv-ovs-dpdk-sriov-hci"
 # controller-0 as-is and be consumed by the `deploy-va.sh` script.
 # Please note, all paths are on the controller-0, meaning managed by the
 # Framework. Please do not edit them!
-_arch_repo: "/home/zuul/src/github.com/openstack-k8s-operators/architecture"
+_arch_repo: "{{ cifmw_architecture_repo }}"
```
fultonj, line 8: I think this would be safer.
```yaml
_arch_repo: "{{ cifmw_architecture_repo | default('/home/zuul/src/github.com/openstack-k8s-operators/architecture') }}"
```

fultonj (commented):
Could you put more info in the commit message?

scenarios/reproducers/dt-nfv-ovs-dpdk-sriov-hci.yml
pablintino: I said the same privately. +1
This is the way I saw in DS people is using to make this parameter overridable. I bet to stick two one of the two flavors of declaring it, like it was here and your suggestion, that is widely used too.

rdiazcam: done, thanks

PR #2401, opened 2024-09-27
Ignore non-vars file in fetch_compute_facts

pablintino:
Some files may be backups of previous versions, like the networking-environment-definition if the mapper is called twice. To avoid trying to load old backed up content load only the default files (yaml and json).

### DIFF
--- a/hooks/playbooks/fetch_compute_facts.yml
+++ b/hooks/playbooks/fetch_compute_facts.yml
@@ -26,6 +26,10 @@
     - name: Load parameters
       ansible.builtin.include_vars:
         dir: "{{ item }}"
+        ignore_unknown_extensions: true
+        extensions:
+          - yaml
+          - yml
       loop:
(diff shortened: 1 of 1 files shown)


pablintino:
recheck

hooks/playbooks/fetch_compute_facts.yml
```diff
@@ -26,6 +26,7 @@
     - name: Load parameters
       ansible.builtin.include_vars:
         dir: "{{ item }}"
+        ignore_unknown_extensions: true
```
frenzyfriday, line 29: lgtm, but should we add `extensions: ` instead to make it more controllable?

pablintino: I had that in one of the patches of the base change. I'll read it to support YAML only, it's true that we usually don't load json that's in the list by default so I'm fine.

lewisdenny (approved):
/approve

PR #2402, opened 2024-09-27
Improve adoption coverage for openstack-operator

jistr:
This should help prevent breakages like [1].

[1] https://github.com/openstack-k8s-operators/openstack-operator/pull/1103

Until now we ran the adoption job only when bumping the component operators, but we should run it also when we change the interface (CRDs) or the behavior of the openstack-operator itself.

Related: [OSPCIX-504](https://issues.redhat.com//browse/OSPCIX-504)


rebtoor (approved):
/lgtm

cescgina:
recheck

PR #2405, opened 2024-09-27
Use Ceph monitoring (not RGW) network for Dashboard

fultonj:
Use the existing `cifmw_cephadm_monitoring_network` parameter to configure the Ceph dashboard instead of the `cifmw_cephadm_rgw_network` parameter. This was not detected earlier because the Ceph playbook set them both to the Ceph public_network, but we recently added a feature to run RGW on a separate network [1].
Also, add retries/delay to the dashboard HTTP check and mange if HTTPS should be used.

[1] https://github.com/openstack-k8s-operators/ci-framework/pull/2301

Jira: https://issues.redhat.com/browse/OSPCIX-506

(cherry-picked from f6e464505a653befe9ee5c458974b6c4737c8129)

### DIFF
--- a/playbooks/ceph.yml
+++ b/playbooks/ceph.yml
@@ -465,6 +465,9 @@
       ansible.builtin.import_role:
         name: cifmw_cephadm
         tasks_from: post
+      vars:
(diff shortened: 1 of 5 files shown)


fmount (approved):
/lgtm

PR #2410, opened 2024-09-30
Added cifmw_run_test_role: test_operator variable to antelope and mas…

sauragar:
…ter edpm baremetal job


arxcruz:
Please update commit message. Also, link for testproject with depends-on on this patch.

PR #2411, opened 2024-09-30
Move test-operator CR name variables from vars to defaults folder

ciecierski:
Currently the names of the resources created with test-operator role were defined by variables in the vars folder of test-operator role. We want to move those vars to defaults folder to enable their easier customization. The main reason why this name customization might be useful is to be able to run tempest multiple times without doing cleanup of CRs for every tempest run.

### DIFF
--- a/roles/test_operator/README.md
+++ b/roles/test_operator/README.md
@@ -26,6 +26,7 @@ Execute tests via the [test-operator](https://openstack-k8s-operators.github.io/
 
 ## Tempest specific parameters
 * `cifmw_test_operator_tempest_registry`: (String) The registry where to pull tempest container. Default value: `quay.io`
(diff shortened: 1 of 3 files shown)


lpiwowar (commented):
Overall looks good to me :+

If it is about enabling multiple runs of the test-operator role within a single job without the need of running a cleanup then can you please add the renaming for the test-operator-logs pod (the same way as here [1])? 

[1] [here](https://github.com/openstack-k8s-operators/ci-framework/pull/2412/files#diff-271aad5b65accb578d5d5d380be2797f6d0b22c56f0cb8fadd6be99c89e2ac77R152)

ciecierski:
Closing in favor of https://github.com/openstack-k8s-operators/ci-framework/pull/2412

PR #2412, opened 2024-09-30
Enable running the test-operator role multiple times

lpiwowar:
It was not possible to run the test-operator role multiple times without cleaning up all resources created by the first role run, such as the logs pod created for the logs collection.

This patch makes it possible to run the test operator role multiple times without cleaning up the resources (e.g., after the update in an update job). It is possible because:

- test-operator-logs pod now has unique name for each job run
- The variable cifmw_test_operator_*_name is now exposed. By specifying a unique name, we can guarantee no name collisions between the test-operator related resources (logs pods, instances of CRs)

If you want to run test-operator role twice within a single job then make sure that for the second run you specify a value for the cifmw_test_operator_*_name. For example:

  cifmw_test_operator_tempest_name: "post-update-tempest-tests"

### DIFF
--- a/roles/test_operator/README.md
+++ b/roles/test_operator/README.md
@@ -25,6 +25,7 @@ Execute tests via the [test-operator](https://openstack-k8s-operators.github.io/
 * `cifmw_test_operator_privileged`: (Boolean) Spawn the test pods with `allowPrivilegedEscalation: true` and default linux capabilities. This is required for certain test-operator functionalities to work properly (e.g.: `extraRPMs`, certain set of tobiko tests). Default value: `true`
 
 ## Tempest specific parameters
+* `cifmw_test_operator_tempest_name`: (String) Value used in the `Tempest.Metadata.Name` field. The value specifies the name of some resources spawned by the test-operator role. Default value: `tempest-tests`
 * `cifmw_test_operator_tempest_registry`: (String) The registry where to pull tempest container. Default value: `quay.io`
 * `cifmw_test_operator_tempest_namespace`: (String) Registry's namespace where to pull tempest container. Default value: `podified-antelope-centos9`
 * `cifmw_test_operator_tempest_container`: (String) Name of the tempest container. Default value: `openstack-tempest`
@@ -48,7 +49,7 @@ Execute tests via the [test-operator](https://openstack-k8s-operators.github.io/
   apiVersion: test.openstack.org/v1beta1
   kind: Tempest
   metadata:
-    name: tempest-tests
+    name: "{{ cifmw_test_operator_tempest_name }}"
     namespace: "{{ cifmw_test_operator_namespace }}"
   spec:
     containerImage: "{{ cifmw_test_operator_tempest_image }}:{{ cifmw_test_operator_tempest_image_tag }}"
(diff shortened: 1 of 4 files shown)


roles/test_operator/README.md
```diff
@@ -25,6 +25,7 @@ Execute tests via the [test-operator](https://openstack-k8s-operators.github.io/
 * `cifmw_test_operator_privileged`: (Boolean) Spawn the test pods with `allowPrivilegedEscalation: true` and default linux capabilities. This is required for certain test-operator functionalities to work properly (e.g.: `extraRPMs`, certain set of tobiko tests). Default value: `true`
 
 ## Tempest specific parameters
+* `cifmw_test_operator_tempest_name`: (String) Value used in the Tempest.Metadata.Name field. The value specifies name of some resources spawned by the test-operator role. Default value: `tempest-tests`
```
lewisdenny, line 28:
```suggestion
* `cifmw_test_operator_tempest_name`: (String) Value used in the Tempest.Metadata.Name field. The value specifies the name of some resources spawned by the test-operator role. Default value: `tempest-tests`
```

ditto for each of the cifmw_test_operator_*_name entries.

lewisdenny (changes requested):
Hi @lpiwowar 

Would it be better to use an randomly generated ID for each run regardless? This way a user would not need to manually specify a name for the second third run etc. 

If the current implementation is best I have a suggestion to add a new section to the README for how to run the role multiple times, just with the details you mentioned in the commit message would be fine.

lpiwowar:
Hi @lewisdenny :wave:, thank you for the review!

> Would it be better to use an randomly generated ID for each run regardless?

Adding the random suffix won't eliminate the need to specify a different name for each execution of the test-operator role. The name as specified in the README is used in `Metadata.Name`. This value impacts a bunch of resources that get created during the test execution. Adding the random suffix will make the logs and collected artifacts slightly more difficult to read. Also, I think adding the random suffix would go against the principle of idempotency. LMKWYT.

> If the current implementation is best I have a suggestion to add a new section to the README

Ok, I'll update the README  :+1:

roles/test_operator/README.md
lpiwowar: Thank you, done.

lewisdenny:
> Hi @lewisdenny 👋, thank you for the review!
> 
> > Would it be better to use an randomly generated ID for each run regardless?
> 
> Adding the random suffix won't eliminate the need to specify a different name for each execution of the test-operator role. The name as specified in the README is used in `Metadata.Name`. This value impacts a bunch of resources that get created during the test execution. Adding the random suffix will make the logs and collected artifacts slightly more difficult to read. Also, I think adding the random suffix would go against the principle of idempotency. LMKWYT.
> 
> > If the current implementation is best I have a suggestion to add a new section to the README
> 
> Ok, I'll update the README 👍

Thank you for the detailed explanation mate. That along with the update README make it clear to me +1 from me.

pablintino:
recheck

tosky:
Does this allow us to run test-operator twice or multiple times?

ciecierski (approved):
looks good to me

lpiwowar:
> Does this allow us to run test-operator twice or multiple times?

@tosky: This will allow to run the test-operator role multiple times within a single job [1]. 

[1] [here](https://github.com/openstack-k8s-operators/ci-framework/pull/2412/files#diff-53ae92ac766af9a25eba7952c80edba18bab29ea3a272761dbb73a69bf517053R206)

lpiwowar:
recheck

Failure not caused by the PR -> https://logserver.rdoproject.org/12/2412/962fb5ad878675d0a65a2e2123984efbe5438fbd/github-check/cifmw-crc-podified-edpm-baremetal/8911c36/job-output.txt

PR #2415, opened 2024-10-01
Add swift network to NetConfig for va-hci-base and restore default

lmiccini:
This commit restores networking-definition.yml to its pre-swift state and overrides cifmw_networking_definition in va-hci-base.


lmiccini:
obsolete

PR #2424, opened 2024-10-04
Wait for InstallPlan to complete when installing the OSP operators

pablintino:
Till now we weren't waiting for the OSP operators to be completely installed what may cause some race conditions. This PR ensure the InstallPlan is completed before proceeding consuming the deployed operators.


abays (approved):
/lgtm
/approve

PR #2430, opened 2024-10-07
Do not run control plane testing while openstackclient is updated.

sathlan:
To test the control plane during update we leverage the
openstackclient pod.

While the control plane is updated, the openstackclient pod is updated
as well. That means that during a short period of time (less than 2
seconds in the tests we have run) the openstackclient might be
unavailable.

To prevent false positive we:
1. make sure the error was not linked to the openstackclient shutting
down (code 137)
2. or, verify if the error message match the expected one for such
error.

Closes: https://issues.redhat.com/browse/OSPRH-10546

### DIFF
--- a/roles/update/templates/workload_launch_k8s.sh.j2
+++ b/roles/update/templates/workload_launch_k8s.sh.j2
@@ -1,8 +1,53 @@
 #!/usr/bin/bash
-
+set +x
 
 export KUBECONFIG="{{ cifmw_openshift_kubeconfig }}"
 export PATH="{{ cifmw_path }}"
 
-cat "{{ cifmw_update_artifacts_basedir }}/workload_launch.sh" | \
-    oc rsh -n openstack openstackclient env WKL_MODE=sanityfast bash
+OS_POD_TIMEOUT={{ cifmw_update_openstackclient_pod_timeout }}
+WAIT=0
+
+# Temporary file where to put the error message, if any.
+ERROR_FILE=/tmp/cifmw_update_ctl_testing_current_ouput.txt
+rm -f "${ERROR_FILE}"
+
+while [ $((WAIT++)) -lt ${OS_POD_TIMEOUT} ]; do
+    set -o pipefail             # Make sure we get the failure, as tee
+                                # will always succeed.
+    cat "{{ cifmw_update_artifacts_basedir }}/workload_launch.sh" | \
+        oc rsh -n openstack openstackclient env WKL_MODE=sanityfast bash 2>&1 | tee "${ERROR_FILE}"
+    RC=$?
+    set +o pipefail
+    if [ "${RC}" -eq 137 ]; then
+       # When the command is interrupted by the restart of the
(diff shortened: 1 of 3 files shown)


ciecierski:
recheck

roles/update/templates/workload_launch_k8s.sh.j2
```diff
@@ -4,5 +4,41 @@
...
+                continue
+            fi
+        fi
+        # Error is not related to the the openstackclient not being
+        # available. We rethrow it.
+        exit ${RC}
+    else
+        # No error.
+        exit 0
+    elif [ "${RC}" -eq 137 ]; then
```
ciecierski, line 33: It looks to me that you meant `if` here, not `elif`.

roles/update/templates/workload_launch_k8s.sh.j2
```diff
@@ -4,5 +4,41 @@
...
+                sleep 1
+                continue
+            fi
+        fi
+        # Error is not related to the the openstackclient not being
+        # available. We rethrow it.
+        exit ${RC}
+    else
+        # No error.
+        exit 0
```
ciecierski, line 32: `    . Expected an expression, got 'end of print statement'`
Missing` fi` ?

roles/update/templates/workload_launch_k8s.sh.j2
sathlan: The problem was that you cannot have `elif` *after* the `else` statement.

roles/update/templates/workload_launch_k8s.sh.j2
sathlan: Same issue as above.

roles/update/templates/workload_launch_k8s.sh.j2
odyssey4me: If the error code is 137, does it fire for both of these conditions?

I would have thought that in an if/elif you would go from most specific to least specific in terms of order because only the first one will fire, but maybe you know something I don't?

sathlan: Good point.  I've refactored the entire conditional section to be more readable.

ciecierski:
recheck

roles/update/templates/workload_launch_k8s.sh.j2
```diff
@@ -1,8 +1,53 @@
...
+        fi
+        # Error is not related to the the openstackclient not being
+        # available. We rethrow it.
+        exit ${RC}
+    fi
+    # No error.
+    exit 0
+done
+
+# We only reach this code if we we timeout while retrying to
```
ciecierski, line 50: Based on the context of pr and script
```suggestion
# We only reach this code if we hit timeout while retrying to
```

odyssey4me (approved):

ciecierski:
@pablintino Could you add required lgtm flags?

PR #941, opened 2023-12-13
[edpm_deploy]Allow passing extra nova-compute config

gibizer:
The install_yamls edpm_deploy_prep target can take an config file path via DATAPLANE_EXTRA_NOVA_CONFIG_FILE ENV variable. Expose this capability via the new cifmw_edpm_deploy_nova_compute_extra_config ansible variable. This way the zuul job definitions can define the following to pass a config snippet to nova-compute:

  vars:
    cifmw_edpm_deploy_nova_compute_extra_config:
      debug: True

Depends-On: https://github.com/openstack-k8s-operators/install_yamls/pull/658

As a pull request owner and reviewers, I checked that:
- [ ] Appropriate testing is done and actually running
- [ ] Appropriate documentation exists and/or is up-to-date:
  - [ ] README in the role
  - [ ] Content of the docs/source is reflecting the changes

PR #2434, opened 2024-10-08
[edpm_deploy_baremetal] Allow passing extra nova-compute config

karelyatin:
Similar to what done for edpm_deploy[1], required changes on install_yamls side already covered with [2]

Needed by: https://review.rdoproject.org/r/c/testproject/+/54904

[1] https://github.com/openstack-k8s-operators/ci-framework/pull/941
[2] https://github.com/openstack-k8s-operators/install_yamls/pull/729

### DIFF
--- a/roles/edpm_deploy_baremetal/tasks/main.yml
+++ b/roles/edpm_deploy_baremetal/tasks/main.yml
@@ -64,6 +64,30 @@
     name: 'install_yamls_makes'
     tasks_from: 'make_edpm_baremetal_compute'
 
+- name: Pass through extra nova-compute config
+  when:
+    - cifmw_edpm_deploy_baremetal_nova_compute_extra_config is defined
+    - cifmw_edpm_deploy_baremetal_nova_compute_extra_config | length > 0
+  tags:
+    - always
+  vars:
+    _cifmw_edpm_deploy_baremetal_nova_extra_config_file: "{{ cifmw_edpm_deploy_baremetal_basedir }}/nova-extra-config.conf"
+  block:
+    - name: Create the config file
+      ansible.builtin.copy:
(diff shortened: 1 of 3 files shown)


karelyatin:
Tested in https://review.rdoproject.org/r/c/testproject/+/54904

roles/edpm_deploy_baremetal/tasks/main.yml
```diff
@@ -64,6 +64,28 @@
     name: 'install_yamls_makes'
     tasks_from: 'make_edpm_baremetal_compute'
 
+- name: Pass through extra nova-compute config
+  when: cifmw_edpm_deploy_baremetal_nova_compute_extra_config is defined
```
lewisdenny, line 68: question(blocking): Seems this when clause is always going to pass due to cifmw_edpm_deploy_baremetal_nova_compute_extra_config being set by default being set to ""

Is this what you intend? I would have expected something like: 

```
- name: Pass through extra nova-compute config
  when: 
    - cifmw_edpm_deploy_baremetal_nova_compute_extra_config is defined
    - cifmw_edpm_deploy_baremetal_nova_compute_extra_config | length > 0
 ```

lewisdenny (changes requested):
Hi @karelyatin 

Thank you for the patch!

roles/edpm_deploy_baremetal/tasks/main.yml
karelyatin: Yes right, can avoid running it always, the end result is empty conf so was kind of noop, but having this skipped should be better. Will update, thx.

lewisdenny (approved):
/lgtm

Thanks for updating @karelyatin !

PR #2445, opened 2024-10-11
Don't consider non-ocp vms for deploy-ocp

cescgina:
When running in zuul, if we keep the non-ocp vms in the libvirt layout
when running from different executors, the config_drive crashes because
the user data changes. This change filters out the non-ocp vms from the
libvirt layout to avoid this problem.

### DIFF
--- a/deploy-ocp.yml
+++ b/deploy-ocp.yml
@@ -30,25 +30,24 @@
   hosts: "{{ cifmw_target_host | default('localhost') }}"
   gather_facts: true
   tasks:
-    # Inject "start: false" in the layout to not start any VM yet.
-    # Starting the VM will be done later, either by the tool deploying
-    # OSP, or the one deploy RHOSO.
-    # VM initial configuration, when managed, is done using cloud-init.
-    - name: Ensure no VM is started when we create them during this run
+    # Filter out non-ocp vms, that are created, but not started by the
+    # create-infra plabyook.
+    # The rest are created by the create-infra playbook, and managing
+    # them from here can cause problems in some environments, e.g
+    # when running each playbook from a different zuul vm
+    - name: Filter out non ocp vms from libvirt layout
       vars:
-        _no_start: >-
-          {% set _vms = {}                                          -%}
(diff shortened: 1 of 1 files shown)


cescgina:
recheck

deploy-ocp.yml
```diff
@@ -30,17 +30,17 @@
...
-    - name: Ensure no VM is started when we create them during this run
+    # Set amount: 0 for all vms except the ocp ones.
+    # The rest are created by the create-infra playbook, and managing
+    # them from here can cause problems in some environments, e.g
+    # when running each playbook from a different zuul vm
+    - name: Ensure only ocp vms are managed here
       vars:
         _no_start: >-
           {% set _vms = {}                                          -%}
           {% for _type in _cifmw_libvirt_manager_layout.vms.keys()  -%}
```
pablintino, line 41: Using an internal var of the libvirt_manager in a playbook that just consumes the role is kind of a hack.

deploy-ocp.yml
```diff
@@ -30,17 +30,17 @@
...
           {%   if _type is not match('^ocp.*')                      -%}
-          {%     set _ = _vms.update({_type: {'start': false}})     -%}
+          {%     set _ = _vms.update({_type: {'amount': 0 }})       -%}
```
pablintino, line 43: If the idea of the playbook is to reuse a layout that may contains stuff that is not only OCPs isn't safer create an intermediary layout with only the ocps and pass it to the role? No one warranties that the libvirt_manager won't do something with those hosts even if the amount is 0 in the future. Removing them from the layout looks like a less error prone/more maintainable way of passing the layout.

pablintino (changes requested):

deploy-ocp.yml
cescgina: fair enough, that was my first version for this PR, let me recover it

lewisdenny (approved):
/approve

ciecierski (approved):

PR #2446, opened 2024-10-11
Ensure crc dnsmasq server entries are not duplicated

pablintino:
Replace the blockinline by lineinfile to avoid injecting multiple times the same server in the dnsmasq config.


lewisdenny:
recheck

lewisdenny (approved):
/approve

[Suggestion non-blocking] This whole playbook is a mess, a better task name or comment could help here. Something that explains this will add the provided dns forwarders unless already configured.

PR #2448, opened 2024-10-11
[test-operator] Expose SELinuxLevel parameter

lpiwowar:
This commit exposes SELinuxLevel parameter for the test-operator CRs.
This parameter is required for the RWX PVCs to work properly.

### DIFF
--- a/roles/test_operator/README.md
+++ b/roles/test_operator/README.md
@@ -23,6 +23,7 @@ Execute tests via the [test-operator](https://openstack-k8s-operators.github.io/
 * `cifmw_test_operator_storage_class`: (String) Value for `storageClass` in generated Tempest or Tobiko CRD file. Defaults to `"lvms-local-storage"` only if `cifmw_use_lvms` is True, otherwise it defaults to `"local-storage"`.
 * `cifmw_test_operator_delete_logs_pod`: (Boolean) Delete tempest log pod created by the role at the end of the testing. Default value: `false`.
 * `cifmw_test_operator_privileged`: (Boolean) Spawn the test pods with `allowPrivilegedEscalation: true` and default linux capabilities. This is required for certain test-operator functionalities to work properly (e.g.: `extraRPMs`, certain set of tobiko tests). Default value: `true`
+* `cifmw_test_operator_selinux_level`: (String) Specify SELinux level that should be used for pods spawned with the test-operator. Note, that `cifmw_test_operator_privileged: true` must be set when this parameter has non-empty value. Default value: `s0:c478,c978`
 
 ## Tempest specific parameters
 * `cifmw_test_operator_tempest_registry`: (String) The registry where to pull tempest container. Default value: `quay.io`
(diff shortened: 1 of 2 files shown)


lpiwowar:
recheck

pablintino:
We need to start planning the rework of the interface. It's starting to look messy and too tied to the operator CR.
/approve

roles/test_operator/README.md
```diff
@@ -23,6 +23,7 @@ Execute tests via the [test-operator](https://openstack-k8s-operators.github.io/
 * `cifmw_test_operator_storage_class`: (String) Value for `storageClass` in generated Tempest or Tobiko CRD file. Defaults to `"lvms-local-storage"` only if `cifmw_use_lvms` is True, otherwise it defaults to `"local-storage"`.
 * `cifmw_test_operator_delete_logs_pod`: (Boolean) Delete tempest log pod created by the role at the end of the testing. Default value: `false`.
 * `cifmw_test_operator_privileged`: (Boolean) Spawn the test pods with `allowPrivilegedEscalation: true` and default linux capabilities. This is required for certain test-operator functionalities to work properly (e.g.: `extraRPMs`, certain set of tobiko tests). Default value: `true`
+* `cifmw_test_operator_selinux_level`: (String) Specify SELinux level that should be used for pods spawned with the test-operator. Note, that `cifmw_test_operator_privileged: true` must be set when this parameter has non-empty value. Default value: `s0:c478,c978`
```
lewisdenny, line 26: Is these two vars are must be set together than an assert would be strongly recommended. 

A pattern like this [1] where a validate task file is called at the beginning of main.yml to assert`if cifmw_test_operator_selinux_level | length > 0 then cifmw_test_operator_privileged | bool` for example. 

[1] https://github.com/openstack-k8s-operators/ci-framework/blob/fd9420beeae007e0d2f051d13ade634ce3152c43/roles/ci_network/tasks/main.yml#L28-L34

lewisdenny (changes requested):

roles/test_operator/README.md
pablintino: @lewisdenny there's another point of view. Why should we enforce a test-operator behavior here in the framework? Is the operator already doing that check? If the answer is no maybe the operator, that is the one that owns the CR and the responsibility of validating it should be the one that validates it. If the answer is yes, nice, the validation is already in place and the CR will fails as soon as it's applied.

lpiwowar: @lewisdenny, @pablintino the test-operator does not verify the SELinuxLevel at the moment but I can promise you it will. I have a [WIP PR here](https://github.com/openstack-k8s-operators/test-operator/pull/228) that unifies the verification for `privileged: ` and other parameters.

For now I'm not updating this PR. Let me know if the promise of me adding the logic to the test-operator is enough for you. Otherwise, I'll update this PR.

lewisdenny: I don't mind where the enforcement is, I just anticipate a question in the future like "Hey I set `cifmw_test_operator_selinux_level` variable in the **CI-Framework** and it's not working" and asserts are cheep.

If the validation will be added to the test-operator that will satisfy me :)

lewisdenny (approved):
/lgtm

PR #2449, opened 2024-10-14
Enable COO in osasinfra DT

mandre:
This is a requirement for telemetry that we're adding to the OSASINFRA DT via a separate architecture PR.


lewisdenny (approved):
/approve

PR #2454, opened 2024-10-14
DNM Check binding dnsmasq listen-address to 0.0.0.0

danpawlik:
There is an issue, that the DNS requests are redirected to the cloud provider to resolve domain: api-int.crc.testing. Normally it is right that it want to resolve, but the dnsmasq should take the query and reply, however this is not the case.
Let's set the dnsmasq listen-address to be accessible to all addresses on that host.


danpawlik:
recheck

roles/libvirt_manager/tasks/create_networks.yml
```diff
@@ -197,7 +197,8 @@
     cifmw_dnsmasq_network_name: "{{ item.name }}"
     cifmw_dnsmasq_network_state: present
     cifmw_dnsmasq_network_definition: "{{ item }}"
-    cifmw_dnsmasq_network_listen_addresses: "{{ _dns_address[item.name] }}"
+    cifmw_dnsmasq_network_listen_addresses:
```
lewisdenny, line 200: Please drop this file from this PR

roles/ci_network/templates/local_domain.conf.j2
```diff
@@ -1,5 +1,5 @@
 # This file is generated by ci-framework tool
-listen-address={{ config.listen_addresses | join(',') }}
+listen-address=0.0.0.0
```
lewisdenny, line 2: Please drop this file from this PR

roles/ci_network/tasks/apply-dnsmasq.yml
```diff
@@ -60,8 +60,7 @@
   vars:
     data:
       listen_addresses:
-        - "127.0.0.1"
-        - "{{ ansible_default_ipv4.address }}"
+        - "0.0.0.0"
```
lewisdenny, line 63: Please drop this file from this PR

ci/playbooks/multinode-customizations.yml
```diff
@@ -108,16 +108,32 @@
...
+
             - name: Configure dnsmasq listen-address to listen on both br-ex and ci-private-network
               when:
                 - not _dnsmasq.stat.exists
               become: true
               ansible.builtin.lineinfile:
                 path: "{{ _dnsmasq_config }}"
-                insertafter: '^listen-address='
-                line: "listen-address={{ _crc_default_net_ip | ansible.utils.ipaddr('address') }}"
+                insertbefore: '^expand-hosts'
```
lewisdenny, line 126: Is there a reason it needs to be before expand-hosts? if not, please just add it to EOF like line 95
```suggestion
                insertafter: "EOF"
```

ci/playbooks/multinode-customizations.yml
```diff
@@ -108,16 +108,32 @@
...
+                line: |
+                  listen-address=0.0.0.0
+                  listen-address=::
                 validate: "{{ _validate }}"
 
+            - name: Print current config
```
lewisdenny, line 132: Please remove this after your testing.

lewisdenny (changes requested):

ci/playbooks/multinode-customizations.yml
danpawlik: of course ;)

lewisdenny:
recheck
infra issues: `Error: initializing source docker://quay.io/podified-antelope-centos9/openstack-iscsid:current-podified: pinging container registry quay.io: Get \\\"https://quay.io/v2/\\\": dial tcp: lookup quay.io:`

danpawlik:
Probably that change will require additional task that was proposed https://review.rdoproject.org/r/c/config/+/55035/4/roles/prepare-crc-extracted/tasks/dnsmasq-nm.yaml#80

which contains:
```
        - name: Set the bind interfaces
          become: true
          ansible.builtin.lineinfile:
            path: /etc/dnsmasq.d/crc-dnsmasq.conf
            line: "interface={{ item }}"
          loop: "{{ ansible_facts['interfaces'] }}"
          when: "{{ item.startswith('enp') }} or {{ item.startswith('eth') }}"
```
It is just a PoC.

PR #2456, opened 2024-10-14

drosenfe:
Add generation of xml results file to validations role. This will allow polarion status reporting for validations.


drosenfe:
recheck

bshephar (approved):
If this is working to get the data you need into Polarion then it lgtm. Just some follow up tidying that we can do, but realistically it's orthogonal to this change anyway. I don't want to distract from the objective of the PR with that conversation here, this change just draws attention to that requirement for me.

lewisdenny (approved):
/approve

JFYI we have  `roles/polarion` for uploading the polarion xml once ready.

bshephar:
> /approve
> 
> JFYI we have `roles/polarion` for uploading the polarion xml once ready.

I think ultimately, what might be nice as a follow up to this PR is a new task file in the polarion role that will take some Ansible variable inputs, and use a jinja2 template to format those results into the required XML for polarion. Then instead of all the XML handling we're doing in this PR, we can simply call import tasks_from that new polarion  file and it can handle to XML side of things.

But since this is only going to impact the EDPM validations, it's fairly limited impact if something were to stop working for some reason. I'm in favor of merging it to keep progress going, and maybe we can backlog a Epic to tidy up the XML side of things.

hjensas (approved):

roles/validations/tasks/edpm/hotfix.yml
```diff
@@ -6,86 +6,96 @@
...
-    output_dir: "{{ cifmw_validations_basedir }}/artifacts"
-    script: >-
-      oc get osdpns -n {{ cifmw_validations_namespace }} --no-headers -o custom-columns=":metadata.name"
-  register: deployed_nodeset_name
+- name: Set validator status to passed
+  ansible.builtin.set_fact:
+    validator_status: "passed"
+
+- name: Validation block
+  block:
```
pablintino, line 14: Now that we have a block why should we repeat 
```yaml
      environment:
        PATH: "{{ cifmw_path }}"
```
over and over?

roles/validations/tasks/xmlcreate.yml
```diff
@@ -0,0 +1,58 @@
+---
+- name: Get validation start time from system
```
pablintino, line 2: Why not use the `now(fmt='%s.%f')`? It's the Ansible way of doing this.
A simple cast to `float` will convert it to a numeric timestamp

roles/validations/tasks/xmlcreate.yml
```diff
@@ -0,0 +1,58 @@
...
+  ansible.builtin.command:
+    cmd: date +%Y-%m-%d-%H:%M:%S.%3N
+  register: individual_validation_start_time
+
+- name: Run a validation
+  ansible.builtin.include_tasks: "{{ item }}"
```
pablintino, line 8: A task file called xmlcreate.yml that runs a validation task is a bit confusing, it does way more than generating an XML.

roles/validations/tasks/xmlcreate.yml
```diff
@@ -0,0 +1,58 @@
...
+
+- name: Get validation end time from system
+  ansible.builtin.command:
+    cmd: date +%Y-%m-%d-%H:%M:%S.%3N
```
pablintino, line 12: ditto.

roles/validations/tasks/main.yml
```diff
@@ -56,5 +68,34 @@
...
+    cmd: date +%Y-%m-%d-%H:%M:%S.%3N
+  register: validations_end_time
+
+- name: Calculate total validations run time
+  ansible.builtin.set_fact:
+    validations_run_time: "{{((validations_end_time.stdout | to_datetime('%Y-%m-%d-%H:%M:%S.%f')) - (validations_start_time.stdout | to_datetime('%Y-%m-%d-%H:%M:%S.%f'))).total_seconds()}}"
+
+- name: Write first line of xml results file
+  ansible.builtin.shell: |
+    mkdir -p "{{ cifmw_validations_xml_status_file_dir }}"
```
pablintino, line 86: ` mkdir -p "{{ cifmw_validations_xml_status_file_dir }}"` Why using a shell to do something Ansible is designed for?

roles/validations/tasks/main.yml
```diff
@@ -56,5 +68,34 @@
...
+    echo "<testsuite errors=\"{{ validations_errors }}\" failures=\"{{ validations_failed }}\" name=\"\" tests=\"{{ validations_executed }}\" time=\"{{ validations_run_time }}\">" >> {{ cifmw_validations_xml_status_file_dir }}/validations_results.xml
```
pablintino, line 87: Generating XML with bash is asking for trouble. There's not nice way to generate XML in Ansible, but using jinja2 would be a step forward.

roles/validations/tasks/xmlcreate.yml
```diff
@@ -0,0 +1,58 @@
...
+  ansible.builtin.set_fact:
+    xml_line_list: "{{ xml_line_list + new_xml_line_list_entry }}"
+  when: '"passed" in validator_status'
+
+- name: Add testcase and failure reason to xml line list when a validation fails
+  when: '"failed" in validator_status'
+  block:
+    - name: Make new xml line list entry
+      ansible.builtin.set_fact:
+        new_xml_line_list_entry: ['  <testcase classname="validations.{{item | dirname}}" name="{{item | basename}}" time="{{ individual_validator_run_time }}">']
```
pablintino, line 38: Details like the indentation handling you do here is one of the reasons I added this comment:
https://github.com/openstack-k8s-operators/ci-framework/pull/2456/files#diff-745f651c7f88527b40991eca9fbe69b880b93734ee5a24b140a3fcac8becac2aR87

roles/validations/tasks/edpm/hotfix.yml
drosenfe: Moved the the three environment lines so they applied to the whole block and deleted the rest.

roles/validations/tasks/xmlcreate.yml
drosenfe: Updated to use: now(fmt='%s.%f')

roles/validations/tasks/xmlcreate.yml
drosenfe: Renamed to run_validation.yml

roles/validations/tasks/xmlcreate.yml
drosenfe: Used now(fmt='%s.%f') instead

roles/validations/tasks/main.yml
drosenfe: Used ansible.builtin.file instead

roles/validations/tasks/main.yml
drosenfe: @pablintino Is this comment a goal or a requirement? Before I started this review I asked on the proj-create-integration-pipeline-jobs about creating xml files and the answer was:   it could be even a simple set of echo "" , but if you want to use something easy and more structured you could use https://pypi.org/project/junitparser/

pablintino: @drosenfe It's a suggestion to improve the maintainability of the role, specially given that I don't see a molecule test that ensure the rendered XML matches the expectations. If you and all the ones involved in this role are fine I'm ok with the change.

drosenfe: I looked quite a lot at  junitparser. Doing a fake test from command line this is the output it gives:

<?xml version="1.0" encoding="utf-8"?>
<testsuites>
        <testsuite name="validations" tests="3" errors="0" failures="0" skipped="0" time="1.8">
                <testcase name="hugepages_and_reboot.yml" classname="validations.edpm" time="0.5"/>
                <testcase name="hotfix.yml" classname="validations.edpm" time="0.6"/>
                <testcase name="scaledown.yml" classname="validations.edpm" time="0.7"/>
        </testsuite>
</testsuites>

This different in a few ways from the current xml files generated by tempest

- it uses tabs instead of spaces (could be changed with sed after creation)
- it has an extra xml header (could be deleted after creation)
- the test cases have order: name, classname, time while tempest xml files have order: classname, name, time
- the testsuite order for tests, errors, failures, skipped is different

What does anyone think about using junitparser? There were some advantages. It sums the test cases, failures, and total time so that the validations roles does not need too track those.

tosky: From an XML point of view:
* tabs vs spaces don't matter, they are all whitespaces (https://www.w3.org/TR/xml/#sec-common-syn)
* the XML header is correct, it doesn't need to be removed. It's part of valid XML files
* the order of attributes in an XML tag (so both testcase and testsuite) doesn't matter.
You don't even need to use pretty=True when exporting the XML file, the file is still a valid XML file (it can be rendered for humans by the viewer).

drosenfe: Latest version uses junitparser to generate xml file

roles/validations/tasks/xmlcreate.yml
drosenfe: Latest version uses junitparser to generate xml

drosenfe:
read the docs test was failing because the branch needed this commit: https://github.com/openstack-k8s-operators/ci-framework/pull/2526   I didn't update the branch correctly and it got do-not-merge/contains-merge-commits error. When I tried to fix that it got even worse. I will create a new review.

PR #2464, opened 2024-10-17
[hci_prepare] Ensure generated resources are applied in the proper NS

pablintino:
The namespace field in the generated CRs is wronly placed. That, and the fact we call apply without specifying the namespace may lead to the resource to be applied to the wrong namespace.

Jira: https://issues.redhat.com/browse/OSPRH-10332

### DIFF
--- a/roles/hci_prepare/tasks/common.yml
+++ b/roles/hci_prepare/tasks/common.yml
@@ -14,13 +14,9 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
-cifmw_hci_prepare_basedir: "{{ cifmw_basedir | default(ansible_user_dir ~ '/ci-framework-data') }}"
-cifmw_hci_prepare_dataplane_dir: "{{ cifmw_basedir }}/artifacts/manifests/kustomizations/dataplane"
-cifmw_hci_prepare_dryrun: false
-cifmw_hci_prepare_skip_load_parameters: false
-cifmw_hci_prepare_ceph_secret_path: "/tmp/k8s_ceph_secret.yml"
-cifmw_hci_prepare_enable_storage_mgmt: true
-cifmw_hci_prepare_enable_repo_setup_service: true
-cifmw_hci_prepare_storage_mgmt_mtu: 1500
-cifmw_hci_prepare_storage_mgmt_vlan: 23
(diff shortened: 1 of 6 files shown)


roles/hci_prepare/tasks/common.yml
```diff
@@ -0,0 +1,22 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+- name: Set common facts
+  ansible.builtin.set_fact:
+    __cifmw_hci_prepare_namespace: >-
```
lewisdenny, line 19:
```suggestion
    _cifmw_hci_prepare_namespace: >-
```

You have your brain in python mode :P

lewisdenny (changes requested):

roles/hci_prepare/tasks/common.yml
pablintino: Fixed.

lewisdenny (approved):
/approve

frenzyfriday:
Do we need an additional var _cifmw_hci_prepare_namespace defined on the job defs now?

lewisdenny:
> Do we need an additional var _cifmw_hci_prepare_namespace defined on the job defs now?

Good question, no, this patch doesn't change the default behavior, it just allows cifmw_install_yamls_defaults to not be needed, when not provided we now have a default :)

If you want to override, which we don't by default, then you could set cifmw_hci_prepare_namespace in the job def but it's not needed.

frenzyfriday:
> > Do we need an additional var _cifmw_hci_prepare_namespace defined on the job defs now?
> 
> Good question, no, this patch doesn't change the default behavior, it just allows cifmw_install_yamls_defaults to not be needed, when not provided we now have a default :)
> 
> If you want to override, which we don't by default, then you could set cifmw_hci_prepare_namespace in the job def but it's not needed.

Ah, right right. This patch is setting the value for the _cifmw_hci_prepare_namespace, I missed that. Thanks!

PR #2465, opened 2024-10-18
[OSPCIX-532] Set volume.catalog_type to volumev3

lpiwowar:
This patch in tempest changed the default value of the catalog_type for the cinder service to block-storage [1]. In RHOSO however, we seem to be still using volumev3.

This patch ensures that the default value in ci-framework deployed environment is still volumev3.

[1] https://review.opendev.org/c/openstack/tempest/+/930296

### DIFF
--- a/roles/tempest/vars/main.yml
+++ b/roles/tempest/vars/main.yml
@@ -4,3 +4,4 @@ cifmw_tempest_tempestconf_profile_default:
   overrides:
     identity.v3_endpoint_type: public
     compute-feature-enabled.dhcp_domain: ''
+    volume.catalog_type: volumev3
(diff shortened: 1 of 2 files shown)


lpiwowar:
@openstack-k8s-operators/cifmw-squad I think this is ready to be merged. But I think we should get a confirmation from somebody from storage that it is ok thing to do.

ASBishop (approved):
/lgtm

As discussed in slack, the upstream cinder-operator will be updated to specify 'block-storage' when upstream operators move on from antelope to (presumably) epoxy. Meanwhile, this PR ensures we can used upstream tempest.

PR #2468, opened 2024-10-21
Update OWNERS file to be more granular

lewisdenny:
By using OWNERS_ALIASES we can have teams own there own roles by adding
an OWNERS file into the directory. This patch starts by allowing two
shiftstack members to review patches that are in roles/shiftstack/**

---
Todo: 
- [ ] Find a home for the missing users


lewisdenny:
Closing as we are going to go with GitHubs CODEOWNERS file after all.

PR #2469, opened 2024-10-21
[polarion] Remove results_merged.xml files

kopecmartin:
The results_merged.xml files are created only for polarion reporting purposes, they are basically temporary files. Their existance after the polarion report is completed, is irrelevant as the original files that were combined to create these merged files still exist.

For logs gathering purposes it's better if we avoid duplication and remove the merged files after they served their purpose.


lewisdenny (approved):
/approve

/cc @openstack-k8s-operators/cifmw-squad

PR #2471, opened 2024-10-21
Enable RHCS 7 in 17 TripleO deployment

fmount:
For adoption, we need to align ci-framework to deploy `Ceph 7` with `17.1`

Jira: https://issues.redhat.com/browse/OSPRH-7886

### DIFF
--- a/hooks/playbooks/adoption_deploy_ceph.yml
+++ b/hooks/playbooks/adoption_deploy_ceph.yml
@@ -143,6 +143,19 @@
             state: present
           loop: "{{ _tripleo_nodes }}"
 
+        # NOTE: TripleO has the hardcoded --yes-i-know option that is not valid anymore
+        # in RHCS 7. TripleO does not receive any new patch both upstream and downstream
+        # (it is a retired project), hence the only option we have is to patch the
+        # current code to not have that line.
+        - name: Patch Ceph bootstrap task
+          become: true
+          delegate_to: "osp-undercloud-0"
+          ansible.builtin.lineinfile:
+            path: "/usr/share/ansible/roles/tripleo_cephadm/tasks/bootstrap.yaml"
+            state: absent
+            regexp: '--yes-i-know'
+            backup: true
+
         - name: Deploy ceph
           delegate_to: "osp-undercloud-0"
           vars:
@@ -152,7 +165,7 @@
               --ntp-server {{ cifmw_adoption_osp_deploy_ntp_server }}
               --ceph-spec ceph_spec.yaml
               --network-data {{ _network_data_file_dest }}
-              --cephadm-default-container
+              --container-image-prepare {{ ansible_user_dir }}/containers-prepare-parameters.yaml
               --output {{ ansible_user_dir }}/deployed_ceph.yaml
           cifmw.general.ci_script:
             chdir: "{{ ansible_user_dir }}"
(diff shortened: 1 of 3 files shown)


roles/adoption_osp_deploy/tasks/prepare_undercloud.yml
```diff
@@ -54,6 +54,17 @@
         script: "{{ _container_prepare_cmd }}"
       when: cifmw_adoption_osp_deploy_scenario.container_prepare_params is not defined
 
+    # Adoption requires Ceph 7 (Reef) as a requirement. Instead of performing a Ceph
```
fmount, line 57: @cescgina this is supposed to be the equivalent of https://github.com/openstack-k8s-operators/install_yamls/pull/930.

However, this is not enough, because due to a bz that I will link, I need to change something else in the TripleO task (see https://github.com/openstack-k8s-operators/install_yamls/pull/930).

Where is deployed Ceph in this role?

cescgina: @fmount thanks for looking into this. This playbook https://github.com/openstack-k8s-operators/ci-framework/blob/main/hooks/playbooks/adoption_deploy_ceph.yml deploys ceph in this role (it's called as a hook)

fmount: Thank you! I'm going to update that playbook in my next iteration to this patch!

roles/adoption_osp_deploy/tasks/prepare_undercloud.yml
```diff
@@ -54,6 +54,17 @@
...
+    # upgrade from 6 (the default) to 7, let's try to deploy 7 in greenfield
+    - name: Patch containers-prepare-parameters Ceph container images
+      when: cifmw_adoption_osp_deploy_scenario.container_prepare_params is not defined
```
fmount, line 60: @cescgina do you think I need this `when` condition or I can patch this file regardless of the variable?

cescgina: you do not need the `when`, as the file should be defined regardless. However, I just realized that I forgot to implement that task for the path when that when does not apply, I'll post a quick PR for that

cescgina: https://github.com/openstack-k8s-operators/ci-framework/pull/2474

fmount: Thank you, makes sense. Also do you know how we can test this patch? (e.g. a testproject where we replace install_yamls w/ ci-framework driven deploy?)

cescgina: yes, I can add it to my testproject

fmount: Thank you!

hooks/playbooks/adoption_deploy_ceph.yml
```diff
@@ -143,6 +143,18 @@
             state: present
           loop: "{{ _tripleo_nodes }}"
 
+        # NOTE: TripleO has the hardcoded --yes-i-know option that is not valid anymore
+        # in RHCS 7. TripleO does not receive any new patch both upstream and downstream
+        # (it is a retired project), hence the only option we have is to patch the
+        # current code to not have that line.
+        - name: Patch Ceph bootstrap task
```
fmount, line 150: @marios FYI as done for `install_yamls`, we need to patch the bootstrap command here as well.

marios: it seems so much more civilised in ansible ;)

cescgina:
@fmount jfyi I rebased the PR because it had a conflict and it was preventing me testing it

cescgina (approved):
I tested this internally and it worked well

fultonj (approved):
/approve
/lgtm

PR #2474, opened 2024-10-21
Copy container-prepare file if defined in scenario

cescgina:
If the adoption source scenario contains a 'container_prepare_params'
file to use, copy to the undercloud instead of generating it.

### DIFF
--- a/roles/adoption_osp_deploy/tasks/prepare_undercloud.yml
+++ b/roles/adoption_osp_deploy/tasks/prepare_undercloud.yml
@@ -54,6 +54,19 @@
         script: "{{ _container_prepare_cmd }}"
       when: cifmw_adoption_osp_deploy_scenario.container_prepare_params is not defined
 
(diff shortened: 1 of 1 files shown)


roles/adoption_osp_deploy/tasks/prepare_undercloud.yml
```diff
@@ -54,6 +54,19 @@
...
 
+    - name: Copy containers-prepare-parameters if needed
+      vars:
+        _container_prapare_path: >-
+          {{
+            [cifmw_adoption_source_scenario_path,
+            cifmw_adoption_osp_deploy_scenario.container_prepare_params
+            ] | path_join
+          }}
+      cifmw.general.copy:
```
pablintino, line 65: I don't think this module exists.

cescgina: thanks Pablo fixed the module name

PR #2475, opened 2024-10-21
[ci_nmstate] DNS generation and empty hosts fixes

pablintino:
Fixing ci_nmstate to not generate DNS config if not DNS were configured. If an instance that is a k8s one has no configuration we won't try to provision it using raw nmstate.


lewisdenny (changes requested):
Please add `cifmw_ci_nmstate_olm_operator_group_patch` to the role readme

lewisdenny (approved):
/approve

PR #2476, opened 2024-10-21
[ci_nmstate] Conditionally call ci_nmstate

pablintino:


lewisdenny (approved):
/approve

frenzyfriday:
I got the part about task: Configure hosts networking using nmstate , but what is the task: Load Networking Environment Definition for?

lewisdenny:
> I got the part about task: Configure hosts networking using nmstate , but what is the task: Load Networking Environment Definition for?

Configure hosts networking using nmstate is just moved earlier in the playbook. That task calls a task file from the network_mapper role that creates the cifmw_networking_env_definition variable used by the framework and by the ci_nmstate role to generate the nncp resource. 

The ci_nmstate role calls the same task file to create the cifmw_networking_env_definition var but it's skipped if already defined, as the call to ci_nmstate is conditional we need it in both places.

PR #2478, opened 2024-10-21
Force pvcreate in CI

fmount:
Looks like there are issues with disk wipe in CI [1]. This double force the PV creation, and it overwrites any LVM metadata or partition tables without further warnings.

Jira: https://issues.redhat.com/browse/OSPCIX-540


fmount:
Abandoning as not required by CI anymore.

PR #2482, opened 2024-10-22
Add HSM Prep role

vakwetu:


roles/hsm_prep/README.md
```diff
@@ -0,0 +1,66 @@
...
+* Upload those images to a private repository for use in setting up a CI job.
+* Create any required config to be mounted by the barbican images to interact with the HSM
+
+For the Lunasa, we expect some preparatory steps to be completed prior to execution in order for the
+role to complete successfully.
+* We expect a VM that contains the Lunasa client software and that is registered as a lunasa client.
+  This VM should contain the following contents:
+  * The contents of the minimal linux client in a zipped tar file.
+  * The lunasa binaries that need to be added to the image under a specified directory.
+  * The lunasa HSM server cert.
```
dmendiza, line 17: Since you're going to tar up the linux-minimal client, why not also add the binaries there too?

roles/hsm_prep/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+- name: Set up Luna
+  ansible.builtin.include_role: hsm_prep
+  vars:
+    cifmw_hsm_admin_password: "<HSM admin password>"
+    cifmw_hsm_server_ip: "IP of HSM"
+    cifmw_hsm_client_ip: "IP of the client - this could be the hypervisor where the Openshift nodes run"
+    cifmw_hsm_luna_clientvm_ip: "IP for the pre-configured running lunaclient"
+    cifmw_hsm_luna_clientvm_user: "User to log into the pre-configured running lunaclient"
+    cifmw_hsm_luna_clientvm_password: "password to log into the pre-configured running lunaclient"
```
dmendiza, line 34: Did you consider using an SSH key instead of a password?

roles/hsm_prep/README.md
```diff
@@ -0,0 +1,66 @@
...
+* `cifmw_hsm_working_dir`: (String) Working directory to store artifacts.  Default value: `/tmp/hsm-prep-working-dir`
+* `cifmw_hsm_client_ip`: (String) ip address or hostname of the client VM
+
+### Luna Parameters
+* `cifmw_hsm_luna_clientvm_ip`: (String) ip address or hostname of the luna client VM
+* `cifmw_hsm_luna_clientvm_user`: (String) user login for the luna client VM
+* `cifmw_hsm_luna_clientvm_password`: (String) user password for the luna client VM
+* `cifmw_hsm_luna_minclient_src`: (String) Location of linux minimal client tarball on the luna client VM. Default value: `/opt/data/Linux-Minimal-Client.tar.gz`
+* `cifmw_hsm_luna_binaries_src`: (String) Location of the luna binaries on the luna client VM. Default value: `/opt/data/bin`
+* `cifmw_hsm_luna_server_cert_src`: (String) Location of HSM server cert on the luna client VM.  Default value: ` /usr/safenet/lunaclient/bin/`
```
dmendiza, line 62: Why put the cert in a `bin` folder?  It makes more sens t put it in `/opt/data` with the rest of the pre-provisioned files.

roles/hsm_prep/README.md
vakwetu: yeah - I suppose we could do that.  six and a half dozen ..

roles/hsm_prep/README.md
```diff
@@ -0,0 +1,63 @@
...
+    cifmw_hsm_server_ip: "IP of HSM"
+    cifmw_hsm_client_ip: "IP of the client - this could be the hypervisor where the Openshift nodes run"
+    cifmw_hsm_luna_partition: "HSM partition for th client to join"
+
+## Parameters
+
+### HSM Details
+* `cifmw_hsm_hsmtype`: (String) The type of HSM required.  Currently, only "luna" is supported. Default value: `luna`
+* `cifmw_hsm_admin_user`: (String) The user to log into the HSM.  Default value: `admin`
+* `cifmw_hsm_admin_password`: (String) The password to log into the HSM.
```
dmendiza, line 42: I think these two are only relevant for Lunas.

roles/hsm_prep/files/image_add_luna_minimal_client.sh
dmendiza, line 1: Why not use the script from the repo that is passed in as a parameter (`cifmw_hsm_barbican_operator_repo`)? Otherwise we're going to have to double up maintenance to the script since changes there would also have to be made here.

vakwetu: I plan to.  This is just here temporarily until that script merges.

vakwetu:
This is OK to close.  While this PR was useful as a first pass, we've since moved  a lot of this logic to separate repo.

PR #2484, opened 2024-10-23
Do not mount non-existing controller private key

lpiwowar:
Some set of tests, for example those in whitebox neutron tempest
plugin, requires a private key that allows them to connect to the
controller node.

This PR ensures that the private key is mounted only when it does
exist.

This change is required because of:
- [OSPCIX-546](https://issues.redhat.com/browse/OSPCIX-546)

[OSPCIX-546](https://issues.redhat.com//browse/OSPCIX-546) further indirectly blocks:
- [OSPCIX-545](https://issues.redhat.com/browse/OSPCIX-545)
- [OSPCIX-532](https://issues.redhat.com/browse/OSPCIX-532)
- [OSPCIX-543](https://issues.redhat.com/browse/OSPCIX-543)

### DIFF
--- a/roles/test_operator/tasks/tempest-tests.yml
+++ b/roles/test_operator/tasks/tempest-tests.yml
@@ -81,38 +81,47 @@
                       }}}, recursive=true)
           }}
 
-- name: Ensure a secret for the cifmw private key file exists
-  when:
-    - not cifmw_test_operator_dry_run | bool
-    - cifmw_test_operator_tempest_ssh_key_secret_name is not defined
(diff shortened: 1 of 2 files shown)


lpiwowar:
recheck

```
The connection to the server oauth-openshift.apps-crc.testing was refused - did you specify the right host or port?
Can not login to the CRC cluster. Sleeping...
Checking login to the cluster 56...
WARNING: Using insecure TLS client config. Setting this option is not supported!
```

PR #2492, opened 2024-10-25
libvirt_manager extra nets - firewall zone

hjensas:
Add support in the libvirt_manager role to overload the nmstate config passed in `cifmw_libvirt_manager_extra_network_configuration` with a firewall zone hint using key name `cifmw_firewall_zone` for an interface in the nmstate interfaces section.

Example:
```
  cifmw_libvirt_manager_extra_network_configuration:
    interfaces:
      - name: vlan10
        cifmw_firewall_zone: libvirt
        type: vlan 
        ...
```
The key is filtered before passing the data to the ci_nmstate role.

When defined a task following the task calling ci_nmstate will run the appropriate firewall-cmd to change the zone for the interfaces.

Jira: [OSPRH-10966](https://issues.redhat.com//browse/OSPRH-10966)


lewisdenny (approved):
/lgtm

PR #2494, opened 2024-10-29
Update Molecule config for local use

lewisdenny:
- Change container name to instance to match what runs in CI so scenarios can be written to run locally and in CI without tweaks
- Remove unused TRIPLEO_ANSIBLE overrides
- Remove unused YAML anchors

### DIFF
--- a/.config/molecule/config_podman.yml
+++ b/.config/molecule/config_podman.yml
@@ -5,20 +5,17 @@ driver:
 log: true
 
 platforms:
-  - name: centos
-    hostname: centos
-    image: ${TRIPLEO_ANSIBLE_MOLECULE_IMAGE:-"ubi9/ubi-init"}
+  - name: instance
+    hostname: instance
+    image: "ubi9/ubi-init"
     registry:
-      url: ${TRIPLEO_ANSIBLE_MOLECULE_REGISTRY:-"registry.access.redhat.com"}
-    dockerfile: Dockerfile.j2
-    pkg_extras: python*setuptools
-    command: ${TRIPLEO_ANSIBLE_MOLECULE_COMMAND:-"/sbin/init"}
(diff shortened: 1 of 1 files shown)


lewisdenny:
recheck

lewisdenny:
recheck

lewisdenny:
recheck

lewisdenny:
recheck
added hold

lewisdenny:
recheck

lewisdenny:
@openstack-k8s-operators/cifmw-squad Hey folks, requesting an override for this patch -_-

.config/molecule/config_podman.yml
```diff
@@ -5,42 +5,30 @@ driver:
...
   env:
     ANSIBLE_STDOUT_CALLBACK: yaml
-    ANSIBLE_ROLES_PATH: "${ANSIBLE_ROLES_PATH}:${HOME}/zuul-jobs/roles"
-    ANSIBLE_LIBRARY: "${ANSIBLE_LIBRARY:-/usr/share/ansible/plugins/modules}"
-    ANSIBLE_FILTER_PLUGINS: "${ANSIBLE_FILTER_PLUGINS:-/usr/share/ansible/plugins/filter}"
   config_options:
     defaults:
       remote_tmp: /tmp
-
-scenario:
```
frenzyfriday, line 42: lgtm, but I am trying to get some context:
what was this scenario being used for?

lewisdenny: So how the Molecule test works is it merges the configs together, each test under roles/<rolename>/molecule/* has it's own partial config. The idea being that `config_podman.yml` holds the common config for all the tests and then the test overrides/sets what is needed for the role. For example setting hostvars. 

Scenarios is not something that should be defined as a default but left to each test to set themselves :)

PR #2497, opened 2024-10-29
[os_net_setup] Add --availability-zone-hint to network creation

eduolivares:
This PR adds the option to configure external networks for certain AZs. In order to do this, the list `availability_zone_hints` should be configured as port of `cifmw_os_net_setup_config`.

### DIFF
--- a/roles/os_net_setup/defaults/main.yml
+++ b/roles/os_net_setup/defaults/main.yml
@@ -8,6 +8,12 @@ cifmw_os_net_setup_config:
     is_default: true
     provider_network_type: flat
     provider_physical_network: datacentre
+    # availability_zone_hints is a list of AZs that should have been configured in neutron controlplane). Example:
+    # availability_zone_hints:
+    #   - zone-1
+    #   - zone-2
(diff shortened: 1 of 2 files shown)


evallesp:
(non-blocking) suggestion: I see some vars used in the template that are missing of some context and format.
Unsure what's the format of availability_zone_hints but I might have a lot of missing context.

WDYT about adding that to the default file: https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/os_net_setup/defaults/main.yml to see an example?

I'm thinking if it might make sense to rework eventually in the README file to briefly explain each of the possibilities of:  cifmw_os_net_setup_config

Apart of this: lgtm

lewisdenny (changes requested):
I agree with @evallesp 

Please add availability_zone_hints to the role defaults, others are missing but they can be added in a follow up if we get time to refactor this role.

eduolivares:
recheck

pablintino:
recheck

lewisdenny (approved):
/lgtm

Thanks @eduolivares ++

PR #2485, opened 2024-10-23
Add missing node groups for multicell nodeset

bogdando:
JIRA [OSPRH-6548](https://issues.redhat.com/browse/OSPRH-6548)

Required-By: https://github.com/openstack-k8s-operators/data-plane-adoption/pull/517

PR #2498, opened 2024-10-29
Add missing molecule jobs for roles

cescgina:
The adoption_osp_deploy and ci_dcn_sites roles are missing molecule jobs, which makes the cifmw-pod-zuul-files check fail in another PR[1]. This change adds the output of `make role_molecule` so the molecule job list is up-to-date.

[1] https://github.com/openstack-k8s-operators/ci-framework/pull/2485


lewisdenny (approved):
/lgtm

PR #2500, opened 2024-10-30
Dynamically rename Tempest resource post-update

lkuchlan:
Renamed the Tempest resource name after updates using the "cifmw_test_operator_tempest_name" parameter.
This change enables the test operator role to run multiple times consecutively without requiring resource cleanup, improving test isolation and efficiency.


lpiwowar (approved):
Looks good to me! :+1:

lewisdenny (approved):
/approve

PR #2459, opened 2024-10-15
Add hook to wait for compute service

hjensas:
This is a workaround.

The openstackdataplane deployment can be marked as complete even if nova is not fully running yet.

Add a hook to run command ...
  `openstack compute service list --service nova-compute -f value | wc -l`
... and wait for the line count to match the expected number of compute services.

This hook will allow us to insert a wait-and-check for compute services being up before running `discover_hosts`.

PR #2501, opened 2024-10-30
Retry mechanism for discover hosts

rdiazcam:
This retry mechanism is intended to raise an error if no computes are
    found after the dataplane deployment, preventing futher execution of
    tasks (which depend on a successful dataplane deployment)[0].
    Limitation: a discover_hosts execution with --strict option is
    considered successful (exit code 0) when an unmapped host is
    discovered so if for example we have two hosts to discover and only
    one is found no error will be thrown.
    
    [0] https://issues.redhat.com/browse/OSPRH-11052

### DIFF
--- a/playbooks/06-deploy-architecture.yml
+++ b/playbooks/06-deploy-architecture.yml
@@ -276,6 +276,11 @@
         -n openstack
         nova-cell0-conductor-0
         nova-manage cell_v2 discover_hosts --verbose
+        --strict
+      register: discover_result
+      until: discover_result.rc == 0
+      retries: 3
+      delay: 30
 
 - name: Run post_deploy hooks
   when:
--- a/roles/edpm_deploy/tasks/main.yml
+++ b/roles/edpm_deploy/tasks/main.yml
@@ -178,6 +178,11 @@
       oc rsh
       --namespace={{ cifmw_install_yamls_defaults['NAMESPACE'] }}
       nova-cell0-conductor-0 nova-manage cell_v2 discover_hosts --verbose
+      --strict
+    register: discover_result
+    until: discover_result.rc == 0
+    retries: 3
+    delay: 30
 
 - name: Validate EDPM
   when: cifmw_edpm_deploy_run_validation | bool
--- a/roles/edpm_deploy_baremetal/tasks/main.yml
+++ b/roles/edpm_deploy_baremetal/tasks/main.yml
@@ -297,3 +297,8 @@
       oc rsh
       -n {{ cifmw_install_yamls_defaults['NAMESPACE'] }}
       nova-cell0-conductor-0 nova-manage cell_v2 discover_hosts --verbose
+      --strict
+    register: discover_result
+    until: discover_result.rc == 0
+    retries: 3
+    delay: 30


lewisdenny:
Thank you for the patch @rdiazcam, I would love input from @gibizer about the use of the --strict argument.

Do we ever have a reason to deploy without compute nodes?

playbooks/06-deploy-architecture.yml
```diff
@@ -276,6 +276,11 @@
         -n openstack
         nova-cell0-conductor-0
         nova-manage cell_v2 discover_hosts --verbose
+        --strict
```
gibizer, line 279: From the doc:
> If --strict is specified, the command will only return 0 if an unmapped host was discovered and mapped successfully.

So this change means that the discover_hosts will be called until at least 1 host is discovered. This could help a bit with the race condition between starting the nova-compute service by the DataPlaneDeployment, letting the nova-compute start up and report the first status to the DB, and between the ci_framework running the discover_hosts command to discover those computes. 

However it will not be the final solution due to the following. If there are more than one compute in the deployment then discovering one of them is enough for this change to move forward. Which means maybe not all the compute will be discovered.

The final solution requires either of the two directions:
* a) do not let the DataPlaneDeployment finish until the nova-compute not just started but also successfully reported its first status to the DB. 
* b) retry discover_hosts until not just one but all the deployed compute nodes are discovered.

gibizer:
I suggest to pull this into the mainline of ci framework instead https://github.com/openstack-k8s-operators/ci-framework/pull/2459/files

lewisdenny:
@gibizer  My understanding was the content in the hook was only a workaround, a hook is the correct place for a workaround but if this is how you/nova team want the discover_hosts logic to work then I guess we should update our call to discover_hosts to use the retry logic as you suggest.

@rdiazcam  Would you like to update you PR to match the logic? My initial thoughts would be creating a tiny role that handles discover_hosts would be a nice route to go.

gibizer:
> @gibizer My understanding was the content in the hook was only a workaround, a hook is the correct place for a workaround but if this is how you/nova team want the discover_hosts logic to work then I guess we should update our call to discover_hosts to use the retry logic as you suggest.

I think the long term solution is to have a proper healthcheck endpoint implemented for nova-compute and let the container starting logic in edpm_ansible waiting for nova-compute to report a healthy state before returning (including the compute node creation). This way when the DataPlaneDeployment returns Ready=True, all the computes are in a discoverable state. However this still fairly fare from the product perspective as the work is ongoing upstream in https://blueprints.launchpad.net/nova/+spec/per-process-healthchecks (and probably won't be fully ready in Epoxy either).

So while the retry in https://github.com/openstack-k8s-operators/ci-framework/pull/2459 is a workaround, it is not something we will be able to remove soon. Keeping it as an optional hook makes the problem widespread as not all job has this workaround enabled and creates a load of triaging CI issues due to this race condition. I suggest add the logic of the workaround to the main codepath of the ci-framework, or at least make this workaround enabled by default in every job that does EDPM deployment.

> 
> @rdiazcam Would you like to update you PR to match the logic? My initial thoughts would be creating a tiny role that handles discover_hosts would be a nice route to go.

PR #2503, opened 2024-10-30
[ci_netconfig] Remove unused and outdated role

pablintino:
This role was part of the old effort to have a common strategy to deploy VA/DTs and install_yamls deployments but the approach drastically changed during its development and it has never been used nor updated to use the networking mapper.


lewisdenny (approved):
I don't see any mention up or downstream:
`rg --glob '!**/roles/ci_netconfig/**' --glob '!**/molecule.yaml' --glob '!**/zuul.d/**' ci_netconfig`

I also see no mention at all after this patch is applied!

/approve

PR #2504, opened 2024-10-30
[ci_metallb] Remove unused and outdated role

pablintino:
This role was part of the old effort to have a common strategy to deploy VA/DTs and install_yamls deployments but the approach drastically changed during its
development and it has never been used nor updated to use the networking mapper.


lewisdenny (approved):
/approve

No usage downstream and no mentions with this PR applied.

frenzyfriday:
/lgtm

quick search didnt show this being used anywhere

PR #2515, opened 2024-11-04
Skip watcher containers for downstream

raukadah:
Watcher related packages got added in RDO antelope but not available in OSP-18 downstream.

https://github.com/openstack-k8s-operators/tcib/pull/215 adds watcher container in tcib. Since downstream tcib jobs install the tcib rpm from main branch so it may break downstream container build and component job.

So in order to avoid breakage, we are skip building watcher related containers.

Note: It also fixes the exclude conditional to check existence of release key in exclude var to avoid following error:
    ```
    ''dict object'' has no attribute ''master'''
    ```

Depends-On: https://github.com/openstack-k8s-operators/tcib/pull/215

### DIFF
--- a/roles/build_containers/defaults/main.yml
+++ b/roles/build_containers/defaults/main.yml
@@ -55,3 +55,12 @@ cifmw_build_containers_install_from_source: false
 # cifmw_build_containers_dockerfile_labels
 # cifmw_build_containers_extra_config
 # cifmw_build_containers_build_timeout
+
+# Note(Chandankumar): Remove it once watcher related packages are available downstream.
+cifmw_build_containers_exclude_containers:
+  rhos18:
+    rhel9:
+      - watcher-base
+      - watcher-api
+      - watcher-applier
+      - watcher-decision-engine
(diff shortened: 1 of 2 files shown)


raukadah:
recheck https://softwarefactory-project.io/zuul/t/rdoproject.org/builds?job_name=cifmw-crc-podified-edpm-baremetal&project=openstack-k8s-operators/ci-framework seems to be almost green!

roles/build_containers/defaults/main.yml
```diff
@@ -55,3 +55,12 @@ cifmw_build_containers_install_from_source: false
 # cifmw_build_containers_dockerfile_labels
 # cifmw_build_containers_extra_config
 # cifmw_build_containers_build_timeout
+
+# Note(Chandankumar): Remove it once watcher related packages are available downstream.
+cifmw_build_containers_exclude_containers:
```
lewisdenny, line 60: Please add this var to the README with an example. 

Rather than defining this temporary list in the role defaults, can it be set in the job variables of the D/S job that calls the role.

lewisdenny (changes requested):
Please add this var to the README

roles/build_containers/defaults/main.yml
lewisdenny, line 60: cifmw_build_containers_exclude_containers is documented in the README as a list but in fact it's a nested dictionary.

raukadah: Thank you @lewisdenny for the review, I have proposed a mr downstream to achieve the same.

raukadah: Good finding, I will update this pull request to address this change and remove watcher skip from here. thank you!

PR #2516, opened 2024-11-05
[BGP] ipv4routes can be defined for netconfig

eduolivares:
In case of octavia, ipv4routes will be needed to connect to the management network.

### DIFF
--- a/roles/ci_gen_kustomize_values/templates/bgp_dt01/network-values/values.yaml.j2
+++ b/roles/ci_gen_kustomize_values/templates/bgp_dt01/network-values/values.yaml.j2
@@ -67,6 +67,13 @@ data:
         - end: {{ range.end }}
           start: {{ range.start }}
 {%    endfor %}
+{%    if network.tools.netconfig.ipv4_routes | default([]) | length > 0 %}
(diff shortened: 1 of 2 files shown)


roles/ci_gen_kustomize_values/templates/bgp_dt01/network-values/values.yaml.j2
```diff
@@ -67,6 +67,13 @@ data:
         - end: {{ range.end }}
           start: {{ range.start }}
 {%    endfor %}
+{%    if network.tools.netconfig.ipv4_routes | default(false) %}
```
pablintino, line 70: Netconfig has no routes options yet
https://github.com/openstack-k8s-operators/ci-framework/blob/e9e103760350dbc675b9a03d8952aa0054f7fb07/plugins/module_utils/net_map/networking_env_definitions.py#L163

eduolivares: @pablintino, thanks for your comment.
I have updated the PR, although I'm not sure if the solution is so simple.

PR #2518, opened 2024-11-06
Use multiple Ceph secrets in ci_dcn_site role

fultonj:
Without this patch the ci_dcn_site role creates
one secret with all Ceph keys and distributes
all of these keys to all DCN sites. This patch
makes the role create one secret for each DCN
site and only that secret is used at that site.
The secret for each DCN site only has the keys
that are required. This will make the testing
consistent with the documentation.

Additional secrets for Ceph configuration are
created and `extraMounts` for both the control
plane and each dataplane is extended to include
the extra secrets so that not all of the Ceph
keys are distributed.

Jira: https://issues.redhat.com/browse/OSPRH-6603


fultonj:
This has been verified in a downstream job

PR #2530, opened 2024-11-12
Add more Ceph OSDs for adoption context

fmount:
In adoption, we need to run scenario tempest tests when ceph is configured as a backend for manila/glance/cinder. This patch increases the number of OSDs to not fail because of not having enough space or something related.

Jira: https://issues.redhat.com/browse/OSPRH-11223


katarimanojk:
Looks good to me.

cescgina:
tested the PR downstream, deployment was successful with 3 osd per host

PR #2532, opened 2024-11-12
Add task for security tlse

millevy:


roles/validations/tasks/security/invoke_playbooks.yml
```diff
@@ -0,0 +1,10 @@
+- name: Invoke the automation playbook (dataplane certs) on controller-0
+               environment:
```
hjensas, line 2: indentation is wrong. `environment` and `cifmw.general.ci_script` must be at same indentation as `name`.

galaxy.yml
```diff
@@ -71,7 +71,7 @@ tags:
 # L(specifiers,https://python-semanticversion.readthedocs.io/en/latest/#requirement-specification). Multiple version
 # range specifiers can be set and are separated by ','
 dependencies:
-  'git+https://github.com/containers/ansible-podman-collections': 'master'
+  'git+https://github.com/containers/ansible-podman-collections': 'main'
```
hjensas, line 74: I think if you need to re-sync the fork and rebase on main. This change already merged.

hjensas (changes requested):

millevy:
lgtm

lewisdenny (changes requested):
Hi @millevy 

Have you tested this in a job downstream with it working?

If you have a jira card for this work, please add it to the commit message. 

@bshephar qq, does the validations role support injecting validations dynamically? This use case is pretty interesting where the DFG is holding the logic in there own repo. 

In the D/S job, they will add their repo for zuul to check it out on the controller but it would be nice to be able to define a var that the validations role can read and execute resulting in D/S only validations only being D/S.

millevy:
@lewisdenny It does work , see my testproject patch. I am dealing with a failure from the deployment side now, but  I did come at some point ,where the tests ran in my deployment.
https://gitlab.cee.redhat.com/ci-framework/ci-framework-testproject/-/merge_requests/863

roles/validations/tasks/security/invoke_tlse_playbooks.yml
```diff
@@ -0,0 +1,11 @@
+- name: Invoke the automation playbook (dataplane certs) on controller-0
+  environment:
+    PATH: "{{ cifmw_path }}"
+  cifmw.general.ci_script:
+    executable: "/bin/bash"
+    output_dir: "{{ cifmw_basedir }}/artifacts"
+    script: |
+      set -ex -o pipefail
+      cd "{{ ansible_user_dir }}/src/gitlab.cee.redhat.com/OSP-DFG-security/automation"
```
pablintino, line 9: We should avoid leaking internal IT details like out gitlab hostname.

pablintino (changes requested):

roles/validations/tasks/security/invoke_tlse_playbooks.yml
hjensas: This one is leaked already 

In the `openstack-k8s-operators` org:
https://github.com/search?q=org%3Aopenstack-k8s-operators%20gitlab.cee.redhat.com&type=code

In `openshift` org:
https://github.com/search?q=org%3Aopenshift%20gitlab.cee.redhat.com&type=code

pablintino: "nice" :-1:

roles/validations/tasks/security/invoke_tlse_playbooks.yml
```diff
@@ -0,0 +1,11 @@
...
+      ansible-playbook -vv playbooks/renew_internal_cert_outer.yml || echo "renew_internal_cert_outer failed, continuing..."
+      ansible-playbook -vv playbooks/data_plane_cert_testing_with_delete.yml
```
pablintino, line 11: If this repo was a collection it would be much simpler to integrate

pablintino:
> Hi @millevy
> 
> Have you tested this in a job downstream with it working?
> 
> If you have a jira card for this work, please add it to the commit message.
> 
> @bshephar qq, does the validations role support injecting validations dynamically? This use case is pretty interesting where the DFG is holding the logic in there own repo.
> 
> In the D/S job, they will add their repo for zuul to check it out on the controller but it would be nice to be able to define a var that the validations role can read and execute resulting in D/S only validations only being D/S.

@lewisdenny @bshephar Hi, based on the touches I'm doing to that role for the XML reporting I guess the answer is no, but, it should be relatively easy to do. The only thing I guess it would be tricky is to catch the error message.
No matter, what I think it would be ideal is to directly import the playbook, but that approach has other caveats:

- The called playbooks should follow the same host patterns we use
- I have the feeling that is not that easy to catch an error in an ansible import_playbook task without breaking the entire execution.

PR #2533, opened 2024-11-12
Reduce compute resources in osasinfra DT

eurijon:
Smaller RAM and disk allocation for compute nodes in osasinfra DT is required for running in bare metal with 250 GB of RAM and 890 GB of disk.

### DIFF
--- a/scenarios/reproducers/dt-osasinfra.yml
+++ b/scenarios/reproducers/dt-osasinfra.yml
@@ -45,8 +45,8 @@ cifmw_networking_mapper_definition_patches_01:
 
 # HCI requires bigger size to hold OCP on OSP disks
 cifmw_block_device_size: 100G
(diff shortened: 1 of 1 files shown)


lewisdenny (approved):
/lgtm

frenzyfriday:
Hey @eurijon when you are ready for this PR to merge you can remove the draft flag and it should auto merge after that

eurijon:
Hey folks, thanks for reviewing. I want to keep this PR on hold as it allows to test osasinfra DT deployments on servers with smaller compute resources, but we don't want to change the default values for now. Is it ok to keep this as draft or should I add a label so there is no need to review and merge?

eurijon:
Closing as it was for testing purposes and those params can be overridden from the job definition.

PR #2537, opened 2024-11-13
[DCN] Disable RGW in the DCN scenario

sbekkerm:
The RGW should be disabled in DCN scenario


fultonj:
We never want to deploy RGW on edge nodes. We could deploy RGW on central nodes in az0, but that won't work in our CI with the way ceph is called from dcn.yml playbook. This is a good workaround for now.

I'm +2 on this.

lewisdenny (approved):
/lgtm

PR #2538, opened 2024-11-13

drosenfe:
Create an xml file of validations results. This is necessary for polarion posting. Note: this is finishing PR https://github.com/openstack-k8s-operators/ci-framework/pull/2456 which I messed up. Thought it would be easier to just create a new PR.


lewisdenny:
Hi @drosenfe @pablintino, I see there is an alturnitive patch submitted for this: https://github.com/openstack-k8s-operators/ci-framework/pull/2560

Which route will we go?

drosenfe:
@lewisdenny  @pablintino   Was on PTO last week. I'm fine with using https://github.com/openstack-k8s-operators/ci-framework/pull/2560  I've been testing that one.  Last comment on slack was that in the xml file being generated the failure reason is not quoted like the other fields and doesn't start with message=   Do we want to change that?

pablintino:
Superseded by https://github.com/openstack-k8s-operators/ci-framework/pull/2560

PR #2541, opened 2024-11-15
Add nodeset2 for scenarios using 2 nodesets

mnietoji:
In NFV, we have some compute nodes belonging to the same deployment that are different and we need 2 nodesets to be able to deploy

### DIFF
--- a/scenarios/reproducers/dt-nfv-ovs-dpdk-sriov-2nodesets.yml
+++ b/scenarios/reproducers/dt-nfv-ovs-dpdk-sriov-2nodesets.yml
@@ -1,0 +1,117 @@
+---
+cifmw_architecture_scenario: "ovs-dpdk-sriov-2nodesets"
+
+# Automation section. Most of those parameters will be passed to the
+# controller-0 as-is and be consumed by the `deploy-va.sh` script.
+# Please note, all paths are on the controller-0, meaning managed by the
+# Framework. Please do not edit them!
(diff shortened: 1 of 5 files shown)


pablintino:
recheck

roles/ci_gen_kustomize_values/templates/common/edpm-nodeset2-values/values.yaml.j2
pablintino, line 1: As you have a template for your own VA this file is not used. I wouldn't push it.,

scenarios/reproducers/dt-nfv-ovs-dpdk-sriov-2nodesets.yml
```diff
@@ -0,0 +1,117 @@
+---
+cifmw_architecture_scenario: "ovs-dpdk-sriov-2nodesets"
+
+# Automation section. Most of those parameters will be passed to the
+# controller-0 as-is and be consumed by the `deploy-va.sh` script.
+# Please note, all paths are on the controller-0, meaning managed by the
+# Framework. Please do not edit them!
+_arch_repo: "/home/zuul/src/github.com/openstack-k8s-operators/architecture"
```
pablintino, line 8: Check other similar files, we allow the user to override this by using:
```yaml
_arch_repo: "{{ cifmw_architecture_repo | default('/home/zuul/src/github.com/openstack-k8s-operators/architecture') }}"
```

mnietoji:
recheck

PR #2543, opened 2024-11-15
Check if openstack user exists before trying to create it

fmount:
Jira: https://issues.redhat.com/browse/OSPRH-11669

### DIFF
--- a/roles/cifmw_cephadm/tasks/post.yml
+++ b/roles/cifmw_cephadm/tasks/post.yml
@@ -63,7 +63,16 @@
       vars:
         ceph_command: "radosgw-admin"
 
+    - name: Check if Ceph user exists
+      ignore_errors: true
+      become: true
+      ansible.builtin.command:
+        cmd: >-
+          {{ cifmw_cephadm_ceph_cli }} user info --uid openstack
+      register: cifmw_ceph_s3_user
+
     - name: Ceph rgw - create access/secret key
+      when: cifmw_ceph_s3_user.rc > 0
       become: true
       ansible.builtin.command:
         cmd: >-


roles/cifmw_cephadm/tasks/post.yml
```diff
@@ -63,7 +63,15 @@
       vars:
         ceph_command: "radosgw-admin"
 
+    - name: Check if Ceph user exists
+      become: true
```
pablintino, line 67: Is become needed?

fmount: yeah, behind the scenes the ceph client needs root access, but I didn't add it at block level because we don't need it in all tasks

roles/cifmw_cephadm/tasks/post.yml
```diff
@@ -63,7 +63,15 @@
...
+      ansible.builtin.command:
+        cmd: >-
+          {{ cifmw_cephadm_ceph_cli }} user info --uid openstack
+      register: cifmw_ceph_s3_user
+
     - name: Ceph rgw - create access/secret key
+      when: "no user info saved in {{ cifmw_ceph_s3_user.stdout }}"
```
fultonj, line 74: It will be in `stderr` not `stdout`  and we also need an `ignore_errors: true` or similar so that if the new task fails (and it will fail the first time) then it does not stop the play with the error below.

```
    TASK [cifmw_cephadm : Check if Ceph user exists _raw_params={{ cifmw_cephadm_ceph_cli }} user info --uid openstack] ***
    Friday 15 November 2024  12:30:33 -0500 (0:00:00.080)       0:05:04.139 *******
    fatal: [compute-vqt0nogt-0]: FAILED! => changed=true
      cmd:
      - podman
      - run
      - --rm
      - --net=host
      - --ipc=host
      - --volume
      - /etc/ceph:/etc/ceph:z
      - --volume
      - /home/ceph-admin/assimilate_ceph.conf:/home/assimilate_ceph.conf:z
      - --entrypoint
      - radosgw-admin
      - registry-proxy.engineering.redhat.com/rh-osbs/rhceph:7-453
      - --fsid
      - 1a1b33b2-e12d-5489-9646-51bfade45372
      - -c
      - /etc/ceph/ceph.conf
      - -k
      - /etc/ceph/ceph.client.admin.keyring
      - user
      - info
      - --uid
      - openstack
      delta: '0:00:09.977484'
      end: '2024-11-15 17:30:43.756108'
      msg: non-zero return code
      rc: 22
      start: '2024-11-15 17:30:33.778624'
      stderr: 'could not fetch user info: no user info saved'
      stderr_lines: <omitted>
      stdout: ''
      stdout_lines: <omitted>
```

fmount: ok, I think we should probably switch to `ansible.builtin.shell` instead of using `ansible.builtin.command`

frenzyfriday:
recheck

content provider failed

fultonj (approved):
/approve
/lgtm

PR #2545, opened 2024-11-15
Update uni05 nodesets to allow for new cell

jamepark4:
Update templates for uni05 epsilon to create two nodeset templates that map to specific cells (cell1 and cell2 respectively).


sdatko (approved):
/lgtm

PR #2546, opened 2024-11-18
[kustomize_deploy] Include symlinks when user kustomize dir is set

eduolivares:
With this patch, not only regular files but also symlinks are taken into account.

### DIFF
--- a/roles/kustomize_deploy/tasks/generate_base64_patches_from_tree.yml
+++ b/roles/kustomize_deploy/tasks/generate_base64_patches_from_tree.yml
@@ -8,14 +8,22 @@
         paths: "{{ cifmw_architecture_user_kustomize_base_dir }}"
         patterns: "*.yml,*.yaml"
         recurse: true
+        file_type: "{{ item }}"
       register: _cifmw_kustomize_deploy_yaml_patches_find
+      loop:
+        - file
+        - link
       delegate_to: localhost
 
     - name: Fetch the content of the kustomize patches
       ansible.builtin.slurp:
         path: "{{ item.path }}"
       register: _cifmw_kustomize_deploy_yaml_patches_slurp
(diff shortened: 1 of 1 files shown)


roles/kustomize_deploy/tasks/generate_base64_patches_from_tree.yml
```diff
@@ -8,14 +8,27 @@
...
+            _cifmw_kustomize_deploy_yaml_patches_find | default([])
+          }}
+      loop: "{{ _kustomize_deploy_yaml_files.results | default([]) }}"
+
     - name: Fetch the content of the kustomize patches
       ansible.builtin.slurp:
         path: "{{ item.path }}"
       register: _cifmw_kustomize_deploy_yaml_patches_slurp
-      loop: "{{ _cifmw_kustomize_deploy_yaml_patches_find.files | default([]) }}"
+      loop: "{{ _cifmw_kustomize_deploy_yaml_patches_find | default([]) }}"
```
pablintino, line 31: I think you can get rid of the new task, the one before this one and make this one like this:
```yaml
    - name: Fetch the content of the kustomize patches
      ansible.builtin.debug:
        msg: "{{ item.path }}"
      register: _cifmw_kustomize_deploy_yaml_patches_slurp
      loop: >-
        {{
          _kustomize_deploy_yaml_files.results |
          map(attribute='files') | flatten
        }}

```

eduolivares: I just tested it and it looks much better. Thanks, @pablintino

lewisdenny (approved):
/approve

The default file_type for ansible.builtin.find is 'file'

This patch loops over both file and link file types, the loop returns two .results, each containing a `filles` list.

`Fetch the content of the kustomize patches` task guards against there being no files or links returned by looping over the returned files list and defaulting to an empty list. 

It then filters the results pulling out the files attribute returns a flat list of files found.

PR #2549, opened 2024-11-18
Support new OpenStack initialization resource for RHOSO install

abays:
We have a new upcoming way to install and deploy the OSP operators. OLM will only install one CRD for OSP, which is an initialization resource. A CR must be created for that CRD, which then, when created, causes all the remaining OSP CRDs to be installed and also deploys their respective operators.

### DIFF
--- a/roles/edpm_prepare/tasks/main.yml
+++ b/roles/edpm_prepare/tasks/main.yml
@@ -119,6 +119,34 @@
           --namespace={{ cifmw_install_yamls_defaults['OPERATOR_NAMESPACE'] }}
           --for=jsonpath='{.status.phase}'=Complete --timeout=20m
 
+- name: Check if the OpenStack initialization CRD exists
+  when:
+    - not cifmw_edpm_prepare_dry_run
+  kubernetes.core.k8s_info:
+    kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
+    api_key: "{{ cifmw_openshift_token | default(omit) }}"
+    context: "{{ cifmw_openshift_context | default(omit) }}"
+    kind: CustomResourceDefinition
+    name: openstacks.operator.openstack.org
+  register: _cifmw_kustomize_deploy_olm_osp_operator_openstack_crd_out
+  tags:
+    - control-plane
+
+- name: OpenStack meta-operator initialization, if necessary
+  when:
+    - not cifmw_edpm_prepare_dry_run
+    - _cifmw_kustomize_deploy_olm_osp_operator_openstack_crd_out is defined
+    - _cifmw_kustomize_deploy_olm_osp_operator_openstack_crd_out.resources is defined
+    - (_cifmw_kustomize_deploy_olm_osp_operator_openstack_crd_out.resources | length) > 0
+  vars:
+    make_openstack_init_env: "{{ cifmw_edpm_prepare_common_env }}"
+    make_openstack_init_dryrun: "{{ cifmw_edpm_prepare_dry_run }}"
+  ansible.builtin.include_role:
+    name: 'install_yamls_makes'
+    tasks_from: 'make_openstack_init'
+  tags:
+    - control-plane
+
   # Note(chkumar): Keeping set_openstack_containers role
   # till we migrate this task to update_containers role
 - name: Update OpenStack Services containers Env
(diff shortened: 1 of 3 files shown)


abays:
I have tested this with VA1 and a custom build of the new OpenStack operator that uses the new paradigm.  The deployment was succesful, even all the way through Tempest.

lewisdenny (commented):
Hi @abays 

I assume we will need to support the new and old method of installing RHOSO? 

If so, could you please explain how that will be handled as we don't really version ci-framework.

abays:
> Hi @abays
> 
> I assume we will need to support the new and old method of installing RHOSO?
> 
> If so, could you please explain how that will be handled as we don't really version ci-framework.

That's a good point.  CIFMW will still need to deploy older releases.  So I will need to add some sort of conditional logic that detects whether the new `Openstack` CRD is available and then branches from there.

abays:
I have tested this a variety of ways with VA1 in my environment:

1. Old OpenStack operator index with new `architecture` PR
2. Old OpenStack operator index with old `architecture` repo (current `main`, the default)
3. New OpenStack operator index with new `architecture` PR
4. New OpenStack operator index with old `architecture` repo (current `main`, the default)

1-3 succeeded as expected.  #4 failed as expected, because the new index requires the creation of an `OpenStack` initialization CR, and the VA will lack this without the new `architecture` PR.

tl;dr this _should_ be safe to merge now based on my test results

lewisdenny:
/lgtm

Thank you @abays <3

abays:
I also have changes for the "lightweight" 3-node CRC deployment, but I cannot get CRC to work in that approach (though it has nothing to do with my changes).  CRC VM just boots into an empty console and thus fails CIFMW validation checks.

morucci:
recheck

dprince:
code here looks good to me. Thanks abays

roles/edpm_prepare/tasks/main.yml
```diff
@@ -119,6 +119,35 @@
...
+    - control-plane
+
+- name: OpenStack meta-operator initialization, if necessary
+  when:
+    - not cifmw_edpm_prepare_dry_run
+    - _cifmw_kustomize_deploy_olm_osp_operator_openstack_crd_out is defined
+    - _cifmw_kustomize_deploy_olm_osp_operator_openstack_crd_out.resources is defined
+    - (_cifmw_kustomize_deploy_olm_osp_operator_openstack_crd_out.resources | length) > 0
+  vars:
+    make_openstack_env: "{{ cifmw_edpm_prepare_common_env |
```
pablintino, line 142: Typo: `make_openstack_env` should be `make_openstack_init_env`

abays:
The latest failures are indeed caused by my changes, but I am confused how they are happening.  The issue is that `make openstack_init` and `make openstack_deploy_prep` both now call `make openstack_repo`.  `make openstack_repo` fails if it was called before, because the existing Git clone will be on the local filesystem and it errors out while complaining about the dir not being empty.  But in the current job, `make openstack_init` should _not_ be being called, because we check if the new `OpenStack` initialization CRD exists and only actually call `make openstack_init` if it does [1].  Yet, somehow the logs show that CIFMW is running `openstack_init`:

```
TASK [install_yamls_makes : Run openstack_init output_dir={{ cifmw_basedir|default(ansible_user_dir ~ '/ci-framework-data') }}/artifacts, chdir=/home/zuul/src/github.com/openstack-k8s-operators/install_yamls, script=make openstack_init, dry_run={{ make_openstack_init_dryrun|default(false)|bool }}, extra_args={{ dict((make_openstack_init_env|default({})), **(make_openstack_init_params|default({}))) }}] ***
Wednesday 18 December 2024  14:28:52 -0500 (0:00:00.025)       0:08:29.117 **** 
[WARNING]: conditional statements should not include jinja2 templating
delimiters such as {{ }} or {% %}. Found: {{ make_openstack_init_until |
default(true) }}
Follow script's output here: /home/zuul/ci-framework-data/logs/ci_script_005_run_openstack.log
changed: [localhost]
```

Maybe there's something wrong with [1], but it succeeded earlier this week and I haven't changed it other than to rename [2]...but maybe that did it?

[1] https://github.com/abays/ci-framework/blob/8e613342016c94a6bc90df0c3c5fa238d6e26367/roles/edpm_prepare/tasks/main.yml#L122-L148
[2] https://github.com/abays/ci-framework/blob/8e613342016c94a6bc90df0c3c5fa238d6e26367/roles/edpm_prepare/tasks/main.yml#L142-L143

abays:
recheck

abays:
recheck

frenzyfriday (approved):

PR #2553, opened 2024-11-19
[polarion] Allow users pass extra custom-fields

kopecmartin:
Jump has --custom-fields argument that is set here in the polarion role. This commit allows users pass extra custom fields by introducing cifmw_polarion_jump_extra_custom_fields variable.

### DIFF
--- a/roles/polarion/tasks/main.yml
+++ b/roles/polarion/tasks/main.yml
@@ -110,6 +110,18 @@
       register: merged_xml_files
 
     - name: Update Test Runs
+      vars:
+        _opts:
+          cijoburl: "{{ cifmw_zuul_url }}/{{ zuul.build }}"
+          jobteststage: "{{ merged_xml_files.files[ loop_idx ].path | dirname | basename }}"
+        _custom_fields: "{{ _opts | combine(cifmw_polarion_jump_custom_fields) }}"
+        _custom_fields_string: >-
+          {{
+            _custom_fields.keys() |
+            zip(_custom_fields.values()) |
+            map('join', '=') |
+            join(',')
+          }}
       ansible.builtin.shell:
         chdir: "{{ cifmw_polarion_jump_repo_dir }}"
         cmd: >-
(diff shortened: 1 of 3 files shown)


roles/polarion/tasks/main.yml
```diff
@@ -119,7 +119,7 @@
           --testrun-title={{ cifmw_polarion_testrun_title }}-{{ loop_idx }}
           --xml-file={{ item.path }}
           --update_testcases={{ cifmw_polarion_update_testcases | default(false) }}
-          --custom-fields cijoburl={{ cifmw_zuul_url }}/{{ zuul.build }},jobteststage={{ merged_xml_files.files[ loop_idx ].path | dirname | basename }}
+          --custom-fields {{ cifmw_polarion_jump_extra_custom_fields | default('') }}cijoburl={{ cifmw_zuul_url }}/{{ zuul.build }},jobteststage={{ merged_xml_files.files[ loop_idx ].path | dirname | basename }}
```
pablintino, line 122: This is not the way we usually do this kind of "defaulting". To try to be consistent with the rest of the project I suggest to add `cifmw_polarion_jump_extra_custom_fields` to the default vars of the role. Document it in the readme.

pablintino: BTW it sounds easier to consume if instead of plain string we expose a dict. If so, the task may look like this:

```yaml
    - name: Update Test Runs
      vars:
        _opts:
          cijoburl: "{{ cifmw_zuul_url }}/{{ zuul.build }}"
          jobteststage: "{{ merged_xml_files.files[ loop_idx ].path | dirname | basename }}"
        _custom_fields: "{{ _opts | combine(cifmw_polarion_jump_extra_custom_fields) }}"
        _custom_fields_string: >-
          {{
            _custom_fields.keys() |
            zip(_custom_fields.values()) |
            map('join', '=') |
            join(',')
          }}
      ansible.builtin.shell:
        chdir: "{{ cifmw_polarion_jump_repo_dir }}"
        cmd: >-
          source "{{ cifmw_polarion_jump_repo_dir }}/jump-venv/bin/activate" &&
          {{ cifmw_polarion_jump_repo_dir }}/jump-venv/bin/python jump.py
          --testrun-id={{ cifmw_polarion_testrun_id }}
          --testrun-title={{ cifmw_polarion_testrun_title }}-{{ loop_idx }}
          --xml-file={{ item.path }}
          --custom-fields {{ _custom_fields_string }}
          {{ cifmw_polarion_jump_extra_vars | default ('') }}
      loop: "{{ merged_xml_files.files }}"
      loop_control:
        index_var: loop_idx
      register: jump_script

```

kopecmartin: thank you for the review, that looks definitely more elegant and professional!

PR #2556, opened 2024-11-19
kuttl: fix repo cloning, fetch the locally built operators

tosky:
When running kuttl tests using install_yamls, operators code are cloned before running the job.
All credits to Douglas Viroel who investigated and fixed a similar issue downstream.

Also, make sure that locally built operator (for example, thanks to the openstack-k8s-operators-content-provider job, or anything else which populates cifmw_operator_build_output) are used.
This is a bit more complex than what EDPM jobs do (see the edpm_prepare role), where only the openstack-operator-index is overridden, likely because here there may be more operators being built.

### DIFF
--- a/ci/playbooks/kuttl/run-kuttl-tests.yml
+++ b/ci/playbooks/kuttl/run-kuttl-tests.yml
@@ -3,11 +3,23 @@
   ansible.builtin.include_vars:
     dir: "{{ cifmw_basedir }}/artifacts/parameters"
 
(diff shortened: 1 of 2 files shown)


tosky:
recheck

tosky:
(and please don't squash when merging)

tosky:
For the record, this is being tested in https://github.com/openstack-k8s-operators/manila-operator/pull/339

fmount (approved):
/lgtm

PR #2563, opened 2024-11-21
Add hook to enable horizon

ashu-011:
This PR adds hooks to enable horizon service which is not enabled by default in a crc deploy . This is needed as we are trying to extend test-operator zuul gating job to include horizon testing .

### DIFF
--- a/hooks/playbooks/control_plane_horizon.yml
+++ b/hooks/playbooks/control_plane_horizon.yml
@@ -1,0 +1,28 @@
+---
+- name: Kustomize ControlPlane for horizon service
+  hosts: "{{ cifmw_target_hook_host | default('localhost') }}"
+  gather_facts: false
+  tasks:
+    - name: Ensure the kustomizations dir exists
+      ansible.builtin.file:
+        path: "{{ cifmw_basedir }}/artifacts/manifests/kustomizations/controlplane"
+        state: directory
+        mode: "0755"
+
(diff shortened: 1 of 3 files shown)


scenarios/centos-9/horizon.yml
```diff
@@ -0,0 +1,5 @@
+---
+pre_deploy:
+  - name: 80 80 80 80 80 80 80 80 Kustomize OpenStack CR
```
hjensas, line 3: nit: Why the repeated `80` here ?

hooks/playbooks/control_plane_horizon.yml
```diff
@@ -0,0 +1,28 @@
...
+  tasks:
+    - name: Ensure the kustomizations dir exists
+      ansible.builtin.file:
+        path: "{{ cifmw_basedir }}/artifacts/manifests/kustomizations/controlplane"
+        state: directory
+        mode: "0755"
+
+    - name: Create kustomize yaml to enable Horizon
+      ansible.builtin.copy:
+        dest: "{{ cifmw_basedir }}/artifacts/manifests/kustomizations/controlplane/66-horizon-kustomization.yaml"
```
pablintino, line 14: Why is the hook named 80 but the file is 60? I'd name it the same for consistency.

ashu-011: Done , thanks

PR #2566, opened 2024-11-21
Kuttl multinode: bump the timeout

tosky:
It seems the test is timing out more often then not. Let's try to see if the execution time is the only issue.


tosky:
recheck

tosky:
[cifmw-multinode-kuttl](https://softwarefactory-project.io/zuul/t/rdoproject.org/build/7787e525414a45b799e502e44f2709bd) took 2h25m and it passed. Maybe we need to bump from 9000 to 9600 just to account for some variations. Rechecking to check the time needed.

PR #2570, opened 2024-11-25
Enable sushy-emulator in create-infra.yml

hjensas:
For adoption of the ironic service we need a Virtual BMC for spinning up test workloads on the OSP 17.x cloud, and to validate workload operations after adoption.

This change ensure the controller is started at the end of create-infra, and does the neccecary steps to apply and validate the sushy-emulator running as a POD on the controller node.

I have switched the use of `ansible.builtin.package` to `ansible.builtin.dnf` in some places to avoid issues with hostvars lookups raising undefined variable errors.

### DIFF
--- a/create-infra.yml
+++ b/create-infra.yml
@@ -75,3 +75,66 @@
       ansible.builtin.import_role:
         name: "libvirt_manager"
         tasks_from: "deploy_layout.yml"
+
+    - name: Gather ansible_user_id
+      ansible.builtin.setup:
+        gather_subset:
+          - user_id
+          - user_dir
+
+    - name: Ensure controllers vms are started and reachable
+      when: groups.controllers | default([]) | length > 0
+      vars:
+        _cifmw_libvirt_manager_layout:
+          vms:
+            controllers:
+              start: true
+              disk_file_name: "dummy"
+        cifmw_libvirt_manager_all_vms: >-
+          {%- set vms = {} -%}
+          {%- for vm in groups.controllers | default([]) -%}
+            {%- set _ = vms.update({vm: 'controllers'}) -%}
+          {%- endfor -%}
+          {{ vms }}
+        ansible_libvirt_pools: {}
+      ansible.builtin.include_role:
+        name: "libvirt_manager"
+        tasks_from: "start_vms.yml"
+        apply:
+          delegate_to: "{{ cifmw_target_host | default('localhost') }}"
+
+    # This bootstraps the controller-0 node, and RedFish virtual BMC is
+    # spawned if cifmw_use_sushy_emulator is enabled.
+    - name: Bootstrap sushy-emulator (RedFish Virtual BMC) on controller-0
+      when:
+        - "'controllers' in groups"
+        - "'controllers' in group_names"
+        - cifmw_use_sushy_emulator | default(false) | bool
+      delegate_to: controller-0
+      become: true
+      vars:
+        cifmw_sushy_emulator_hypervisor_target: "{{ cifmw_target_host }}"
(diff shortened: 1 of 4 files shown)


hjensas:
recheck

create-infra.yml
```diff
@@ -75,3 +75,100 @@
...
+        name: "libvirt_manager"
+        tasks_from: "start_vms.yml"
+        apply:
+          delegate_to: "{{ cifmw_target_host | default('localhost') }}"
+
+    # This bootstraps the controller-0 node, and RedFish virtual BMC is
+    # spawned if cifmw_use_sushy_emulator is enabled.
+    - name: Gather controller-0 facts
+      when: cifmw_use_sushy_emulator | default(true) | bool
+      delegate_to: controller-0
```
pablintino, line 139: So, with this, we are enforcing users of this playbook to have a controller-0 instance. Can we make it conditional? Like, if I need sushy -> Enforce there is controller-0

create-infra.yml
```diff
@@ -75,3 +75,100 @@
...
+        _group_nodes: >-
+          {%- set group = {} -%}
+          {%- set _ = group.update({_group_name: _nodes}) -%}
+          {{ group }}
+      ansible.builtin.set_fact:
+        _vm_groups: >-
+          {{ _vm_groups | default({}) | combine(_group_nodes) }}
+      loop: "{{ _reproducer_inventory_files.results }}"
+      loop_control:
+        loop_var: "_inventory_file"
```
pablintino, line 113: Parsing an Ansible inventory directly, is never a good idea, there are more than one way of adding an instance to a group :) I'd suggest to use https://docs.ansible.com/ansible/latest/reference_appendices/special_variables.html#term-groups. That is the one we use for the networking mapper.

create-infra.yml
hjensas: I'm not sure I got it right, but I added conditions on:
```
        - "'controllers' in groups"
        - "'controllers' in group_names"
```
I also change the `cifmw_use_sushy_emulator` default to `false`.

Let me know, will these changes allow it to work without controller-0?

create-infra.yml
hjensas: yeah, I copied this from here https://github.com/openstack-k8s-operators/ci-framework/blob/b794aa220f6290399c6314ab7a85a11bf2f9caed/deploy-osp-adoption.yml#L46 since I saw the `_vm_groups` fact was used when starting vms here: https://github.com/openstack-k8s-operators/ci-framework/blob/b794aa220f6290399c6314ab7a85a11bf2f9caed/roles/adoption_osp_deploy/tasks/main.yml#L35

I changed it to drop the entried read of reproducer inventory files.
*lmknwyt*

hjensas:
recheck

pablintino (commented):
Just a simple comment. It looks good to me already.

create-infra.yml
```diff
@@ -75,3 +75,75 @@
...
+    # This bootstraps the controller-0 node, and RedFish virtual BMC is
+    # spawned if cifmw_use_sushy_emulator is enabled.
+    - name: Gather controllers facts
+      when:
+        - "'controllers' in groups"
+        - "'controllers' in group_names"
+      ansible.builtin.setup:
+        gather_subset:
+          - '!all'
+          - min
```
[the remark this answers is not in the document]
hjensas: It also works if I remove that setup task completely, so I did that.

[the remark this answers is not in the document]
pablintino: @hjensas nice, thanks for trying to cleanup.

PR #2571, opened 2024-11-26
Add job to run kuttl without install_yamls

cescgina:
This change attempts to make the approach that nova-operator uses to run
kuttl more generic, so it can be used in other operators that follow
a similar patter in their tests, like watcher-operator.

Related: OSPRH-11424


zuul.d/kuttl_multinode.yaml
```diff
@@ -71,3 +71,15 @@
       commands_after_kuttl_run:
         - oc get pv
         - oc get all
+
+- job:
+    name: cifmw-multinode-kuttl-operator-target
+    parent: cifmw-base-multinode-kuttl
```
SeanMooney, line 77: there isnte really a need to make it multi node.

in the nova one i override the noeset to be single node 

https://github.com/openstack-k8s-operators/nova-operator/blob/main/.zuul.yaml#L41
the parent is a multi node job but there is no benefit to multi node here

i would suggest adding 

 nodeset: centos-9-medium-crc-extracted-2-39-0-3xl 

and updating the name.

cescgina:
tested the job definition by changing the nova-operator kuttl job to use it in https://github.com/openstack-k8s-operators/nova-operator/pull/902 , job is passing https://github.com/openstack-k8s-operators/nova-operator/pull/902#issuecomment-2506181957

cescgina:
also got a green run from the watcher-operator version https://softwarefactory-project.io/zuul/t/rdoproject.org/build/6ed0fe8ce7ed433ca20ec7bbd71e9095 via https://github.com/openstack-k8s-operators/watcher-operator/pull/14

ci/playbooks/kuttl/deploy-deps.yaml
```diff
@@ -0,0 +1,164 @@
...
+          }}
+        cifmw_edpm_prepare_operators_build_output: "{{ operators_build_output }}"
+        cifmw_edpm_prepare_make_openstack_env: "{{ _local_operators_indexes | combine(_openstack_operator_images) }}"
+
+    - name: Deploy openstack Operators
+      block:
+        - name: detect if openstack operator is installed
+          environment:
+            KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+            PATH: "{{ cifmw_path | default(ansible_env.PATH) }}"
```
pablintino, line 105: Why not passing this environment vars as part of the block to avoid repeating the same thing 4 times?

ci/playbooks/kuttl/deploy-deps.yaml
```diff
@@ -0,0 +1,164 @@
...
+      loop:
+        - "{{ cifmw_basedir }}/artifacts/parameters"
+        - "/etc/ci/env"
+      loop_control:
+        label: "{{ item }}"
+
+    - name: Check we have some compute in inventory
+      ansible.builtin.set_fact:
+        has_compute: >-
+          {% set ns = namespace(found=false) -%}
```
pablintino, line 38: No need to use a namespace, a regular var is enough. Moreover, this can be done without jinja2 scripts with plain ansible-jinja2 filters:
```yaml
hostvars.keys() | select('search', '^compute.*') | length > 0

```

ci/playbooks/kuttl/deploy-deps.yaml
```diff
@@ -0,0 +1,164 @@
...
+
+- name: Build dataset hook
+  hosts: "{{ cifmw_target_host | default('localhost') }}"
+  gather_facts: false
+  connection: local
+  tasks:
+    - name: Assert that operator_name is set
+      ansible.builtin.assert:
+        that:
+          - operator_name is defined
```
pablintino, line 24: If we move this to line 8 we can safe some time/resources/bandwith if the assertion is not met.

ci/playbooks/kuttl/deploy-deps.yaml
```diff
@@ -0,0 +1,164 @@
+---
+- name: Run ci_framework bootstrap playbook
+  ansible.builtin.import_playbook: "../../../playbooks/01-bootstrap.yml"
+
+- hosts: "{{ cifmw_target_host | default('localhost') }}"
+  name: install dev tools
```
pablintino, line 6: An un-capitalized playbook name looks strange.

ci/playbooks/kuttl/deploy-deps.yaml
```diff
@@ -0,0 +1,164 @@
...
+          - crc_ci_bootstrap_networks_out is defined
+          - "'crc' in crc_ci_bootstrap_networks_out"
+          - "'default' in crc_ci_bootstrap_networks_out['crc']"
+
+    - name: Ensure we have needed bits for compute when needed
+      when:
+        - has_compute | bool
+      ansible.builtin.assert:
+        that:
+          - "'compute' in crc_ci_bootstrap_networks_out"
```
pablintino, line 58: Why enforcing the instance name? I mean, you already used a different criteria to check if a compute is available, based on a regex. I'd vote to stick to a fixed name or pick the first one that is called `^compute.*`.

ci/playbooks/kuttl/deploy-deps.yaml
cescgina: actually we don't set a compute in the nodeset for this job. I carried this from the nova kuttl job, but it must be there for a different reason, since the kuttl nodeset does not have compute nodes. I'm removing this task

ci/playbooks/kuttl/deploy-deps.yaml
cescgina: same as above, removing this tasks as we won't have computes for kuttl

ci/playbooks/kuttl/deploy-deps.yaml
cescgina: good catch, done

ci/playbooks/kuttl/deploy-deps.yaml
cescgina: done, thanks

ci/playbooks/kuttl/deploy-deps.yaml
```diff
@@ -0,0 +1,136 @@
...
+          delay: 10
+
+        - name: Wait for OpenStack operator to get installed
+          when: openstack_operator_subscription.stdout == ""
+          ansible.builtin.command:
+            cmd: >-
+              oc wait InstallPlan {{ cifmw_edpm_prepare_wait_installplan_out.stdout }}
+              --namespace=openstack-operators
+              --for=jsonpath='{.status.phase}'=Complete --timeout=20m
+
```
SeanMooney, line 121: we may want to have a hook here to install operators that are not part of the meta operator or other customisation.

i.e. we need to be able to install watcher operator here after the OpenStack operator is installed.

the "make kuttl-test-prep" should not actually depend on the operator its testing being installed so this will still work as we can add a pre-run playbook to the job that inherits form this so its not a blocker.

we have a hookpoint as a result of that but just wanted to highlight this.

SeanMooney (approved):
i have one comemnt inline but i bleive this is more or less good as is.

we can hook this via a zuul pre-run playbook in a child job but wanted to see if people tough having an actually ci framework hook was warranted or not.

ci/playbooks/kuttl/deploy-deps.yaml
cescgina: just to illustrate how we do it in watcher-operator here is the hook to install it https://github.com/openstack-k8s-operators/watcher-operator/pull/14/files#diff-108978819c05ae183d88ec87959c2341a94cfc3f9465e3aeee82d554217b4f58R56

SeanMooney: ya that pretty clean so we likely don't need a hook point here.
we could always add one later if we hit issues with the current approch.

SeanMooney: the one thing we need to be careful of is the watcher operator will edepnd on CRD defined in the main operator bundels. i.e. the Mariadb CRDs so I'm not sure if the post_infa hook is after the oepsntack operators are installed.

with that said the side effect will be that they watcher-oeprator will fail until the other CRDs are defined but that is likely ok. it should heal when the rest of the operators are installed.

PR #2572, opened 2024-11-26
[shiftstack]Enable the option to run the role isolated

rlobillo:
In order to allow users to run the shiftstack role without being part of
the full reproducer playbook, below changes are needed:

- Add missing kubeconfig param
- Load .bashrc while running the ansible-navigator. That's needed
for reading the envvars that will be used for writting the junit report.
- Enable the option to exclude artifacts to be gathered.

### DIFF
--- a/roles/shiftstack/defaults/main.yml
+++ b/roles/shiftstack/defaults/main.yml
@@ -26,6 +26,7 @@ cifmw_shiftstack_client_pod_name: "shiftstackclient-{{ cifmw_shiftstack_project_
 cifmw_shiftstack_client_pod_namespace: "openstack"
 cifmw_shiftstack_client_pvc_manifest: "{{ cifmw_shiftstack_client_pod_name }}_pvc.yml"
 cifmw_shiftstack_cluster_name: "ostest"
+cifmw_shiftstack_exclude_artifacts_regex: "openshift-install"
 cifmw_shiftstack_installation_dir: "{{ cifmw_shiftstack_basedir }}/installation"
 cifmw_shiftstack_manifests_dir: "{{ cifmw_shiftstack_basedir }}/manifests"
 cifmw_shiftstack_project_name: "shiftstack"
--- a/roles/shiftstack/tasks/test_config.yml
+++ b/roles/shiftstack/tasks/test_config.yml
@@ -31,6 +31,7 @@
         namespace: "{{ cifmw_shiftstack_client_pod_namespace }}"
         pod_name: "{{ cifmw_shiftstack_client_pod_name }}"
         command: >-
+          source .bashrc &&
           cd shiftstack-qa &&
           ansible-navigator run playbooks/{{ cifmw_shiftstack_run_playbook }} -e @jobs_definitions/{{ testconfig }} -e ocp_cluster_name={{ cifmw_shiftstack_cluster_name }} -e user_cloud={{ cifmw_shiftstack_project_name }}
       ansible.builtin.include_tasks: exec_command_in_pod.yml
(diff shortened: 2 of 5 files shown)


pablintino:
recheck

rlobillo:
Tested on https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/59184/rehearse-59184-periodic-ci-shiftstack-ci-release-4.16-e2e-openstack-rhoso/1862537103728971776

roles/shiftstack/defaults/main.yml
```diff
@@ -34,4 +34,5 @@ cifmw_shiftstack_qa_repo: "https://review.gerrithub.io/shiftstack/shiftstack-qa"
 cifmw_shiftstack_run_playbook: "ocp_testing.yaml"
 cifmw_shiftstack_sc: "local-storage"
 cifmw_shiftstack_shiftstackclient_artifacts_dir: "/home/cloud-admin/artifacts"
+cifmw_shiftstack_exclude_artifacts_regex: "openshift-install"
```
eurijon, line 37: Hey Ramón, even if it's just a cosmetic thing, we're following an alphabetical order for these vars definition, can you please follow the same approach?
Can you please also add the param to [ci-framework/roles/shiftstack/README.md](https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/shiftstack/README.md)?

eurijon (commented):
It looks good! just added a minor comment

roles/shiftstack/defaults/main.yml
rlobillo: Thanks for your comments! Fixed

rlobillo:
recheck

rlobillo:
@pablintino would you mind to review this one? Thank you!

rlobillo:
recheck

roles/shiftstack/tasks/test_config.yml
```diff
@@ -53,7 +54,7 @@
         KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
       ansible.builtin.command:
         cmd: >
-          oc rsync -n {{ cifmw_shiftstack_client_pod_namespace }}
+          oc rsync -n {{ cifmw_shiftstack_client_pod_namespace }} --exclude={{ cifmw_shiftstack_exclude_artifacts_regex }}
```
pablintino, line 57: Does it make sense to have this new parameter as optional? If so, I think we should render it only if not empty.

rlobillo: Addressing your comment. Thanks for the input!

roles/shiftstack/tasks/test_config.yml
```diff
@@ -48,12 +49,19 @@
 
   always:
     - name: "Copy the artifacts from the pod '{{ cifmw_shiftstack_client_pod_name }}'"
+      vars:
+        _exclude_options: >-
+          {{
+            '--exclude=' ~ cifmw_shiftstack_exclude_artifacts_regex
+            if cifmw_shiftstack_exclude_artifacts_regex | default('') != ''
```
eurijon, line 56: Just realized that if `cifmw_shiftstack_exclude_artifacts_regex` is optional now, we can set it to empty for the molecule test (in [roles/shiftstack/molecule/default/converge.yml](https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/shiftstack/molecule/default/converge.yml)) as there is no openshift-install artifacts to exclude in this test.

rlobillo: Good one! Added

rlobillo:
recheck

lewisdenny (approved):
/lgtm

PR #2574, opened 2024-11-27
Override Ceph version

fmount:
We used to support `N+1` server version with external `Ceph`, and this policy is still adopted. 
For this reason we need a way to do early testing in CI.
This patch adds the ability to override the `CentOS` `Ceph` repository and
install the target `cephadm` release passed as input.

Jira: https://issues.redhat.com/browse/OSPRH-10666

### DIFF
--- a/zuul.d/edpm_multinode.yaml
+++ b/zuul.d/edpm_multinode.yaml
@@ -343,6 +343,13 @@
         dashboard_enabled: true
         cephfs_enabled: true
         ceph_nfs_enabled: true
+      # Override the Ceph container tag and deploy Squid
+      cifmw_cephadm_container_tag: "v19"
+      # Override the Ceph Tools repo and install cephadm Squid
+      cifmw_cephadm_repository_override: true
+      cifmw_cephadm_version: "squid"
+      cifmw_cephadm_prepare_host: true
+
     files:
       - ^hooks/playbooks/control_plane_ceph_backends.yml
       - ^hooks/playbooks/control_plane_hci_pre_deploy.yml
--- a/roles/cifmw_cephadm/defaults/main.yml
+++ b/roles/cifmw_cephadm/defaults/main.yml
@@ -140,3 +140,10 @@ cifmw_cephadm_update_log_commands:
     cmd: "log last cephadm"
 cifmw_cephadm_wait_for_dashboard_retries: 10
 cifmw_cephadm_wait_for_dashboard_delay: 20
+# Repository parameters
+cifmw_cephadm_repository_override: false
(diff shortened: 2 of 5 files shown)


fmount:
early testing looks promising: https://logserver.rdoproject.org/74/2574/26cb85495cc7cb381c195455a94245e3dfa53355/github-check/podified-multinode-hci-deployment-crc/6aef5e6/controller/ci-framework-data/logs/ceph/ceph_version.log

zuul.d/edpm_multinode.yaml
```diff
@@ -343,6 +343,15 @@
         dashboard_enabled: true
         cephfs_enabled: true
         ceph_nfs_enabled: true
+      # Override the Ceph container image and deploy Squid
+      cifmw_cephadm_container_ns: "quay.io/ceph"
```
fmount, line 347: @fultonj @tosky I was thinking to switch this `multinode-hci-edpm` to `Ceph Squid`, so we have a way to do early testing upstream. 
1. Does it make sense to do this here? It would help to perform a deployment sanity check, like we are going to do in devstack (but in that case I'm not sure we're going to backport squid to Antelope) 
2. If you're positive about Squid, do we want to add here a stable tag (e.g. `v19.2`) to avoid testing the `latest` build?

I can simply remove this part here and only target something downstream. I wanted to get your opinion and feedback first because I'm potentially going to do the same thing in `install_yamls` [1] and target this new Ceph versions for tempest jobs w/ Glance, Cinder and Manila.

[1] https://github.com/openstack-k8s-operators/install_yamls/pull/952

fmount:
recheck

fmount:
@fultonj from internal testing looks good and we can land this patch if you're ok with it.

zuul.d/edpm_multinode.yaml
```diff
@@ -343,6 +343,15 @@
         dashboard_enabled: true
         cephfs_enabled: true
         ceph_nfs_enabled: true
+      # Override the Ceph container image and deploy Squid
```
pablintino, line 346: Do we really need the values here that are matching the defaults? Not really agains it but this means that we need to touch a few places to bump the version.

fmount: mmm I wanted a place where we can have a well defined list of parameters that allow us to test the next version.
For this reason I can revert back the defaults under `defaults/main.yaml` to match `Reef`, and leave here, at job level, the overrides to install `Squid`.
How does it sounds?

fultonj (approved):
/approve
/lgtm

fmount:
recheck

fmount:
recheck

PR #2575, opened 2024-11-27
Pin dependencies to avoid deps break CI

pablintino:
Till this commit we were not pinning the dependencies, resulting in break jobs as soon as a dependency changed underneath without notice. This commit pins all the dependencies to given known versions.


lewisdenny (changes requested):
Thank you @pablintino 

What about dogpile.cache and ansible-core in test-requirements.txt? Can we not pin them?

I also verified we can use ansible-core 2.15, I was concerned because we we are still using RHEL 9 default Python of 3.9, but ansible-core 2.15 supports 3.9 and is the latest release we can use:
https://github.com/ansible/ansible/blob/c89e95163fb1dceddf7abbefbb2e008aab9bef3c/setup.cfg#L40

pablintino:
> Thank you @pablintino
> 
> What about dogpile.cache and ansible-core in test-requirements.txt? Can we not pin them?
> 
> I also verified we can use ansible-core 2.15, I was concerned because we we are still using RHEL 9 default Python of 3.9, but ansible-core 2.15 supports 3.9 and is the latest release we can use: https://github.com/ansible/ansible/blob/c89e95163fb1dceddf7abbefbb2e008aab9bef3c/setup.cfg#L40

doh! I missed those two. The ansible-core was already pinned in the common file with 2.15.3, that we can definitely use.

lewisdenny:
recheck

lewisdenny:
recheck

pablintino:
> @pablintino I think there is a change with the Ansible version that is effecting how the vars/default vars are picked up in the Molecule tests :/
> 
> ```
> RUNNING HANDLER [dnsmasq : Restart dnsmasq] ************************************
> Wednesday 04 December 2024  23:25:30 -0500 (0:00:00.774)       0:06:05.970 ****
> fatal: [instance]: FAILED! =>
>   msg: |-
>     The conditional check 'cifmw_libvirt_manager_dependency_packages | difference(ansible_facts.packages.keys()) | length == 0' failed. The error was: error while evaluating conditional (cifmw_libvirt_manager_dependency_packages | difference(ansible_facts.packages.keys()) | length == 0): 'cifmw_libvirt_manager_dependency_packages' is undefined. 'cifmw_libvirt_manager_dependency_packages' is undefined
> 
>     The error appears to be in '/home/zuul/src/github.com/openstack-k8s-operators/ci-framework/roles/dnsmasq/handlers/main.yml': line 17, column 3, but may
>     be elsewhere in the file depending on the exact syntax problem.
> 
>     The offending line appears to be:
> 
> 
>     - name: Restart dnsmasq
>       ^ here
> ```

@lewisdenny Guess what, it only fails in the first run. Manually running it for another time makes the test pass.

pablintino:
recheck

frenzyfriday:
/lgtm
The testproject, as I remember, passed in the end

frenzyfriday (approved):

lewisdenny (approved):
/approve

PR #2578, opened 2024-11-28
[kustomize_deploy] Make wait conditions optional

pablintino:
In some ocassions, like when applying more than one nodesets the user may not want to wait for the first one to be applied and wait for all of them at once when all are applied.
This change is about not forcing the user, for no reason, to run a command to wait for a condition after some `oc apply` has been done.

### DIFF
--- a/roles/kustomize_deploy/tasks/execute_step.yml
+++ b/roles/kustomize_deploy/tasks/execute_step.yml
@@ -5,8 +5,6 @@
       - stage is defined
       - stage_id is defined
       - stage['path'] is defined
-      - stage['wait_conditions'] is defined
-      - stage['wait_conditions'] | length > 0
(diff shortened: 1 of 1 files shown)


roles/kustomize_deploy/tasks/execute_step.yml
```diff
@@ -40,7 +38,7 @@
         quiet: true
       ignore_errors: true  # noqa: ignore-errors
       register: _builtin_wait_cond_result
-      loop: "{{ stage.wait_conditions }}"
+      loop: "{{ stage.wait_conditions | default([]) }}"
```
hjensas, line 41: Would this execute faster if a condition was used on the block? 
Instead of running each task with an empty iteration, all the tasks in the block is skipped?

PR #2579, opened 2024-12-02
[test-operator][tobiko] Change downloaded images path

lpiwowar:
This patch changes the default path for downloaded images used for
tobiko testing from ~/.downloaded-images/ubuntu-minimal to
/usr/local/share.

This change is needed because we have to free /var/lib/tobiko from any
files that are present prior the container execution. This allows us to
mount ephemeral-storage to /var/lib/tobiko and use this directory as the
only point where write actions are allowed when readOnlyRootFilesystem:
true.

### DIFF
--- a/roles/test_operator/vars/main.yml
+++ b/roles/test_operator/vars/main.yml
@@ -35,7 +35,7 @@ cifmw_test_operator_tobiko_default_conf:
     test_runner_timeout: 14400.0
   ubuntu:
     interface_name: enp3s0
-    image_file: ~/.downloaded-images/ubuntu-minimal
+    image_file: /usr/local/share/ubuntu-minimal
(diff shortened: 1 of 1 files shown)


lpiwowar:
/hold

This PR requires review from: @eduolivares

eduolivares:
looks good to me. I would just add "Depends-On: https://github.com/openstack-k8s-operators/tcib/pull/233"

slawqo (approved):

lpiwowar:
@eduolivares there is cyclic dependency between those two PRs. We will have to schedule both PRs to be merged ideally at the same time.

eduolivares (approved):

frenzyfriday (approved):

PR #2580, opened 2024-12-02
Move test-operator to openstack-operators namespace

lpiwowar:
The test-operator role was installing test-operator into the openstack
namespace when it should be installed in the openstack-operators
namespace together with all the other namespaces. The installation into
the openstack namespace was needed when in the beginning the
test-operator was not capable of working correctly from other namespace
than openstack. Now the test-operator should be capable of working
correctly from the openstack-operators namespace.


lpiwowar:
recheck

lpiwowar:
recheck

lpiwowar:
no longer needed

PR #2583, opened 2024-12-04
Fix create-infra.yml sushy-emulator

hjensas:
The condition: `controllers'' in group_names` cause it to always skip the tasks to set up sushy-emulator.

  false_condition: '''controllers'' in group_names'
  skip_reason: Conditional result was False

Remove that part of the when condition for the block.


pablintino (commented):
/approve

PR #2585, opened 2024-12-04
Reprocuder: write ironic_nodes.yaml 

hjensas:
In the reproducer role's `generate_bm_info.yaml` add another task
to write the YAML required to enroll nodes in Openstack ironic. The
ironic file is only written if `cifmw_reproducer_ironic_node_name_prefix`
is set.

The `cifmw_reproducer_ironic_node_name_prefix` is used to filter from
libvirt_manager_bm_info_data so that only the nodes where the name
startswith the value of this option is added to the ironic_nodes.yaml.

Related: [OSPRH-12093](https://issues.redhat.com//browse/OSPRH-12093)

### DIFF
--- a/roles/reproducer/tasks/generate_bm_info.yml
+++ b/roles/reproducer/tasks/generate_bm_info.yml
@@ -121,3 +121,33 @@
     dest: "{{ cifmw_reproducer_controller_basedir }}/parameters/baremetal-info.yml"
     content: "{{ _content | to_nice_yaml }}"
     mode: "0644"
+
+- name: Output ironic_nodes to file
+  when: cifmw_reproducer_ironic_node_name_prefix
+  ansible.builtin.copy:
+    dest: "{{ cifmw_basedir }}/parameters/ironic_nodes.yaml"
+    content: |
+      {% set _ironic_nodes = [] %}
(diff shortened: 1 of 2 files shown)


roles/reproducer/tasks/generate_bm_info.yml
```diff
@@ -121,3 +121,33 @@
...
     content: "{{ _content | to_nice_yaml }}"
     mode: "0644"
+
+- name: Output ironic_nodes to file
+  when: cifmw_reproducer_ironic_node_name_prefix
+  ansible.builtin.copy:
+    dest: "{{ cifmw_basedir }}/parameters/ironic_nodes.yaml"
+    content: |
+      {% set _ironic_nodes = [] %}
+      {% for node in libvirt_manager_bm_info_data.keys() if node.startswith(cifmw_reproducer_ironic_node_name_prefix) %}
```
pablintino, line 131: I'm not too sure how this line will work. `cifmw_reproducer_ironic_node_name_prefix:` usually means/is equivalent to `cifmw_reproducer_ironic_node_name_prefix: null`, and passing null/None to `startswith` is not allowed.

hjensas: If cifmw_reproducer_ironic_node_name_prefix is `null` / `none` the task won't run - the `when` condition will be false and the task is skipped.

lewisdenny (approved):
/lgtm

PR #2592, opened 2024-12-10
Revert "Set Cinder customServiceConfig globally in DCN DT"

fultonj:
Reverts openstack-k8s-operators/ci-framework#2582

Sorry 2582 wasn't ready to merge yet.


hjensas (approved):

krcmarik (approved):

fultonj:
I don't think this patch caused a regression. I think the issue was in an earlier version of this instead and I've since fixed it.

https://github.com/openstack-k8s-operators/architecture/pull/459

So I'm putting this revert on hold and retesting with the above.

PR #2593, opened 2024-12-11
adoption_osp_deploy - os-net-config ironic

hjensas:
Set up the ironic network interface on osp controller nodes if the network is defined in the scenario.

By default it will map to `nic3`, but it can be overriden in the adoption scenario by setting `os_net_config_ironic_iface` in the scenario defenition.


lewisdenny:
recheck

lewisdenny (approved):
/approve

PR #2594, opened 2024-12-11
[test_operator] Common default values for container images

eduolivares:
test-operator currently supports 4 test frameworks: tempest, tobiko, horizontests and ansibletests. The relevant variables required to download their corresponding container images had to be configured separately, but usually were configured with common values. With this PR, default values for registry, namespace and tag can be configured and they apply to the 4 test frameworks.

### DIFF
--- a/roles/test_operator/defaults/main.yml
+++ b/roles/test_operator/defaults/main.yml
@@ -38,14 +38,18 @@ cifmw_test_operator_storage_class: "{{ cifmw_test_operator_storage_class_prefix 
 cifmw_test_operator_delete_logs_pod: false
 cifmw_test_operator_privileged: true
 cifmw_test_operator_selinux_level: "s0:c478,c978"
+# default testfw registry, namespace and tag can be overriden per testw (tempest, tobiko, horizontest and ansibletest)
+cifmw_test_operator_default_registry: quay.io
+cifmw_test_operator_default_namespace: podified-antelope-centos9
+cifmw_test_operator_default_image_tag: current-podified
 
 # Section 2: tempest parameters - used when run_test_fw is 'tempest'
 cifmw_test_operator_tempest_name: "tempest-tests"
-cifmw_test_operator_tempest_registry: quay.io
-cifmw_test_operator_tempest_namespace: podified-antelope-centos9
+cifmw_test_operator_tempest_registry: "{{ cifmw_test_operator_default_registry }}"
+cifmw_test_operator_tempest_namespace: "{{ cifmw_test_operator_default_namespace }}"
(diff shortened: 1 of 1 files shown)


lpiwowar (changes requested):
@eduolivares if we want to keep this in the `defaults` can you update the `README.md` please?

Otherwise looks good to me!:) :+1:

eduolivares:
> Also, cc @eshulman2
> 
> Do you think Ella this will break your PR? -> #2374
> 
> I think we should wait with merging this before Ella tells us her opinion.

I have taken a look at #2374 and wow, it's a big change.
However, I think #2594 won't affect much to it. Because #2374 will resolve any conflicts during task "Overwrite global_vars with stage_vars".
But, yes, let's see what @eshulman2 thinks before approving #2594

roles/test_operator/README.md
```diff
@@ -24,6 +24,9 @@ Execute tests via the [test-operator](https://openstack-k8s-operators.github.io/
...
+* `cifmw_test_operator_default_registry`: (String) Default registry for all the supported test frameworks (tempest, tobiko, horizontest and ansibletest) to pull their container images. It can be overridden at test framework level.  Defaults to `quay.io`
+* `cifmw_test_operator_default_namespace`: (String) Default registry's namespace for all the supported test frameworks (tempest, tobiko, horizontest and ansibletest) to pull their container images. It can be overridden at test framework level.  Defaults to `podified-antelope-centos9`
+* `cifmw_test_operator_default_image_tag`: (String) Default tag for all the supported test frameworks (tempest, tobiko, horizontest and ansibletest) to pull their container images. It can be overridden at test framework level.  Defaults to `current-podified`
 * `cifmw_test_operator_stages`: (List) List of dictionaries defining the stages that should be used in the test operator role. List items options are:
   * `name`: (String) The name of the stage. The name must be unique.
   * `type`: (String) The framework name you would like to call, currently the options are: tempest, ansibletest, horizontest, tobiko.
```
lpiwowar, line 32: **issue (non-blocking):** we could update this section of the README as well:

Important note! Only variables with the following structure can be used to override inside a stage: `cifmw_test_operator_[test-operator CR name]_[parameter name]`. ~~the only exception to that is `cifmw_test_operator_concurrency` which can not be override inside a stage.~~

lpiwowar (commented):
Looks good to me! :+1: 

Only one small change request in the README as the newly introduced parameters `cifmw_test_operator_default_*` can not be overridden inside of `cifmw_test_operator_stages`.

lpiwowar (approved):
Looks good to me! :+1:

lpiwowar:
recheck

PR #2596, opened 2024-12-12
Add hypervisor param to shiftstack playbook execution

eurijon:
For being able to resolve shift-on-stack apps endpoints (routes) hostnames from RHOSO pods it's required to add the DNS record entry in then dnsmasq service running on the hypervisor.

i.e.:
```
# cat /etc/cifmw-dnsmasq.d/addresses.conf 
address=/apps.ocp.openstack.lab/x.y.z.v
address=/apps.ostest.shiftstack.local/x.y.z.s      <<-----
```

It needs to be done after shift-on-stack cluster is installed as the installer is creating the floating IP for the apps endpoints, and the FIP is not known in advance.

For that purpose the cifmw_target_host param is passed so the playbook running on the shiftstackclient pod is able to reach the hypervisor.

### DIFF
--- a/hooks/playbooks/control_plane_horizon.yml
+++ b/hooks/playbooks/control_plane_horizon.yml
@@ -1,0 +1,28 @@
+---
+- name: Kustomize ControlPlane for horizon service
+  hosts: "{{ cifmw_target_hook_host | default('localhost') }}"
+  gather_facts: false
+  tasks:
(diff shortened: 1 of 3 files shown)


eurijon:
Continues on https://github.com/openstack-k8s-operators/ci-framework/pull/2626

PR #2597, opened 2024-12-12
[DCN] Update network values template

sbekkerm:
Explicitly set the route table ID to prevent random assignment, which can lead to the unexpected re-creation of VLAN interfaces.

Related Issue: [OSPRH-9899](https://issues.redhat.com//browse/OSPRH-9899)


hjensas (commented):
/lgtm

PR #266, opened 2023-06-07
Add support to skip registry login

Sandeepyadav93:
With this commit,  we do the following:-

1) Add a variable `cifmw_openshift_setup_skip_internal_registry_login` in openshift_setup role to skip registry login. This default to false 

2) For edpm deployment and baremetal deployment jobs in CI we pull needed containers from content-provider jobs, so we are passing `cifmw_openshift_setup_skip_internal_registry_login: true` via scenario file to skip internal registry login.

Recently, we are seeing lot of random failures during internal registry login so skiping this task as we don't consume anything from internal registry.

[1] https://logserver.rdoproject.org/21/221/3e2f992b0875a82a4178c52ec87dc22142eef674/github-check/dataplane-operator-crc-podified-edpm-baremetal/4d77948/job-output.txt

As a pull request owner and reviewers, I checked that:
- [x] Appropriate testing is done and actually running
- [x] Appropriate documentation exists and/or is up-to-date:
  - [x] README in the role
  - [x] Content of the docs/source is reflecting the changes

PR #2604, opened 2024-12-13
[test-operator] Move away from Jobs to Pods

lpiwowar:
With this PR [1] the test operator dropped the usage of OCP Jobs for
spawning of the test pods.

This change updates the test-operator role so that it works with the
new version of the test-operator that spawns test pods directly through
the OCP Pods object.

[1] https://github.com/openstack-k8s-operators/test-operator/pull/266

### DIFF
--- a/roles/test_operator/tasks/tempest-tests.yml
+++ b/roles/test_operator/tasks/tempest-tests.yml
@@ -29,7 +29,7 @@
       tempest_list_allowed:
         yaml_file: "{{ cifmw_test_operator_artifacts_basedir }}/list_allowed.yml"
         groups: "{{ cifmw_test_operator_default_groups }}"
-        job: "{{ cifmw_test_operator_job_name | default(omit) }}"
+        job: "{{ cifmw_test_operator_instance_name | default(omit) }}"
       register:
         list_allowed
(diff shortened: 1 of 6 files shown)


lpiwowar:
/hold

This PRs need to be merged at the same time. The idea is to first get /approved and /lgtm for both of them and then remove /hold at the same time:

Move away from Jobs to Pods  #266
https://github.com/openstack-k8s-operators/ci-framework/pull/2604

lpiwowar:
recheck

roles/test_operator/tasks/tempest-tests.yml
```diff
@@ -29,7 +29,7 @@
       tempest_list_allowed:
         yaml_file: "{{ cifmw_test_operator_artifacts_basedir }}/list_allowed.yml"
         groups: "{{ cifmw_test_operator_default_groups }}"
-        job: "{{ cifmw_test_operator_job_name | default(omit) }}"
+        job: "{{ cifmw_test_operator_instance_name | default(omit) }}"
```
kstrenkova, line 32: There is a job variable. Is this the correct behavior?

PR #2605, opened 2024-12-14
Add dynamic DCB configuration to Juniper switches

abhiramnarayana:
- Implements forwarding classes, classifiers and traffic control profiles.
- Applies DCB settings to interfaces dynamically.
- Adds scheduler and scheduler map configurations.


rdiazcam:
lgtm

PR #2606, opened 2024-12-16
NFV OVS DPDK and SRIOV workload changes

Jaganathancse:
This change is to add workload with both ovs dpdk
and sriov ports and ping testing for both ports
during update.

### DIFF
--- a/roles/update/templates/workload_launch.sh.j2
+++ b/roles/update/templates/workload_launch.sh.j2
@@ -42,8 +42,8 @@ function os_cmd {
 
 function set_vm_ip {
     ## assign floating ip or external ip
-    local workload_sriov={{workload_sriov|default(false) | bool | ternary("True", "")}}
-    if [ -n "${workload_sriov}" ]; then
+    local fixed_ip={{ $1|default(false) | bool }}
+    if [ -n "${fixed_ip}" ]; then
         EXTERNAL_IP=$(os_cmd port show "${SRIOV_PORT}" -f yaml -c fixed_ips | awk '/ip_address/{print $3;exit}')
         VM_IP=${EXTERNAL_IP}
     else
@@ -84,7 +84,7 @@ cleanup_on_exit() {
 
 function generate_traffic {
     if [ -z "${VM_IP}" ]; then
-        set_vm_ip
+        set_vm_ip $1
     fi
     if ! ${SSH} -i "${SSH_KEY_FILE}" cirros@${VM_IP} test -e ${IPERF_BIN}; then
         NS_FIXUP=false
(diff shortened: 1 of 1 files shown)


Jaganathancse:
planning to update this Pr.

Jaganathancse:
recheck

roles/update/templates/workload_launch.sh.j2
```diff
@@ -510,7 +542,16 @@ if [[ "${MODE}" == "workload_traffic" ]]; then
     echo "export CINDER_VOL_ID=${CINDER_VOL_ID}" >> ~/workload_suffix
     echo "export INSTANCE_FIP=${INSTANCE_FIP}" >> ~/workload_suffix
     echo "export VM_IP=${VM_IP}" > ~/vm_ip.sh
-    generate_traffic
+    local workload_sriov={{workload_sriov|default(false) | bool | ternary("True", "")}}
```
danpawlik, line 545: why not just: 
```
    local workload_sriov={{workload_sriov|default(false) | bool  }}
    local workload_dpdk={{workload_dpdk|default(false) | bool }}
```

roles/update/templates/workload_launch.sh.j2
```diff
@@ -42,8 +42,8 @@ function os_cmd {
 
 function set_vm_ip {
     ## assign floating ip or external ip
-    local workload_sriov={{workload_sriov|default(false) | bool | ternary("True", "")}}
-    if [ -n "${workload_sriov}" ]; then
+    local fixed_ip={{ $1|default(false) | bool }}
```
danpawlik, line 45: how it should work? Here I see bash + ansible in jinja2 template.

danpawlik (changes requested):
the scripts seems to have jinja2 and shell script mixed. Does they work?

Jaganathancse:
> @Jaganathancse please update the PR or leave bot to close it.

Will update this PR and its mandatory. 
Please also review it.

roles/update/templates/workload_launch.sh.j2
sdatko: +1 this is mixing two worlds,, it is not gonna work

Jaganathancse:
> the scripts seems to have jinja2 and shell script mixed. Does they work?

followed existing code only, just added the scenario for instance with both ovs dpdk and sriov ports

roles/update/templates/workload_launch.sh.j2
Jaganathancse: ok thanks, will test and update it.

PR #2607, opened 2024-12-16
Consider that the zuul executor may be an IP

pablintino:
Exposed a variable to pass the API endpoint without computing it too.

### DIFF
--- a/ci/playbooks/multinode-autohold.yml
+++ b/ci/playbooks/multinode-autohold.yml
@@ -20,10 +20,16 @@
       block:
         - name: Fetch existing autoholds from zuul
           vars:
+            _zuul_host: >-
+              {{
+                zuul.executor.hostname
+                if (zuul.executor.hostname is ansible.utils.ip) else
+                (zuul.executor.hostname | split('.'))[1:] | join('.')
+              }}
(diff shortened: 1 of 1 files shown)


ci/playbooks/multinode-autohold.yml
```diff
@@ -20,10 +20,16 @@
       block:
         - name: Fetch existing autoholds from zuul
           vars:
+            _zuul_host: >-
+              {{
+                zuul.executor.hostname
+                if (zuul.executor.hostname is ansible.utils.ip) else
+                (zuul.executor.hostname | split('.'))[1:] | join('.'))
```
hjensas, line 27: I think the last parantheses must be removed?

```
AnsibleError: template error while templating string: unexpected ')'. String: {{ zuul.executor.hostname if (zuul.executor.hostname is ansible.utils.ip) else (zuul.executor.hostname | split('.'))[1:] | join('.')) }}. unexpected ')'
```

hjensas (changes requested):

lewisdenny (approved):
/approve

hjensas (approved):

PR #2608, opened 2024-12-16
[ci_gen_kustomize_values] Allow reject vars patches

pablintino:


roles/ci_gen_kustomize_values/tasks/generate_snippets.yml
```diff
@@ -77,11 +86,22 @@
...
+      {% set _final_list = []                                                  -%}
+      {% for entry in (_cifmw_gen_kustomize_values_reject_extra_expressions +
+                      cifmw_ci_gen_kustomize_values_remove_keys_expressions)   -%}
+      {%   if entry is string                                                  -%}
+      {%     set _ = _final_list.append(entry)                                 -%}
+      {%   else                                                                -%}
+      {%     set _ = _final_list.extend(entry)                                 -%}
+      {%   endif                                                               -%}
+      {% endfor                                                                -%}
+      {{ _final_list | unique }}
```
hjensas, line 99: I think we can just concatenate the lists, and then `| flatten | unique`?
```
{{ 
  (_cifmw_gen_kustomize_values_reject_extra_expressions +
   cifmw_ci_gen_kustomize_values_remove_keys_expressions)
  | flatten | unique  
}}
```

pablintino: @hjensas the approach I took allows the user to pass individual regexes, thus the if/else

pablintino: @hjensas Good catch man, i didn't thought about using `flatten` this time.

lewisdenny (approved):
/approve

PR #2611, opened 2024-12-17
Log Ceph health before starting Ceph upgrade

fultonj:
The Ceph Upgrade tasks in the cifmw_cephadm role will fail before the upgrade starts if the health status is WARN or ERR. We have had the job fail in CI but we do not know why. The task should log the Ceph health before starting the upgrade so that CI results will give the job owner more insight into why the job failed.


fultonj:
@fmount @katarimanojk

The `component-compute-edpm-update-rhel9-rhoso18.0-crc-ceph` job failed 3 times in a row on HEALTH_WARN but we don't know why. My initial plan was just to get the debug information and I think we're save to merge that part of this patch.

For the second part of this patch I took away failure if health is only in WARN. What do you think of that? If we know the cluster can survive and upgrade when in WARN then I think it makes sense to keep that change, but if we know it can't then I'll take it out. I thought we should at least get a run of the job with both changes to start debugging the problem though.

roles/cifmw_cephadm/tasks/ceph_upgrade.yml
```diff
@@ -18,12 +18,15 @@
   ansible.builtin.set_fact:
     ceph_health: "{{ ceph.stdout | from_json }}"
 
-- name: Fail if health is HEALTH_WARN || HEALTH_ERR
+- name: Log ceph health for debug purposes
```
fmount, line 21: If we proceed even when Ceph is `HEALTH_WARN`, do we care about printing this msg? When the process is over (if it doesn't fail) we trigger logs gathering (L67). Basically I think it doesn't get us useful information showing that the cluster is `HEALTH_OK` vs `HEALTH_WARN`.

fmount:
> @fmount @katarimanojk
> 
> The `component-compute-edpm-update-rhel9-rhoso18.0-crc-ceph` job failed 3 times in a row on HEALTH_WARN but we don't know why. My initial plan was just to get the debug information and I think we're save to merge that part of this patch.
> 
> For the second part of this patch I took away failure if health is only in WARN. What do you think of that? If we know the cluster can survive and upgrade when in WARN then I think it makes sense to keep that change, but if we know it can't then I'll take it out. I thought we should at least get a run of the job with both changes to start debugging the problem though.

What is weird is that I didn't expect that job to perform a ceph minor update. It was added in `unigamma` only, where we know everything about the status of Ceph in terms of sizing, while we don't have control on the `component-compute-edpm-update` job.
I'm wondering if the right choice would be to not perform such update in that job, but I'm also ok to still keep it and proceed even if the status is in `HEALTH_WARN`. However, as per cephadm code, there might be chances to get another failure when new OSDs containers are rolled out, because of the `HEALTH_WARN` and potential data movement (scrub, recovery, etc) happening at that time.

roles/cifmw_cephadm/tasks/ceph_upgrade.yml
katarimanojk: +1, if we go ahead without failing for HEALTH_WARN, we may avoid printing 'ceph -s -f json' here as we have same info collected in the logs at the end.

fmount: However, looking at the last log it was sufficient, looking at the warning msg, to figure out the reason of the failed upgrade.
+1 to keep this task and only failing if `HEALTH_ERR`!

fultonj:
@fmount should I update this to only log the debug info and fail the update for health warn?

roles/cifmw_cephadm/tasks/ceph_upgrade.yml
fultonj: I think it gets us useful information.

We didn't know why the job was failing before. We just knew it was in health warn. Now we know it's because OSD `count 2 < osd_pool_default_size 3`. 

If the job succeeds it will we gather and print the info but if it fails we don't get that information.

> When the process is over (if it doesn't fail) we trigger logs gathering (L67)

"if it doesn't fail" is the problem that part of this patch will solve.

It helps more than it harms.

fmount:
> > What is weird is that I didn't expect that job to perform a ceph minor update. It was added in `unigamma` only, where we know everything about the status of Ceph in terms of sizing, while we don't have control on the `component-compute-edpm-update` job. I'm wondering if the right choice would be to not perform such update in that job, but I'm also ok to still keep it and proceed even if the status is in `HEALTH_WARN`. However, as per cephadm code, there might be chances to get another failure when new OSDs containers are rolled out, because of the `HEALTH_WARN` and potential data movement (scrub, recovery, etc) happening at that time.
> 
> per our discussion we now know that `cifmw_ceph_update: false` was not set. Maybe we should default it to false and only override it in Gama? since the job was getting run when we didn't want it to.

Yeah, it's already `false` and set to true in the unigamma variables. However, if other jobs reference the same variables, they should provide their own overrides.

fmount (approved):
/lgtm

PR #2618, opened 2024-12-18
Add configuration variables for control plane testing during update.

sathlan:
Add a number of variables to control various aspects of the control
plane testing.  Using those new varibles we can configure control
plane testing settings from the job definition.

Increased the default time we wait for the last vm to be created and
destroyed as 5 minutes (the previous timeout) was slightly too short.
It's now 7 minutes.

Closes: https://issues.redhat.com/browse/OSPRH-12349


ciecierski (approved):

sathlan:
recheck

PR #2619, opened 2024-12-18
[test_operator][tobiko] Download default customized guest image

eduolivares:
With recent changes in tobiko 0.8.4, the tobiko pods will download an already customized image instead of using official ubuntu images that were customized during tobiko execution.


roles/test_operator/defaults/main.yml
```diff
@@ -151,6 +151,7 @@ cifmw_test_operator_tobiko_version: master
 cifmw_test_operator_tobiko_pytest_addopts: null
 cifmw_test_operator_tobiko_prevent_create: null
 cifmw_test_operator_tobiko_num_processes: null
+cifmw_test_operator_tobiko_advanced_image_url: "https://softwarefactory-project.io/ubuntu-minimal-customized-enp3s0"
```
pablintino, line 154: Care to update the README to point to what's expected from it and the associated value?

eduolivares: :facepalm: in a moment

lewisdenny:
recheck

lewisdenny (approved):
/approve

Please take out of draft status when you would like it to merge :)

PR #2620, opened 2024-12-20
Add the hypervisors to the inventory

pablintino:
For use cases like ShiftOnStack the deployment may need to tweak the hypervisor.
We did not have a way to tell the deployment how to reach the hypervisor so this commit exposes the hypervisor Ansible instance to each host and creates a hypervisors group in the generated inventory.

### DIFF
--- a/roles/libvirt_manager/templates/all-inventory.yml.j2
+++ b/roles/libvirt_manager/templates/all-inventory.yml.j2
@@ -20,6 +20,20 @@ virtual machines.
         ansible_ssh_common_args: "-o StrictHostKeyChecking=no -J {{ ansible_user | default(ansible_user_id) }}@{{ _hostname }}"
 {% endif %}
 {% endfor %}
+hypervisors:
+  hosts:
+{% set _hypervisors = (
+      ( _cifmw_libvirt_manager_layout.vms.values() | selectattr('target', 'defined') | map(attribute='target') | list)
+      + [inventory_hostname] ) | unique %}
+{% for _hypervisor in _hypervisors %}
+    {{ _hypervisor }}:
+{% if hostvars[_hypervisor]['ansible_host'] is defined %}
+{% set _hostname = hostvars[_hypervisor]['ansible_host'] %}
+        ansible_host: "{{ _hostname }}"
+        ansible_ssh_private_key_file: "~/.ssh/ssh_{{ _hostname }}"
+        vars:
+{% endif %}
+{% endfor %}
 localhosts:
   hosts:
     localhost:
--- a/roles/libvirt_manager/templates/inventory.yml.j2
+++ b/roles/libvirt_manager/templates/inventory.yml.j2
@@ -14,4 +14,5 @@
 {% else                                                     %}
       ansible_ssh_private_key_file: ~/.ssh/id_cifw
 {% endif                                                    %}
+      cifmw_hypervisor_host: "{{ _cifmw_libvirt_manager_layout.vms[host.key].target | default(inventory_hostname) }}"
 {% endfor %}
--- a/roles/reproducer/tasks/configure_controller.yml
+++ b/roles/reproducer/tasks/configure_controller.yml
@@ -179,6 +179,7 @@
         marker: "## {mark} {{ _ssh_host }}"
         block: |-
           Host {{ _ssh_host }} {{ hostvars[host]['inventory_hostname'] }}
+            Hostname {{ _ssh_host }}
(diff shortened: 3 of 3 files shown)


hjensas:
So shift-stack needs to re-wire the datacenter networking, add install more RAM servers and/or modify the routers or the DNS configuration? This seems like a problem with separation of concerns? My worry here is that we it will be harder to move this automation to a different platform (cloud or hardware) in the future.

Would it not be possible to pre-seed whatever is required in the hypervisor with the "module" that creates the infrastructure? (libvirt-manager)

pablintino:
> So shift-stack needs to re-wire the datacenter networking, add install more RAM servers and/or modify the routers or the DNS configuration? This seems like a problem with separation of concerns? My worry here is that we it will be harder to move this automation to a different platform (cloud or hardware) in the future.
> 
> Would it not be possible to pre-seed whatever is required in the hypervisor with the "module" that creates the infrastructure? (libvirt-manager)

@hjensas no, shift-stack needs to add a DNS entry in the hypervisor that can only be added after the payload OCP cluster is deployed, that's is, after they deploy OpenShift on top of RHOSO.
May it imply a separation of concerns problem? Sure, but I don't think there's a quick way to accomplish this in the framework that is not this one.

hjensas:
> > So shift-stack needs to re-wire the datacenter networking, add install more RAM servers and/or modify the routers or the DNS configuration? This seems like a problem with separation of concerns? My worry here is that we it will be harder to move this automation to a different platform (cloud or hardware) in the future.
> > Would it not be possible to pre-seed whatever is required in the hypervisor with the "module" that creates the infrastructure? (libvirt-manager)
> 
> @hjensas no, shift-stack needs to add a DNS entry in the hypervisor that can only be added after the payload OCP cluster is deployed, that's is, after they deploy OpenShift on top of RHOSO. May it imply a separation of concerns problem? Sure, but I don't think there's a **quick** way to accomplish this in the framework that is not this one.

ack, I am guessing but this must be for the API and Ingress addresses on a provider network. It would be possible to provide the addresses in install-config - but they may have a valid reason to explicitly test without setting them statically.

/lgtm

eurijon:
It's been tested from https://github.com/openstack-k8s-operators/ci-framework/pull/2596 and the hostname is correctly retrieved:

```
2024-12-24 05:31:23,460 p=25595 u=zuul n=ansible | TASK [shiftstack : Debug hypervisor ansible_host msg=hypervisor ansible_host: {{ hostvars[hostvars['controller-0']['cifmw_hypervisor_host']]['ansible_host'] }} ] ***
2024-12-24 05:31:23,460 p=25595 u=zuul n=ansible | Tuesday 24 December 2024  05:31:23 -0500 (0:00:00.376)       1:04:03.311 ****** 
2024-12-24 05:31:23,527 p=25595 u=zuul n=ansible | ok: [localhost] => 
  msg: 'hypervisor ansible_host: titan19.lab.eng.tlv2.redhat.com '
```

For context, we set the API and Ingress FIPs in the install-config.yaml but those FIPs are created before cluster provisioning (outside ci-framework). The FIPs could be reserved, but that requires additional logic in ci-framework for FIP management, and significant changes in the shift-on-stack automation too. On the other hand, the Ingress hostnames depend on the cluster name, which can be different for each cluster. That's why adding the DNS entry in the hypervisor after shift-on-stack cluster deployment seems the most reasonable way.

lewisdenny (approved):
/approve

eurijon:
@pablintino hey, is it ok to cherry-pick this PR to `18.0-fr1` branch?

ciecierski:
With addition of hypervisor to the inventory there is problem in adoption_osp_deploy https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/adoption_osp_deploy/tasks/main.yml#L28 

`
TASK [Wait for SSH on started VMs] *********************************************
changed: [hypervisor]

TASK [Ensure we get SSH on nodes] **********************************************
FAILED - RETRYING: [hypervisor]: Ensure we get SSH on nodes (60 retries left).
FAILED - RETRYING: [hypervisor]: Ensure we get SSH on nodes (59 retries left).
ok: [hypervisor]

TASK [Start vm hypervisor] *****************************************************
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: VMNotFound: virtual machine cifmw-hypervisor not found
fatal: [hypervisor]: FAILED! => {"changed": false, "msg": "virtual machine cifmw-hypervisor not found"}
`

PR #2628, opened 2025-01-02
KUTTL: don't bail out immediately, save the test reports

tosky:
Execute all KUTTL tests without bailing out at the first error. Instead make sure that the tests are executed and their JUnitXML reports are collected if available.
Move the accept/fail condition at the end, following the same pattern used elsewhere when running tests.


tosky:
recheck

tosky:
recheck

tosky:
As the kuttl tests executed here don't generate JUnitXML reports yet, I've tested the change here: https://github.com/openstack-k8s-operators/glance-operator/pull/678

PR #2629, opened 2025-01-03
Adoption ironic hook - IPA var/lib/ironic/httpboot

hjensas:
For inspection we need to add the ironic-python-agent to the /var/lib/ironic/httpboot folder. Update the hook to install the package on groups - osp-controllers and osp-underclouds, and copy the file to the appropriate location.

### DIFF
--- a/hooks/playbooks/adoption_ironic_post_oc.yml
+++ b/hooks/playbooks/adoption_ironic_post_oc.yml
@@ -17,32 +17,83 @@
 - name: OSP 17 - Ironic post overcloud
   hosts: "{{ cifmw_target_host | default('localhost') }}"
   tasks:
-    - name: Gather ansible_user_dir from undercloud
-      delegate_to: osp-undercloud-0
+    - name: Gather ansible_user_dir from OSP nodes
+      delegate_to: "{{ item }}"
       ansible.builtin.setup:
         gather_subset:
           - user_dir
+      loop:
+        - controller-0
+        - osp-undercloud-0
+        - osp-controller-0
+        - osp-controller-1
+        - osp-controller-2
 
-    - name: Glance images - ironic-python-agent
-      delegate_to: osp-undercloud-0
+    - name: Install ironic-python-agent images package
       block:
         - name: Install the ironic-python-agent images package
           become: true
+          delegate_to: "{{ item }}"
           ansible.builtin.dnf:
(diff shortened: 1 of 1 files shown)


lewisdenny (approved):
/approve

hooks/playbooks/adoption_ironic_post_oc.yml
```diff
@@ -17,32 +17,83 @@
...
-    - name: Gather ansible_user_dir from undercloud
-      delegate_to: osp-undercloud-0
+    - name: Gather ansible_user_dir from OSP nodes
+      delegate_to: "{{ item }}"
       ansible.builtin.setup:
         gather_subset:
           - user_dir
+      loop:
+        - controller-0
+        - osp-undercloud-0
```
pablintino, line 27: Do we have a way to get these hosts from a group? Just in case we add/remove more.

hjensas: I don't think so, I tried to do this without a loop - i.e by doing `when: inventory_hostname in group["group_name"]` style condition - it failed on looking up the group. I think that approach would execute much faster ...


 ... side note - I think it would be much more efficient elsewhere as well - for example these "delegate_to" loops would be beneficial to run in parallel:

* https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/adoption_osp_deploy/tasks/prepare_overcloud.yml#L113
* https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/adoption_osp_deploy/tasks/prepare_overcloud.yml#L194
* https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/adoption_osp_deploy/tasks/prepare_overcloud.yml#L206
* https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/adoption_osp_deploy/tasks/prepare_overcloud.yml#L217
* https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/adoption_osp_deploy/tasks/prepare_overcloud.yml#L236

pablintino: Yeah, sometime ago I had a look at the Ansible internals and iirc delegate_to only works with hosts, not groups.

frenzyfriday (approved):
/lgtm

PR #2632, opened 2025-01-07
[repo_setup] Download rhos-release using kerberos

pablintino:
If the RPM name points to a URL we now use a custom plugin to fetch the content.
If the endpoint challenges the plugins with SPNEGO authentication and a kerberos ticket is present the plugin will authenticate itself using the ticket.

### DIFF
--- a/plugins/modules/url_request.py
+++ b/plugins/modules/url_request.py
@@ -1,0 +1,192 @@
+#!/usr/bin/python
+
+# Copyright: (c) 2025, Pablo Rodriguez <pabrodri@redhat.com>
+# Apache License Version 2.0 (see LICENSE)
+
+__metaclass__ = type
+
+DOCUMENTATION = r"""
+---
+module: url_request
+short_description: Downloads/fetches the content of a SPNEGO secured URL
+extends_documentation_fragment:
+    - files
+
+description:
+- Downloads/fetches the content of a SPNEGO secured URL
+- A kerberos ticket should be already issued
+
+author:
+  - Adrian Fusco (@afuscoar)
+  - Pablo Rodriguez (@pablintino)
+
+options:
+  url:
+    description:
+      - The URL to retrieve the content from
+    required: True
+    type: str
+  verify_ssl:
+    description:
+      - Enables/disables using TLS to reach the URL
+    required: False
+    type: bool
+    default: true
+  dest:
+    description:
+      - Path to the destination file/dir where the content should be downloaded
+      - If not provided the content won't be written into disk
+    required: False
+    type: str
+
+"""
+
+EXAMPLES = r"""
+- name: Get some content
+  url_request:
+    url: "http://someurl.local/resource"
+    dest: "{{ ansible_user_dir }}/content.raw"
+    mode: "0644"
+  register: _fetched_content
+
+- name: Show base64 content
+  debug:
+    msg: "{{ _fetched_content.response_b64 }}"
+"""
+
+RETURN = r"""
+status_code:
+    description: HTTP response code
+    type: int
+    returned: returned request
+content_type:
+    description: HTTP response Content-Type header content
+    type: str
+    returned: returned request
+headers:
+    description: HTTP response headers
+    type: dict
+    returned: returned request
+response_b64:
+    description: Returned content base64 encoded
+    type: str
+    returned: successful request
+response_json:
+    description: Returned content as a dict
+    type: str
+    returned: successful request that returned application/json
+path:
+    description: Written file path
+    type: str
+    returned: successful request
+"""
+
+import base64
+import os.path
+import re
+
+from ansible.module_utils.basic import AnsibleModule
+
+
+try:
+    from requests import get
+
+    python_requests_installed = True
+except ImportError:
+    python_requests_installed = False
+try:
+    from requests_kerberos import HTTPKerberosAuth, OPTIONAL
+
+    python_requests_kerberos_installed = True
+except ImportError:
+    python_requests_kerberos_installed = False
+
+
+def main():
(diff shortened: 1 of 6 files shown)


roles/repo_setup/tasks/rhos_release.yml
```diff
@@ -1,8 +1,29 @@
...
+      ansible.builtin.file:
+        path: "{{ cifmw_repo_setup_rhos_release_path }}"
+        state: directory
+        mode: "0755"
+
+    - name: Download the RPM
+      cifmw.general.url_request:
+        url: "{{ cifmw_repo_setup_rhos_release_rpm }}"
+        dest: "{{ cifmw_repo_setup_rhos_release_path }}/rhos-release.rpm"
+        mode: "0644"
```
frenzyfriday, line 15: Is the mode parameter considered in the plugin? I did not see where it is explicitly implemented? Is it something that is done by default?

frenzyfriday (approved):
lgtm

roles/repo_setup/tasks/rhos_release.yml
pablintino: @frenzyfriday It is handled here: https://github.com/openstack-k8s-operators/ci-framework/pull/2632/files#diff-946a5845b8481f6141ff3ac2d799db7405291377fde48342f8c0377f62fbb6f3R175-R180

roles/repo_setup/tasks/rhos_release.yml
```diff
@@ -1,8 +1,29 @@
...
+  when: cifmw_repo_setup_rhos_release_rpm is url
+  block:
+    - name: Create download directory
+      ansible.builtin.file:
+        path: "{{ cifmw_repo_setup_rhos_release_path }}"
+        state: directory
+        mode: "0755"
+
+    - name: Download the RPM
+      cifmw.general.url_request:
```
evallesp, line 12: (blocking) question: Have you tried to use uri_module with gssapi as True? I'm unsure if we need the custom module.

roles/repo_setup/tasks/rhos_release.yml
```diff
@@ -1,8 +1,29 @@
 ---
+- name: Download RHOS Release if the given rpm is a URL
+  when: cifmw_repo_setup_rhos_release_rpm is url
+  block:
```
evallesp, line 4: (non-blocking) Question: Do we need to kinit using a keytab? is that done somewhere?

plugins/modules/url_request.py
```diff
@@ -0,0 +1,190 @@
...
+                filename = re.findall(
+                    "filename=(.+)", response.headers["content-disposition"]
+                )
+                dest = filename[0] if filename else None
+            elif os.path.exists(dest) and os.path.isdir(dest):
+                dest = None
+
+            if not dest:
+                module.fail_json(
+                    msg="Destination points to a directory and the filename cannot be retrieved from the response"
```
evallesp, line 168: typo: Should this be "or" instead of "and"?

plugins/modules/url_request.py
```diff
@@ -0,0 +1,190 @@
...
+                module.fail_json(msg=f"Error with the JSON response: {str(e)}")
+
+        if "dest" in module.params:
+            dest = module.params["dest"]
+            if (
+                os.path.exists(dest)
+                and os.path.isdir(dest)
+                and "content-disposition" in response.headers
+            ):
+                filename = re.findall(
```
evallesp, line 159: pPraise Like a lot, so if the file exists and its folder fails unless there's content-disposition then it'll be saved to the file pointing to that header.
If not, then it's see regular and basic behavior that everyone would eexpect

Took me some time to understand this, and I'm unsure if some comments there would be good for the module maintenance.

plugins/modules/url_request.py
```diff
@@ -0,0 +1,190 @@
...
+        argument_spec=module_args, supports_check_mode=True, add_file_common_args=True
+    )
+
+    if not python_requests_installed:
+        module.fail_json(msg="requests required for this module.")
+
+    if not python_requests_kerberos_installed:
+        module.fail_json(msg="requests_kerberos required for this module.")
+
+    if module.check_mode:
```
evallesp, line 126: (non-blocking) question: Could you add here some details of how this would help to test the module? IIUC, this should allow us to test the module without doing any action, but I'm unsure if we would need to not support check mode.

evallesp (commented):
LGTM in general. I have some questions and a suggestion of using a builtin module.
In any case, I like the custom module, and I don't have a problem of maintain it's usage instead my suggestion.

roles/repo_setup/tasks/rhos_release.yml
adrianfusco: I tried in the past and I couldn't make it work because the negotiation that should be done in the background.
Ref: https://github.com/ansible/ansible/issues/32231#issuecomment-515504531

roles/repo_setup/tasks/rhos_release.yml
pablintino: Yes, it's up to the user to do so. In our case, it's done in the base zuul job.

plugins/modules/url_request.py
pablintino: I don't think so, if dest points to a directory but we were able to extract the filename we won't throw the error.

plugins/modules/url_request.py
pablintino: Added some additional comments.

pablintino:
recheck

pablintino:
recheck

lewisdenny (approved):
/approve

Nice work guys

PR #2633, opened 2025-01-08
Add retry to cephadm install

jirimacku:
The cephadm package install is failing occasionally because of repo temporary issues. Let's add retries.

### DIFF
--- a/roles/cifmw_cephadm/tasks/install_cephadm.yml
+++ b/roles/cifmw_cephadm/tasks/install_cephadm.yml
@@ -30,6 +30,10 @@
   ansible.builtin.dnf:
     name: cephadm
     state: present
+  register: task_result
+  retries: "{{ cifmw_cephadm_wait_install_retries }}"
+  delay: "{{ cifmw_cephadm_wait_install_delay }}"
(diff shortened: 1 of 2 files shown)


roles/cifmw_cephadm/tasks/install_cephadm.yml
```diff
@@ -30,6 +30,10 @@
   ansible.builtin.dnf:
     name: cephadm
     state: present
+  register: task_result
+  retries: 8
```
fmount, line 34: maybe not important, just out of curiosity: how did we come up w/ these numbers? is there any best practice to set defaults? Would it be good to define two variables under `roles/cifmw_cephadm/default/main.yaml` with those numbers, so we can eventually pass them as overrides to the jobs?
Apart from this small potential change, I'm not against this patch and we can land it.

jirimacku: Thanks, good point. I've just added the default values.

fmount (approved):
/approve
/lgtm

fmount (approved):
/approve
/lgtm

PR #2637, opened 2025-01-09
Define a new EDPM role for installing nvidia driver on nodes

sbauza:
In order to do nvidia vGPU testing, we need to deploy a specific package on EDPM nodes and do some other actions.
This is a multi-stage role requring a reboot between the two stages hence the two defined phases.

### DIFF
--- a/playbooks/nvidia-mdev-phase1.yml
+++ b/playbooks/nvidia-mdev-phase1.yml
@@ -1,0 +1,73 @@
+---
+# Copyright 2024 Red Hat, Inc.
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+
+- name: Gather the list of EDPM computes
+  hosts: "{{ cifmw_target_hook_host | default('localhost') }}"
+  gather_facts: false
+  tasks:
+    - name: Fetch OSP BMO nodesets
+      environment:
+        KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+        PATH: "{{ cifmw_path }}"
+      ansible.builtin.command:
+        cmd: >-
+          oc get OpenStackBaremetalSet -n "{{ namespace|default('openstack') }}"  -o yaml
+      register: _osp_bmo_nodsets_oc_out
+
+    - name: Craft the BM hosts list
+      ansible.builtin.set_fact:
+        _bmo_provisioned_hosts: >-
+          {% set hosts = [] -%}
+          {% set nodesets = (_osp_bmo_nodsets_oc_out.stdout | from_yaml)['items'] | default([]) -%}
+          {% for spec in nodesets | map(attribute='spec') -%}
+          {%   for host_key, host_val in spec.baremetalHosts.items() -%}
+          {%     set _ = hosts.append(
+            {
+            'name': host_key,
+            'ip': host_val['ctlPlaneIP'] | ansible.utils.ipaddr('address'),
+            'user': spec.cloudUserName,
+            'group': host_key | split('-') | first + 's'
+          }) -%}
+          {%   endfor -%}
+          {% endfor -%}
+          {{ hosts }}
+
+    - name: Add OSP BMO nodesets to Ansible
+      ansible.builtin.add_host:
+        name: "{{ item.name }}"
+        groups: "{{ item.group }}"
+        ansible_ssh_user: "{{ item.user }}"
+        ansible_host: "{{ item.ip }}"
+        ansible_ssh_private_key_file: "{{ ansible_user_dir }}/.ssh/id_cifw"
+        ansible_ssh_extra_args: '-o StrictHostKeyChecking=no'
+      loop: "{{ _bmo_provisioned_hosts }}"
+
+    - name: Wait for the instance to boot
+      delegate_to: "{{ item.name }}"
+      ansible.builtin.wait_for_connection:
+        sleep: 2
+        timeout: 600
+      loop: "{{ _bmo_provisioned_hosts }}"
+
+- name: Run the Nvidia phase 1 role
+  hosts: edpms
+  tasks:
+    - name: Run phase1
+      ansible.builtin.import_role:
+        name: edpm_nvidia_mdev_prepare
+        # As a reminder, at the end of phase1, the compute will reboot
+        tasks_from: phase1
--- a/roles/edpm_nvidia_mdev_prepare/README.md
+++ b/roles/edpm_nvidia_mdev_prepare/README.md
@@ -1,0 +1,23 @@
+# edpm_nvidia_mdev_prepare
+
+This Ansible role prepares an EDPM node by configuring it for nvidia mediated
+devices usage by installing from an external location a RPM package containing
+the nvidia driver.
+
+## Privilege escalation
+`become` will be set to True during the phase1 task in order to blacklist
+the nouveau driver, regenerate the initramfs and then install the nvidia RPM
+package. During phase2, we'll also create system files.
+
+## Parameters
+
+* `cifmw_edpm_nvidia_mdev_prepare_disable_nouveau`: (boolean) Whether to disable the `nouveau` kernel driver. Default: `true`.
+
+* `cifmw_edpm_nvidia_mdev_prepare_driver_url`: (string) The location of the proprietary nvidia driver RPM package so that we can install it. Can be any URI supported by DNF.
+
+* `cifmw_edpm_nvidia_mdev_prepare_package_name`: (string) The installed package name, which can't be inferred from the package name, usually set as `NVIDIA-vGPU-rhel`.
+
+* `cifmw_edpm_nvidia_mdev_prepare_sriov_devices`: (list) List of PCI addresses corresponding to the nvidia physical SR-IOV GPUs that require virtual functions creation. Leave it defaulted to ALL if you want to enable SR-IOV VFs for all your nvidia GPUs.
+
+# Should we start the service for automatically creating the VFs ?
+* `cifmw_edpm_nvidia_mdev_prepare_sriov_manage_start`: (boolean) Whether you want the virtual functions to be created upon reboot. Default: `true`.
--- a/roles/edpm_nvidia_mdev_prepare/handlers/main.yml
+++ b/roles/edpm_nvidia_mdev_prepare/handlers/main.yml
@@ -1,0 +1,15 @@
+---
+# Copyright Red Hat, Inc.
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
--- a/roles/edpm_nvidia_mdev_prepare/tasks/cleanup.yml
+++ b/roles/edpm_nvidia_mdev_prepare/tasks/cleanup.yml
@@ -1,0 +1,19 @@
+---
+# Copyright Red Hat, Inc.
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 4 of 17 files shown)


sbauza:
recheck

sbauza:
recheck

slagle:
> How would customers do this? Would they not use a custom service interface, or the bootstrap command interface so that this type of package install is handled using the products interfaces?
> 
> Here is an example custom service for repo setup: https://github.com/openstack-k8s-operators/install_yamls/blob/main/devsetup/edpm/services/dataplane_v1beta1_openstackdataplaneservice_reposetup.yaml
> 
> At first glance, I am very much against this change. We should use the products features to do this kind of thing.
> 
> @slagle - wdyt?

I think we'd expect customers to use the product interfaces for this type of thing. If they can't or need to run ad-hoc ansible for some reason, we want to understand that use case to see if it's something we want to add to the product.

Looking at some of the tasks (such as "Regenerate initramfs" and "Create a systemd unit file that will enable SRIOV VFs"), aren't these things that a customer would need to do as well when installing this driver?

Do we intend to manually document these steps, or do we intend to provide supported ansible content that customers can use? If it's the ansible content, then that should be in edpm-ansible, with an enabling service in openstack-operator.

However, if we intend to only support the driver itself, and not any of the installation steps, then I suppose  ci-framework can use ad-hoc methods to put it in place. That does seem more error prone though.

hjensas:
@sbauza @jamepark4 - would it be possible to link the doc's with the manual instructions that you are referring too?

My opinion here is, if our customers cannot achieve this by using the [documented](https://docs.redhat.com/en/documentation/red_hat_openstack_services_on_openshift/18.0/html/customizing_the_red_hat_openstack_services_on_openshift_deployment/assembly_customizing-the-data-plane#con_data-plane-services_custom_dataplane)[1] interface for a customizing the dataplane then we need to improve those interfaces.

I understand that this is an area of - every customer/env is different - so shipping the exact code does not make sense, but still customers will have to do it. I fail to understand why we choose to directly interact via Ansible with the EDPM nodes in ci-framework, instead of utilizing the opportunity to ensure the products interfaces for custom dataplane services can be used.


[1] https://docs.redhat.com/en/documentation/red_hat_openstack_services_on_openshift/18.0/html/customizing_the_red_hat_openstack_services_on_openshift_deployment/assembly_customizing-the-data-plane#con_data-plane-services_custom_dataplane

SeanMooney:
> @sbauza @jamepark4 - would it be possible to link the doc's with the manual instructions that you are referring too?
> 
> My opinion here is, if our customers cannot achieve this by using the [documented](https://docs.redhat.com/en/documentation/red_hat_openstack_services_on_openshift/18.0/html/customizing_the_red_hat_openstack_services_on_openshift_deployment/assembly_customizing-the-data-plane#con_data-plane-services_custom_dataplane)[1] interface for a customizing the dataplane then we need to improve those interfaces.
> 
> I understand that this is an area of - every customer/env is different - so shipping the exact code does not make sense, but still customers will have to do it. I fail to understand why we choose to directly interact via Ansible with the EDPM nodes in ci-framework, instead of utilizing the opportunity to ensure the products interfaces for custom dataplane services can be used.
> 
> [1] https://docs.redhat.com/en/documentation/red_hat_openstack_services_on_openshift/18.0/html/customizing_the_red_hat_openstack_services_on_openshift_deployment/assembly_customizing-the-data-plane#con_data-plane-services_custom_dataplane

we are doing this in the ci framework repo because we cannot support automated installation in production. we have discuss this in the past but the there are business and leagaage reasons why we cant include the driver or automate the download.

As with ceph the customistaion fo the EDPM node to install the nvidia drivers will be left to the customer to download form the nvida portal themselves and install.

it was an intentional decision  not to use the ability to define custom dataplane service to do this, it is possible to do but not something redhat can support in a customer env.

in past releases we have worked around the installation of the driver in ci by building a overcloud image to use for GPU testing with the content pre installed.
that is not a workflow we are really supporting with the new installer which is why this is being done as a post host porvisoing role.

hjensas:
> > @sbauza @jamepark4 - would it be possible to link the doc's with the manual instructions that you are referring too?
> > My opinion here is, if our customers cannot achieve this by using the [documented](https://docs.redhat.com/en/documentation/red_hat_openstack_services_on_openshift/18.0/html/customizing_the_red_hat_openstack_services_on_openshift_deployment/assembly_customizing-the-data-plane#con_data-plane-services_custom_dataplane)[1] interface for a customizing the dataplane then we need to improve those interfaces.
> > I understand that this is an area of - every customer/env is different - so shipping the exact code does not make sense, but still customers will have to do it. I fail to understand why we choose to directly interact via Ansible with the EDPM nodes in ci-framework, instead of utilizing the opportunity to ensure the products interfaces for custom dataplane services can be used.
> > [1] https://docs.redhat.com/en/documentation/red_hat_openstack_services_on_openshift/18.0/html/customizing_the_red_hat_openstack_services_on_openshift_deployment/assembly_customizing-the-data-plane#con_data-plane-services_custom_dataplane
> 
> we are doing this in the ci framework repo because we cannot support automated installation in production. we have discuss this in the past but the there are business and leagaage reasons why we cant include the driver or automate the download.
> 
> As with ceph the customistaion fo the EDPM node to install the nvidia drivers will be left to the customer to download form the nvida portal themselves and install.
> 

I understand that.

> it was an intentional decision not to use the ability to define custom dataplane service to do this, it is possible to do but not something redhat can support in a customer env.
> 

I am not going to continue the argument. :) 
But the fact that there is choice made here to not utilize a custom dataplane service makes me wonder if the custom service interface not being hard to use, or if ci-framework is lacking something to make it easy to add a custom dataplane service.

hjensas (commented):
/lgtm

SeanMooney:
a custom dataplen service was not used because we did not want to build a custom ansible runner image and we did not want to put the playbook inline.

we also wanted to follow the patten used for ceph and the existing hci_prepare role.

galaxy.yml
```diff
@@ -8,7 +8,7 @@ namespace: cifmw
 name: general
 
 # The version of the collection. Must be compatible with semantic versioning
-version: 1.0.0
+version: 1.0.0+15f75128
```
pablintino, line 11: Please, restore this file to the original one, with 1.0.0 as version.

playbooks/nvidia-mdev-phase1.yml
pablintino, line 1: Is this a hook? If so, maybe we should move it to the hooks dir along with the phase2 counterpart.

roles/edpm_nvidia_mdev_prepare/README.md
pablintino, line 1: The readme is empty and this, in general is a hard stop for us, please, fill the readme.

roles/edpm_nvidia_mdev_prepare/tasks/phase1.yml
```diff
@@ -0,0 +1,87 @@
...
+      options nouveau modeset=0
+    force: false
+  when:
+    - cifmw_edpm_nvidia_mdev_prepare_disable_nouveau | bool
+  register: _blacklist_nouveau
+
+- name: Make sure that we defined the driver URL
+  ansible.builtin.assert:
+    that:
+      - cifmw_edpm_nvidia_mdev_prepare_driver_url is defined
```
pablintino, line 33: There's a default value in the role. It will be always defined.
I suggest you to remove the default value (or comment out it in the defaults yaml).

roles/edpm_nvidia_mdev_prepare/tasks/phase1.yml
```diff
@@ -0,0 +1,87 @@
...
+  ansible.builtin.package_facts:
+    manager: auto
+
+- name: Install nvidia driver RPM either from path or URL
+  become: true
+  ansible.builtin.dnf:
+    name: "{{ cifmw_edpm_nvidia_mdev_prepare_driver_url }}"
+    state: present
+    disable_gpg_check: true
+  when: cifmw_edpm_nvidia_mdev_prepare_package_name not in ansible_facts.packages
```
pablintino, line 47: Strange condition, so, if the package is already installed we do nothing with the URL. In that case, why should we enforce the user to provide the URL?

roles/edpm_nvidia_mdev_prepare/tasks/phase1.yml
```diff
@@ -0,0 +1,87 @@
...
+      become: true
+      ansible.builtin.file:
+        path: /var/lib/openstack/reboot_required/nvidia_mdev_reboot
+        state: touch
+        mode: "0600"
+    # Since the EDPM role isn't installed, we can't call it
+    # Removing it for now
+    # - name: Call edpm_reboot role
+      # ansible.builtin.include_role:
+      #   name: edpm_reboot
```
pablintino, line 87: This will never happen. Why should we keep this commented code?

galaxy.yml
sbauza: Ack

playbooks/nvidia-mdev-phase1.yml
sbauza: This playbook is intended to be executed in a downstream VA which is called by a job, hence the playbook. @jamepark4 do you think we could move it into the hooks dir ?

roles/edpm_nvidia_mdev_prepare/tasks/phase1.yml
sbauza: Excellent point.

roles/edpm_nvidia_mdev_prepare/tasks/phase1.yml
sbauza: We only want to install the package from an external source if the package is already present locally hence the conditional.

pablintino:
@sbauza Hi, I'm ok with your answers, let me know when you have the patch ready with the content you agreed to change and I'll give my approval.

roles/edpm_nvidia_mdev_prepare/tasks/phase1.yml
jamepark4: If the edpm role is never installed at this point and calling reboot-os as a dataplaneservice is currently blocked, can we replace this with something like `ansible.builtin.reboot` for the interim?

jamepark4: Another alternative could be using bmh to reboot the host via idrac e.g. oc `annotate bmh -n openshift-machine-api compute-0 reboot.metal3.io='{"mode": "hard"}'`

sbauza: I just removed the commented code, as I think it leaves the role user to make sure the EDPM node needs to be rebooted when they want.

sbauza (commented):
changes are made based on your comments

playbooks/nvidia-mdev-phase1.yml
```diff
@@ -0,0 +1,73 @@
...
+  hosts: "{{ cifmw_target_hook_host | default('localhost') }}"
+  gather_facts: false
+  tasks:
+    - name: Fetch OSP BMO nodesets
+      environment:
+        KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+        PATH: "{{ cifmw_path }}"
+      ansible.builtin.command:
+        cmd: >-
+          oc get OpenStackBaremetalSet -n "{{ namespace|default('openstack') }}"  -o yaml
```
lewisdenny, line 28:
```suggestion
          oc get OpenStackBaremetalSet -n "{{ namespace|default('openstack') }}" -o yaml
```

roles/edpm_nvidia_mdev_prepare/tasks/phase1.yml
```diff
@@ -0,0 +1,82 @@
...
+    mode: "0644"
+    content: |-
+      blacklist nouveau
+      options nouveau modeset=0
+    force: false
+  when:
+    - cifmw_edpm_nvidia_mdev_prepare_disable_nouveau | bool
+  register: _blacklist_nouveau
+
+- name: Make sure that we defined the driver URL
```
lewisdenny, line 30: I would move this task to be the first that runs(above "Blacklist nouveau"), we should do our asserts at the beginning of the taskfile

roles/edpm_nvidia_mdev_prepare/tasks/phase1.yml
```diff
@@ -0,0 +1,82 @@
...
+
+- name: Enforce a reboot to ensure that we have the driver loaded
+  block:
+    - name: Create directory required by edpm-reboot role
+      become: true
+      ansible.builtin.file:
+        path: /var/lib/openstack/reboot_required/
+        state: directory
+        mode: "0755"
+    - name: Create required file to enforce a reboot
```
lewisdenny, line 77: Please add a newline between these two tasks

roles/edpm_nvidia_mdev_prepare/files/.gitkeep
lewisdenny, line 1: Please delete

roles/edpm_nvidia_mdev_prepare/handlers/main.yml
lewisdenny, line 1: Please delete

roles/edpm_nvidia_mdev_prepare/tasks/cleanup.yml
lewisdenny, line 1: Please delete

roles/edpm_nvidia_mdev_prepare/tasks/main.yml
lewisdenny, line 1: I would either delete this file or add a single debug task that explains this role should not be ran standalone.

roles/edpm_nvidia_mdev_prepare/vars/main.yml
lewisdenny, line 1: Please delete

lewisdenny (changes requested):
Really nice work on this role, nice and clean, thank you for the molecule test and nice README <3

PR #2642, opened 2025-01-10
compress text artifacts

morucci:
This change proposes to compress text files where size above 1MB.

Toolings/libs/browsers are able to deflate on the fly, thus for fair/substainable use of storage better to compress.

Some investigation shown that this simple step can save hundreds of MB of data for a build. In the case of a CI artifacts storage system and network bandwidth usage that non negligible.

### DIFF
--- a/ci/playbooks/collect-logs.yml
+++ b/ci/playbooks/collect-logs.yml
@@ -138,6 +138,19 @@
             src: "{{ ansible_user_dir }}/{{ zuul.project.src_dir }}/docs_build"
             dest: "{{ ansible_user_dir }}/zuul-output/logs/docs_build"
 
+        - name: Compress artifacts
+          become: true
+          ansible.builtin.shell:
+            chdir: "{{ ansible_user_dir }}/zuul-output/logs/"
+            cmd: |
+              echo "Before compression:"
+              du -sh
+              echo "Start compression"
+              echo "Compressing text files upper to 1MB ..."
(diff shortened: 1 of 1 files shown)


frenzyfriday (commented):
sorry, which folder is the PR compressing for example from https://logserver.rdoproject.org/59e/rdoproject.org/59e086b87aaf4d24a85df8da5b69a283/ ?

frenzyfriday:
Reopening to optimize zuul logs storage space

frenzyfriday (approved):

morucci:
> sorry, which folder is the PR compressing for example from https://logserver.rdoproject.org/59e/rdoproject.org/59e086b87aaf4d24a85df8da5b69a283/ ?

That's not a specific folder but text files larger than 1MB that are compressed. For instance "packagemanifests" in  https://logserver.rdoproject.org/93a/rdoproject.org/93aefc2071b14b5aa159b571f9d70d6b/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/ has been compressed (from 140MB to 17MB).

morucci:
See stdout_lines in https://softwarefactory-project.io/zuul/t/rdoproject.org/build/93aefc2071b14b5aa159b571f9d70d6b/console#11/0/19/controller where the log shows that we went from 350MB of logs to 190MB.

dasm:
@morucci are you happy with the change? Can we make it `ready for review`?

PR #2643, opened 2025-01-13
Ensure EPEL is reinstalled

pablintino:
Some roles like repo_setup may have altered the state of /etc/yum.repos.d/ without removing the EPEL rpm leading to yum/dnf to think the repos are already installed.
Enforcing its deletion ensures the rpm always deploys the repository files.


frenzyfriday (approved):

evallesp:
LGTM

evallesp (approved):

PR #2644, opened 2025-01-14
Make requests_kerberos optional if endpoint is not secured

pablintino:
If the url to fetch using uri_request is not secured (it does not return 401/403) make requests_kerberos optional.


evallesp (approved):
LGTM

PR #2647, opened 2025-01-14
[repo_setup] Use a role scoped copy of url_request

pablintino:
Since the role seems to be used from zuul without a nested playbook this role needs to have all of its dependencies self contained and not depending on global modules. To avoid copy/pasting the content of the module a symlink is used.
This change is required for zuul to call this role from its executor.


frenzyfriday (approved):

evallesp (approved):
lgtm

PR #2651, opened 2025-01-16
Remove ctlplane_vip from undercloud control plane bridge

ciecierski:


ciecierski:
recheck

frenzyfriday (approved):
I do not have the context behind why ctlplane_vip is removed but lgtm as per the commit msg + sanity check (it is not being used anywhere else)
/lgtm

hjensas (changes requested):
I do not agree with this change. The undercloud should have two VIPs.
The issue here is conflicting addresses used in openstack-k8s-operators/data-plane-adoption HCI scenarion.

[1] Is the undercloud's VIP on the ctlplane network.
[2] Is the overcloud's VIP on the ctlplane network.

These addresses cannot be the same, the undercloud VIP is hosted on the undercloud node. The overcloud VIP is controlled by pacemaker cluster on the overcloud nodes.

In the uni01alpha scenario in data-plane-adoption you will see that undercloud ctlplane VIP is 192.168.122.101 and overcloud VIP is 192.168.122.99 - to avoid the overlap.

The VIP's on the undercloud is controller by undercloud.conf options:
`undercloud_public_host`, `undercloud_admin_host`. Looking at [3] it seems these are set to `nthhost(122)` and `nthhost(123)`  - which does not match what we define as VIP in data-plane-adoption scenario.

[1] https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/scenarios/hci.yaml#L24
[2] https://github.com/openstack-k8s-operators/data-plane-adoption/blob/main/scenarios/hci/vips_data.yaml#L20
[3] https://github.com/openstack-k8s-operators/ci-framework/blob/6295c4e3f419db89bbccef61eddebecfaaf37161/roles/adoption_osp_deploy/tasks/prepare_undercloud.yml#L137C1-L142C69

PR #2548, opened 2024-11-18
Adoption osp deploy/uni01alpha

hjensas:
These changes are required for https://github.com/openstack-k8s-operators/data-plane-adoption/pull/706 to deploy.

PR #2653, opened 2025-01-16
Fix adoption_vars, only one defaultRoute

hjensas:
With [PR#2548](https://github.com/openstack-k8s-operators/ci-framework/pull/2548) support was added for multiple edpm nodesets, these may not have the same networks attached - i.e Compute and Networkere nodes have different network needs. When introducing this support the `networks` was made dynamic, and the `defaultRoute` was set for every network.

Only one network can have the `defaultRoute`, this caused an error when deploying the openstackdataplanenodeset.

Restore the behaviour prior to https://github.com/openstack-k8s-operators/ci-framework/pull/2548 - which would only set `defaultRoute` for the `ctlplane` network.

Also, filter the cifmw_libvirt_manager_pub_net when generating networks
for edpm_nodes. cifmw_libvirt_manager_pub_net (typically named ocpbm) is
a ci-framework "management" network and should not be included in the
EDPM nodeset configs.

Jira: [OSPRH-13013](https://issues.redhat.com//browse/OSPRH-13013)

### DIFF
--- a/roles/adoption_osp_deploy/defaults/main.yml
+++ b/roles/adoption_osp_deploy/defaults/main.yml
@@ -25,3 +25,6 @@ cifmw_adoption_osp_deploy_repos:
   - openstack-17.1-for-rhel-9-x86_64-rpms
   - fast-datapath-for-rhel-9-x86_64-rpms
   - rhceph-7-tools-for-rhel-9-x86_64-rpms
+
+cifmw_adoption_osp_deploy_adoption_vars_exclude_nets:
+  - "{{ cifmw_libvirt_manager_pub_net | default('ocpbm') }}"
--- a/roles/adoption_osp_deploy/templates/adoption_vars.yaml.j2
+++ b/roles/adoption_osp_deploy/templates/adoption_vars.yaml.j2
@@ -29,11 +29,13 @@ edpm_nodes:
     ansible:
       ansibleHost: {{ node_nets.networks.ctlplane.ip_v4 }}
     networks:
-    {% for net in node_nets.networks.keys() %}
-      - defaultRoute: true
-        fixedIP: {{ node_nets.networks[net].ip_v4 }}
+    {% for net in node_nets.networks.keys() if net not in cifmw_adoption_osp_deploy_adoption_vars_exclude_nets %}
+      - fixedIP: {{ node_nets.networks[net].ip_v4 }}
         name: {{ net }}
         subnetName: subnet1
+{% if net == 'ctlplane' %}
+        defaultRoute: true
+{% endif %}
     {% endfor %}
     {% endfor %}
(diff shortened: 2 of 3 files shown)


roles/adoption_osp_deploy/templates/adoption_vars.yaml.j2
```diff
@@ -30,10 +30,12 @@ edpm_nodes:
       ansibleHost: {{ node_nets.networks.ctlplane.ip_v4 }}
     networks:
     {% for net in node_nets.networks.keys() %}
```
ciecierski, line 32: make sure ocpbm network is not added to networks

hjensas: PR updated, it will now filter on `cifmw_libvirt_manager_pub_net` - i.e ocpbm network.

hjensas:
recheck

roles/adoption_osp_deploy/templates/adoption_vars.yaml.j2
ciecierski: I got error: 'AnsibleUndefinedVariable: ''cifmw_libvirt_manager_pub_net'' is undefined.
cifmw_libvirt_manager_pub_net is not defined during deploy-osp-adoption.yml playbook run, but only during create-infra.yml

hjensas: oh, ok - I changed the approch. Added a default var in adoption_osp_deploy `cifmw_adoption_osp_deploy_adoption_vars_exclude_nets` with value:
` - "{{ cifmw_libvirt_manager_pub_net | default("ocpbm") }}`

I think ci-framework uses `public` as the default for `cifmw_libvirt_manager_pub_net` but I think is is more commonly defined as "ocpbm" ...

ciecierski: Thank you for the change. It works for me now.

ciecierski (approved):

roles/adoption_osp_deploy/defaults/main.yml
```diff
@@ -25,3 +25,6 @@ cifmw_adoption_osp_deploy_repos:
   - openstack-17.1-for-rhel-9-x86_64-rpms
   - fast-datapath-for-rhel-9-x86_64-rpms
   - rhceph-7-tools-for-rhel-9-x86_64-rpms
+
+cifmw_adoption_osp_deploy_adoption_vars_exclude_nets:
```
pablintino, line 29: An entry to the README to document it and understand its meaning would be nice.

hjensas: Thanks @pablintino, I have updated the README.

ciecierski (approved):

pablintino (approved):

PR #2655, opened 2025-01-16
Remove auto draft GH action

pablintino:
After GA the amount of PRs that this repo receives is not that high, and more importantly, almost all are legit ones that requires review, so usually we are removing the draft immediately or asking the user to do so.


frenzyfriday (approved):
/lgtm

evallesp (approved):
LGTM

PR #2659, opened 2025-01-16
Log overcloud installation to log

holser:
In case of failure it would be nice to have a separate log for overcloud deployment in order to be able to analyze it in case of failure. This patch saves overcloud deployment to the log and spot a few last lines to console.


bogdando (approved):
/lgtm

ciecierski (approved):

holser:
@pablintino without "| tee -a overcloud_install.log" all output is stdout only which is saved to ansible.log only. When job fails it's not enough to analyse it what went wrong during OSP installation.

PR #2660, opened 2025-01-16
Unset resources section in workflows

lpiwowar:
There is a bug in Tempest which prevents us to set a sane upper
limit to the amount of memory Tempest can consume [1]. Until the fix
reaches the upstream build of the openstack-tempest-all image we
should not be setting any resource limits to the tempest pods.

This commit is ensuring that we are unsetting the default values set
on the side of the test-operator in the workflow section of Tempest CR.

[1] https://bugs.launchpad.net/tempest/+bug/2088074


lpiwowar:
Duplicate of #2661

PR #2661, opened 2025-01-16
Unset resources section in workflows

lpiwowar:
There is a bug in Tempest which prevents us to set a sane upper limit to the amount of memory Tempest can consume [1]. Until the fix reaches the upstream build of the openstack-tempest-all image we should not be setting any resource limits to the tempest pods.

This commit is ensuring that we are unsetting the default values set on the side of the test-operator in the workflow section of Tempest CR.

[1] https://bugs.launchpad.net/tempest/+bug/2088074
Related-Issue: https://issues.redhat.com/browse/OSPCIX-630

### DIFF
--- a/roles/test_operator/tasks/tempest-tests.yml
+++ b/roles/test_operator/tasks/tempest-tests.yml
@@ -173,3 +173,28 @@
               test_operator_cr |
               combine({'spec': {'workflow': overriden_workflow}}, recursive=true)
           }}
+
+- name: Make sure resources are not set for worklfow step
+  when:
+    - not cifmw_test_operator_dry_run | bool
+    - stage_vars_dict.cifmw_test_operator_tempest_workflow | list | length > 0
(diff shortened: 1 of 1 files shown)


lpiwowar:
Testing here -> https://github.com/openstack-k8s-operators/test-operator/pull/273

lpiwowar:
Tested here -> https://github.com/openstack-k8s-operators/test-operator/pull/273

We can see that the `resources` section got unset even in `workflow` section -> https://logserver.rdoproject.org/73/273/fb7ec89d375795c972ada38b456aa3317248a600/github-check/podified-multinode-edpm-deployment-crc-test-operator/9868c05/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack/crs/tempests.test.openstack.org/tempest-tests-tempest.yaml

PR #2663, opened 2025-01-17
[test-operator] Move away from Jobs to Pods

lpiwowar:
With this PR [1] the test operator dropped the usage of OCP Jobs for
spawning of the test pods.

This change updates the test-operator role so that it works with the
new version of the test-operator that spawns test pods directly through
the OCP Pods object.

[1] https://github.com/openstack-k8s-operators/test-operator/pull/266


lpiwowar:
Ups

PR #2670, opened 2025-01-20
[DNM] Testing manual install plan.

sathlan:
Just testing the deployment of operator set to manual, to be able to
deploy old version..


roles/kustomize_deploy/tasks/authenticate_registry.yml
```diff
@@ -0,0 +1,71 @@
...
+
+- name: Update OpenShift secret with the new .dockerconfigjson
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  ansible.builtin.command: >-
+    oc set data secret/pull-secret -n openshift-config
+    --from-file=.dockerconfigjson=/home/zuul/authfile.txt
+
+- name: Read username and password and create Kubernetes secret
```
ciecierski, line 32: This looks fine if we want to provide interface in ci-framework, but for your awareness we have update_edpm_services_playbook hook in downstream  which you can use too.

PR #2674, opened 2025-01-21
[test_operator] Avoid overwriting previous changes in tempest workflows

eduolivares:
A couple of blocks at the end of the tempest-tests.yml tasks override the originally configured tempest workflows when certain conditions match.
Before this PR, the second block overrode the changes performed in the first block.

### DIFF
--- a/roles/test_operator/tasks/tempest-tests.yml
+++ b/roles/test_operator/tasks/tempest-tests.yml
@@ -173,6 +173,11 @@
               test_operator_cr |
               combine({'spec': {'workflow': overriden_workflow}}, recursive=true)
           }}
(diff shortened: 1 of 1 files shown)


lpiwowar (changes requested):
/lgtm

Thank you for the fix! Tested locally with and without the fix and with the current version when `overrides` is set in a workflow step it is not updated with the proxy_host_address config option needed by wntp.

lpiwowar (approved):

lpiwowar:
My previous request for a change was a mistake. The PR LGTM!

frenzyfriday (approved):
/approve

PR #2675, opened 2025-01-21
Fix after_kuttl commands so that the output is not truncated

cescgina:
The commands run after kuttl were simply redirecting the output and not
appending to the file, so if there were multiple commmands run, only the
output of the last one was kept. Instead, append to the file so the
output of all commands is visible.

Depends-On: https://github.com/openstack-k8s-operators/install_yamls/pull/999


cescgina:
recheck

raukadah:
/lgtm

thank you @cescgina for fixing it!

PR #2676, opened 2025-01-22
Lower the priority of component repo when gating repo is used

raukadah:
Currently the prirority of gating and component repo is 1.
If the package is in both repos. The package under test will get installed from any repo. It makes the code.eng or opendev gerrit review depends-on testing faulty as not sure from which repo package is picked up.
    
In order to fix that, when gating and componenet repo is used, lower the prirority of component repo so that package get pulled from gating repo.

Related-Jira: https://issues.redhat.com/browse/OSPRH-11948


evallesp (approved):
LGTM

frenzyfriday (approved):

pablintino (approved):
/approve
/lgtm

PR #2680, opened 2025-01-25
Adds gh action to check for DNM labels

frenzyfriday:
This patch should go together with https://github.com/openshift/release/pull/60916 (doesnt matter after or before)

This action checks if the PR has `do-not-merge/hold` or `do-not-merge/work-in-progress`. If yes it fails.
When this action is set as "required" in the repo settings this prevents accidental merge if these labels are set.

The contains() function does not support regexes or arrays - hence the use of "or"

### DIFF
--- a/.github/workflows/check-dnm-label.yml
+++ b/.github/workflows/check-dnm-label.yml
@@ -1,0 +1,23 @@
+---
+name: Check for DNM labels
+
+on:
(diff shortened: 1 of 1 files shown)


pablintino (approved):

lewisdenny (approved):
/lgtm

PR #2682, opened 2025-01-27
[DNM] Fix ipv6 condition

gthiemonge:


roles/libvirt_manager/tasks/create_networks.yml
```diff
@@ -130,11 +130,9 @@
     _name: "{{ item | replace('-', '_') }}"
     _no_prefix_name: "{{ _name | regex_replace('cifmw[-_]','') }}"
     _ipv6: >-
-      {{
-        (ansible_facts[_name].ipv6 is defined) |
-        ternary(ansible_facts[_name].ipv6 |
-                selectattr('scope', 'equalto', 'global') | first, {})
-      }}
+      {% if ansible_facts[_name].ipv6 is defined -%}
```
hjensas, line 133: What issue are you fixing?

There is no information in the commit message, no link to issue/jira ticket explaining what the problem is.
To me, it looks like you are doing the same thing using a different "style"? If there was a commit message explaining what/why I probably would not have needed to ask these questions. :)

gthiemonge: it's just a DNM patch that tries to fix an issue I have in my env.
It looks like the syntax: "var is defined| ternary(var, other_var)" is not correct because ternary always evaluates both parameter.
The patch works for me now, but I'm not sure it covers all the cases
I'll remove the DNM tag if I find a clear/better solution

lewisdenny: Shifting to draft status due to this comment @gthiemonge , please update the PR is anything changes :)

PR #2687, opened 2025-01-28
Remove rgw default zone and realm

fmount:
We don't test `RGW` `multisite` in `OpenStack`. From `Squid+` a default realm is not deployed anymore, resulting in `RGW` deployment failures. This patch removes both default realm and zone from the `RGW` spec.


silvacarloss:
/lgtm
Thank you Francesco

fmount:
recheck

fmount:
recheck

fmount:
Reopened here: https://github.com/openstack-k8s-operators/ci-framework/pull/2876

PR #2694, opened 2025-01-30
[openshift_setup] Add support for overriding dependent operators CatalogSource

lewisdenny:
This is needed to easily test pre-release versions of OCP. OpenStack
depends on a few operators, these operators don't easily expose
pre-release versions for installation. A hack like this patch allows a
user to create a new CatalogSource for the current GA version of OCP and
patch the dependent operators to install from it.

JIRA: https://issues.redhat.com/browse/OSPRH-13706

### DIFF
--- a/roles/kustomize_deploy/tasks/install_operators.yml
+++ b/roles/kustomize_deploy/tasks/install_operators.yml
@@ -186,6 +186,14 @@
           status: "True"
         wait_timeout: 300
 
+    - name: Create catalog source and switch dependent operators to consume it
+      when:
+        - cifmw_openshift_setup_enable_operator_catalog_override |
+          default(false) | bool
+      ansible.builtin.import_role:
+        name: openshift_setup
+        tasks_from: patch_dependent_operators_source.yml
+
     - name: Wait for controller-manager pods
       kubernetes.core.k8s_info:
         kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
@@ -198,6 +206,7 @@
           type: Ready
           status: "True"
         wait_timeout: 300
+
     - name: Wait for webhook-server pods
       kubernetes.core.k8s_info:
         kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
--- a/roles/openshift_setup/tasks/patch_dependent_operators_source.yml
+++ b/roles/openshift_setup/tasks/patch_dependent_operators_source.yml
@@ -1,0 +1,58 @@
+---
+# HACK: (Lewis) This taskfile is used to test pre-release OCP versions
+# by intalling the previous OCP release's catalog source for the
+# dependent operators to use and overriding their subscriptions
+
+- name: Write catalog source
+  kubernetes.core.k8s:
+    kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
+    state: present
+    definition:
+      apiVersion: operators.coreos.com/v1alpha1
+      kind: CatalogSource
+      metadata:
+        name: "{{ cifmw_openshift_setup_operator_override_catalog_name }}"
+        namespace: "{{ cifmw_openshift_setup_operator_override_catalog_namespace }}"
+      spec:
+        displayName: "{{ cifmw_openshift_setup_operator_override_catalog_name }}"
+        image: "{{ cifmw_openshift_setup_operator_override_catalog_image }}"
+        publisher: CI-Framework
+        sourceType: grpc
+        updateStrategy:
+          registryPoll:
+            interval: 10m
+
+- name: Patch metallb operator subscription to use custom CatalogSource
+  kubernetes.core.k8s:
+    state: patched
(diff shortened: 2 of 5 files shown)


roles/openshift_setup/tasks/patch_dependent_operators_source.yml
```diff
@@ -0,0 +1,58 @@
+---
+# HACK: (Lewis) This taskfile is used to test pre-release OCP versions
+# by intalling the previous OCP release's catalog source for the
+# dependent operators to use and overriding there subscriptions
```
lewisdenny, line 4:
```suggestion
# dependent operators to use and overriding their subscriptions
```

frenzyfriday:
It would be good to add the new variables in the readme for the openshift_setup role

frenzyfriday (approved):
Non blocking: It would be good to also add the new variables in the readme for the openshift_setup role

bshewale:
LGTM

roles/kustomize_deploy/tasks/install_operators.yml
```diff
@@ -186,6 +186,14 @@
...
         wait_timeout: 300
 
+    - name: Create catalog source and switch dependent operators to consume it
+      when:
+        - cifmw_openshift_setup_enable_operator_catalog_override |
+          default(false) | bool
+      ansible.builtin.import_role:
+        name: openshift_setup
+        tasks_from: patch_dependent_operators_source.yml
+
```
pablintino, line 196: Do we really need this call here? 
If we follow what's done in that role what I'd expect is just another `include_tasks` at the end with `cifmw_openshift_setup_enable_operator_catalog_override` as a conditional and the proper default in the role defaults file.
The role runs **always** and before this one so we should be safe.

pablintino (changes requested):

roles/kustomize_deploy/tasks/install_operators.yml
lewisdenny: I can't validate currently because downstream CI infra is having an outage but my understanding is when the openshift_setup role runs, the subscriptions do not yet exist. 

Yes I could create the CatalogSource when openshift_setup runs but I wanted to keep the logic together. 

I could also create the subscriptions rather than patching them and maybe OLM wouldn't complain but this is already a nasty hack that I'm trying to minimise the footprint. 

I can verify once downstream CI infra is back.

roles/openshift_setup/tasks/patch_dependent_operators_source.yml
```diff
@@ -0,0 +1,58 @@
...
+        image: "{{ cifmw_openshift_setup_operator_override_catalog_image }}"
+        publisher: CI-Framework
+        sourceType: grpc
+        updateStrategy:
+          registryPoll:
+            interval: 10m
+
+- name: Patch metallb operator subscription to use custom CatalogSource
+  kubernetes.core.k8s:
+    state: patched
```
evallesp, line 27: (non-blocking) question:
I see that patched state in kubernetes.core.k8s exists silently (do not raise an error) if the resource to be patched doesn't exist.

Would it make sense to define a failing message? Make it sense to continue the execution?

evallesp (commented):
LGTM, minor request before approving.

roles/openshift_setup/tasks/patch_dependent_operators_source.yml
evallesp: Even further, is it possible that resource to be missing?

lewisdenny: Yes that's correct, if either subscription are missing the patch will not be applied and will silently continue. 

I don't mind this behavior because if the operator and subscription are not present it's due to something changing in the OpenStack product and I don't want to fail on that.

roles/kustomize_deploy/tasks/install_operators.yml
```diff
@@ -186,6 +186,14 @@
           status: "True"
         wait_timeout: 300
 
+    - name: Create catalog source and switch dependent operators to consume it
+      when:
+        - cifmw_openshift_setup_enable_operator_catalog_override |
```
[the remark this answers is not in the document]
lewisdenny: Done <3

roles/kustomize_deploy/tasks/install_operators.yml
lewisdenny: Yes, this is correct ^ 

The openshift_setup role runs before the openstack operators install plain is applied meaning the subscriptions we need to patch don't exist.

pablintino: ok ok, reality check here, thanks for checking

pablintino (approved):

lewisdenny:
recheck

lewisdenny:
recheck

evallesp (approved):
LGTM

PR #2695, opened 2025-01-30
[test_operator][tobiko.conf] ubuntu section replaced with advanced_vm

eduolivares:
Section ubuntu from tobiko.conf file has been replaced with advanced_vm. Coming soon: ubuntu images will be replaced with fedora images.


eduolivares:
recheck

lewisdenny (changes requested):
Hey @eduolivares could you please provide some context for this change for people that are not familiar with Tobiko?

Is there a change in Tobiko you can link to for example.

PR #2699, opened 2025-01-31
Update build_openstack_packages to support downstream gitlab

raukadah:
This pr:
    * Since downstream gerrit repos is moved to gitlab. Zuul works with gitlab and it clones the gitlab repos in 
       job   workspace. Use the zuul clonned content instead of clonning from gerrit or gitlab. As zuul already checks 
       out the source code. Let's reuse it for dlrn purpose by updating packages.yml.
    * Drop branch creation task as it is no longer needed.
    * For RHOS-18, downstream distgit uses rhoso-18 not rhos. It also updates the project.ini to reflect the same.
    
Note: While running build_openstack_packages in containers using make run_ctx_molecule, we are hitting following error in rpmbuild.log in CI and in local environment.
```
DEBUG: unshare(67239936) failed, falling back to unshare(131072)
ERROR: Namespace unshare failed.
 ```
Based on this[ bug](https://bugzilla.redhat.com/show_bug.cgi?id=1336750#c3) , By passing `--cap-add=SYS_ADMIN` fixes the issue.
    
Signed-off-by: Chandan Kumar (raukadah) <raukadah@gmail.com>
    
Related-Jira: https://issues.redhat.com/browse/OSPRH-13660

### DIFF
--- a/roles/build_openstack_packages/tasks/run_dlrn.yml
+++ b/roles/build_openstack_packages/tasks/run_dlrn.yml
@@ -138,45 +138,29 @@
         refspec: "{{ _change.refspec }}"
         version: 'FETCH_HEAD'
 
-    - name: "Clone downstream Openstack {{ project_name_mapped.stdout }}" # noqa: name[template]
+    - name: "Update packages.yml to use zuul repo for {{ project_name_mapped.stdout }}" # noqa: name[template], command-instead-of-module
+      vars:
+        _repo_name: '{{ _change.host | split("://") | last }}/{{ _change.project }}'
+        _old_content: 'osp-patches: git@{{ _change.host | split("//") | last }}:{{ _change.project }}.git'
+        _new_content: 'osp-patches: file://{{ ansible_user_dir }}/{{ zuul.projects[_repo_name].src_dir }}'
       when:
         - cifmw_bop_openstack_project_path | length == 0
         - not repo_status.stat.exists
         - "'host' in _change"
         - "'redhat.com' in _change.host"
-      block:
-        - name: Clone the downstream project
-          ansible.builtin.git:
-            repo: '{{ _change.host }}/gerrit/{{ _change.project }}'
-            dest: '{{ cifmw_bop_build_repo_dir }}/DLRN/data/{{ project_name_mapped.stdout }}-tested'
-            refspec: "{{ _change.refspec }}"
-            version: 'FETCH_HEAD'
-
-        - name: Update the branch name for local DLRN clone # noqa: command-instead-of-module
-          ansible.builtin.command:
-            cmd: "git checkout -b {{ item }}"
-          args:
-            chdir: "{{ cifmw_bop_build_repo_dir }}/DLRN/data/{{ project_name_mapped.stdout }}-tested"
-          loop:
-            - "{{ cifmw_bop_osp_release }}-rhel-9-trunk"
-            - "{{ cifmw_bop_osp_release }}-trunk-patches"
-
-        - name: Update the project reference in packages.yml # noqa: command-instead-of-module
-          vars:
-            _old_content: 'osp-patches: ssh://{{ _change.host | split("//") | last }}:22/{{ _change.project }}'
-            _new_content: 'osp-patches: file://{{ cifmw_bop_build_repo_dir }}/DLRN/data/{{ project_name_mapped.stdout }}-tested'
-          ansible.builtin.command:
-            cmd: >-
-              sed -i -e 's|{{ _old_content }}|{{ _new_content }}|g' packages.yml
-          args:
-            chdir: "{{ cifmw_bop_build_repo_dir }}/DLRN/{{ cifmw_bop_rdoinfo_repo_name }}"
+      ansible.builtin.command:
+        cmd: >-
+          sed -i -e 's|{{ _old_content }}|{{ _new_content }}|g' packages.yml
+      args:
+        chdir: "{{ cifmw_bop_build_repo_dir }}/DLRN/{{ cifmw_bop_rdoinfo_repo_name }}"
 
     - name: Find the last commit from the clonned repo
       register: _commit
       vars:
+        _repo_name: '{{ _change.host | split("://") | last }}/{{ _change.project }}'
         _project_path: >-
           {%- if cifmw_bop_osp_release is defined -%}
-          {{ cifmw_bop_build_repo_dir }}/DLRN/data/{{ project_name_mapped.stdout }}-tested
+          {{ ansible_user_dir }}/{{ zuul.projects[_repo_name].src_dir }}
           {%- else -%}
           {{ cifmw_bop_build_repo_dir }}/DLRN/data/{{ project_name_mapped.stdout }}
           {%- endif -%}
--- a/roles/build_openstack_packages/templates/projects.ini.j2
+++ b/roles/build_openstack_packages/templates/projects.ini.j2
@@ -40,7 +40,8 @@ custom_preprocess={{ cifmw_bop_build_repo_dir }}/DLRN/scripts/set_nvr.sh,{{ cifm
 info_files=osp.yml
 versions_url=https://trunk.rdoproject.org/centos9-{{ cifmw_bop_openstack_release }}/current-podified/versions.csv
 downstream_source_git_branch={{ cifmw_bop_osp_release }}-trunk-patches
-downstream_distro_branch={{ cifmw_bop_osp_release }}{{ '.0' if '.' not in cifmw_bop_osp_release else '' }}-rhel-9-trunk
+# downstream_distro_branch needs to be rhoso not rhos.
+downstream_distro_branch={{ cifmw_bop_osp_release | replace('rhos-', 'rhoso-') }}{{ '.0' if '.' not in cifmw_bop_osp_release else '' }}-rhel-9-trunk
 downstream_tag=osp-{{ cifmw_bop_osp_release.split('-')[1] }}{{ '.0' if '.' not in cifmw_bop_osp_release else '' }}
 downstream_distgit_key=osp-distgit
 downstream_source_git_key=osp-patches
--- a/Makefile
+++ b/Makefile
@@ -167,6 +167,7 @@ run_ctx_molecule: ci_ctx ## Run molecule check in a container
 	podman run \
 		--rm \
 		--security-opt label=disable \
+		--cap-add=SYS_ADMIN \
 		-v ${PWD}:/opt/sources \
 		--user root \
 		-e MOLECULE_CONFIG=${MOLECULE_CONFIG} \

### CODE REVIEW

**Comment by lewisdenny** [comment]
Hi Chandan, could you please add a commit message to explain this change

**Review by lewisdenny — CHANGES_REQUESTED** [review]
Hi @raukadah, could you please add a commit message to explain this change

**Comment by raukadah** [comment]
> Hi @raukadah, could you please add a commit message to explain this change

Done!

File: roles/build_openstack_packages/tasks/run_dlrn.yml [hunk 1]
```diff
@@ -145,25 +145,15 @@
...
-            chdir: "{{ cifmw_bop_build_repo_dir }}/DLRN/data/{{ project_name_mapped.stdout }}-tested"
-          loop:
-            - "{{ cifmw_bop_osp_release }}-rhel-9-trunk"
-            - "{{ cifmw_bop_osp_release }}-trunk-patches"
+            version: "{{ cifmw_bop_osp_release }}-trunk-patches"
 
         - name: Update the project reference in packages.yml # noqa: command-instead-of-module
           vars:
-            _old_content: 'osp-patches: ssh://{{ _change.host | split("//") | last }}:22/{{ _change.project }}'
+            _old_content: 'osp-patches: git@{{ _change.host | split("//") | last }}:{{ _change.project }}.git'
```
SeanMooney, line 156: thecnically the git protocal is just ssh by another name but ya this makes sense

roles/build_openstack_packages/templates/projects.ini.j2
```diff
@@ -40,7 +40,8 @@ custom_preprocess={{ cifmw_bop_build_repo_dir }}/DLRN/scripts/set_nvr.sh,{{ cifm
 info_files=osp.yml
 versions_url=https://trunk.rdoproject.org/centos9-{{ cifmw_bop_openstack_release }}/current-podified/versions.csv
 downstream_source_git_branch={{ cifmw_bop_osp_release }}-trunk-patches
-downstream_distro_branch={{ cifmw_bop_osp_release }}{{ '.0' if '.' not in cifmw_bop_osp_release else '' }}-rhel-9-trunk
+# downstream_distro_branch needs to be rhoso not rhos.
+downstream_distro_branch={{ cifmw_bop_osp_release | replace('rhos', 'rhoso') }}{{ '.0' if '.' not in cifmw_bop_osp_release else '' }}-rhel-9-trunk
```
SeanMooney, line 44: {{ '.0' if '.' not in cifmw_bop_osp_release else '' }}

so that is transformign 18 to 18.0  but it it was already 18.0.4 it would not modify it
makes sense

SeanMooney (commented):
/lgtm

roles/build_openstack_packages/tasks/run_dlrn.yml
```diff
@@ -145,25 +145,15 @@
         - "'host' in _change"
         - "'redhat.com' in _change.host"
       block:
-        - name: Clone the downstream project
+        - name: Clone the downstream project from zuul clonned repo
```
pablintino, line 148: IMHO Tying this role even more to zuul is a strategic mistake

roles/build_openstack_packages/tasks/run_dlrn.yml
```diff
@@ -145,25 +145,15 @@
...
           ansible.builtin.git:
-            repo: '{{ _change.host }}/gerrit/{{ _change.project }}'
+            repo: '{{ ansible_user_dir }}/src/{{ _change.host | split("://") | last }}/{{ _change.project }}'
```
pablintino, line 150: If we are using zuul as input, why not pick the source from it's metadata? It's already a git repo, so no need to checkout.

roles/build_openstack_packages/tasks/run_dlrn.yml
```diff
@@ -145,25 +145,15 @@
...
-
-        - name: Update the branch name for local DLRN clone # noqa: command-instead-of-module
-          ansible.builtin.command:
-            cmd: "git checkout -b {{ item }}"
-          args:
-            chdir: "{{ cifmw_bop_build_repo_dir }}/DLRN/data/{{ project_name_mapped.stdout }}-tested"
-          loop:
-            - "{{ cifmw_bop_osp_release }}-rhel-9-trunk"
-            - "{{ cifmw_bop_osp_release }}-trunk-patches"
+            version: "{{ cifmw_bop_osp_release }}-trunk-patches"
```
pablintino, line 152: Hardcoding `-trunk-patches` is just asking for future maintenance burden.

pablintino (changes requested):

roles/build_openstack_packages/tasks/run_dlrn.yml
SeanMooney: i strongly disagree.

the fact that this role was ever cloning code is a huge red flag.

our ci is meant to depend on zuul providing repos with all relevant patches pre merged by the zuul executor.

the current logic for making depend on work after the fact is flawed and only exist because of the fact we do not use zuul to prepare all source repos.

i was originally going to go further and ask chandan to remove the clonign fo repos form external sources entirely and only use the zuul provided source code which is the upstream best practice for opendev zuul.

we make it a hard error by default if devstack ever clones a repo for example as it breaks that pardime.

part of the reason we write jobs to let zuul clone the relvent code is to ensure all jobs in any given build set are executing on the same code.

we do get a similar effect by using content providers but its still better to avoid the extra network load and ci time required to have each job clone form upstream/downstream repos during the build.

roles/build_openstack_packages/tasks/run_dlrn.yml
raukadah: Done!

roles/build_openstack_packages/tasks/run_dlrn.yml
raukadah: We construct zuul_change_list like this
```
2025-02-04 10:14:38.069014 | controller | TASK [build_openstack_packages : Print Zuul change list var=cifmw_bop_change_list] ***
2025-02-04 10:14:38.069019 | controller | Tuesday 04 February 2025  10:14:38 -0500 (0:00:00.120)       0:05:51.056 ******
2025-02-04 10:14:38.069030 | controller | ok: [localhost] =>
2025-02-04 10:14:38.162668 | controller |   cifmw_bop_change_list:
2025-02-04 10:14:38.162698 | controller |   - branch: rhos-18.0-trunk-patches
2025-02-04 10:14:38.162704 | controller |     change: '4'
2025-02-04 10:14:38.162709 | controller |     host: https://gitlab.cee.redhat.com
2025-02-04 10:14:38.162714 | controller |     project: eng/openstack/python-watcherclient
2025-02-04 10:14:38.162718 | controller |     refspec: refs/changes/4/4/a1224a5f2a947bd3d7216b59e74f03492aa145ec
```
Based on above data, we perform further operations like clonning the repo and building the package using DLRN.
Since for downstream we are going to rely on zuul clonned repo (in future, we may add the same process for upstream gerrit and github). Here, We need to construct the repo name so that we can get the source of zuul metadata. I tried to modify the code on same line. I hope address the above concern.

roles/build_openstack_packages/tasks/run_dlrn.yml
raukadah: I am agreeing with Sean here. If an role's user populate the repos just like zuul does and construct zuul like vars then the role will work flawlessly.

raukadah:
recheck unknown failure on  cifmw-molecule-build_openstack_packages

roles/build_openstack_packages/tasks/run_dlrn.yml
pablintino: Maybe my message was way too short, cause I'm not pushing to clone the repos by ourselves, I agree with both of you on that. I was just saying that, if the code is already checked out by an external tool (not the cifmw, zuul, for example) why we need to checkout? In this case, if the job is properly configured the repos should be already present in the instance, if a depends-on exists zuul already patched that checked out content, etc. With this checkout here we may loose the depends-on, for example. So, tl;dr: Why we do not just copy the content from the directory that zuul uses?

roles/build_openstack_packages/tasks/run_dlrn.yml
pablintino: I was referring to, why do we craft the src directory assuming that zuul clones using that format instead of using the `src_dir` property that zuul uses to expose where's the content cloned?

roles/build_openstack_packages/tasks/run_dlrn.yml
karelyatin: Just noticed this PR and marked mine to be closed in favor of this https://github.com/openstack-k8s-operators/ci-framework/pull/2719.
Agree we should rely on zuul(or ci-framework if repos are copied from local to controller vm by it) cloned repos whether it's opendev gerrit, github or gitlab. That would also handle if there are multiple PRs in depends-on on for same project.
If the repo do not already exist already then we attempt the fetch/clone, i.e the case when zuul(or ci-framework repo copy) is not used.

raukadah: Thank you @SeanMooney @pablintino @karelyatin for the input and explaining in detail. In the current pr, I have used zuul provided repos. It is working fine downstream in testproject.

In order to use zuul provided repos for upstream also, I will do a followup pr. Currently we need to merge this pr to unblock downstream

roles/build_openstack_packages/tasks/run_dlrn.yml
raukadah: @pablintino Thank you for explaining it. I will address this change in a follow up pr as In order to use zuul src_dir, I need to rewrite the `Print Zuul change list` task and so that it should work with follow up tasks.

roles/build_openstack_packages/tasks/run_dlrn.yml
raukadah: I chatted with @karelyatin , For right now, we need to keep supporting both the workflow. For downstream, I will keep existing workflow and add using zuul clonned repos.

karelyatin: Follow ups for upstream or downstream additional support fine for me if the current version is working.
/lgtm

roles/build_openstack_packages/templates/projects.ini.j2
karelyatin: the only issue here with replace, if cifmw_bop_osp_release set to rhose, replace will result into bad string. so we change rhos- -> rhoso- ? or just not touch it here and change wherever cifmw_bop_osp_release is set.

raukadah: if we check all the openstack project dist git. They have both branches rhos-18.0-trunk-patches and rhoso-18.0-trunk-patches. But for watcher , it has rhoso-18.0-trunk-patches. That's why I went via this.

karelyatin: but this parameter is for distgit branches which ends with rhel-9-trunk, rhos- is legacy, rhoso- is current, watcher must be created recently so have only the current one.
issue is with rhos-> rhoso replacement, "rhos-" -> "rhoso-" i.e including dash should handle the concern i raised above.

roles/build_openstack_packages/tasks/run_dlrn.yml
raukadah: @pablintino @karelyatin Hello, I just checked the build_openstack_packages role. we have `cifmw_bop_openstack_project_path` var to build openstack_packages from local clone. Here is the example for the same. https://github.com/openstack-k8s-operators/ci-framework/tree/main/roles/build_openstack_packages#run-the-role and related tasks: https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/build_openstack_packages/tasks/parse_and_build_pkgs.yml#L46

For zuul case, We are constructing the zuul change list and clonning all the changes again instead of using zuul provided one.. I think we can drop the clonning specific changes and use the zuul provided repos for all the things. What do you say?

roles/build_openstack_packages/templates/projects.ini.j2
tosky: @karelyatin where do you see that rhos- prefix ix legacy? IIRC it is not.

roles/build_openstack_packages/tasks/run_dlrn.yml
karelyatin: that var can be used for only single project build, while change list for multiple, so useful but not complete

raukadah:
recheck

raukadah:
recheck

lewisdenny (approved):

PR #2712, opened 2025-02-04
WIP - increase timeout for ssh authorized_keys

saneax:
seeing failure here on timeouts, however when debugging was able to ssh into and populate authorized_keys
https://paste.opendev.org/show/beM3Mai0MwTe4OOwLb7L/

### DIFF
--- a/roles/libvirt_manager/tasks/manage_vms.yml
+++ b/roles/libvirt_manager/tasks/manage_vms.yml
@@ -63,7 +63,7 @@
       set -o pipefail;
       cat ~/.ssh/authorized_keys |
       ssh -v core@{{ vm_con_name }} "cat >> ~/.ssh/authorized_keys"
-  retries: 5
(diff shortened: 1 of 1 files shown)


roles/libvirt_manager/tasks/manage_vms.yml
```diff
@@ -63,7 +63,7 @@
       set -o pipefail;
       cat ~/.ssh/authorized_keys |
       ssh -v core@{{ vm_con_name }} "cat >> ~/.ssh/authorized_keys"
-  retries: 5
+  retries: 20
```
lewisdenny, line 66: Any reason for picking 20 retries rather than something like 10?

lewisdenny (changes requested):

saneax:
the actual issue resolved itself and was due to missing basic ssh crypt libraries

PR #2714, opened 2025-02-05
[url_request] Use the system CA bundle before using the bundled one

pablintino:
requests by default uses the certifi CA bundle to validate SSL certificates. If requests was installed using the system package manager the package may point to the system CA bundle but if not it points to certifi that may not have all the CAs installed in the system.

This approach tries to use the system CA bundle if possible.

### DIFF
--- a/plugins/modules/url_request.py
+++ b/plugins/modules/url_request.py
@@ -109,14 +109,32 @@ except ImportError:
     python_requests_gssapi_installed = False
 
 
-def _get_auth_module(module, url, verify_ssl):
+def _get_verify(verify: bool):
+    # If SSL verification is disabled do nothing
+    if not verify:
+        return False
+    # If the REQUESTS_CA_BUNDLE env var is present
+    # let requests handle picking the CA bundle from the env
+    if "REQUESTS_CA_BUNDLE" in os.environ:
+        return True
+    # Before defaulting to the requests/certifi CA bundle try to
+    # use the system one if available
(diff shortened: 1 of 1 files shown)


plugins/modules/url_request.py
```diff
@@ -109,14 +109,32 @@
...
 
 
-def _get_auth_module(module, url, verify_ssl):
+def _get_verify(verify: bool):
+    # If SSL verification is disabled do nothing
+    if not verify:
+        return False
+    # If the RESQUEST_CA_BUNDLE env var is present
+    # let requests handle picking the CA bundle from the env
+    if "RESQUEST_CA_BUNDLE" in os.environ:
```
frenzyfriday, line 118: there may be a typo in the var name? `REQUESTS_CA_BUNDLE` ?

evallesp (commented):
LGT; but needs to handle @frenzyfriday comment.

plugins/modules/url_request.py
lewisdenny: same on line 116

pablintino:
Good catch guys, fixed! @frenzyfriday @evallesp

evallesp (approved):

frenzyfriday (approved):

lhh (approved):
This is a reasonable workaround and avoids every script from having to install the right version of requests _or_ set the REQUESTS_CA_BUNDLE variable, but yet, it still allows the use of that envvar.

PR #2717, opened 2025-02-06
Add "tobiko" namespace to be gathered by the openstack-must-gather

slawqo:
This namespace is used by Tobiko to run e.g. background ping POD (and other PODs with background tasks in the future) and it would be nice to have logs from those PODs gathered in the CI job's results.


eduolivares (approved):

pablintino (approved):

fmount (approved):
/lgtm

PR #2721, opened 2025-02-07
Add Ansible variable cifmw_openshift_setup_apply_marketplace_fix

danpawlik:
On some images (especially on legacy CRC images e.g. 2.30.0), there is no need to run the openshift-marketplace workaround which is removing all the pods from the namespace.
That step might be problematic on old deployments, due the ImagePullPolicy is set to "Always", so after removing the pods when the credentials are vanished, pods would be not recreated.


pablintino:
I'd say, let's accept this change, create and create a task to delete everything if nothing breaks.

frenzyfriday (approved):
Do we need to set it to true in some jobs (that needs it right now) as soon as this PR merges?

danpawlik:
@frenzyfriday I think it can be disabled now, no need to re-run similar command as it is done already in base job.

bshewale (approved):

PR #2727, opened 2025-02-12
Change the method of installing test-operator

kstrenkova:
This patch introduces a new way of installing test-operator. Before, we wanted test-operator role to install the test-operator, but now test-operator will be installed the same way as other operators - using openstack-operator.

Additionally, there is a need to ensure we can change which version of test-operator is used. This will be necessary for having a way to prevent big issues by changing the versions temporarily.


roles/env_op_images/tasks/main.yml
```diff
@@ -86,6 +86,21 @@
...
+      when:
+      - cifmw_env_op_test_operator_image is defined
+      ansible.builtin.set_fact:
+        cifmw_openstack_service_images_content: >-
+          {{
+            cifmw_openstack_service_images_content |
+            default({}) |
+            combine(
+              {
+                "RELATED_IMAGE_TEST_OPERATOR_MANAGER_IMAGE_URL" : {{ cifmw_env_op_test_operator_image }}
```
lpiwowar, line 99: **question:**  What about using the already existing `cifmw_test_operator_index` parameter?

roles/env_op_images/defaults/main.yml
```diff
@@ -21,3 +21,4 @@
 cifmw_env_op_images_dir: "{{ cifmw_basedir | default(ansible_user_dir ~ '/ci-framework-data') }}"
 cifmw_env_op_images_file: operator_images.yaml
 cifmw_env_op_images_dryrun: false
+cifmw_env_op_test_operator_image: ""
```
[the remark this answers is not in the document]
kstrenkova: This ties to my previous question of how the variables and defaults are used. If it is possible like that, I would love to put it there :smile:

kstrenkova:
I had a discussion with @lpiwowar. I will change the logic of the patch tomorrow, I am working on it.

roles/test_operator/tasks/main.yml
```diff
@@ -23,63 +23,29 @@
...
-          nodeSelector: "{{ cifmw_test_operator_node_selector | default(omit) }}"
-  when: not cifmw_test_operator_dry_run | bool
+    - name: Patch test-operator version in CSV
+      ansible.builtin.command:
+        cmd: >-
+          oc patch ClusterServiceVersion {{ (_csvs_out.stdout | from_json).items[0].metadata.name }}
+          --type='json'
+          -p='[{"op": "replace", "path": "/spec/install/spec/deployments/0/spec/template/spec/containers/0/env/{{ _image_index }}/value", "value": " {{ cifmw_test_operator_index }} "}]'
 
 - name: Wait until the test-operator csv is present
```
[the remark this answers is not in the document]
kstrenkova: Than you for the review! You are right there is no test-operator csv anymore. I have pushed changes that take openstack-operator into consideration. I have still not tested it locally as I seem to not make it work, but at least the logic i pushed for now.

lpiwowar:
Overall looks good to me, I have just two recommendation:

1. Test the PR with modified [test-operator image](https://github.com/openstack-k8s-operators/test-operator/blob/7e625ca596376ad7cc27dcc5c73752a8b1bd8b6c/.zuul.yaml#L13) in the test-operator repository. This is useful to verify that this PR works even when we request a specific test-operator image in the job definition.
2. I would personally try to use `kubernetest.core.k8s` and `kubernetest.core.k8s_json_patch` modules for some tasks you are modifying.

kstrenkova:
Thank you for the review! I have changed most of the tasks to use `kubernetest.core.k8s` module. The only one that I believe is not possible to translate and should stay in shell is getting the index of the image. I will add the PR with modified test-operator image soon to test the change. I tested it locally and it worked so I have high hopes for the patch working.

kstrenkova:
TODO for me is to remove cleanup from the test_operator role, as the new method doesn't create any resources that need cleaning. I would like to commit it as a second commit, so I will wait a bit until the first commit works as expected.

lpiwowar:
**issue (blocking):** 
I've noticed one thing in the [test PR](https://github.com/openstack-k8s-operators/test-operator/pull/280). Test-operator bundle is being passed there instead of the test-operator index. Unless there seems to be no other solution then the appropriate thing would probably be to rename `cifmw_test_operator_index` -> `cifmw_test_operator_bundle` and modify the usage of the original variable (`cifmw_test_operator_index`) both upstream and downstream.

You should consult this with somebody from the ciops team I believe they will have an opinion about this as well:).

But nice update! :tada:  It seems to be working ([here](https://logserver.rdoproject.org/80/280/c992a54cae3796135edeb6b40c70144638f70cee/github-check/podified-multinode-edpm-deployment-crc-test-operator/90aaeb0/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack-operators/pods/test-operator-controller-manager-7d6ccbf597-lvwjl/test-operator-controller-manager-7d6ccbf597-lvwjl-describe)).

kstrenkova:
I have tested this change in test-operator PR [https://github.com/openstack-k8s-operators/test-operator/pull/280/files] and the logs show that the correct image was used for test-operator-controller-manager here -> [https://logserver.rdoproject.org/80/280/00b096a722543328be52a9cd546e16d1e1035fad/github-check/podified-multinode-edpm-deployment-crc-test-operator/b89a06e/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather/namespaces/openstack-operators/pods/test-operator-controller-manager-78cbb5dbd7-6rjng/test-operator-controller-manager-78cbb5dbd7-6rjng-describe](url). I think this patch is ready for review,

kstrenkova:
recheck

roles/test_operator/tasks/main.yml
```diff
@@ -23,107 +23,99 @@
...
+      kubernetes.core.k8s_json_patch:
+        kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
+        api_key: "{{ cifmw_openshift_token | default(omit)}}"
+        context: "{{ cifmw_openshift_context | default(omit)}}"
+        kind: ClusterServiceVersion
+        api_version: operators.coreos.com/v1alpha1
+        namespace: "{{ cifmw_test_operator_controller_namespace }}"
+        name: "{{ openstack_operator_csv_name }}"
+        patch:
+          - path: "/spec/install/spec/deployments/0/spec/template/spec/containers/0/env/{{ image_index.stdout }}/value"
```
pablintino, line 75: Can't we use the OpenstackVersion CR for this?

kstrenkova: What do you mean exactly? :smile: The thought behind this is, that if we want to temporarily change test-operator version specifically, this is the backdoor for such change. But normally it would not be run.

kstrenkova:
I tried to get full green checks with branch up-to-date, but too there are too many changes being added. I hope it can also get merged with out-of-date base branch :sweat_smile:

lpiwowar (commented):
I leave the LGTM up to the ci-framework folks but from my point of view this looks good. I saw two PRs where this change was tested and both of them proved the change worked:
- https://github.com/openstack-k8s-operators/test-operator/pull/280 (tested on Feb 18th)
- https://github.com/openstack-k8s-operators/test-operator/pull/282 (tested on Feb 20th)

roles/test_operator/tasks/main.yml
pablintino: @kstrenkova In the OpenstackVersion the user can set the versions of the services to deploy so I was suggesting that maybe setting it there (or adding the support to do so) would be a better approach than directly patching the CSV.

lpiwowar: @pablintino can I mark this conversation as Resolved for now so that this PR can be merged?

```
A conversation must be resolved before this pull request can be merged. 
```

bshewale (approved):

PR #2729, opened 2025-02-13
Add fqdn to the list of host patterns in ssh config

katarimanojk:
Ceph orchestrator hosts are configured with fqdn names, ceph adoption tasks try to access the fqdn and fails with ssh issues.

Good to have the fqdn (hostname + .localdomain) in the ssh config to avoid ssh issues.

### DIFF
--- a/roles/libvirt_manager/tasks/manage_vms.yml
+++ b/roles/libvirt_manager/tasks/manage_vms.yml
@@ -26,6 +26,7 @@
         - "{{ vm }}"
         - "{{ _type_id }}"
         - "{{ _type_id }}.utility"
+        - "{{ _hostname }}.localdomain"
         - "{{ _type_id }}.{{ inventory_hostname }}"
         - "{{ _hostname }}"
         - "{{ vm_con_name }}"


fmount:
@lewisdenny @pablintino any side effect with this additional entry? I'm wondering if we want to conditionally add this line, or it doesn't hurt to always keep it.
The reason behind this I suppose comes from [1][2].

[1] https://github.com/python/cpython/blob/2fbc9861c3b1325095019b3df6f3b5b0a1235f1f/Lib/socket.py#L793-L817 
[2] https://github.com/ceph/ceph/blob/main/src/cephadm/cephadmlib/net_utils.py#L245

lewisdenny:
> @lewisdenny @pablintino any side effect with this additional entry? I'm wondering if we want to conditionally add this line, or it doesn't hurt to always keep it. The reason behind this I suppose comes from [1][2].
> 
> [1] https://github.com/python/cpython/blob/2fbc9861c3b1325095019b3df6f3b5b0a1235f1f/Lib/socket.py#L793-L817 [2] https://github.com/ceph/ceph/blob/main/src/cephadm/cephadmlib/net_utils.py#L245

I'm not aware of any impact but I would need to see some downstream tests ran with this patch before merging, a run of the ceph adoption job showing the patch fixes the issue and a run of a uni job to confirm it doesn't break anything.

katarimanojk:
@fmount @lewisdenny ceph adoption test project job successful results here: https://softwarefactory-project.io/zuul/t/rdoproject.org/build/f4577b29aa304132930e4172d14c1455

hjensas (approved):

katarimanojk:
recheck

fmount (approved):
/lgtm

fmount:
@katarimanojk you might need to rebase

pablintino (approved):

PR #2733, opened 2025-02-14
[bop]Donot use dlrn --local for downstream

raukadah:
When we use dlrn --local, then dlrn will use local git reporegardless of version.csv info.
But for downstream, we need to follow version.csv to build downstream packages properly. that's why this pr is removing --local from dlrn command.


raukadah:
recheck

lewisdenny (approved):
@evallesp could you please give this on a review :)

amoralej (approved):
lgtm

PR #2734, opened 2025-02-14
Remove --local option from dlrn command

amoralej:


frenzyfriday (commented):
Could you pls provide a bit more context behind the patch?

amoralej:
We finally followed a different approach, although this may make sense. I'm closing this PR and will reopen if needed.

PR #2736, opened 2025-02-18
[ci_gen_kustomize_values] Remove BGP templates

eduolivares:
Templates under the "bgp" directory implemented the first approach to test BGP on Podified setups, with all OCP nodes running on a single rack and EDPM nodes running on separate racks.
This configuration was replaced by bgp_dt01, which deploys OCP nodes and EDPM nodes splitted in 3 racks. The content of "bgp" became obsole and it is removed with this patch.


pablintino (approved):

lewisdenny (approved):
lgtm, thanks for updating the codeowners @eduolivares

PR #2738, opened 2025-02-19
Update COO sub to stable channel

vyzigold:
COO released v 1.0.0 and the development channel disappeared. We need to switch to the stable channel.

### DIFF
--- a/roles/openshift_obs/defaults/main.yml
+++ b/roles/openshift_obs/defaults/main.yml
@@ -24,7 +24,7 @@ cifmw_openshift_obs_definition:
     labels:
       operators.coreos.com/observability-operator.openshift-operators: ""
   spec:
-    channel: development
+    channel: stable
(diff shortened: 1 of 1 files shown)


vyzigold:
recheck test-operator double install

vyzigold:
this is failing on the test-operator double install as well as the 
```
The error was: 'list object' has no attribute 'trunk_id'. 'list object' has no attribute 'trunk_id'
```
which I've seen discussed on slack. Neither failure is connected to the PR, let's wait what comes out from the slack discussion.

vyzigold:
recheck I think a workaround should already be in place for the previous issue, so let's see.

vyzigold:
recheck `Error from server (NotFound): deployments.apps "nmstate-operator" not found`

lewisdenny (approved):

pablintino (approved):

PR #2739, opened 2025-02-19
[kustomize_deploy] Add name field in architecture steps

pablintino:
Till now, to add a kustomize patch users refered to the step they wanted to patch by a numeric ID or an ID autogenerated based on the output build path. To give users more control, this change allows users to add a name field in each step in the automation field making patching a bit easier.

### DIFF
--- a/roles/kustomize_deploy/tasks/execute_step.yml
+++ b/roles/kustomize_deploy/tasks/execute_step.yml
@@ -15,7 +15,15 @@
 - name: Assert all wait conditions are valid
   vars:
     _stage_name_id: "stage_{{ stage_id }}"
-    _stage_name: "stage_{{ stage.build_output | basename | splitext | first }}"
+    _stage_name: >-
+      {{
+        (
+          stage.name |
+          default(
+            'stage_' + (stage.build_output | basename | splitext | first)
+          )
+        ) | regex_replace('[^a-zA-Z\d_-]', '_')
+      }}
     _custom_conditions: >-
       {{
         cifmw_architecture_wait_condition[_stage_name] |
@@ -57,7 +65,15 @@
 - name: Group tasks under the same tags
   vars:
     _stage_name_id: "stage_{{ stage_id }}"
-    _stage_name: "stage_{{ stage.build_output | basename | splitext | first }}"
+    _stage_name: >-
+      {{
+        (
+          stage.name |
+          default(
+            'stage_' + (stage.build_output | basename | splitext | first)
+          )
+        ) | regex_replace('[^a-zA-Z\d_-]', '_')
+      }}
     _tag_name_id: "deploy_architecture_{{ _stage_name_id }}"
     _tag_name: "deploy_architecture_{{ _stage_name }}"
     _skip_tags: >-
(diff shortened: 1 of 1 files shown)


roles/kustomize_deploy/tasks/execute_step.yml
```diff
@@ -15,7 +15,15 @@
 - name: Assert all wait conditions are valid
   vars:
     _stage_name_id: "stage_{{ stage_id }}"
-    _stage_name: "stage_{{ stage.build_output | basename | splitext | first }}"
+    _stage_name: >-
```
hjensas, line 18: Would we not need to do the same in L60 below. `name: Group tasks under the same tags`

roles/kustomize_deploy/tasks/execute_step.yml
hjensas, line 1: Would it make sense to follow up with something for hooks as well? In task `- name: "Executing pre_stage hooks for {{ stage.path }}"` only `_stage_name_id` is honored.

roles/kustomize_deploy/tasks/execute_step.yml
```diff
@@ -15,7 +15,15 @@
...
+      {{
+        (
+          stage.name |
+          default(
+            'stage_' + (stage.build_output | basename | splitext | first)
```
hjensas, line 23: so, what this means is that we can - even without this change do:

Would do the same:
```
cifmw_architecture_user_kustomize:
  {{ stage.name }}
    service-values:
       foo: bar
```

Because, we get the value of 'name' key for a stage in this task below?

```
    - name: "Generate values.yaml for {{ stage.path }}"
      when:
        - _val.src_file is defined
      vars:
        _name: "{{ _val['name'] }}"
```

hjensas: oh, I think I get the `_val['name']` now.
In architecture we have:

```
      - path: examples/dt/uni01alpha/edpm
        wait_conditions:
          - >-
            oc -n openstack wait openstackdataplanenodeset
            openstack-edpm
            --for condition=SetupReady
            --timeout=10m
        values:
          - name: edpm-nodeset-values
            src_file: values.yaml
        build_output: edpm.yaml
```

We can patch using:
```
cifmw_architecture_user_kustomize:
  # (stage.build_output | basename | splitext | first)
  edpm:                                       
    edpm-nodeset-values:
      patchy: patch
```

?

roles/kustomize_deploy/tasks/execute_step.yml
pablintino: Good catch, fixed.

roles/kustomize_deploy/tasks/execute_step.yml
pablintino: It totally makes sense. I'll post an update on Monday to consider both IDs for hooks, I'll need some minor changes in the run_hook role to support passing two patterns for the hook vars.

roles/kustomize_deploy/tasks/execute_step.yml
pablintino: Oh, keep in mind `_val['name']` refers to the name of the `values`, not the step, in that case `edpm-deployment-values`. The name field I'm adding goes at the root level of the step, like this:
```yaml
      - path: examples/dt/uni01alpha/edpm
        name: step-name-edpm-nodeset-values # Like this
        wait_conditions:
          - >-
            oc -n openstack wait openstackdataplanenodeset
            openstack-edpm
            --for condition=SetupReady
            --timeout=10m
        values:
          - name: edpm-nodeset-values
            src_file: values.yaml
        build_output: edpm.yaml


```

lewisdenny (approved):

hjensas (approved):

bshewale (approved):

PR #2740, opened 2025-02-19
Multi-cell adoption support for osp-deploy plugin

bogdando:
Adapt templating vars for multi-cell layout, but having only default(cell1) supported yet.

Remove no longer used edpm_computes.

Assume a single default cell exists only (which becomes cell1 after adoption).

Change data formats to become compliant with multi-cell topology.

### DIFF
--- a/roles/adoption_osp_deploy/templates/adoption_vars.yaml.j2
+++ b/roles/adoption_osp_deploy/templates/adoption_vars.yaml.j2
@@ -1,13 +1,14 @@
 #jinja2: trim_blocks:True, lstrip_blocks:True
+{%+ if multi_cell|default(false) +%}
+source_mariadb_ip:
+  default: {{ _controller_1_internalapi_ip }}
+{%+ else +%}
 source_mariadb_ip: {{ _controller_1_internalapi_ip }}
+{%+ endif +%}
+
 source_ovndb_ip: {{ _controller_1_internalapi_ip }}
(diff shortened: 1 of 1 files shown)


bogdando:
This needs further work to align it with [rdo-jobs/playbooks/data_plane_adoption/deploy_tripleo_run_repo_tests.yaml](https://review.rdoproject.org/r/c/rdo-jobs/+/56115/4/playbooks/data_plane_adoption/deploy_tripleo_run_repo_tests.yaml#119).
To make that possible, we should split edpm_nodes into cells specific osdpnses and into [edpm_nodes_networker](https://review.rdoproject.org/r/c/rdo-jobs/+/54113) as the latter is not Compute cells aware and should belong to another non-cell specific osdpns @fyanac @karelyatin @cescgina FYI

bogdando:
> This needs further work to align it with [rdo-jobs/playbooks/data_plane_adoption/deploy_tripleo_run_repo_tests.yaml](https://review.rdoproject.org/r/c/rdo-jobs/+/56115/4/playbooks/data_plane_adoption/deploy_tripleo_run_repo_tests.yaml#119). To make that possible, we should split edpm_nodes into cells specific osdpnses and into [edpm_nodes_networker](https://review.rdoproject.org/r/c/rdo-jobs/+/54113) as the latter is not Compute cells aware and should belong to another non-cell specific osdpns @fyanac @karelyatin @cescgina FYI

It looks that https://github.com/openstack-k8s-operators/ci-framework/pull/2877 should unlbock this one
@hjensas thanks for fixing it!

PR #2741, opened 2025-02-19
[autoholds] Use the krb_requests module to fetch holds

pablintino:
Since the zuul API may be behind a kerberos secured gateway let's consume the endpoint with the module to allow automatic authentication if a kerberos token was already been issued.

JIRA: https://issues.redhat.com/browse/OSPRH-12508

### DIFF
--- a/ci/playbooks/multinode-autohold.yml
+++ b/ci/playbooks/multinode-autohold.yml
@@ -1,6 +1,6 @@
 ---
 - name: "Run ci/playbooks/multinode-autohold.yml"
-  hosts: "{{ cifmw_zuul_target_host | default('all') }}"
+  hosts: controller, primary
   gather_facts: true
   tasks:
     - name: Verify if "success" flag exists after successful tests execution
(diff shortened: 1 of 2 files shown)


pablintino:
recheck

pablintino:
recheck

lewisdenny:
```
openstack router create --external-gateway public zuul-ci-subnet-router-3782195d -f yaml
non-zero return code
ConflictException: 409: Client Error for url: https://network.rdo.mtl2.vexxhost.net/v2.0/routers, Quota exceeded for resources: ['router'].
```

owch

lewisdenny:
recheck

ci/playbooks/multinode-autohold.yml
```diff
@@ -1,6 +1,6 @@
 ---
 - name: "Run ci/playbooks/multinode-autohold.yml"
-  hosts: "{{ cifmw_zuul_target_host | default('all') }}"
+  hosts: controller, primary
```
lewisdenny, line 3: Could you help me understnad this change. 
AFAIU `cifmw_zuul_target_host` is only used for reproducer jobs and usually targets localhost which doesn't make sense so good to remove.

What is `primary`? This one doesn't make sense to me because this taskfile basically creates a skip file on the host if there is an autohold created for the job. 

But the cleanup script only runs on `Controller` so it's only checking there for the skip file.

lewisdenny (changes requested):

ci/playbooks/multinode-autohold.yml
lewisdenny: Forgot the link: https://review.rdoproject.org/r/plugins/gitiles/config/+/refs/heads/master/playbooks/crc/bootstraped-ci-network-cleanup.yml

pablintino: @lewisdenny Hi, the change is mainly because I observed here https://softwarefactory-project.io/zuul/t/rdoproject.org/build/3244b3583d2b41c9b1fe18d58a3aebe2/console that the playbook is running in all the hosts of the run, but obviously we want it to be run only in the controller node, as is the only one that may have the kerberos token. It has zero utility to run it in the compute and/or crc.
Why primary? Some adoption nodesets, specially in DS, use primary instead of controller to refer to the ansible controller, thus this addition. I discovered this when I created the pre-run that does the kinit.

pablintino:
recheck

pablintino:
recheck

ci/playbooks/multinode-autohold.yml
lewisdenny: Ah, that makes sense mate, annoying we have some divergence on the "controller" name.

lewisdenny (approved):

pablintino:
recheck

frenzyfriday (approved):

PR #2761, opened 2025-02-25
Change Ceph Dashboard port to 8443

fultonj:
Update edpm_ceph_hci_pre role since the default port for the dashboard was wrong as per the ceph docs.

https://docs.ceph.com/en/latest/mgr/dashboard/#host-name-and-port

If we're going to change this port in the product (see depends-on) then we should update how we test it in ci-framework.

Depends-On: https://github.com/openstack-k8s-operators/edpm-ansible/pull/888
Jira: https://issues.redhat.com/browse/OSPRH-14300


fmount (approved):
/lgtm

pablintino (approved):

PR #2765, opened 2025-02-27
Add job for testing watcher in the master pipeline

cescgina:
Since watcher-operator uses master and not antelope container, create
a job to run in the master pipeline to ensure new content does not break
watcher.


raukadah (approved):
/lgtm

lewisdenny (approved):
Thanks @cescgina <3

PR #2768, opened 2025-03-01
Limit ncclient version to avoid failures

rsafrono:
CI jobs are failing trying to deploy recent version (0.6.18) of ncclient


rsafrono:
Recent version of ncclient that was created today (0.6.19) does not cause failures during ci job run, so we can deploy safely without this MR. Closing

PR #2769, opened 2025-03-03
Use provider ifcfg for os-net-config

hjensas:
In the role adoption_osp_deploy use the `ifcfg` provider for os-net-config.

NOTE! This only fixes one of the issue described in Jira.

Jira: [OSPRH-14449](https://issues.redhat.com//browse/OSPRH-14449)


hjensas:
This is invalid, wrong repos used on test machine.  os-net-config-18.0.1-18.0.20241215115912.22b9a65.el9ost came from rhoso-18.0-for-rhel-9-x86_64-rpms

PR #2771, opened 2025-03-04
Add wait to prevent race condition

kstrenkova:
Some jobs show an error that could be appearing thanks to a race condition. This patch aims to prevent this case by first waiting for a pod to start before running the next task.

### DIFF
--- a/roles/test_operator/tasks/run-test-operator-job.yml
+++ b/roles/test_operator/tasks/run-test-operator-job.yml
@@ -111,6 +111,7 @@
         api_key: "{{ cifmw_openshift_token | default(omit)}}"
         context: "{{ cifmw_openshift_context | default(omit)}}"
         state: present
+        wait: true
         definition:
           apiVersion: v1
           kind: Pod


kstrenkova:
Had to rebase. Thank you @tosky for lgtm, I hope you can add it after the checks pass again.

roles/test_operator/tasks/run-test-operator-job.yml
```diff
@@ -111,6 +111,7 @@
         api_key: "{{ cifmw_openshift_token | default(omit)}}"
         context: "{{ cifmw_openshift_context | default(omit)}}"
         state: present
+        wait: true
```
evallesp, line 114: (blocking) WDYT about adding a wait_timeout value?

kstrenkova: Thank you for the review :smile: I would stay true to how it was done in other tasks. If you look at `Start tests` task it combines `state: present` and `wait: true` and it works without problems. If we add wait_timeout it will be really small anyways because the error this patch will fix is created by less than a second difference.
Does this reasoning sound okay to you or do you want to add the timeout due to other reasons?

evallesp (approved):

adrianfusco (approved):

PR #2776, opened 2025-03-05
Module crawl_n_mask to mask secrets in yaml files

amartyasinha:
This PR aims to fix https://issues.redhat.com/browse/OSPRH-14524

- crawl_n_mask is a new module which is being introduced in CIFMW.
- Module requires only two arguments,`path` (type: path) and `isdir` (type: bool, default: False).
- This module can process yaml file directly or it can even crawl and find yaml files within provided directory.
- Masking of remaining log files yet to be done as part of separate PR.
- Instead of reinventing the wheel, borrowed the mask script from [openstack-must-gather](https://github.com/openstack-k8s-operators/openstack-must-gather/tree/main/pyscripts) and modified according to our requirement.

PR tested: Secrets in[ yaml files](https://softwarefactory-project.io/zuul/t/rdoproject.org/build/650f887c360a4607acc2860598d68ed2/log/ci-framework-data/artifacts/manifests/openstack/input/kustomization.yaml) are getting masked through this PR.


scripts/crawl_n_mask.py
```diff
@@ -0,0 +1,264 @@
...
+            self.path.endswith((tuple(["yaml", "yml"])))
+            or os.path.basename(self.path).split(".")[0] in ALLOWED_YAML_FILES
+        ):
+            self._mask_yaml()
+        # elif self.path.endswith("log"):
+        #     self._maskLogFile()
+        return True
+
+    def _mask_yaml(self) -> None:
+        s = self._readYaml()
```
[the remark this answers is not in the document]
evallesp: (non-bloking) nit: I'd try to avoid one character variable names.

scripts/crawl_n_mask.py
```diff
@@ -0,0 +1,264 @@
...
+    # parse the provided options
+    OPTS = parse_opts(sys.argv)
+
+    if OPTS.dir is not None and os.path.exists(OPTS.dir):
+        # craw through the provided directly and then
+        # process eligible files individually
+        crawl(OPTS.dir)
+
+    if OPTS.path is not None and os.path.exists(OPTS.path):
+        SecretMask(OPTS.path).mask()
```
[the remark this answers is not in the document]
amartyasinha: > (non-blocking) question: Don't we need to check if this file is not in excluded_file_ext_regex ?

Yes, that's something I missed. I'll add it.

scripts/crawl_n_mask.py
```diff
@@ -0,0 +1,264 @@
...
+    )
+    opts = parser.parse_args(argv[1:])
+    return opts
+
+
+if __name__ == "__main__":
+    # parse the provided options
+    OPTS = parse_opts(sys.argv)
+
+    if OPTS.dir is not None and os.path.exists(OPTS.dir):
```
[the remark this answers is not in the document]
amartyasinha: It works for both -d and --dir. I've checked it.

scripts/crawl_n_mask.py
```diff
@@ -0,0 +1,264 @@
...
+        Read and Load the yaml file
+        for processing.
+        """
+        try:
+            assert self.path is not None
+            with open(self.path, "r") as f:
+                return yaml.safe_load(f)
+        except (FileNotFoundError, yaml.YAMLError) as e:
+            print(f"Error while reading YAML: {e}")
+            # sys.exit(-1)
```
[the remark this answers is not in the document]
amartyasinha: Earlier, I was planning to process all log files. But now, will only process yaml logs in this MR. So, it's no longer required and will remove it.

scripts/tests/test_crawl_n_mask.py
```diff
@@ -0,0 +1,68 @@
...
+        """
+        For each file present in the tests/samples we:
+        - Load the file by reading the yaml definition
+        - Process using the SecreMask module
+        - assert the content of the secret is
+          different
+        """
+        for root, _, files in os.walk(SAMPLE_DIR):
+            for f in files:
+                print("Processing file %s" % f)
```
evallesp, line 42: (blocking) suggestion: I'd rather remove this print. This would make unnecessary noise imho.

scripts/tests/test_crawl_n_mask.py
```diff
@@ -0,0 +1,68 @@
...
+        """
+        utility function to load a sample yaml file.
+        """
+        try:
+            assert path is not None
+            with open(path, "r") as f:
+                return list(yaml.safe_load_all(f))
+        except (FileNotFoundError, yaml.YAMLError) as e:
+            print(f"Error while reading YAML: {e}")
+            # sys.exit(-1)
```
evallesp, line 29: (blocking) suggestion: I'd remove this leftover.

scripts/tests/test_crawl_n_mask.py
```diff
@@ -0,0 +1,68 @@
...
+SAMPLE_DIR = "samples"
+
+
+class TestSecretMask(unittest.TestCase):
+    """
+    The class that implements basic tests for
+    SecretMask.
+    """
+
+    def _read_yaml_sample(self, path) -> Optional[Union[list, None]]:
```
evallesp, line 19: (blocking) suggestion: Here it's returning None, but in the caller method we don't check if it's list or None.
I'd suggest just returning List, and if something goes wrong, catching that from the error.
In any case, this is a test case, so I'd not expect to check if the file is there, cause if not, the test would fail in any case.

Could you check the part you're trying to cover, but you don't need to test the actual test?

So here I'd just return the list object without any try except clause or the possibility to return None.

scripts/tests/test_crawl_n_mask.py
```diff
@@ -0,0 +1,68 @@
...
+            for f in files:
+                print("Processing file %s" % f)
+                actual = self._read_yaml_sample(os.path.join(root, f))
+                expected = copy.deepcopy(actual)
+                # Mask secret by processing the data section
+                # of the yaml file we got
+                SecretMask(os.path.join(root, f))._process_list(expected)
+
+                """
+                files are named secret{1, 2, 3, ... N}: for these secrets
```
evallesp, line 50: (blocking) typo: Actually first one is type as sample1

scripts/crawl_n_mask.py
```diff
@@ -0,0 +1,275 @@
...
+    for root, _, files in os.walk(path, onerror=handle_error):
+        if any(excluded in root for excluded in EXCLUDED_DIRS):
+            continue
+
+        for f in files:
+            if re.search(excluded_file_ext_regex, f) is None:
+                SecretMask(os.path.join(root, f)).mask()
+
+
+def handle_error(e):
```
evallesp, line 233: (blocking) suggestion: Somehow we need to handle better the errors.

My suggestion is to groom all the try except blocks, avoid returning None if something went wrong and directly in the except, after logging the problem (or logging the problem here) sys exist 1. So ansible caller could check the return status by checking the return code.

Also, I'm thinking if it'd be good to move this as ansible module. So you install it and you can use it directly instead of ansible.comand: /bin/bash crawl_n_mask.py

This module would receive path and/or and/or folder.

scripts/tests/test_crawl_n_mask.py
```diff
@@ -0,0 +1,68 @@
...
+                we expect a change in their content because sensitive
+                data has been masked; the sample file named "nochange.yaml",
+                instead, is the one that does not contain any sensitive data,
+                hence no masking is applied and we expect the original data
+                content being the same as the processed one.
+                """
+                if "nochange" in f:
```
evallesp, line 57: I'd create two tests here, one when there's something to mask and another when there's no something.
Also I'm thinking the idea to test a bit more in detail for each of the tests in the script.

But first of all, let's check if we want to create an ansible module instead of a script.

evallesp (commented):
I'm thinking the idea of moving this to an ansible module like:  krb_request module. check the file commit history: https://github.com/openstack-k8s-operators/ci-framework/commits/main/plugins/modules/krb_request.py

scripts/crawl_n_mask.py
```diff
@@ -0,0 +1,264 @@
...
+excluded_file_ext_regex = r"(^.*(%s).*)" % "|".join(EXCLUDED_FILE_EXT)
+
+# regex of keys which will be checked against every key
+# as in yaml files, we have data in format <key> = <value>
+# if a key is sensitive, it will be found using this regex
+key_regex = r"(%s)\d*$" % "|".join(PROTECT_KEYS)
+regexes = [gen_regex, con_regex]
+
+
+class SecretMask:
```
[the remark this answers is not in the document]
amartyasinha: As of now, we are going to have a script handy. The initial idea was to run the script against the generated logs.

scripts/crawl_n_mask.py
```diff
@@ -0,0 +1,264 @@
...
+
+    def __init__(self, path: Optional[Any] = None) -> None:
```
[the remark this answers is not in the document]
amartyasinha: Done changes

scripts/crawl_n_mask.py
```diff
@@ -0,0 +1,264 @@
...
+        self._writeYaml(s)
+
+    def _apply_regex(self, value: str) -> str:
+        """
+        For each string value passed as argument, try
+        to match the pattern according to the provided
+        regexes and mask any potential sensitive info.
+        """
+        for pattern in regexes:
+            value = re.sub(pattern, r"\1{}".format(MASK_STR), value, flags=re.I)
```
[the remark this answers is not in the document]
amartyasinha: A large string might have multiple secrets. That's why it is checking for all regex patterns using for loop.

scripts/crawl_n_mask.py
```diff
@@ -0,0 +1,264 @@
...
+            elif isinstance(v, dict):
+                self._applyMask(v)
+
+    def _readYaml(self) -> Optional[Union[Dict[str, Any], list]]:
+        """
+        Read and Load the yaml file
+        for processing.
+        """
+        try:
+            assert self.path is not None
```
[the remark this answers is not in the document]
amartyasinha: yes, cleaned it up.

scripts/crawl_n_mask.py
amartyasinha: we have some yaml files in logs which are not yaml. If we do sys.exit(-1), then the script will stop when it find such files. Unless those files are not fixed, it is better to return None instead of exiting the script.

scripts/tests/test_crawl_n_mask.py
amartyasinha: I've borrowed the script and test cases from openstack-must-gather. The idea was to keep things simple. We can keep improving test cases and the script once we decide how we will run it over our logs dir.

scripts/tests/test_crawl_n_mask.py
amartyasinha: aah, my bad! renamed that

amartyasinha (commented):
Worked on the review of @evallesp

lewisdenny:
Hey @amartyasinha just wondering if you considered using a tool like https://github.com/gitleaks/gitleaks

It's already in use in Prow, implemented by the openshift-ci team. Maybe you know a reason we can't use this project like it won't detect the types of secrets we have or something.

amartyasinha:
> Hey @amartyasinha just wondering if you considered using a tool like https://github.com/gitleaks/gitleaks
> 
> It's already in use in Prow, implemented by the openshift-ci team. Maybe you know a reason we can't use this project like it won't detect the types of secrets we have or something.

@lewisdenny Thanks for suggesting gitleaks. Just checked it out and it's amazing. I scanned a log directory and it was able to find most of the secrets. We can modify their config to have our own secret detection rules (to cover the detection of remaining secrets).

We have two options now:
- First is to go with the basic Ansible module which will only scan yaml files.

- Another option is to look into gitleaks, create a config according to our requirement, and add a way to mask the secrets (that's something gitleaks doesn't provide).

What's your opinion on that?

lewisdenny:
> > Hey @amartyasinha just wondering if you considered using a tool like https://github.com/gitleaks/gitleaks
> > 
> > It's already in use in Prow, implemented by the openshift-ci team. Maybe you know a reason we can't use this project like it won't detect the types of secrets we have or something.
> 
> @lewisdenny Thanks for suggesting gitleaks. Just checked it out and it's amazing. I scanned a log directory and it was able to find most of the secrets. We can modify their config to have our own secret detection rules (to cover the detection of remaining secrets).
> 
> We have two options now:
> - First is to go with the basic Ansible module which will only scan yaml files.
> 
> - Another option is to look into gitleaks, create a config according to our requirement, and add a way to mask the secrets (that's something gitleaks doesn't provide).
> 
> What's your opinion on that?

My recommendation when it comes to security is to not roll your own. If gitleaks severs our purpose then it may be the safer route but let's hear from other maintainers as well.

amartyasinha:
recheck

plugins/modules/crawl_n_mask.py
```diff
@@ -0,0 +1,359 @@
...
+    Crawler function which will crawl through the log directory
+    and find eligible files for masking.
+    """
+    changed = False
+    for root, _, files in os.walk(path, onerror=handle_walk_errors):
+        if any(excluded in root for excluded in EXCLUDED_DIRS):
+            continue
+
+        for f in files:
+            if re.search(excluded_file_ext_regex, f) is None:
```
danpawlik, line 182: I think here can be:
```
if not re.search(excluded_file_ext_regex, f):
```

plugins/modules/crawl_n_mask.py
```diff
@@ -0,0 +1,359 @@
...
+    type (Iterator) is parsed to list
+    to make in-place change easy.
+    """
+    try:
+        assert file_path is not None
+        with open(file_path, "r") as f:
+            return list(yaml.safe_load_all(f))
+    except (FileNotFoundError, yaml.YAMLError) as e:
+        module.warn("Error opening file: %s" % e)
+    return None
```
danpawlik, line 285: no need to set `return None` . It would be None. It can be just: `return`.

plugins/modules/crawl_n_mask.py
```diff
@@ -0,0 +1,359 @@
...
+    """
+    Read and Load the yaml file for
+    processing. Using yaml.safe_load_all
+    to handle all documents within a
+    single yaml file stream. Return
+    type (Iterator) is parsed to list
+    to make in-place change easy.
+    """
+    try:
+        assert file_path is not None
```
danpawlik, line 280: nit: maybe would look better. Your way is also fine:
```
if not file_path:
    raise("no file provided")
```

plugins/modules/crawl_n_mask.py
amartyasinha: Done

plugins/modules/crawl_n_mask.py
amartyasinha: Done

amartyasinha:
recheck

danpawlik (approved):
looks good, but someone more involved in that issue should approve it.

plugins/modules/crawl_n_mask.py
```diff
@@ -0,0 +1,359 @@
...
+
+    # the AnsibleModule object will be our abstraction working with Ansible
+    # this includes instantiation, a couple of common attr would be the
+    # args/params passed to the execution, as well as if the module
+    # supports check mode
+    module = AnsibleModule(argument_spec=module_args, supports_check_mode=True)
+
+    params = module.params
+    path = params["path"]
+    isdir = params["isdir"]
```
frenzyfriday, line 332: What happens if someone puts a wrong value of `isdir`, like passes a yaml file and sets isdir: true or passes a dir patch and sets isdir false by mistake?

amartyasinha: - When a file is passed and isdir is set to true, then crawl() will be called with that file as parameter, and within  crawl(), os.walk() will fail, and will raise an exception through handle_walk_errors().

- If a dir is passed and isdir is set to false, then mask() will be called, and it will return when if condition does not match. Since there is no dedicated validation of isdir and path, it may be confusing.

plugins/modules/crawl_n_mask.py
```diff
@@ -0,0 +1,359 @@
...
+
+    if isdir and os.path.exists(path):
+        # craw through the provided directly and then
+        # process eligible files individually
+        changed = crawl(module, path)
+
+    if (
+        not isdir
+        and os.path.exists(path)
+        and re.search(excluded_file_ext_regex, path) is None
```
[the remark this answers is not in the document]
amartyasinha: done

plugins/modules/crawl_n_mask.py
```diff
@@ -0,0 +1,359 @@
...
+        changed = mask_yaml(module, path)
+    return changed
+
+
+def process_list(lst: list) -> None:
+    for item in lst:
+        if isinstance(item, dict):
+            apply_mask(item)
+        elif isinstance(item, list):
+            process_list(item)
```
frenzyfriday, line 210: Do we need something like:
```
elif isinstance(item, str):
      apply_regex(v)
```
in cases where the list contains just some strings instead of dict/list?

frenzyfriday (approved):

frenzyfriday:
> > > Hey @amartyasinha just wondering if you considered using a tool like https://github.com/gitleaks/gitleaks
> > > It's already in use in Prow, implemented by the openshift-ci team. Maybe you know a reason we can't use this project like it won't detect the types of secrets we have or something.
> > 
> > 
> > @lewisdenny Thanks for suggesting gitleaks. Just checked it out and it's amazing. I scanned a log directory and it was able to find most of the secrets. We can modify their config to have our own secret detection rules (to cover the detection of remaining secrets).
> > We have two options now:
> > 
> > * First is to go with the basic Ansible module which will only scan yaml files.
> > * Another option is to look into gitleaks, create a config according to our requirement, and add a way to mask the secrets (that's something gitleaks doesn't provide).
> > 
> > What's your opinion on that?
> 
> My recommendation when it comes to security is to not roll your own. If gitleaks severs our purpose then it may be the safer route but let's hear from other maintainers as well.

Maybe we use this module as we already have it to prevent immediate leaks and as a follow up patch look into gitleaks?

amartyasinha:
recheck

plugins/modules/crawl_n_mask.py
amartyasinha: As per my findings, secrets are in form `<key>: <value>` within dictionary. We have nested lists and dictionary, but never encountered secret or value directly within list in yaml format logs. That's why I skipped this.

If you have any findings, let me know, and I'll make the changes.

amartyasinha:
recheck

evallesp (commented):
I think in general is quite good, but we need molecule and unit tests here.

plugins/modules/crawl_n_mask.py
```diff
@@ -0,0 +1,359 @@
...
+
+
+def crawl(module, path) -> bool:
+    """
+    Crawler function which will crawl through the log directory
+    and find eligible files for masking.
+    """
+    changed = False
+    for root, _, files in os.walk(path, onerror=handle_walk_errors):
+        if any(excluded in root for excluded in EXCLUDED_DIRS):
```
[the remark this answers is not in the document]
amartyasinha: I think that's what I'm trying to do. In exclude_dirs list, I have `openstack-k8s-operators-openstack-must-gather` as logs of this module are masked within itself. So, we do not want to process it again.
One example where we find this dir in our log is `/4130fbe/controller/ci-framework-data/logs/openstack-k8s-operators-openstack-must-gather`. The partial path matching will help us skipping this dir.

plugins/modules/crawl_n_mask.py
```diff
@@ -0,0 +1,359 @@
...
+    changed = False
+    if (
+        path.endswith((tuple(["yaml", "yml"])))
+        or os.path.basename(path).split(".")[0] in ALLOWED_YAML_FILES
+    ):
+        changed = mask_yaml(module, path)
+    return changed
+
+
+def process_list(lst: list) -> None:
```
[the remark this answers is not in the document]
amartyasinha: Added!

plugins/modules/crawl_n_mask.py
amartyasinha: Added a validation. Now we are explicitly checking if path exists and it's value matches with is dir or not, I'm removing the redundant checking of os.path.exists(path) from L358 and L363

plugins/modules/crawl_n_mask.py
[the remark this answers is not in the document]
amartyasinha: I see. I'll add `/` at the start and end of excluded_dirs items.
PS: as of now, we are only processing `yaml` files. Remaining `json` and `.log` files will be covered in upcoming patches.

amartyasinha:
recheck prow

plugins/modules/crawl_n_mask.py
```diff
@@ -0,0 +1,373 @@
...
+    if os.path.isdir(path) != isdir:
+        module.fail_json(msg=f"Value of isdir/path is incorrect. Please check it")
+
+    # if the user is working with this module in only check mode we do not
+    # want to make any changes to the environment, just return the current
+    # state with no modifications
+    if module.check_mode:
+        module.exit_json(**result)
+
+    if isdir:
```
evallesp, line 358: (non-blocking): I'm unsure why this is necessary.

evallesp (approved):
I'm unsure why isdir is necessary here. I think we should go to calculate it.
In any case lgtm.

frenzyfriday (approved):

PR #2789, opened 2025-03-12
Stop hooks from running multiple times when both stage_id and stage_name finds a valid hook

frenzyfriday:


roles/kustomize_deploy/tasks/execute_step.yml
```diff
@@ -121,7 +121,11 @@
       when:
         - not cifmw_kustomize_deploy_generate_crs_only | bool
       vars:
-        hooks: "{{ stage.pre_stage_run | default([]) }}"
+        hooks: >-
+          {{
+            (stage.post_stage_run | default([]))
```
bshewale, line 126: (stage.post_stage_run | default([])) i think this should be (stage.pre_stage_run | default([]))

frenzyfriday: Sorry, my bad. Updated now.

roles/kustomize_deploy/tasks/execute_step.yml
```diff
@@ -117,11 +117,20 @@
       loop_control:
         label: "{{ item.stat.path | default('none') | basename }}"
 
+# REMOVE BEFORE MERGE
```
bshewale, line 120: As it's ready for review then we can remove this debug statement.

frenzyfriday: Looks like it did not solve the problem: https://sf.apps.int.gpc.ocp-hub.prod.psi.redhat.com/zuul/t/components-integration/build/dfdc88000d1c4d9599d4051e37c622ff I'll change this to WIP

PR #2793, opened 2025-03-14
Change default of expectedFailuresList parameter

kstrenkova:
The expectedFailuresList parameter is used to define which tests are expected to fail so that they are not taken into consideration when marking the job's status. It is useful for specific scenarios, such as flaky tests, but we do not want every job to have this parameter by default. It should be used manually to make its intended use more explicit.

### DIFF
--- a/roles/test_operator/defaults/main.yml
+++ b/roles/test_operator/defaults/main.yml
@@ -138,8 +138,10 @@ cifmw_test_operator_tempest_config:
         {{ stage_vars_dict.cifmw_test_operator_tempest_include_list | default('') }}
       excludeList: |
         {{ stage_vars_dict.cifmw_test_operator_tempest_exclude_list | default('') }}
+      {%- if stage_vars_dict.cifmw_test_operator_tempest_expected_failures_list is defined %}
       expectedFailuresList: |
-        {{ stage_vars_dict.cifmw_test_operator_tempest_expected_failures_list | default('') }}
+        {{ stage_vars_dict.cifmw_test_operator_tempest_expected_failures_list }}
+      {%- endif %}
       concurrency: "{{ cifmw_test_operator_concurrency }}"
       externalPlugin: "{{ stage_vars_dict.cifmw_test_operator_tempest_external_plugin | default([]) }}"
       extraRPMs: "{{ stage_vars_dict.cifmw_test_operator_tempest_extra_rpms | default([]) }}"


frenzyfriday:
recheck

Unrelated failure
```"Unhandled Error" err="k8s.io/client-go/tools/watch/informerwatcher.go:146: Failed to watch *unstructured.Unstructured: failed to list *unstructured.Unstructured: Get \"https://api.crc.testing:6443/apis/dataplane.openstack.org/v1beta1/namespaces/openstack/openstackdataplanedeployments?fieldSelector=metadata.name%3Dedpm-deployment&resourceVersion=100135\": dial tcp: lookup api.crc.testing on 199.204.44.24:53: no such host"```

**Comment by danpawlik** [comment]
@kstrenkova try @frenzyfriday solution, or tune it a littlebit if that will not pass:
```
  kind: Tempest
  metadata:
    name: "{{ stage_vars_dict.cifmw_test_operator_tempest_name }}-{{ _stage_vars.name }}"
  spec:
    tempestRun:
      includeList: |
        {{ stage_vars_dict.cifmw_test_operator_tempest_include_list | default('') }}
      {%- if stage_vars_dict.cifmw_test_operator_tempest_expected_failures_list is defined %}
      expectedFailuresList: |
        {{ stage_vars_dict.cifmw_test_operator_tempest_expected_failures_list  }}
      {%- endif %}
      concurrency: "{{ cifmw_test_operator_concurrency }}"
```

kstrenkova:
> @kstrenkova try @frenzyfriday solution, or tune it a littlebit if that will not pass:
> 
> ```
>   kind: Tempest
>   metadata:
>     name: "{{ stage_vars_dict.cifmw_test_operator_tempest_name }}-{{ _stage_vars.name }}"
>     namespace: "{{ cifmw_test_operator_namespace }}"
>   spec:
>     tempestRun:
>       includeList: |
>         {{ stage_vars_dict.cifmw_test_operator_tempest_include_list | default('') }}
>       {%- if stage_vars_dict.cifmw_test_operator_tempest_expected_failures_list is defined %}
>       expectedFailuresList: |
>         {{ stage_vars_dict.cifmw_test_operator_tempest_expected_failures_list  }}
>       {%- endif %}
>       concurrency: "{{ cifmw_test_operator_concurrency }}"
> ```
I have tried using this kind of approach, but it gives me an error that I am not sure how to fix.
`ERROR! We were unable to read either as JSON nor YAML, these are the errors we got from each:
JSON: Expecting value: line 1 column 1 (char 0)

Syntax Error while loading YAML.
  found character that cannot start any token
`
Do you have any suggestions?

danpawlik:
In that case, I guess you can not use characters that are for jinja2 .
Remove the  jinja2 conditions and just set empty value.

kstrenkova:
Hm, I mean currently it's set as `default('')` and I have tried setting it as `default(omit)` which did not work with the `|` symbol. So maybe there is no solution for this? The point of this patch was to not see the parameter in the jobs if it's not defined. The reason is that the approach is a bit of hacky way to pass flaky tests and we do not want the jobs to have it normally. But if there is no solution I am also okay with closing this.

dasm (approved):

kstrenkova:
I do not think there is a way to exclude the `expectedFailuresList` from job run. The jinja code and the `default(omit)` both do not work for this issue. I am not sure why the changes were approved.

kstrenkova:
I am closing this PR as it seems there is no solution.

PR #2796, opened 2025-03-17
[devscripts] Allow using OCP stable versions

eduolivares:
With this patch, ci-fmw allows to configure devscripts openshift_version values "stable-<majorversion>" (e.g. "stable-4.18") which is dynamically updated every time a new minor version becomes the latest stable.

Example:
https://mirror.openshift.com/pub/openshift-v4/clients/ocp/stable-4.18/


roles/devscripts/tasks/build_config.yml
```diff
@@ -40,6 +40,16 @@
       devscripts_config_patches
     }}
 
+- name: Replace OCP version if "stable-" alias used
+  when:
+    - cifmw_devscripts_config.openshift_version.startswith("stable-")
+  vars:
+    ocp_release_txt: "{{ lookup('ansible.builtin.url', 'https://mirror.openshift.com/pub/openshift-v4/clients/ocp/' ~ cifmw_devscripts_config.openshift_version ~ '/release.txt', split_lines=False) }}"
```
lewisdenny, line 47: Could you please prefix these internal vars with _ like so? they are only used in this one task so we should denote them as internal and we don't have to document them.
```suggestion
    _ocp_release_txt: "{{ lookup('ansible.builtin.url', 'https://mirror.openshift.com/pub/openshift-v4/clients/ocp/' ~ cifmw_devscripts_config.openshift_version ~ '/release.txt', split_lines=False) }}"
```

lewisdenny (changes requested):
Cool patch @eduolivares 

Could you please let us know when you test it downstream and add a note about this feature in to the role README?

eduolivares:
> Cool patch @eduolivares
> 
> Could you please let us know when you test it downstream and add a note about this feature in to the role README?

@lewisdenny thanks for your comments!
The PR has been tested downstream and I have added a note to the role README.

bshewale:
+1 lgtm.

bshewale (approved):

sdatko (approved):
Magnificent piece of code :cake:

lewisdenny (approved):

PR #2797, opened 2025-03-18
Adds GH action to sync branches

frenzyfriday:
This PR is to test the following, but on a dummy branch
- One reusable workflow to actually sync the branches
- One cron action to call ^ periodically
- Runs on Monday 21:00 so that I can test if it works

### DIFF
--- a/.github/workflows/sync_branches_reusable_workflow.yml
+++ b/.github/workflows/sync_branches_reusable_workflow.yml
@@ -1,0 +1,39 @@
+---
+name: Sync a follower branch with Main
+on:
+  workflow_call:
+    inputs:
+      main-branch:
+        required: true
+        type: string
+      follower-branch:
+        required: true
+        type: string
+
+jobs:
+  sync:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+    steps:
+      - name: Checkout main branch
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          ref:
+            ${{ inputs.main-branch }}
+
+      - name: Checkout, rebase and push to follower branch
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          ref:
+            ${{ inputs.follower-branch }}
+      - run: |
+          # Details about the GH action bot comes from
+          # https://api.github.com/users/github-actions%5Bbot%5D
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git rebase origin/${{ inputs.main-branch }}
+          git push origin ${{ inputs.follower-branch }}
--- a/.github/workflows/sync_branches_periodically.yml
+++ b/.github/workflows/sync_branches_periodically.yml
@@ -1,0 +1,12 @@
+---
+name: Periodically sync branches
+on:
+  schedule:
+    - cron: '0 21 * * 1'
+
+jobs:
+  trigger_sync:
+    uses: openstack-k8s-operators/ci-framework/.github/workflows/sync_branches_reusable_workflow.yml@main
+    with:
+      main-branch: main
+      follower-branch: ananya-do-not-use-tmp


udis (approved):

.github/workflows/sync_branches_reusable_workflow.yml
```diff
@@ -0,0 +1,32 @@
...
+        with:
+          fetch-depth: 0
+
+      - name: Update the follower branch
+        run: |
+          git checkout ${{ inputs.main-branch }}
+          git fetch origin
+          git checkout ${{ inputs.follower-branch }}
+          git pull
+          git rebase origin/${{ inputs.main-branch }}
```
jguiditta, line 31: question: what happens if there is a merge conflict here?

bshewale (approved):

.github/workflows/sync_branches_reusable_workflow.yml
frenzyfriday: As I understood no one commits directly to the follower branch. So it is (hopefully) always behind the main one - direct rebasable.

.github/workflows/sync_branches_reusable_workflow.yml
```diff
@@ -0,0 +1,32 @@
...
+
+jobs:
+  sync:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
```
lewisdenny, line 21: Any reason you are not using v4?

.github/workflows/sync_branches_reusable_workflow.yml
```diff
@@ -0,0 +1,32 @@
...
+        with:
+          fetch-depth: 0
+
+      - name: Update the follower branch
```
lewisdenny, line 25: I don't see how this would work? This shell command is going to be ran inside a VM with unconfigured git. 

Here is an example using the checkout@v4 aciton where you can see the git username and email needs to be set to commit, maybe that's not needed just to push but surely some authentication is required which I think using the checkout@v4 action for this step as well may handle. 

 https://github.com/actions/checkout#Push-a-commit-using-the-built-in-token

lewisdenny (changes requested):

.github/workflows/sync_branches_reusable_workflow.yml
frenzyfriday: Right!! I added these two to get the git config to use Github actions bot
```
          git config user.name "github-actions[bot]"
          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
I am assuming Github Actions bot is like a service user maintained by Github. I am following the discussion from here: https://github.com/orgs/community/discussions/26560 and using the email ID they used.

.github/workflows/sync_branches_reusable_workflow.yml
frenzyfriday: thanks, updated

.github/workflows/sync_branches_reusable_workflow.yml
lewisdenny: Okay, I still don't understand how this step will have authentication to push to our repo which is why I suggested to use "actions/checkout@v4" for this step as well but happy to try it.

lewisdenny:
/hold

You can remove the hold when ready to test @frenzyfriday

lewisdenny (dismissed):

bshewale (dismissed):

.github/workflows/sync_branches_reusable_workflow.yml
frenzyfriday: 🤦🏽‍♀️ Sorry, I understood wrong yesterday. Updated now.

lewisdenny (dismissed):

frenzyfriday:
recheck

.github/workflows/sync_branches_reusable_workflow.yml
```diff
@@ -0,0 +1,42 @@
...
+      contents: write
+      pull-requests: write
+    steps:
+      - name: Checkout main branch
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+        ref:
+          ${{ inputs.main-branch }}
+        run: |
```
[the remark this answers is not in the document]
jguiditta: Also, if 'main' is the default branch, you should not need much of this anyway, I don't think. For instance, in another project, the checkout step is simply:
```
steps:
    - uses: actions/checkout@v4
      with:
        fetch-depth: 0
```

.github/workflows/sync_branches_reusable_workflow.yml
```diff
@@ -0,0 +1,42 @@
...
+          git fetch origin
+
+      - name: Checkout, rebase and push to follower branch
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+        ref:
+          ${{ inputs.follower-branch }}
+        run: |
```
jguiditta, line 35: This one will probably also need to be pulled out to the items level, so `- run: |`

.github/workflows/sync_branches_reusable_workflow.yml
```diff
@@ -0,0 +1,42 @@
...
+          # Details about the GH action bot comes from
+          # https://api.github.com/users/github-actions%5Bbot%5D
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git pull
```
jguiditta, line 40: suggestion: I don't think `git pull` should be needed here, as the previous step (once this is restructured) just checked out the follower-branch. It should already be current.

.github/workflows/sync_branches_reusable_workflow.yml
```diff
@@ -0,0 +1,42 @@
...
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+        ref:
+          ${{ inputs.follower-branch }}
+        run: |
+          # Details about the GH action bot comes from
+          # https://api.github.com/users/github-actions%5Bbot%5D
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
jguiditta, line 39: According to the [docs](https://github.com/marketplace/actions/checkout#push-a-commit-to-a-pr-using-the-built-in-token), unless I am misunderstanding, this line should be:
```
git config user.email "{user.id}+{user.login}@users.noreply.github.com"
```

jguiditta (commented):
Sorry, I thought I only had the one comment, but keep finding more things to mention as I read though the checkout action [docs](https://github.com/marketplace/actions/checkout#push-a-commit-to-a-pr-using-the-built-in-token). I have batched up what I think should be the rest of my thoughts here. 
One general suggestion I have is to create your follower branch on your fork to fully test this patch. Then you should be able to see it run in your own fork of the action. I have done this successfully in the past on other projects.

frenzyfriday:
Moving to DNM once more. I tested on a dummy repo: https://github.com/frenzyfriday/frenzyfriday-testproject/blob/stable/.github/workflows/sync_branches_reusable_workflow.yml this worked finally. I found an obvious flaw in my patch - it assumes a /sync-olive comment will trigger it. Where will one comment this? On a random PR? No. So I need to find a better trigger for my workflow.
Also if `Require a pull request before merging` is enabled the workflow will fail. So I need to create a token with limited permission and allow that token as exception to the PR rule.

So, DNM till I do all that.
We can go with these meanwhile:
- https://github.com/openstack-k8s-operators/ci-framework/pull/2808 
- https://github.com/openstack-k8s-operators/ci-framework/pull/2805

.github/workflows/sync_branches_reusable_workflow.yml
frenzyfriday: yes, sorry. Updated now

.github/workflows/sync_branches_reusable_workflow.yml
frenzyfriday: yes, sorry. Updated now

.github/workflows/sync_branches_reusable_workflow.yml
frenzyfriday: If I use "user" that would be the person triggering the workflow, not the GH actions bot, right? I think in the doc they are pointing to how the email ids are generated (from the json output of the api call)

frenzyfriday:
> Moving to DNM once more. I tested on a dummy repo: https://github.com/frenzyfriday/frenzyfriday-testproject/blob/stable/.github/workflows/sync_branches_reusable_workflow.yml this worked finally. I found an obvious flaw in my patch - it assumes a /sync-olive comment will trigger it. Where will one comment this? On a random PR? No. So I need to find a better trigger for my workflow. Also if `Require a pull request before merging` is enabled the workflow will fail. So I need to create a token with limited permission and allow that token as exception to the PR rule.
> 
> So, DNM till I do all that. We can go with these meanwhile:
> 
> * [One time manual rebase to sync stable with master #2808](https://github.com/openstack-k8s-operators/ci-framework/pull/2808)
> * [One time manual rebase to sync stable with master #2805](https://github.com/openstack-k8s-operators/ci-framework/pull/2805)

Updated some bits. At least the cron sync for olive should work now (while I focus on service users etc)

.github/workflows/sync_branches_reusable_workflow.yml
jguiditta: Hmm, you may be right! And I see the run against your own repo passed, so I think we can likely go with what you have here.

jguiditta (approved):

danpawlik (approved):

frenzyfriday:
> Team email would just look better ;)

This is the email id of the github actions bot. I am not sure if I can change it to something specific

bshewale (approved):

PR #2798, opened 2025-03-18
Disable DLRN reporting in testproject

arxcruz:
Right now, all testproject are reporting to dlrn, no matter it is required or not, this is overloading the dlrn service making it timeout. Ideally, we should not have zuul variables here, but it is the easier way to do it rather than update all job definition.


evallesp:
FTR: Might be related with DLRN timeouts, not sure here. But better disabled. Thanks

roles/dlrn_report/tasks/main.yml
```diff
@@ -14,11 +14,21 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+- name: Warning if DLRN is not running
+  debug:
+  msg: "DLRN will not be executed if it is a testproject, make sure you set cifmw_dlrn_report_force_report to true if you want to report to DLRN"
+  when:
+    - ('testproject' in zuul.project.name) and (cifmw_dlrn_report_force_report | bool)
```
sdatko, line 21: If I understand correctly, this condition is wrong.
You want it to display for testproject, but when the force variable is set to false, right?
Then the value of force variable should be `not`ed I think?

Right now in testproject without force variable (i.e. using default value) the whole task will be skipped.
`[(testproject – true) and (force false)] => false`

roles/dlrn_report/tasks/main.yml
```diff
@@ -14,11 +14,21 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+- name: Warning if DLRN is not running
+  debug:
+  msg: "DLRN will not be executed if it is a testproject, make sure you set cifmw_dlrn_report_force_report to true if you want to report to DLRN"
```
sdatko, line 19: syntax error?

The msg should be nested under `debug`.
Also please use `ansible.builtin.debug` instead of just `debug`.

roles/dlrn_report/tasks/main.yml
```diff
@@ -14,11 +14,21 @@
...
   block:
     - name: Install dlrnapi-client
       ansible.builtin.import_tasks: install.yml
 
     - name: Report DLRN results
       ansible.builtin.import_tasks: dlrn_report_results.yml
+  when:
+    - zuul is defined
+    - zuul.project is defined
+    - ('testproject' in zuul.project.name) and not (cifmw_dlrn_report_force_report | bool)
```
sdatko, line 33: and in opposition to my previous comment, here there should not be `not` in condition probably...?

sdatko (changes requested):

frenzyfriday (approved):

frenzyfriday (approved):

PR #2804, opened 2025-03-20
Add content-provider variant for FR2

sdatko:
Some required projects do not have 18.0-fr2 branch, so we override them to checkout the main branch by default. When preparing a build for a matching branch, Zuul will merge the jobs definitions and the override will happen only for FR2.

It would be still possible to use Depends-On to test specific changes in the projects that are listed as required ones.


bshewale (approved):

bshewale:
Do we have any testproject ?

evallesp (approved):

PR #2809, opened 2025-03-20
Correct content provider variant for FR2

sdatko:


sdatko:
So... tests for main passed: https://softwarefactory-project.io/zuul/t/rdoproject.org/buildset/f5782cece3c04e65bb259690311fc18a

But my jobs for 18.0-fr2 still did not trigger...

Investigating further...

zuul.d/base.yaml
```diff
@@ -293,6 +326,17 @@
       push_registry: quay.rdoproject.org
       quay_login_secret_name: quay_nextgen_zuulgithubci
 
+- job:
+    name: cifmw-base-crc-openstack
+    parent: cifmw-base-crc
+    branches: ^18.0-fr2
```
frenzyfriday, line 332: Are we expecting branches with branch names like `<some_prefix>-18.0.fr2` and trying to exclude those in repos where `cifmw-base-crc-openstack` runs?
I am asking because we are using plain old `branches: 18.0-fr2` in the other jobs. 
Both `branches: 18.0-fr2` and `^18.0-fr2` would work here of course but later seeing the minor regex difference someone may wonder what's the difference between branch names in the repos where `cifmw-base-crc-openstack` runs.

zuul.d/base.yaml
```diff
@@ -204,6 +222,20 @@
               tenant:
                 ip: 172.19.0.100
                 config_nm: false
+
+# Variant for FR2: some projects do not have 18.0-fr2 branch, so we override...
+# (when preparing a build for matching branch, Zuul will merge the definitions)
+- job:
+    name: cifmw-podified-multinode-edpm-base-crc
+    parent: base-extracted-crc
+    branches: 18.0-fr2
```
frenzyfriday, line 231: When we have fr3 in the future , we would change these jobs directly to `18.0-fr[2,3]` or something? (Provided of course we dont have fr3 branches for cifmw and repo-setup - which we will most likely not have)
Or do we want to make these something like `18.0-fr(?![01]\b)\d` already?

zuul.d/base.yaml
```diff
@@ -109,6 +110,22 @@
...
       cifmw_artifacts_basedir: "{{ cifmw_basedir | default(ansible_user_dir ~ '/ci-framework-data') }}"
 
+- job:
+    name: openstack-k8s-operators-content-provider
+    parent: openstack-k8s-operators-content-provider-base
+
+# Variant for FR2: some projects do not have 18.0-fr2 branch, so we override...
+# (when preparing a build for matching branch, Zuul will merge the definitions)
+- job:
+    name: openstack-k8s-operators-content-provider
```
frenzyfriday, line 120: sorry, what would happen if instead of making a base job we did:
```
    name: openstack-k8s-operators-content-provider
    parent: cifmw-base-minimal
    branches: 18.0-fr2
    required-projects:
      - name: openstack-k8s-operators/ci-framework
        override-checkout: main
      - name: openstack-k8s-operators/repo-setup
        override-checkout: main
```
I mean why didnt we need to make a base job in the other scenarios?

frenzyfriday (commented):
Sorry, I have too many questions 😅

zuul.d/base.yaml
sdatko: There is a Zuul bug when merging the required projects between variants when there occur repo listed without override-checkout and the job is triggered against branch that do not exist in one of the required projects. The bug do not occur when using parent-child relation and the override behaves as expected (surprisingly).

zuul.d/base.yaml
sdatko: The answer is: I don't know.

Please note this is a draft and I am just playing with things – no decisions made, I am looking for anything that may be the best solution, there are incomplete pieces of code here.

zuul.d/base.yaml
sdatko: That might be one option, but it is not up to me. I was just experimenting, without any final decisions.

Also, please take into account that we may not use negative lookaheads in Zuul, as RE2 engine do not support them.

frenzyfriday: ohh, ack, that makes sense thanks.

sdatko:
The state of the art for now here is:

I finally understood what is wrong and I am done with the testing, but the change to fix the problem for all projects may be too big – I summarized my observations in OSPNW-998 in case someone would like to pick this up and continue / decide what should be the final solution.

PR #2812, opened 2025-03-20
Add IBM nodesets that will enforce spawning CI on IBM hosts

danpawlik:
Some CI jobs would be executed directly on the IBM private cloud until Zuul CI will not make round robin between all providers.


frenzyfriday:
Hey, is this patch connected to some other patch that forces these nodesets on some jobs? Also are you putting up this PR to decrease nodeset failures?

frenzyfriday (approved):

lewisdenny (approved):

PR #2815, opened 2025-03-21
Revert "[bop]Donot use dlrn --local for downstream"

amoralej:
I am proposing a new approach for this. I have proposed a slight change in the behavior of the dlrn downstream driver [1] which avoids the issue with the versions.csv and gives us the advantadges of using --local (i.e. we don't need to worry about git repos remotes, etc...)

[1] https://softwarefactory-project.io/r/c/DLRN/+/33366

This reverts commit cd2198d97e92e0d696a94c9d269eb4e9ee1ba219.


lewisdenny:
WDYT @evallesp  <3

amoralej:
I'm closing this as the problem we had was fixed in a different way.

PR #2817, opened 2025-03-21
Change yq with jq for image filtering

danpawlik:
The yq tool might not be installed on every system, so more safe for use would be jq.

### DIFF
--- a/roles/env_op_images/tasks/main.yml
+++ b/roles/env_op_images/tasks/main.yml
@@ -60,22 +60,22 @@
         oc get ClusterServiceVersion
         -l operators.coreos.com/openstack-operator.openstack-operators
         --all-namespaces
-        -o yaml |
-        yq e '
+        -o json |
+        jq -r '
         [.items[]? |
         .spec.install.spec.deployments[]? |
         .spec.template.spec.containers[]? |
         .env[]? |
         select(.name? | test("^RELATED_IMAGE")) |
-        select(.name == "*manager*" or .name == "*MANAGER*") |
-        {(.name): .value}]'
+        select(.name | contains("MANAGER")) |
+        {(.name): .value} ]'
       register: _sa_images_content
       args:
         executable: /bin/bash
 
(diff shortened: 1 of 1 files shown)


danpawlik:
recheck

ciecierski (approved):
lgtm, tested on my OCP setup

ciecierski (approved):
Lgtm

frenzyfriday (approved):

roles/env_op_images/tasks/main.yml
```diff
@@ -60,22 +60,22 @@
...
+        -o json |
+        jq -r '
         [.items[]? |
         .spec.install.spec.deployments[]? |
         .spec.template.spec.containers[]? |
         .env[]? |
         select(.name? | test("^RELATED_IMAGE")) |
-        select(.name == "*manager*" or .name == "*MANAGER*") |
-        {(.name): .value}]'
+        select(.name | contains("MANAGER")) |
```
sdatko, line 70: We don't care about lower and upper cases here anymore?

Otherwise I would expect here something like `|ascii_upcase` involved, or `test("manager"; "i")` instead... (tested on https://play.jqlang.org)

danpawlik: so from the list all are uppercase

danpawlik:
recheck

sdatko (approved):

roles/env_op_images/tasks/main.yml
sdatko: ack, thanks for clarifying – I am all fine then with this :+1:

roles/env_op_images/tasks/main.yml
```diff
@@ -61,7 +61,7 @@
         -l operators.coreos.com/openstack-operator.openstack-operators
         --all-namespaces
         -o yaml |
```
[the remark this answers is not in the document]
sdatko: Done

PR #2818, opened 2025-03-21
Revert "Use zuul clonned repos for upstream in build_openstack_packages role" and revers zuul changes list

amoralej:
This is breaking builds with downstream driver in two ways:

- Not using --local requires the repos to have proper remote definition.
- pre-run scripts as patch_rebaser [1] also require proper remotes.

For the first, we may handle it via change in dlrn [2], but for the second one I'm not sure about the best way.

[1] https://github.com/release-depot/patch_rebaser/
[2] https://softwarefactory-project.io/r/c/DLRN/+/33366

This reverts commit f79ca6f89ae08948f74e34fdd32a691a7cef9f06.

After this revert we still have problems when building multiple in-flight patches in any form:

1. Having multiple depends-on on the same package does not work.
2.  Piling multiple reviews and adding the depends-on to the top one is
   also not working.

This PR is also fixing (2) by reversing the order of the related zuul changes to  find and process the top change for the repo.

### DIFF
--- a/roles/build_openstack_packages/tasks/parse_and_build_pkgs.yml
+++ b/roles/build_openstack_packages/tasks/parse_and_build_pkgs.yml
@@ -1,6 +1,6 @@
 ---
 - name: Parse Zuul changes
-  with_items: "{{ zuul['items'] }}"
+  with_items: "{{ zuul['items'] | reverse | list }}"
   when:
     - zuul is defined
     - "'change_url' in item"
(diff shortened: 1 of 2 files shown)


amoralej:
@raukadah @karelyatin check this, please. Feel free to propose alternatives

amoralej:
Testing in https://github.com/openstack-k8s-operators/watcher-operator/pull/120

lewisdenny (approved):

amartyasinha (approved):
lgtm

PR #2827, opened 2025-03-26
[DNM][TEST] Switch kuttl job on 4.18 without using the base-crc-cloud

bshewale:
job


bshewale:
Not needed as we have fix ready for the failure.

PR #2828, opened 2025-03-26
Start virtqemud service instead of checking for the sock path

eshulman2:


roles/libvirt_manager/tasks/clean_layout.yml
```diff
@@ -2,10 +2,15 @@
...
-    path: "/var/run/libvirt/virtqemud-sock"
+- name: Populate service facts
+  ansible.builtin.service_facts:
+
+- name: Start virtqemud socket service
+  ansible.builtin.service:
+    name: virtqemud.socket
+    state: started
+    enabled: yes
+  when: ansible_facts['services']['virtqemud.socket']['status'] | default('not-found') != 'not-found'
```
danpawlik, line 13: wondering if that condition would be not better:
```
"'virtqemud.socket' not in ansible_facts.services"
```

roles/libvirt_manager/tasks/clean_layout.yml
```diff
@@ -22,7 +27,7 @@
       cifmw_libvirt_manager_dependency_packages |
       difference(ansible_facts.packages.keys()) |
       length == 0
-    - _virtqemud.stat.exists
+    - ansible_facts['services']['virtqemud.socket']['status'] | default('not-found') != 'not-found'
```
danpawlik, line 30: same here

roles/libvirt_manager/tasks/clean_layout.yml
eshulman2: I would say that no because the ansible module doesn't list the socket service only the main service and we would like to make sure the service exist before trying to start the socket and interact with it.

roles/libvirt_manager/tasks/clean_layout.yml
eshulman2: same as before

jamepark4:
Tested on nova hybrid environment downstream and no longer seeing any issues on my side using this PR.

roles/libvirt_manager/tasks/clean_layout.yml
```diff
@@ -2,10 +2,19 @@
...
+
+- name: Start virtqemud socket service
+  ansible.builtin.service:
+    name: "{{ item }}"
+    state: started
+    enabled: yes
+  loop:
+    - virtqemud.service
+    - virtqemud.socket
+  when: ansible_facts['services']['virtqemud.service']['status'] | default('not-found') != 'not-found'
```
dasm, line 16: This line is not required. Based on [the module description](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/service_module.html#parameter-state) the start action will be executed only when it's not already running.
> `started`/`stopped` are idempotent actions that will not run commands unless necessary.

roles/libvirt_manager/tasks/clean_layout.yml
```diff
@@ -22,7 +31,7 @@
       cifmw_libvirt_manager_dependency_packages |
       difference(ansible_facts.packages.keys()) |
       length == 0
-    - _virtqemud.stat.exists
+    - ansible_facts['services']['virtqemud.service']['status'] | default('not-found') != 'not-found'
```
dasm, line 34: Shouldn't it be already passing this conditional, based on [the line modified above](https://github.com/openstack-k8s-operators/ci-framework/pull/2828/files#diff-3193dfb4f3068ac49015aceb156d84037f57116dcd60094d44809a8a9c4eaf68R11)?

dasm (commented):
Minor questions.
Thank you @eshulman2 !

roles/libvirt_manager/tasks/clean_layout.yml
eshulman2: true but it is not true if it is not installed/doesn't exist

roles/libvirt_manager/tasks/clean_layout.yml
eshulman2: only if the service exist, same problem if it wasn't installed/deleted for some reason

roles/libvirt_manager/tasks/clean_layout.yml
eshulman2: generally I agree but I feel this also solves the issue of it not being started

eshulman2:
@dasm can you please review and vote so we can get it in?

roles/libvirt_manager/tasks/clean_layout.yml
danpawlik: I checked it out locally and am surprised myself

danpawlik (approved):

amartyasinha (approved):

PR #2829, opened 2025-03-26
DNM Moving cifmw-crc-podified-edpm-baremetal job to use just IBM cloud

danpawlik:
From time to time we are out of the resources in one of our cloud providers. The IBM private cloud provider is used rary, especially for that job.
Let's try to use just this provider to unlock other CI jobs that needs to be running on Vexxhost provider.


danpawlik:
it needs to fail. Can not do job in both providers.

PR #2837, opened 2025-03-27
Change base molecule nodeset to use just IBM providers

danpawlik:
The other jobs require multi node deployment that mostly will work on Vexxhost. To unlock resources for other job, let's move molecule job to IBM.


danpawlik:
recheck

frenzyfriday (approved):

dasm (approved):

PR #2841, opened 2025-03-27
Monitor changes during update.

sathlan:
This collect changes in:
- Pod in openstack and openstack-operators namespace;
- containers on compute nodes;
- rpm on compute nodes;
- current step of the update;
and put it them in a timeline of event during the update process.

It creates a file `{{ cifmw_update_artifacts_basedir
}}/update_timeline.log` which is usually
`/home/zuul/ci-framework-data/tests/update/update_timeline.log` where
all the changes detected are logged in a `diff` format:

```diff
-keystone-cron-29050981-jvphr Succeeded registry.redhat.io/rhoso/openstack-keystone-rhel9a256:3fa3d690c31e9ec3146042609671f570cc6522a8ac68f87b5a82cc447cfd698d 2025-03-27T07:01:00
+keystone-cron-29051161-vt4zr Succeeded registry.redhat.io/rhoso/openstack-keystone-rhel9a256:3fa3d690c31e9ec3146042609671f570cc6522a8ac68f87b5a82cc447cfd698d 2025-03-27T10:01:00
```

Here we have change in the stage and a restart of the keystone-cron
container on the compute.

Furthermore during initialization of the logs, it collects the local
time on all host to make it easier to find the time delta between the
events on the controller-0 and the different nodes of the deployment.

The monitoring is done every
`cifmw_update_resources_monitoring_interval` seconds which is 10 by
default.

Resolves-Part-Of: [OSPRH-15060](https://issues.redhat.com//browse/OSPRH-15060)


danpawlik:
@sathlan please fix issues raised in cifmw-molecule-update

playbooks/update.yml
```diff
@@ -19,6 +28,11 @@
   hosts: "{{ cifmw_target_host | default('localhost') }}"
   gather_facts: false
   tasks:
+    - name: Set update step to Update Repo and OpenStack Services Containers
+      ansible.builtin.command:
+        cmd: >
+          {{ cifmw_basedir }}/tests/update/update_step.sh
+          Update Repo and OpenStack Services Containers
```
danpawlik, line 35: nit: maybe in the single quote will be better?

roles/update/molecule/default/converge.yml
```diff
@@ -20,5 +20,13 @@
   vars:
     ansible_user_dir: "{{ lookup('env', 'HOME') }}"
     cifmw_update_run_dryrun: true
-  roles:
-    - role: "update"
+    cifmw_openshift_kubeconfig: "{{ lookup('env', 'HOME') }}/.kube/config"
```
danpawlik, line 23: not so sure here.
I think it should be using same value as in other roles: `cifmw_openshift_kubeconfig: "{{ ansible_user_dir }}/.crc/machines/crc/kubeconfig"`.
The `.kube/.config might not exists`

roles/update/tasks/init_monitoring.yml
```diff
@@ -0,0 +1,55 @@
...
+  no_log: "{{ cifmw_nolog | default(true) | bool }}"
+
+- name: Start monitoring
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  ansible.builtin.shell:
+    cmd: >-
+      {{ cifmw_update_artifacts_basedir }}/monitor_resources_changes.sh
+      -a monitor -t {{ cifmw_update_resources_monitoring_interval }} &
```
danpawlik, line 54: that might not work. 
Try using: `async`

sathlan: Oki, this was a annoying one.

So using `async: 0` means that `pool` becomes relevant.  If too short the parent process get kill and the script as well.  When too long, it should be fine, but it depends on the fact the we stay in the same playbook.

Overall, I change the script to become a daemon, so no need to play with `async` and `poll`.  I though about using `nohup` which should be there, but eventually decided to not intruced the dependency.

playbooks/update.yml
sathlan: This avoid long line, would rather keep it as it is.

playbooks/update.yml
```diff
@@ -85,15 +109,38 @@
...
+    - name: Set update step to End of Update Role
+      ansible.builtin.command:
+        cmd: >
+          {{ cifmw_basedir }}/tests/update/update_step.sh
+          End of Update Role
+    - name: Stop monitoring
+      block:
+        - name: Verify monitoring pid file
+          ansible.builtin.stat:
+            path: "{{ cifmw_update_artifacts_basedir }}/monitor_resources_change.pid"
```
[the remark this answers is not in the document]
sathlan: Fixed.

roles/update/README.md
```diff
@@ -19,4 +19,6 @@ Role to run update
 * `cifmw_update_reboot_test`: (Boolean) Activate the reboot test after update. Default to `False`.
 * `cifmw_update_ansible_ssh_private_key_file`: (String) Define the path to the private key file used for the compute nodes.
 * `cifmw_update_wait_retries_reboot`: (Integer) Number of retries to wait for a compute node reboot. One retry is done every five seconds. Default to 60, so five minutes.
+* `cifmw_update_resources_monitoring_interval`: (Integer) Interval, in second, between two resources monitor during update. Default to 10 seconds.
+
```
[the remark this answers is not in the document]
sathlan: Fixed.

roles/update/templates/monitor_resources_changes.sh.j2
```diff
@@ -0,0 +1,278 @@
...
+get_pod_info() {
+    local namespace=$1
+    # Proctect against non working OpenShift environment
+    if [ ! -s "${KUBECONFIG}" ]; then
+        echo "Kubeconfig non existant, skipping tests: '${KUBECONFIG}'" \
+             2>>${SLOG_FILE}
+        return
+    fi
+    oc get pods -n $namespace -o json 2>>${SLOG_FILE} | \
+        jq -r '.items[] | "\(.metadata.name) \(.status.phase) \(.spec.containers[] | .image) \(.status.startTime)"' \
```
ciecierski, line 143: Just a few improvments:
- Use oc builtin templating mechanism instead of parsing with jq
- Add column names
- Startime is fine, but I prefer to have human readable format in logs too

roles/update/templates/monitor_resources_changes.sh.j2
```diff
@@ -0,0 +1,278 @@
...
+TIMELINE_LOG_FILE="${BASE_DIR}/update_timeline.log"
+
+# Where to find the inventory to connect to the compute.
+CI_INVENTORY="{{ ansible_user_dir }}/ci-framework-data/artifacts/zuul_inventory.yml"
+
+# Log files
+UPDATE_STAGE_FILE="${BASE_DIR}/current_update_stage.log"
+
+# OpenShift variables
+OS_NAMESPACES=("openstack-operators" "openstack")
```
ciecierski, line 28: Use variables from roles

roles/update/templates/monitor_resources_changes.sh.j2
sathlan: > Use oc builtin templating mechanism instead of parsing with jq

I tested this:

```bash
oc get pods -n "openstack" -o json | jq -r '.items[] | "\(.metadata.name) \(.status.phase) \(.spec.containers[] | .image) \(.status.startTime)"' | sort > a
oc get pods -n "openstack" -o go-template --template='{{range .items}}{{.metadata.name}} {{.status.phase}} {{range .spec.containers}}{{.image}} {{end}}{{.status.startTime}}{{"\n"}}{{end}}'|sort > b
```
and diffing `a` and `b`:

```diff
< placement-7684599dd-q78qv Running registry.redhat.io/rhoso/openstack-placement-api-rhel9@sha256:46a80b56e06ddc8777541d6ad5c6c011b99c3a2ee80a337de93309177d54590d 2025-04-14T03:20:20Z
< placement-7684599dd-q78qv Running registry.redhat.io/rhoso/openstack-placement-api-rhel9@sha256:46a80b56e06ddc8777541d6ad5c6c011b99c3a2ee80a337de93309177d54590d 2025-04-14T03:20:20Z
< placement-7684599dd-qssl4 Running registry.redhat.io/rhoso/openstack-placement-api-rhel9@sha256:46a80b56e06ddc8777541d6ad5c6c011b99c3a2ee80a337de93309177d54590d 2025-04-14T03:19:43Z
< placement-7684599dd-qssl4 Running registry.redhat.io/rhoso/openstack-placement-api-rhel9@sha256:46a80b56e06ddc8777541d6ad5c6c011b99c3a2ee80a337de93309177d54590d 2025-04-14T03:19:43Z
< placement-7684599dd-xrn6f Running registry.redhat.io/rhoso/openstack-placement-api-rhel9@sha256:46a80b56e06ddc8777541d6ad5c6c011b99c3a2ee80a337de93309177d54590d 2025-04-14T03:19:04Z
< placement-7684599dd-xrn6f Running registry.redhat.io/rhoso/openstack-placement-api-rhel9@sha256:46a80b56e06ddc8777541d6ad5c6c011b99c3a2ee80a337de93309177d54590d 2025-04-14T03:19:04Z
---
> placement-7684599dd-q78qv Running registry.redhat.io/rhoso/openstack-placement-api-rhel9@sha256:46a80b56e06ddc8777541d6ad5c6c011b99c3a2ee80a337de93309177d54590d registry.redhat.io/rhoso/openstack-placement-api-rhel9@sha256:46a80b56e06ddc8777541d6ad5c6c011b99c3a2ee80a337de93309177d54590d 2025-04-14T03:20:20Z
> placement-7684599dd-qssl4 Running registry.redhat.io/rhoso/openstack-placement-api-rhel9@sha256:46a80b56e06ddc8777541d6ad5c6c011b99c3a2ee80a337de93309177d54590d registry.redhat.io/rhoso/openstack-placement-api-rhel9@sha256:46a80b56e06ddc8777541d6ad5c6c011b99c3a2ee80a337de93309177d54590d 2025-04-14T03:19:43Z
> placement-7684599dd-xrn6f Running registry.redhat.io/rhoso/openstack-placement-api-rhel9@sha256:46a80b56e06ddc8777541d6ad5c6c011b99c3a2ee80a337de93309177d54590d registry.redhat.io/rhoso/openstack-placement-api-rhel9@sha256:46a80b56e06ddc8777541d6ad5c6c011b99c3a2ee80a337de93309177d54590d 2025-04-14T03:19:04Z
```
so in the `jq` we get two lines for the two images, while in `oc` we get only one line.  I guess everything on one line is better right ? Let me know and I'll just include the `oc` templating.

> Add column names

given that we are interested in diff from one run to another, the column name will not appear in the output.  Don't see a easy way to solve this.

> Startime is fine, but I prefer to have human readable format in logs too

Beside the diff we also get this in the `update_timeline.log` file:

```diff
# ===> 2025-04-14 03:07:27 - CHANGES DETECTED in openstack-operators:
+openstack-operator-controller-operator-7c756db9c8-6d7wc Pending registry.redhat.io/openshift4/ose-kube-rbac-proxy-rhel9@sha256:27398947432dd34625c7a8c537c273dc0e8af85d2fcce6749519acd379bb64ca 2025-04-14T03:07:25Z
+openstack-operator-controller-operator-7c756db9c8-6d7wc Pending registry.redhat.io/rhoso-operators/openstack-rhel9-operator@sha256:f3620ecac70139e08f92e6a9444e1edd603cf7d1efa64dda19986defab87241b 2025-04-14T03:07:25Z
```

with a clear human readable timestamp.

roles/update/templates/monitor_resources_changes.sh.j2
sathlan: hum ... so we have `cifmw_update_namespace` as a var in the update role ([there](https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/update/defaults/main.yml#L33) but we don't have a local to the role `openstack-operators` variable in the role (unless I missed it)

At the cifmw level the `openstack-operators` namespace name seems pretty much hardcoded.
So, given that `update` doesn't need a specific namespace, I'm not sure we should have any variable for namespaces as `openstack` and `openstack-operators` are pretty much hardcoded right now.

Given that having multiple namespace for `openstack` is coming (is there ?) that still might be something we need to do, but here, not sure.

sathlan (commented):
I could change the `oc` vs `jq` templating  if you thing the output is better.  For the namespaces, not sure.

sathlan:
The structure is the same but collection of logs is now based on `events` from `podman` and `openshift`.  This solve the problem of missing events when using `diff` based loging and pave the way to use that log to infer if the update happen as intended as we don't miss any event.  Support for `rpm` was dropped as it already has it own timeline file in `/var/log/dnf.log` on each host.

pinikomarov:
looks good to me

playbooks/update.yml
```diff
@@ -89,15 +113,38 @@
           vars:
             cifmw_cephadm_fsid: "{{ ceph_fsid }}"
 
+
```
dasm, line 116: nit: extra whiteline.

roles/update/tasks/main.yml
```diff
@@ -44,11 +44,76 @@
     - cifmw_ci_gen_kustomize_values_installplan_approval is defined
     - cifmw_ci_gen_kustomize_values_installplan_approval | lower == 'manual'
 
+- name: Handle OpenStack operator initialization
+  when:
+    - cifmw_ci_gen_kustomize_values_deployment_version is defined
+    - cifmw_ci_gen_kustomize_values_deployment_version in ['v1.0.3', 'v1.0.6']
```
dasm, line 50: Why do you use these specific versions? Please change hardcoded values to configurable arguments.

dasm (changes requested):

playbooks/update.yml
sathlan: Done.

roles/update/tasks/main.yml
sathlan: This is because those are the only versions that will ever need that.  Starting from v1.0.7 we don't need that block.  So we might as well save us one variable and just hard code it.

dasm: Ack. That makes sense.

roles/update/tasks/init_monitoring.yml
```diff
@@ -0,0 +1,57 @@
...
+
+- name: Set update step to {{ cifmw_update_state | default("Starting Update") }}
+  ansible.builtin.command:
+    cmd: >
+      {{ cifmw_update_artifacts_basedir }}/update_event.sh
+      {{ cifmw_update_state | default("Starting Update") }}
+
+- name: Initialize monitoring
+  ansible.builtin.command:
+    cmd: "{{ cifmw_update_artifacts_basedir }}/monitor_resources_changes.sh -a init"
```
dasm, line 43: The `molecule` test is failing on this step, but I don't see why. Maybe I'm missing something, but I don't see you're invocating anything k8s-related.
Probably you're missing configuration in the `roles/update/monocule/converge.yaml`, however I'd like to know why are we seeing this issue on the `-a init` step.
```
    cifmw_openshift_kubeconfig: "{{ ansible_user_dir }}/.crc/machines/crc/kubeconfig"
```

```
2025-05-12 10:08:19.446415 | controller | TASK [update : Initialize monitoring] ******************************************
2025-05-12 10:08:19.446423 | controller | Monday 12 May 2025  10:08:17 -0400 (0:00:02.278)       0:00:04.215 ************
2025-05-12 10:08:19.446433 | controller | fatal: [instance]: FAILED! => changed=true
2025-05-12 10:08:19.446518 | controller |   cmd:
2025-05-12 10:08:19.446527 | controller |   - /home/zuul/ci-framework-data/tests/update/monitor_resources_changes.sh
2025-05-12 10:08:19.446532 | controller |   - -a
2025-05-12 10:08:19.446539 | controller |   - init
2025-05-12 10:08:19.446560 | controller |   delta: '0:00:01.501742'
2025-05-12 10:08:19.446746 | controller |   end: '2025-05-12 10:08:19.422141'
2025-05-12 10:08:19.447419 | controller |   msg: non-zero return code
2025-05-12 10:08:19.447572 | controller |   rc: 1
2025-05-12 10:08:19.447582 | controller |   start: '2025-05-12 10:08:17.920399'
2025-05-12 10:08:19.447594 | controller |   stderr: |-
2025-05-12 10:08:19.447600 | controller |     error: Missing or incomplete configuration info.  Please point to an existing, complete config file:
2025-05-12 10:08:19.447609 | controller |
2025-05-12 10:08:19.447616 | controller |
2025-05-12 10:08:19.449711 | controller |       1. Via the command-line flag --kubeconfig
2025-05-12 10:08:19.449770 | controller |       2. Via the KUBECONFIG environment variable
2025-05-12 10:08:19.491229 | controller |       3. In your home directory as ~/.kube/config
2025-05-12 10:08:19.491250 | controller |
2025-05-12 10:08:19.491258 | controller |     To view or setup config directly use the 'config' command.
```

dasm (approved):

danpawlik (dismissed):

ciecierski (approved):

jistr:
LGTM sans the rebase issue

ciecierski (approved):

PR #2845, opened 2025-03-28
[DNM] Check zuul jobs without router setup

karelyatin:
For clouds where instances directly attached to public network, having a router with default gateway is unnecessary. This to test if we can avoid this and save some quoto on resources.

Depends-On: https://github.com/openstack-k8s-operators/openstack-operator/pull/1379


karelyatin:
recheck

PR #2846, opened 2025-03-28
adoption - use --local-push-destination

hjensas:
`--local-push-destination` sets the registry on the undercloud as the location for container images. This means that director pulls the necessary images from the Red Hat Container Catalog and pushes them to the registry on the undercloud. Director uses this registry as the container image source. To pull directly from the Red Hat Container Catalog, omit this option.

Since we are seeing repeated run failures on pulling containers, lets use the director/undercloud's registry.

Jira: [OSPRH-15000](https://issues.redhat.com//browse/OSPRH-15000)


hjensas:
Seems we don't have the secrets for this ...

PR #2847, opened 2025-03-28
Adds GH action to run branch sync

frenzyfriday:
- One reusable workflow
  - You can pass source and target branch and a deploy key to it
  - This can be called from only another gh action workflow that is merged on the main branch
- One trigger workflow
  - This can be triggered with a rest API call with only a PAT generated by someone who owns the repo/organization
  - You can pass source and target branch through the payload. The default is set to main (and a dummy branch for poc)
  - I tried with a[ bad-person branch](https://github.com/frenzyfriday/frenzyfriday-testproject/tree/bad-person) to see if I can trigger this workflow from a non-main branch - I could not
- Tokens:
  - One deploy_token with permission to push to only the target branch
  - One PAT with permission to write on the repo
  - Push to all branches are restricted. The deploy token is added as a bypass to allow push to the target branch (stable/olive whichever we decide)
 ```
curl -L \
  -X POST \
  -H "Accept: application/vnd.github+json" \
  -H "Authorization: Bearer github_pat_PATPAT" \
  -H "X-GitHub-Api-Version: 2022-11-28" \
  https://api.github.com/repos/frenzyfriday/frenzyfriday-testproject/dispatches \
  -d '{"event_type":"trigger-sync", "client_payload":{"source-branch":"main", "target-branch":"ananya-do-not-use-tmp"}}'
```
Tested on my own repo:
- https://github.com/frenzyfriday/frenzyfriday-testproject/blob/main/.github/workflows/sync_branches_reusable_workflow.yml
- https://github.com/frenzyfriday/frenzyfriday-testproject/blob/main/.github/workflows/sync_branches_with_ext_trigger.yml
- https://github.com/frenzyfriday/frenzyfriday-testproject/actions

### DIFF
--- a/.github/workflows/sync_branches_reusable_workflow.yml
+++ b/.github/workflows/sync_branches_reusable_workflow.yml
@@ -9,6 +9,10 @@ on:
       target-branch:
         required: true
         type: string
+    secrets:
+      ssh-key:
+        description: 'Deploy token write access'
+        required: true
 
 jobs:
   sync-branches:
@@ -20,6 +24,8 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
+          ssh-key: ${{ secrets.ssh-key }}
+          persist-credentials: true
 
       - name: Git config
         run: |
--- a/.github/workflows/sync_branches_with_ext_trigger.yml
+++ b/.github/workflows/sync_branches_with_ext_trigger.yml
@@ -1,18 +1,14 @@
-name: Sync branches with external trigger
-
+---
+name: Sync a target branch with source branch
 on:
-  workflow_dispatch:
-    inputs:
-      source-branch:
-        required: false
-        default: 'main'
-      target-branch:
-        required: false
-        default: 'ananya-do-not-use-tmp'
+  repository_dispatch:
+    types: [trigger-sync]
 
 jobs:
   trigger-sync:
     uses: openstack-k8s-operators/ci-framework/.github/workflows/sync_branches_reusable_workflow.yml@main
     with:
-      source-branch: ${{ inputs.source-branch }}
-      target-branch: ananya-do-not-use-tmp # Hardcoded till testing finishes
+      source-branch: ${{ github.event.client_payload.source-branch }}
+      target-branch: ${{ github.event.client_payload.target-branch }}
+    secrets:
+      ssh-key: ${{ secrets.DEPLOY_KEY }}

### CODE REVIEW

File: .github/workflows/sync_branches_reusable_workflow.yml [hunk 1]
```diff
@@ -4,11 +4,17 @@ on:
...
     inputs:
       source-branch:
-        required: true
+        required: false
         type: string
+        default: main
       target-branch:
-        required: true
+        required: false
+        default: ananya-do-not-use-tmp
```
dprince, line 12: should the default here just be 'ci-stable'

.github/workflows/sync_branches_reusable_workflow.yml
```diff
@@ -20,6 +26,8 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
+          ssh-key: ${{ secrets.ssh-key }}
```
dprince, line 29: why did you need this part? In the push force on the other repos we didn't have to add an explicit ssh-key

dprince:
@frenzyfriday I think what you have here looks good and if the tests work is fine. I had one inline comment on the need for the ssh-key part. Also, it might be possible to simplify the workflow here by combining the reusable workflow into the external_trigger workflow. Esssentially just 1 workflow that is meant to be externally triggered.

.github/workflows/sync_branches_reusable_workflow.yml
frenzyfriday: In ci-framework repo settings I am allowing only the deploy key to have push rights without PR on the stable branch. If I dont pass the ssh key (which is the deploy key) it cannot authenticate as the deploy key for the bypass. 

image

^ These are the things I allowed in the settings

frenzyfriday:
> @frenzyfriday I think what you have here looks good and if the tests work is fine. I had one inline comment on the need for the ssh-key part. Also, it might be possible to simplify the workflow here by combining the reusable workflow into the external_trigger workflow. Esssentially just 1 workflow that is meant to be externally triggered.

There are two reasons why I added the reusabel workflow:
- We want to add a cron job action that calls this workflow to sync the olive branch (which is the one used downstream till the ci-frameowrk pipeline is set up)
- I wanted to add a default branch. We may not need defaults thought, so maybe I can change it back to a single workflow after the ci-framework pipeline is done and we do not need olive anymore

.github/workflows/sync_branches_reusable_workflow.yml
sdatko, line 12: Wouldn't `tmp-sync-test` be a better default name than something starting with `ananya-...` ?

.github/workflows/sync_branches_with_ext_trigger.yml
```diff
@@ -1,18 +1,14 @@
...
+  repository_dispatch:
+    types: [trigger-sync]
 
 jobs:
   trigger-sync:
     uses: openstack-k8s-operators/ci-framework/.github/workflows/sync_branches_reusable_workflow.yml@main
     with:
-      source-branch: ${{ inputs.source-branch }}
+      source-branch: ${{ github.event.client_payload.source-branch }}
       target-branch: ananya-do-not-use-tmp # Hardcoded till testing finishes
```
sdatko, line 12: time to replace hardcoded with variable?

.github/workflows/sync_branches_reusable_workflow.yml
```diff
@@ -4,11 +4,17 @@ on:
   workflow_call:
     inputs:
       source-branch:
```
sdatko, line 6: one side thing ~ it may require some change in the git actions below in the file, but just form a practical perspective – can we rename this parameter to `source-ref` and ensure this works with both commit hash and branch name? The practical thing is, when we will start testing job it will take a few hours before we may trigger the sync workflow, so it would be actually better to specify a specific commit hash here as source instead of just `main`

frenzyfriday: ack, lemme check how I can change this to work with commit hash

.github/workflows/sync_branches_with_ext_trigger.yml
frenzyfriday: updated, sorry, missed this one!

.github/workflows/sync_branches_reusable_workflow.yml
frenzyfriday: Lemme do it in a follow up patch

.github/workflows/check-dnm-label.yml
```diff
@@ -17,7 +17,9 @@ jobs:
     steps:
 
       - name: Fail if PR is DNM
-        if: contains(github.event.pull_request.labels.*.name, 'do-not-merge/hold') || contains(github.event.pull_request.labels.*.name, 'do-not-merge/work-in-progress')
+        if: contains(github.event.pull_request.labels.*.name, 'do-not-merge/hold') ||
+        contains(github.event.pull_request.labels.*.name, 'do-not-merge/work-in-progress' ||
+        contains(github.event.pull_request.labels.*.name, 'do-not-merge/contains-merge-commits' )
```
sdatko, line 22: are those changes somehow related to the sync workflow? :thinking:

sdatko (dismissed):
A quality content for me :hamburger:

danpawlik (dismissed):

.github/workflows/check-dnm-label.yml
frenzyfriday: no, that was supposed to be a different patch 🤦🏽‍♀️ sorry

dprince (approved):

evallesp (approved):

PR #2849, opened 2025-03-31
susy_emulator - allow 8000/tcp if firewall enabled

hjensas:
When deploying sushy_emulator as a pod on, check if the firewalld service is enabled and if it is allow TCP port 8000.

Jira: [OSPRH-15286](https://issues.redhat.com//browse/OSPRH-15286)

### DIFF
--- a/roles/sushy_emulator/tasks/configure_firewalld.yml
+++ b/roles/sushy_emulator/tasks/configure_firewalld.yml
@@ -1,0 +1,31 @@
+---
+# Copyright Red Hat, Inc.
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


hjensas:
... fantastic, something is installing the firewall at a later stage so the issue cannot be fixed here.

```
TASK [sushy_emulator : Open TCP port 8000 in firewalld if firewalld.service is enabled and active port=8000/tcp, zone=default, state=enabled, immediate=True, permanent=True] ***
Monday 31 March 2025  03:56:43 -0400 (0:00:00.519)       0:05:59.730 ********** 
fatal: [hypervisor -> controller-0]: FAILED! => 
  msg: |-
    The conditional check 'firewalld_service_status.status.UnitFileState == 'enabled'' failed. The error was: error while evaluating conditional (firewalld_service_status.status.UnitFileState == 'enabled'): 'dict object' has no attribute 'UnitFileState'. 'dict object' has no attribute 'UnitFileState'
  
    The error appears to be in '/home/zuul/src/github.com/openstack-k8s-operators/ci-framework/roles/sushy_emulator/tasks/configure_firewalld.yml': line 22, column 3, but may
  
    - name: Open TCP port 8000 in firewalld if firewalld.service is enabled and active
      ^ here

```

PR #2851, opened 2025-03-31
Bump adoption multinode jobs timeouts

bogdando:
Change base multinode adoption jobs timeouts from 4h to 5h (max
value for Zuul tenant).

Jira: [OSPCIX-758](https://issues.redhat.com//browse/OSPCIX-758)

### DIFF
--- a/zuul.d/adoption.yaml
+++ b/zuul.d/adoption.yaml
@@ -200,7 +200,7 @@
     name: cifmw-adoption-base-source-multinode
     parent: base-extracted-crc
     abstract: true
-    timeout: 14400
+    timeout: 18000
     attempts: 1
     nodeset: centos-9-multinode-rhel-9-2-crc-extracted-2-39-0-3xl
     roles: &multinode-roles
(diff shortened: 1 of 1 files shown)


dasm:
@bogdando if you could, please add an info to the commit message, that you're changing `4h` timeouts into `5h` timeouts. It will help to understand the impact of jobs.

bogdando:
> @bogdando if you could, please add an info to the commit message, that you're changing `4h` timeouts into `5h` timeouts. It will help to understand the impact of jobs.

@dasm Done

bogdando:
recheck provider

ciecierski:
recheck

zuul.d/adoption.yaml
```diff
@@ -200,7 +200,7 @@
     name: cifmw-adoption-base-source-multinode
     parent: base-extracted-crc
     abstract: true
-    timeout: 14400
+    timeout: 18000
```
karelyatin, line 203: How this relevant for OSPCIX-758? the issue was specific to standalone jobs but this is change base multinode jobs, periodic multinodes already have 18000 timeout set and in check we not running the multinodes ones

karelyatin (changes requested):

zuul.d/adoption.yaml
bogdando: you are right, it is not related to standalone jobs in that CIX, so we don't need this change

PR #2855, opened 2025-04-01
Parse in better way default IP address

danpawlik:
It was wrong that we assumed that the CIDR for the network would be '/24'.


frenzyfriday (approved):

raukadah (approved):
Thanks @danpawlik !

PR #2859, opened 2025-04-01
Rename nodesets to include crc-cloud in the name

danpawlik:
Since OCP 4.18, CI jobs would be using CRC cloud, not CRC extracted, that might be confusing for people outside the project.

### DIFF
--- a/zuul.d/nodeset.yaml
+++ b/zuul.d/nodeset.yaml
@@ -339,13 +339,82 @@
       - name: controller
         label: centos-9-stream-crc-2-39-0-xl
 
-
 #
-# CRC-2.48 (OCP4.18) nodesets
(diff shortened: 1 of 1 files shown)


zuul.d/nodeset.yaml
```diff
@@ -339,33 +339,96 @@
...
+        nodes:
+          - compute-0
+          - compute-1
       - name: ocps
         nodes:
           - crc
 
 - nodeset:
-    name: centos-9-crc-2-48-0-xxl
+    name: centos-9-2x-centos-9-xxl-crc-cloud-ocp-4-18-1-xxl
```
bshewale, line 374: i think you missed medium here. it should be centos-9-medium-2x-centos-9-xxl-crc-cloud-ocp-4-18-1-xxl

frenzyfriday (approved):

PR #2866, opened 2025-04-04
Use correct ansible module to print variable vlaue

bshewale:
Print the variable data/value using `msg` instead of `var`.


danpawlik:
There is var and it should print nice output comparing to what msg will show.

danpawlik:
@bshewale yes. For me it is not needed change :)

dasm (approved):

PR #2868, opened 2025-04-04
Set ServerAlive params on ssh_jumper role

rlobillo:
the shiftstack testing is getting disconnected transparently. We believe this could help.


danpawlik:
@rlobillo improve `Assert SSH host - minimal config` - https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/ssh_jumper/molecule/default/converge.yml#L80 and would be good ;)

rlobillo:
Testing on internal lab setup in progress

danpawlik (approved):

PR #2875, opened 2025-04-07
Switch to OCP 4.18 for periodic-whitebox-neutron-tempest-plugin CI job

danpawlik:
The job is using parent that is related to crc-cloud, where nodeset is set to use crc-extracted and that's wrong.
Let's switch to CRC cloud (OCP 4.18) also here.


amartyasinha (approved):

bshewale (approved):

danpawlik:
recheck

PR #2878, opened 2025-04-08
use ceph ingress service vars

katarimanojk:
Using vars will help in overriding the spec during adoptoin.
Also count:1 will create issues as ingress handles both haproxy and keepalived


fmount (approved):
/lgtm

PR #2880, opened 2025-04-08
Prevent updating openstackclient pod during openstack update

ciecierski:
Openstackclient pod is used during update to start/stop continues workload tests from it. Openstackclient should not be updated during update to not interfere with this tests. Update of the client should happen at the end of update on request.

### DIFF
--- a/roles/update/tasks/main.yml
+++ b/roles/update/tasks/main.yml
@@ -33,8 +33,14 @@
   when:
     - cifmw_update_control_plane_check | bool
     - not cifmw_update_run_dryrun | bool
-  ansible.builtin.shell: |
-    {{ cifmw_update_artifacts_basedir }}/control_plane_test_start.sh
+
+  block:
+    - name: Prevent updating openstackclient during openstack update
+      ansible.builtin.include_tasks: prevent_openstack_client_update.yml
+
+    - name: Start the continuous control plane test
+      ansible.builtin.shell: |
+        {{ cifmw_update_artifacts_basedir }}/control_plane_test_start.sh
 
 
 - name: Set openstack_update_run Makefile environment variables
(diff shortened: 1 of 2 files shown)


roles/update/tasks/main.yml
```diff
@@ -72,8 +78,20 @@
...
     - not cifmw_update_run_dryrun | bool
-  ansible.builtin.command: |
-    {{ cifmw_update_artifacts_basedir }}/control_plane_test_stop.sh
+
+  block:
+    - name: Stop the continuous control plane test
+      ansible.builtin.command: |
+        {{ cifmw_update_artifacts_basedir }}/control_plane_test_stop.sh
+
+    - name: Patch openstackversion to trigger update of openstackclient
```
danpawlik, line 87: nit: maybe `Patch openstackversion to ensure openstackclient version is right`

danpawlik (approved):

sathlan:
As far as I understand this end up not working because the pod was still restarted even if its image was not changed.

(here)[https://github.com/openstack-k8s-operators/ci-framework/commit/bb0126a2c08996b6fe20e2cde894c56f2b4d17ee] I gave a try to this in another way: create a container with all the required file to authenticate against openstack and run it using podman on the controller.  So far this was successful.

frenzyfriday (approved):

jistr:
Decoupling the continuous testing logic from the openstackclient pod sounds a bit better to me than doing special handling with the pod in the testing workflow. (The special handling complicates the workflow in a way that doesn't stem from user needs, rather it stems just from testing choices. Also if we don't propagate the more complex workflow into the docs, the distance between what's documented and what's tested increases.)

danpawlik:
maybe some more additional reviewers: @sathlan or @holser ?

ciecierski:
Closing in favor of  https://github.com/openstack-k8s-operators/ci-framework/pull/2935

PR #2881, opened 2025-04-09
New OLM Parameter for Managing Deployment of Specific Versions

sathlan:
A new parameter, `cifmw_ci_gen_kustomize_values_deployment_version`,
has been introduced for overriding OLM startingCSV values in
kustomize. This also ensures the appropriate architecture directory is
utilized and all necessary subscriptions are deployed based on the
installed version.  This eventually leverages the overlay `kustomize`
feature, hence the new `olm_subscriptions_overlay.yml` that helps set
the right overlay.

When configured, this parameter will stop the update playbook from
executing the `set_openstack_containers` role and the deployment
playbook from executing the `update_container` role. The assumption is
that OLM will be used to update any necessary operators.

New test cases have been added using Molecule for different OpenStack
operator versions: v1.0.3, v1.0.6, and v1.0.7. Note that v1.0.7 is
designed to support all future deployments where only the
openstack-operator subscription is needed.

To assist in testing and verification within a live environment, tasks
for setting up and tearing down OpenShift CRC are provided.

Depends-On: https://github.com/openstack-k8s-operators/architecture/pull/537

Resolves-Part-Of: [OSPRH-15056](https://issues.redhat.com//browse/OSPRH-15056)

### DIFF
--- a/roles/ci_gen_kustomize_values/tasks/olm_subscriptions_overlay.yml
+++ b/roles/ci_gen_kustomize_values/tasks/olm_subscriptions_overlay.yml
@@ -1,0 +1,57 @@
+---
+# Copyright Red Hat, Inc.
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+# Description: This playbook generates the kustomize file that
+# retrieves the appropriate overlay based on
+# cifmw_ci_gen_kustomize_values_deployment_version. It allows for
+# specifying the deployment version when multiple versions are
+# available in the OLM catalog. This is particularly useful for
+# testing updates but can also be used to deploy any version from the
+# OLM catalog.
+
+- name: Set the right overlay for the subscriptions
+  ansible.builtin.set_fact:
+    _cifmw_update_deployment_version_dir: >-
+      {{
+      cifmw_ci_gen_kustomize_values_deployment_version
+      if cifmw_ci_gen_kustomize_values_deployment_version in ['v1.0.3', 'v1.0.6']
+      else
+      'default'
+      }}
+
+- name: Point to the right overlay for OLM when deploying old version
+  vars:
+    _kustomization:
+      components:
+        - "../../../lib/olm-deps"
+        - "../../../lib/olm-openstack-subscriptions/overlays/{{ _cifmw_update_deployment_version_dir }}"
+      resources:
+        - "values.yaml"
+  ansible.builtin.copy:
+    content: |
+      ---
+      {{ _kustomization | ansible.builtin.to_nice_yaml(indent=2) }}
+    dest: "{{ cifmw_ci_gen_kustomize_values_architecture_repo }}/examples/common/olm-subscriptions/kustomization.yaml"
+    mode: "0644"
+  when: _cifmw_update_deployment_version_dir != 'default'
+
+# Pass down the new value for `cifmw_kustomize_deploy_olm_source_files`
+- name: Change directory for the customization file when deploying old version
+  ansible.builtin.set_fact:
+    cifmw_kustomize_deploy_olm_source_files: >-
+      {{
+      cifmw_ci_gen_kustomize_values_architecture_repo
+      }}/examples/common/olm-subscriptions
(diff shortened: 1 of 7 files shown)


roles/kustomize_deploy/tasks/install_operators.yml
```diff
@@ -14,9 +14,20 @@
...
+    tasks_from: olm_subscriptions_overlay.yml
+  when: cifmw_ci_gen_kustomize_values_deployment_version is defined
+
 - name: Generate values.yaml for OLM resources
   vars:
-    cifmw_architecture_scenario: 'common/olm'
+    cifmw_architecture_scenario: >-
+      {{
+      'common/olm'
+      if not cifmw_ci_gen_kustomize_values_deployment_version | default(false) | bool
```
sathlan, line 28: `is defined` better ?

roles/ci_gen_kustomize_values/tasks/olm_subscriptions_overlay.yml
```diff
@@ -0,0 +1,57 @@
...
+# available in the OLM catalog. This is particularly useful for
+# testing updates but can also be used to deploy any version from the
+# OLM catalog.
+
+- name: Set the right overlay for the subscriptions
+  ansible.builtin.set_fact:
+    _cifmw_update_deployment_version_dir: >-
+      {{
+      cifmw_ci_gen_kustomize_values_deployment_version
+      if cifmw_ci_gen_kustomize_values_deployment_version in ['v1.0.3', 'v1.0.6']
```
frenzyfriday, line 30: So this has to be updated if in future new directories are added to https://github.com/openstack-k8s-operators/architecture/pull/537/files#diff-8bff4b03dba49b542ae98d89742a4a850c35ea66d16355a9fe9fa045da7cc5f9 for other versions?
Would it be easier to have like `if cifmw_ci_gen_kustomize_values_deployment_version not in [<something>]`

roles/ci_gen_kustomize_values/templates/common/olm-values/values.yaml.j2
```diff
@@ -4,3 +4,8 @@ data:
 {% if cifmw_ci_gen_kustomize_values_sub_channel is defined %}
     openstack-operator-channel: {{ cifmw_ci_gen_kustomize_values_sub_channel }}
 {% endif %}
+{% if cifmw_ci_gen_kustomize_values_deployment_version is defined %}
+{% if cifmw_ci_gen_kustomize_values_deployment_version not in ['v1.0.3', 'v1.0.6'] %}
```
frenzyfriday, line 8: Same question here, this will have to be updated when there is another version that will have its own overlay?

roles/kustomize_deploy/tasks/install_operators.yml
frenzyfriday: I'd say yes `is defined` will be better, just because everywhere else we are using is defined for this var - uniformity. But  `default(false)` is also fine

sathlan: done.

roles/kustomize_deploy/tasks/install_operators.yml
```diff
@@ -14,9 +14,20 @@
...
+    tasks_from: olm_subscriptions_overlay.yml
+  when: cifmw_ci_gen_kustomize_values_deployment_version is defined
+
 - name: Generate values.yaml for OLM resources
   vars:
-    cifmw_architecture_scenario: 'common/olm'
+    cifmw_architecture_scenario: >-
+      {{
+      'common/olm'
+      if not cifmw_ci_gen_kustomize_values_deployment_version is defined
```
sathlan, line 28: should use `is not defined`

sathlan: Done

roles/ci_gen_kustomize_values/tasks/olm_subscriptions_overlay.yml
```diff
@@ -0,0 +1,57 @@
...
+      else
+      'default'
+      }}
+
+- name: Point to the right overlay for OLM when deploying old version
+  vars:
+    _kustomization:
+      components:
+        - "../../../lib/olm-deps"
```
danpawlik, line 39: nit: wondering, if here we could add absolute path.
I know that in other roles it is done same, so I will just leave a comment for me in the future ;)
I'm trying to work on local deployment for some test and that might be problematic.

danpawlik (approved):
LGTM - code + proposed solution.

roles/ci_gen_kustomize_values/tasks/olm_subscriptions_overlay.yml
sathlan: Yes this is  a copy of the original [examples/common/olm/kustomization.yaml](https://github.com/openstack-k8s-operators/architecture/blob/main/examples/common/olm/kustomization.yaml) where this definition is taken from.

Furthermore that code will only run for `v1.0.3` and `v1.0.6`.  Then we will directly using the new file from [this PR](https://github.com/openstack-k8s-operators/architecture/pull/537/files#diff-cada4dbe2af4cf7df05739d297c27b10d50e8817176838539480cf9bac95efe5) where definition is set.

Overall I think this should stay as it is.  Local testing using `kubectl kustomize` on that directory worked fine in my testing.

frenzyfriday (approved):

PR #2887, opened 2025-04-09
Create bmh compute replace validator

drosenfe:
Add a validator that verifies a faulty bmh compute node may be replaced

### DIFF
--- a/roles/validations/defaults/main.yml
+++ b/roles/validations/defaults/main.yml
@@ -51,3 +51,9 @@ cifmw_validations_xml_status_file_dir: "{{ cifmw_validations_basedir }}/tests/va
 cifmw_validations_edpm_scale_down_hostname: compute-2.ctlplane.example.com
 cifmw_validations_edpm_scale_down_nodename: edpm-compute-2
 cifmw_validations_timeout: 100
+
+# variables needed for bmh compute replacement
+cifmw_validations_bmh_replace_leaf_label: leaf0-1
+cifmw_validations_bmh_spare_leaf_label: leaf0-0
+cifmw_validations_bmh_spare_nodename: edpm-compute-0-0
+cifmw_validations_bmh_spare_hostname: edpm-compute-0-0.ctlplane.openstack.lab
--- a/roles/validations/tasks/edpm/bmh_compute_replace.yml
+++ b/roles/validations/tasks/edpm/bmh_compute_replace.yml
@@ -1,0 +1,196 @@
+# This job tests the functionality of the openstack-operator to replace a
+# bmh compute node.
+#
+# This job was created to satisfy:
+# https://issues.redhat.com/browse/OSPRH-15061
+
+- name: Get name of nodeset containing bmh node to be used as spare node
+  cifmw.general.ci_script:
+    output_dir: "{{ cifmw_validations_basedir }}/artifacts"
+    script: >-
+      oc -n {{ cifmw_validations_namespace }} get bmh {{ cifmw_validations_bmh_spare_leaf_label }} -o jsonpath='{.spec.consumerRef.name}'
+  register: bmh_nodeset_name
+
+- name: Verify bmh node being used as spare is provisioned before scale down
+  cifmw.general.ci_script:
+    output_dir: "{{ cifmw_validations_basedir }}/artifacts"
+    script: >-
+      oc -n {{ cifmw_validations_namespace }} get bmh {{ cifmw_validations_bmh_spare_leaf_label }} -o jsonpath='{.status.provisioning.state}'
+  register: bmh_nodes_before_scale_down
+  failed_when: bmh_nodes_before_scale_down.stdout != "provisioned"
+
+- name: Get compute service list
+  cifmw.general.ci_script:
+    output_dir: "{{ cifmw_validations_basedir }}/artifacts"
+    script: >-
+      oc -n {{ cifmw_validations_namespace }} rsh openstackclient openstack compute service list
+  register: compute_service_list_out
+  until: '"{{ cifmw_validations_bmh_spare_hostname }}" in compute_service_list_out.stdout'
+
+- name: Disable nova-compute for node being removed
+  cifmw.general.ci_script:
+    output_dir: "{{ cifmw_validations_basedir }}/artifacts"
+    script: >-
+      oc -n {{ cifmw_validations_namespace }} rsh openstackclient openstack compute service set {{ cifmw_validations_bmh_spare_hostname }} nova-compute --disable
+  register: compute_service_set_out
+  until: '"Failed" not in compute_service_set_out.stdout'
+
+- name: Get ovn controller id of host to be removed
+  cifmw.general.ci_script:
+    output_dir: "{{ cifmw_validations_basedir }}/artifacts"
+    script: >-
+      oc -n {{ cifmw_validations_namespace }} rsh openstackclient openstack network agent list --host {{ cifmw_validations_bmh_spare_hostname }} | grep "OVN Controller agent"  | awk '{print $2}'
+  register: remove_ovn_id
+
+- name: Delete network agent for compute being removed
+  cifmw.general.ci_script:
+    output_dir: "{{ cifmw_validations_basedir }}/artifacts"
+    script: >-
+      oc -n {{ cifmw_validations_namespace }} rsh openstackclient openstack network agent delete {{ remove_ovn_id.stdout }}
+
+- name: Get compute service id of host to be removed
+  cifmw.general.ci_script:
+    output_dir: "{{ cifmw_validations_basedir }}/artifacts"
+    script: >-
+      oc -n {{ cifmw_validations_namespace }} rsh openstackclient openstack compute service list --host {{ cifmw_validations_bmh_spare_hostname }} -f value -c ID
+  register: remove_compute_service_id
+
+- name: Delete compute service for node being removed
+  cifmw.general.ci_script:
+    output_dir: "{{ cifmw_validations_basedir }}/artifacts"
+    script: >-
+      oc -n {{ cifmw_validations_namespace }} rsh openstackclient openstack compute service delete {{ remove_compute_service_id.stdout }}
+
+- name: Patch nodeset to remove node
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  cifmw.general.ci_script:
+    output_dir: "{{ cifmw_validations_basedir }}/artifacts"
+    script: >-
+      oc patch -n {{ cifmw_validations_namespace }} osdpns/"{{ bmh_nodeset_name.stdout | trim}}" --type=json --patch '[{ "op": "remove", "path": "/spec/nodes/{{ cifmw_validations_bmh_spare_nodename }}" }]'
+
+- name: Wait for nodeset to be SetupReady again
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  cifmw.general.ci_script:
+    output_dir: "{{ cifmw_validations_basedir }}/artifacts"
+    script: >-
+      oc wait osdpns "{{ bmh_nodeset_name.stdout | trim }}"
+      --namespace={{ cifmw_validations_namespace }}
+      --for=condition=SetupReady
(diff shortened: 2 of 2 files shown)


roles/validations/tasks/edpm/bmh_compute_replace.yml
```diff
@@ -0,0 +1,196 @@
...
+# bmh compute node.
+#
+# This job was created to satisfy:
+# https://issues.redhat.com/browse/OSPRH-15061
+
+- name: Get name of nodeset containing bmh node to be used as spare node
+  cifmw.general.ci_script:
+    output_dir: "{{ cifmw_validations_basedir }}/artifacts"
+    script: >-
+      oc -n {{ cifmw_validations_namespace }} get bmh {{ cifmw_validations_bmh_spare_leaf_label }} --no-headers | awk '{print $3}'
```
bshephar, line 11: The awk is ok, but you can also do this with just `oc`:
```
$ oc get bmh --no-headers -o custom-columns=":metadata.name"
bshephar-edpm-compute-01
bshephar-edpm-compute-02
bshephar-edpm-compute-03
```

roles/validations/tasks/edpm/bmh_compute_replace.yml
```diff
@@ -0,0 +1,196 @@
...
+  register: bmh_nodeset_name
+
+- name: Verify bmh node being used as spare is provisioned before scale down
+  cifmw.general.ci_script:
+    output_dir: "{{ cifmw_validations_basedir }}/artifacts"
+    script: >-
+      oc -n {{ cifmw_validations_namespace }} get bmh {{ cifmw_validations_bmh_spare_leaf_label }}
+  register: bmh_nodes_before_scale_down
+  failed_when: '"provisioned" not in bmh_nodes_before_scale_down.stdout'
```
bshephar, line 20:
```suggestion
      oc -n {{ cifmw_validations_namespace }} get bmh {{ cifmw_validations_bmh_spare_leaf_label }} -o jsonpath='{.status.provisioning.state}'
  register: bmh_nodes_before_scale_down
  failed_when: bmh_nodes_before_scale_down != "provisioned"
```

I think that just makes the `failed_when` conditional a little more reliable. Just in case the stdout is something unexpected that just happens to container `provisioned`.

roles/validations/defaults/main.yml
```diff
@@ -51,3 +51,9 @@ cifmw_validations_xml_status_file_dir: "{{ cifmw_validations_basedir }}/tests/va
 cifmw_validations_edpm_scale_down_hostname: compute-2.ctlplane.example.com
 cifmw_validations_edpm_scale_down_nodename: edpm-compute-2
 cifmw_validations_timeout: 100
+
+# variables needed for bmh compute replacement
+cifmw_validations_bmh_replace_leaf_label: leaf0-1
+cifmw_validations_bmh_spare_leaf_label: leaf0-0
```
bshephar, line 57: Is this actually a `label`? Or is it the node name? For example, we typically use `app: openstack` as a label:
```
[m3@osp-df-3 beta]$ oc get bmh -o yaml bshephar-edpm-compute-01 | yq --yaml-output .metadata.labels
app: openstack
openstack-edpm-ipam-osbms-hostname: edpm-compute-0
openstackbaremetalset.openstack.org/name: openstack-edpm-ipam
openstackbaremetalset.openstack.org/namespace: openstack
openstackbaremetalset.openstack.org/uid: c80bb9b7-a04c-4407-bba2-965ce7db1984
```

So, to list nodes using the label we could:
```
[m3@osp-df-3 beta]$ oc get bmh -l app=openstack
NAME                       STATE         CONSUMER               ONLINE   ERROR   AGE
bshephar-edpm-compute-01   provisioned   openstack-edpm-ipam    true             72d
bshephar-edpm-compute-02   available                            false            73d
bshephar-edpm-compute-03   provisioned   openstack-edpm-bootc   true             70d
```

But looking at the `oc get bmh` commands used in this PR below, it seems like we're not searching for the nodes using a label selector using `-l`. If this is indeed the name, maybe `cifmw_validations_bmh_spare_leaf_nodename` ?

roles/validations/tasks/edpm/bmh_compute_replace.yml
```diff
@@ -0,0 +1,196 @@
...
+    script: >-
+      oc -n {{ cifmw_validations_namespace }} rsh openstackclient openstack compute service set {{ cifmw_validations_bmh_spare_hostname }} nova-compute --disable
+  register: compute_service_set_out
+  until: '"Failed" not in compute_service_set_out.stdout'
+
+- name: Get ovn controller id of host to be removed
+  cifmw.general.ci_script:
+    output_dir: "{{ cifmw_validations_basedir }}/artifacts"
+    script: >-
+      oc -n {{ cifmw_validations_namespace }} rsh openstackclient openstack network agent list --host {{ cifmw_validations_bmh_spare_hostname }} | grep "OVN Controller agent"  | awk '{print $2}'
```
bshephar, line 42:
```suggestion
      oc -n {{ cifmw_validations_namespace }} rsh openstackclient openstack network agent list --host {{ cifmw_validations_bmh_spare_hostname }} -f value -c ID
```

That will just give you the value:
```
oc rsh openstackclient openstack network agent list --host edpm-compute-1.ctlplane.example.com -f value -c ID
945d481a-20c4-45d3-8c92-3b2c04ba628f
```

roles/validations/tasks/edpm/bmh_compute_replace.yml
```diff
@@ -0,0 +1,196 @@
...
+  cifmw.general.ci_script:
+    output_dir: "{{ cifmw_validations_basedir }}/artifacts"
+    script: >-
+      oc -n {{ cifmw_validations_namespace }} rsh openstackclient openstack network agent delete {{ remove_ovn_id.stdout }}
+
+- name: Get compute service id of host to be removed
+  cifmw.general.ci_script:
+    output_dir: "{{ cifmw_validations_basedir }}/artifacts"
+    script: >-
+      oc -n {{ cifmw_validations_namespace }} rsh openstackclient openstack compute service list --host {{ cifmw_validations_bmh_spare_hostname }} | grep "nova-compute"  | awk '{print $2}'
```
bshephar, line 55:
```suggestion
      oc -n {{ cifmw_validations_namespace }} rsh openstackclient openstack compute service list --host {{ cifmw_validations_bmh_spare_hostname }} -f value -c ID
```

Same for this one:

```
$ oc rsh openstackclient openstack compute service list --host edpm-compute-1.ctlplane.example.com -f value -c ID
3e29ed72-c314-4c21-bfc4-93b7afda69c0
```

roles/validations/tasks/edpm/bmh_compute_replace.yml
```diff
@@ -0,0 +1,196 @@
...
+      --timeout={{ cifmw_validations_timeout }}m
+
+- name: Patch spare bmh node to change its label to match label of node being replaced
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  cifmw.general.ci_script:
+    output_dir: "{{ cifmw_validations_basedir }}/artifacts"
+    script: >-
+      oc patch -n {{ cifmw_validations_namespace }} bmh/"{{ cifmw_validations_bmh_spare_leaf_label }}" --type=json --patch '[{ "op": "replace", "path": "/metadata/labels/nodeName", "value": "{{ cifmw_validations_bmh_replace_leaf_label }}" }]'
```
bshephar, line 92: Ahh, right, so it is the nodeName, but it's also being used as the label.

roles/validations/tasks/edpm/bmh_compute_replace.yml
```diff
@@ -0,0 +1,196 @@
...
+
+- name: Verify bmh node being used as spare is available after changing label
+  cifmw.general.ci_script:
+    output_dir: "{{ cifmw_validations_basedir }}/artifacts"
+    script: >-
+      oc -n {{ cifmw_validations_namespace }} get bmh {{ cifmw_validations_bmh_spare_leaf_label }}
+  register: bmh_nodes_after_scale_down
+  until: '"available" in bmh_nodes_after_scale_down.stdout'
```
bshephar, line 100:
```suggestion
      oc -n {{ cifmw_validations_namespace }} get bmh {{ cifmw_validations_bmh_spare_leaf_label }} -o jsonpath='{.status.provisioning.state}'
  register: bmh_nodes_after_scale_down
  until: bmh_nodes_after_scale_down == "available"
```

bshephar (commented):
Looks good. I just added a couple of suggestions that might be a bit more reliable than checking stdout for strings. My suggestions just focus on getting the specific information you want from the initial compared to make the conditionals a bit more simplified.

My thoughts are that it will probably result in the jobs being a bit more reliable.

roles/validations/defaults/main.yml
drosenfe: I got label from this review: https://github.com/openstack-k8s-operators/architecture/pull/434/files   It does look like nodeName it is what its called and label is for the whole code block.

drosenfe: The above review also adds a bmhLabelSelector value which is the nodeName. I also see the comment below about this. I'm not certain what I should do. I can name the variable as you would like. Should I go with the suggested: cifmw_validations_bmh_spare_leaf_nodename or leave as it is?

roles/validations/tasks/edpm/bmh_compute_replace.yml
drosenfe: The output of the current command with the awk is:

oc -n openstack get bmh leaf0-1 --no-headers | awk '{print $3}'
nodeset-0

i.e the validator needs the nodeset name and not the compute list. There could be a better option than the awk, but I'm not finding it. Tried this since the nodeset is in the CONSUMER column:

oc get bmh leaf0-1--no-headers -o custom-columns=":metadata.consumer"
but it returned none

If there is a better way to get the nodeset name when there are multiple nodesets then I could change it.

roles/validations/tasks/edpm/bmh_compute_replace.yml
drosenfe: I'll try it. From command line it looks like it should work:
oc -n openstack get bmh leaf0-1 -o jsonpath='{.status.provisioning.state}'
provisioned

roles/validations/tasks/edpm/bmh_compute_replace.yml
drosenfe: Using -f value -c ID only returns one id so it should be OK.

roles/validations/tasks/edpm/bmh_compute_replace.yml
drosenfe: I'll try it. From command line looks like it should work.

roles/validations/tasks/edpm/bmh_compute_replace.yml
drosenfe: Similar to other comment on label vs name. Leaving names the same with same reason.

roles/validations/tasks/edpm/bmh_compute_replace.yml
bshephar: Ahh right, ok. That makes sense:
```
oc -n {{ cifmw_validations_namespace }} get bmh {{ cifmw_validations_bmh_spare_leaf_label } -o jsonpath='{.spec.consumerRef.name}'
```

drosenfe: done

roles/validations/tasks/edpm/bmh_compute_replace.yml
drosenfe: done

roles/validations/tasks/edpm/bmh_compute_replace.yml
drosenfe: done

roles/validations/tasks/edpm/bmh_compute_replace.yml
drosenfe: done

bshephar (approved):

drosenfe:
@danpawlik   This is for a validator that is specific to edpm. The DFG has approved it. Need a second approver. Anyway you could do that?

frenzyfriday (approved):

PR #2889, opened 2025-04-10
[cifmw_external_dns] Check better for the secret

jirimacku:
We see often the task exits the retry loop but we don't have the secrets and the subsequent Populate key task fails.


jirimacku:
Verified with unigamma job.

danpawlik (approved):

frenzyfriday (approved):

PR #2890, opened 2025-04-10
DNM Flattening kuttl job execution

danpawlik:
Executing kuttl job locally is very difficult, due the Ansible execution is nested in many places.

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/2891


danpawlik:
recheck

danpawlik:
That would not work. Many operations needs to be done on ansible host. 
What can be done to improve that whole execution:
- remove parsing zuul inventory file 
- remove at all "connection" to Zuul. Right now, most of the code can not be executed without Zuul/Zuul inventory etc.

PR #2891, opened 2025-04-11
Set UTC timezone on all hosts

danpawlik:
There are different timezone sets on hosts, so during the CI job execution, logs shows different time, which might be confusing.


danpawlik:
I think it needs to be done in config repo, due beginning of the job is done by base job.

frenzyfriday (approved):

frenzyfriday (approved):
Didnt @amartyasinha have a similar patch some time back?
But anyways, lgtm!

danpawlik:
Let's do it in base job to have better time view from the beginning. 
https://review.rdoproject.org/r/c/config/+/57421

PR #2893, opened 2025-04-11
Change hardcoded eth0 value in multinode-customizations play

danpawlik:
The hardcoded value for controller default network interface is making problems on local deployment of kuttl job.
With that patch, few steps would be less to execute.

### DIFF
--- a/ci/playbooks/multinode-customizations.yml
+++ b/ci/playbooks/multinode-customizations.yml
@@ -279,7 +279,9 @@
     - name: Get the default iface connection
       register: controller_default_connection_out
       ansible.builtin.command:
-        cmd: "nmcli -g general.connection device show eth0"
+        cmd: >-
+          nmcli -g general.connection
+          device show {{ cifmw_controller_interface_name | default('eth0') }}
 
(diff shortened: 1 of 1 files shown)


evallesp (approved):
LGTM

ci/playbooks/multinode-customizations.yml
```diff
@@ -279,7 +279,9 @@
     - name: Get the default iface connection
       register: controller_default_connection_out
       ansible.builtin.command:
-        cmd: "nmcli -g general.connection device show eth0"
+        cmd: >-
+          nmcli -g general.connection
+          device show {{ cifmw_controller_interface_name | default('eth0') }}
```
frenzyfriday, line 284: We will need to add this to the job def of the kuttl job in a follow up patch, right?

frenzyfriday (approved):

ci/playbooks/multinode-customizations.yml
danpawlik: I don't think so, but we can add as default value...

PR #2894, opened 2025-04-14
Add junit Ansible callback

danpawlik:
The junit callback might be interesting to display better Ansible tasks execution overview.


danpawlik:
recheck

PR #2895, opened 2025-04-14
[kustomize_deploy] use recursive=True to combine user kustomize values

eduolivares:
Currently ci-fmw supports defining user kustomize values for a certain architecture step referring to that step with stage_\<number\> or <stage_name>.
If those values are provided using the both available methods, they should be combined using recursive=True.

[OSPRH-15775](https://issues.redhat.com//browse/OSPRH-15775)


dasm (approved):
FTR: [recursive doc](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/combine_filter.html#parameter-recursive)

PR #2903, opened 2025-04-15
[OSPNW-910] Add kustomize template for uni04delta-ipv6-adoption

abays:
This is needed to make https://github.com/openstack-k8s-operators/architecture/pull/517 work as intended

### DIFF
--- a/roles/ci_gen_kustomize_values/templates/uni04delta-ipv6-adoption/network-values/values.yaml.j2
+++ b/roles/ci_gen_kustomize_values/templates/uni04delta-ipv6-adoption/network-values/values.yaml.j2
@@ -1,0 +1,147 @@
+---
+# source: uni04delta-ipv6-adoption/network-values/values.yaml.j2
+{% set _ipv = cifmw_ci_gen_kustomize_values_ip_version_var_mapping %}
+{% set ns = namespace(interfaces={},
+                      ocp_index=0,
+                      lb_tools={}) %}
+data:
+{% for host in cifmw_networking_env_definition.instances.keys() -%}
+{%   if host is match('^(ocp|crc).*') %}
+  node_{{ ns.ocp_index }}:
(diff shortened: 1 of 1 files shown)


roles/ci_gen_kustomize_values/templates/uni04delta-ipv6-adoption/network-values/values.yaml.j2
```diff
@@ -0,0 +1,147 @@
...
+    {{ network.network_name }}_ip: {{ network[_ipv.ip_vX] }}
+{%     endfor %}
+{%   endif %}
+{% endfor %}
+
+{% for network in cifmw_networking_env_definition.networks.values() %}
+{% set ns.lb_tools = {} %}
+  {{ network.network_name }}:
+    dnsDomain: {{ network.search_domain }}
+{%  if network.tools is defined and network.tools.keys() | length > 0 %}
```
danpawlik, line 32: that condition should not look like: `if network and 'tools' in network and network.tools.keys() | length > 0`?

danpawlik:
I don't think that yaml is rendered "beautiful" but for sure should be readable by other tool.

abays:
> I don't think that yaml is rendered "beautiful" but for sure should be readable by other tool.

This entire template is a copy-paste of [1].  If someone wants to clean both of them up, I think that can be done in a follow-up.

[1] https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/ci_gen_kustomize_values/templates/uni04delta-ipv6/network-values/values.yaml.j2

frenzyfriday (approved):

roles/ci_gen_kustomize_values/templates/uni04delta-ipv6-adoption/network-values/values.yaml.j2
abays: We'll circle back to this in follow-ups if need be

PR #2905, opened 2025-04-16
[tobiko] Replaced ubuntu advanced guest image with fedora

eduolivares:
With recent tobiko changes, customized fedora images are supported to test QoS, VLANs, etc.


roles/test_operator/defaults/main.yml
```diff
@@ -161,7 +161,7 @@ cifmw_test_operator_tobiko_version: master
 cifmw_test_operator_tobiko_pytest_addopts: null
 cifmw_test_operator_tobiko_prevent_create: null
 cifmw_test_operator_tobiko_num_processes: null
-cifmw_test_operator_tobiko_advanced_image_url: "https://softwarefactory-project.io/ubuntu-minimal-customized-enp3s0"
+cifmw_test_operator_tobiko_advanced_image_url: "https://kaplonski.pl/files/Customized-Fedora-Cloud-Base-Generic-41-1.4.x86_64.qcow2"
```
danpawlik, line 164: <3

danpawlik: In that case, I suggest to consider to make a Github workflow.

eduolivares: what do you mean?
To test this image I was planning to use a tcib job:
https://github.com/openstack-k8s-operators/tcib/pull/249
Although I could create a job for ci-fmw too.

danpawlik: So IMHO we should:
- create a new project in openstack-k8s-operators e.g. test-images
- create github workflows (or just temporary upload @slawqo  image there  and make release) that will do do virt customize + make release in the end.

dasm:
I removed `Ready for Review` label as it clearly has another `DNM` label.

PR #2906, opened 2025-04-16
Update CRDs when changing test-operator version

kstrenkova:
This patch updates the CRDs, if the user wants to patch the test operator version. Until now the CRDs stayed unchanged even after version change, so testing new test-operator parameters was not possible. It is important to point out, that the update will not be triggered in 90% of test-operator use cases. It is primarily used in two situations:
1. When running checks for new PRs in the test-operator repository.    
2. When updating the test-operator version in jobs to prevent failures.

The approach isn’t the cleanest, but alternative solutions are significantly more complex than the benefits they offer.

### DIFF
--- a/roles/test_operator/tasks/main.yml
+++ b/roles/test_operator/tasks/main.yml
@@ -118,6 +118,45 @@
           )
         }}
 
+    - name: Update existing CRDs
+      when: cifmw_test_operator_version is defined
+      block:
+        - name: Delete CRDs created by test-operator
+          kubernetes.core.k8s:
+            kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
+            api_key: "{{ cifmw_openshift_token | default(omit)}}"
+            context: "{{ cifmw_openshift_context | default(omit)}}"
+            kind: CustomResourceDefinition
+            state: absent
+            api_version: v1
+            name: "{{ item }}"
+            namespace: "{{ cifmw_test_operator_namespace }}"
+            wait: true
+            wait_timeout: 600
+          loop:
+            - "{{ cifmw_test_operator_tempest_crd_name }}"
+            - "{{ cifmw_test_operator_tobiko_crd_name }}"
+            - "{{ cifmw_test_operator_ansibletest_crd_name }}"
+            - "{{ cifmw_test_operator_horizontest_crd_name }}"
+
+        - name: Clone test-operator repository and checkout into specified version
+          ansible.builtin.git:
+            repo: "https://github.com/openstack-k8s-operators/test-operator.git"
+            dest: /tmp/test-operator
+            refspec: '+refs/pull/*:refs/heads/*'
+            version: "{{ cifmw_test_operator_version }}"
+            force: true
+
+        - name: Run make generate manifests install from /tmp/test-operator dir
+          environment:
+            KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+            PATH: "{{ cifmw_path | default(ansible_env.PATH) }}"
+          ansible.builtin.shell: >-
+            set -o pipefail;
+            make generate manifests install
+          args:
+            chdir: /tmp/test-operator
+
 - name: Call test stages loop
   when: not cifmw_test_operator_dry_run | bool
   ansible.builtin.include_tasks: stages.yml
--- a/roles/test_operator/README.md
+++ b/roles/test_operator/README.md
@@ -5,8 +5,9 @@ Execute tests via the [test-operator](https://openstack-k8s-operators.github.io/
 ## Parameters
 * `cifmw_test_operator_artifacts_basedir`: (String) Directory where we will have all test-operator related files. Default value: `{{ cifmw_basedir }}/tests/test_operator` which defaults to `~/ci-framework-data/tests/test_operator`
 * `cifmw_test_operator_namespace`: (String) Namespace inside which all the resources are created. Default value: `openstack`
-* `cifmw_test_operator_controller_namespace`: (String) Namespace inside which the test-operator-controller-manager is created. Default value: `openstack-opearators`
+* `cifmw_test_operator_controller_namespace`: (String) Namespace inside which the test-operator-controller-manager is created. Default value: `openstack-operators`
 * `cifmw_test_operator_bundle`: (String) Full name of container image with bundle that contains the test-operator. Default value: `""`
+* `cifmw_test_operator_version`: (String) The commit hash corresponding to the version of test-operator the user wants to use. This parameter is only used when `cifmw_test_operator_bundle` is also set.
 * `cifmw_test_operator_timeout`: (Integer) Timeout in seconds for the execution of the tests. Default value: `3600`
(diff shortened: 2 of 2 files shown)


roles/test_operator/tasks/main.yml
```diff
@@ -118,6 +118,50 @@
...
+        namespace: "{{ cifmw_test_operator_namespace }}"
+        wait: true
+        wait_timeout: 600
+      loop:
+        - tempests.test.openstack.org
+        - tobikoes.test.openstack.org
+        - ansibletests.test.openstack.org
+        - horizontests.test.openstack.org
+
+    - name: Install required packages
```
danpawlik, line 139: Nit: I don't think that's necessary.

roles/test_operator/tasks/main.yml
```diff
@@ -118,6 +118,50 @@
...
+
+    - name: Clone test-operator repository and checkout into specified version
+      ansible.builtin.git:
+        repo: "https://github.com/openstack-k8s-operators/test-operator.git"
+        dest: /tmp/test-operator
+        refspec: '+refs/pull/*:refs/heads/*'
+        version: "{{ cifmw_test_operator_version }}"
+        force: true
+
+    - name: Run make generate manifests install
```
danpawlik, line 155: nit: suggest to add execution path for the title.
E.g.: `- name: Run make generate manifests install from /tmp/test-operator dir`

roles/test_operator/tasks/main.yml
```diff
@@ -118,6 +118,50 @@
...
+
+    - name: Install required packages
+      become: true
+      ansible.builtin.package:
+        state: present
+        name:
+          - git
+          - make
+
+    - name: Clone test-operator repository and checkout into specified version
```
danpawlik, line 147: there should not be a condition like: `when: cifmw_test_operator_version` ?

roles/test_operator/tasks/main.yml
```diff
@@ -118,6 +118,50 @@
...
+      ansible.builtin.git:
+        repo: "https://github.com/openstack-k8s-operators/test-operator.git"
+        dest: /tmp/test-operator
+        refspec: '+refs/pull/*:refs/heads/*'
+        version: "{{ cifmw_test_operator_version }}"
+        force: true
+
+    - name: Run make generate manifests install
+      environment:
```
danpawlik, line 156: there should not be a condition like: `when: cifmw_test_operator_version` ?

roles/test_operator/tasks/main.yml
kstrenkova: Thank you for the review! Good catch, it needs the test-operator version to work.

My thought behind it, was that the whole block is supposed to work only when we have both `cifmw_test_operator_bundle` and `cifmw_test_operator_version`. But now that you mention it, maybe we could run the block with bundle only, if someone doesn't want to update CRDs.

roles/test_operator/tasks/main.yml
kstrenkova: I was curious if it was necessary. But I saw some tasks that installed make, so I left it there. I can try removing it for the next run.

danpawlik (approved):

roles/test_operator/tasks/main.yml
```diff
@@ -118,6 +118,45 @@
...
+            context: "{{ cifmw_openshift_context | default(omit)}}"
+            kind: CustomResourceDefinition
+            state: absent
+            api_version: v1
+            name: "{{ item }}"
+            namespace: "{{ cifmw_test_operator_namespace }}"
+            wait: true
+            wait_timeout: 600
+          loop:
+            - tempests.test.openstack.org
```
arxcruz, line 137: This should be a variable that we can change

kstrenkova: Good catch. I checked and we should be able to use `cifmw_test_operator_tempest_crd_name` and corresponding variables instead of fixed values.

kstrenkova: Okay, so I misunderstood the review. The point was to create a variable with all the CRDs. In my opinion this is not needed as adding a new resource type will be maybe created once in a long while.

roles/test_operator/tasks/main.yml
```diff
@@ -118,6 +118,45 @@
...
+            namespace: "{{ cifmw_test_operator_namespace }}"
+            wait: true
+            wait_timeout: 600
+          loop:
+            - "{{ cifmw_test_operator_tempest_crd_name }}"
+            - "{{ cifmw_test_operator_tobiko_crd_name }}"
+            - "{{ cifmw_test_operator_ansibletest_crd_name }}"
+            - "{{ cifmw_test_operator_horizontest_crd_name }}"
+
+        - name: Clone test-operator repository and checkout into specified version
```
dasm, line 142: I'm wondering, how are we getting the `test-operator` at the first place. Is it shipped as the github repository? Or is it downloaded from the image hub?
This is very risky. Every time when the new change is introduced to the main branch, we might end up with overwriting existing tests. It allows for faster iteration, but it's not safe.

dasm (changes requested):

roles/test_operator/tasks/main.yml
kstrenkova: Thank you for the review! To clarify what is the thought behind this patch, I want to highlight that this task will be not be called frequently. As it is said in the commit message, it is primarily used in two situations:
1. When running checks for new PRs in the test-operator repository.
2. When updating the test-operator version in jobs to prevent failures.

I am aware this approach is not the cleanest, but the complexity of the solution would rise quite a lot with a different working approach. When you take in consideration that the task will not be called frequently, I thinks this solution is a good middle ground. What do you think?

dasm: Your explanation makes perfect sense. I'd love to see an alternative approach. But I understand it might be not feasible.

roles/test_operator/tasks/main.yml
```diff
@@ -118,6 +118,45 @@
...
+            version: "{{ cifmw_test_operator_version }}"
+            force: true
+
+        - name: Run make generate manifests install from /tmp/test-operator dir
+          environment:
+            KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+            PATH: "{{ cifmw_path | default(ansible_env.PATH) }}"
+          ansible.builtin.shell: >-
+            set -o pipefail;
+            make generate manifests install
```
dasm, line 156: @kstrenkova please forgive my ignorance. I assumed `make generate manifests install` is located in [install_yamls](https://github.com/openstack-k8s-operators/install_yamls/tree/main) repository. However I cannot find it. Would you be able to point me where to look for it?
Side-note (maybe it's answered inside of the `make generate`): Should you wait for the manifests to be applied?

kstrenkova: No problem. The `make generate manifests install` is located in the test-operator Makefile. The way this patch is done, is the same way a user would locally prepare and install test-operator. That is why I think it's safe to say it is done sequentially and there is no need to add wait (if I understood you question correctly :D)

dasm (approved):

kstrenkova:
I rebased the change with the current main. I hope all the jobs pass, so we can merge it.

danpawlik (approved):

PR #2907, opened 2025-04-17
DNM: Use $PIP instead of explicit $SYSTEM_PIP

amartyasinha:
This patch is the first step in enforcing the use of venv for cifmw. We want that all deps of cifmw should be within it's venv and should not consume system's pip.

It is important as from python3.12 onwards, use of system wide packages are discouraged, and it is recommended to use venv.
Ref: https://issues.redhat.com/browse/OSPRH-15882

Local test did not face any error due to this change. Will do some tests through jobs on this PR.


amartyasinha:
Closing this for now, as this change should be one of the last steps. Investigating how we can do that. Closing it for now.

amartyasinha:
recheck

danpawlik:
recheck

roles/reproducer/tasks/configure_controller.yml
```diff
@@ -354,8 +354,11 @@
       register: _async_dep_install
       async: 600  # 10 minutes should be more than enough
       poll: 0
-      ansible.builtin.pip:
-        requirements: https://raw.githubusercontent.com/openstack-k8s-operators/ci-framework/main/common-requirements.txt
+      community.general.make:
+        chdir: "{{ ansible_user_dir }}/{{ zuul.projects['github.com/openstack-k8s-operators/ci-framework'].src_dir }}"
+        target: setup_molecule
```
dasm, line 359: Q: Should it be `setup_env` instead of `setup_molecule`?
(I see we usl a lot of `setup_molecule` in places which are not really `molecule`)

amartyasinha: All previous `make` call to install venv and deps were using `setup_molecule`, so I also used the same. Will look into the difference, and update it.

amartyasinha:
recheck

PR #2908, opened 2025-04-18
Fix get cri-o stats script

danpawlik:
The script was not providing stats properly, because the oc binary is located in different place.

### DIFF
--- a/scripts/get-stats.sh
+++ b/scripts/get-stats.sh
@@ -21,12 +21,24 @@ set -x
 
 DURATION_TIME=${DURATION_TIME:-10}
 
-NODE_NAMES=$(/usr/local/bin/oc get node -o name -l node-role.kubernetes.io/worker)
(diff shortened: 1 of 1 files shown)


scripts/get-stats.sh
```diff
@@ -21,12 +21,17 @@ set -x
 
 DURATION_TIME=${DURATION_TIME:-10}
 
-NODE_NAMES=$(/usr/local/bin/oc get node -o name -l node-role.kubernetes.io/worker)
+if ! [ -f "/home/zuul/bin/oc" ]; then
```
dasm, line 24: Instead of doing that, I'd recommend something like:
```
PATH=$PATH:/home/zuul/bin
```
Then you would be able to just call `oc` command without the full path.

amartyasinha (approved):

evallesp (approved):

PR #2909, opened 2025-04-18
Removed variant of cifmw-molecule-tofu job

dasm:
Multiple job definitions with the same name are called variants. These may have different selection criteria which indicate to Zuul that, for instance, the job should behave differently on a different git branch. In this case, we have the same jobs, doing almost the same thing.

Removed duplicated job definition.


ci/config/molecule.yaml
```diff
@@ -98,3 +98,8 @@
 - job:
     name: cifmw-molecule-shiftstack
     nodeset: centos-9-crc-2-48-0-xl-ibm
+- job:
+    name: cifmw-molecule-tofu
+    nodeset: centos-9-crc-2-48-0-xl
```
danpawlik, line 103: nit: I don't think setting nodeset is needed here.

danpawlik (approved):

ci/config/molecule.yaml
dasm: I'm not sure about that. I just left what was initially done [here](https://github.com/openstack-k8s-operators/ci-framework/pull/2909/files#diff-897128cff4fe0d5b2c8aba92a2c8b3ed825cc44ba495b87a1ebe5353d0e42a7fL9)

dasm:
recheck

udis (approved):

PR #2910, opened 2025-04-21
Increase ControlPersist timeout to 300 seconds

eduolivares:
This ControlPersist socket periodically closing seems to be the cause of intermittent MODULE FAULIRE, happening sometimes in downstream CI.
Increase the timeout to make this less likely to happen.

See https://github.com/ansible/ansible/issues/78344

[OSPRH-15947](https://issues.redhat.com//browse/OSPRH-15947)

### DIFF
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -16,4 +16,4 @@ inventory = inventory.yml
 pipelining = True
 any_errors_fatal = True
 [ssh_connection]
-ssh_args = -o ControlMaster=auto -o ControlPersist=60s
+ssh_args = -o ControlMaster=auto -o ControlPersist=300


ansible.cfg
```diff
@@ -16,4 +16,4 @@ inventory = inventory.yml
 pipelining = True
 any_errors_fatal = True
 [ssh_connection]
-ssh_args = -o ControlMaster=auto -o ControlPersist=60s
+ssh_args = -o ControlMaster=auto -o ControlPersist=300
```
cubeek, line 19: The original had "s" appended - I don't know if seconds are the default unit - just wanted to point it out :)

eduolivares: Thanks for pointing it out.
I copied it from:
https://github.com/openstack/openstack-ansible/blob/master/scripts/openstack-ansible.rc#L52

It seems all these "qualifiers" are equivalent to seconds: (none)|s|S
https://man.archlinux.org/man/sshd_config.5.en#TIME_FORMATS

eduolivares:
> nit: you increase it 5x time, not too much? Maybe 2x or 3x is enough? I approve, but something to consider to increase but not so drastically.

You may be right.
I just copied it from openstack-ansible:
https://github.com/openstack/openstack-ansible/blob/master/scripts/openstack-ansible.rc#L52

I assumed it should be fine for our CI, as well as it is fine for opendev.

dasm (approved):

PR #2911, opened 2025-04-21
Clean deployment resources as well as operators

eshulman2:
Allowed cleaning up deployment resources as well as operators.


roles/kustomize_deploy/tasks/cleanup.yml
```diff
@@ -14,14 +14,51 @@
...
+        cifmw_deploy_architecture_steps['stages'] |
+        reverse |
+        selectattr('build_output', 'defined') |
+        map(attribute='build_output') |
+        list
+      }}
+    _stages_crs_path: >-
+      {{
+        [cifmw_kustomize_deploy_kustomizations_dest_dir]
+        | product(_stages_crs)
```
evallesp, line 42: praise: Cool!

roles/kustomize_deploy/tasks/cleanup.yml
```diff
@@ -14,14 +14,51 @@
...
+        | map('join', '/')
+        | unique
+      }}
+    _operators_crs:
```
evallesp, line 46: (blocking) question: Don't we want to maintain: cifmw_kustomize_deploy_cp_dest_file for OSP control-plane resources?

evallesp (commented):
LGTM in general! I have a question of a missing variable that points to a generated CR file.

roles/kustomize_deploy/tasks/cleanup.yml
eshulman2: having that it is part of the files used in the deployment it will be in the generated `_stages_crs_path` var already and if it won't be then there was no controlplane deployed anyway. (unless you are familiar with any usecase in which it is the case in which case I will change it a bit)

evallesp: ack, thanks!

evallesp (dismissed):
LGTM

evallesp (approved):

PR #2912, opened 2025-04-21
Add insecure registries to DT/VA style deployments

jamepark4:
When attempting to create deployments that leverage images generated from tcib jobs, the process will fail due to the fact that the dynamically created registries IP address is not present in the permitted insecure list.  Adding that IP address to the list to allow deployment.


roles/openshift_setup/tasks/main.yml
```diff
@@ -168,6 +168,31 @@
...
+    definition:
+      apiVersion: config.openshift.io/v1
+      kind: Image
+      metadata:
+        name: cluster
+      spec:
+        registrySources:
+          insecureRegistries:
+            - "{{ cifmw_update_containers_registry }}"
+          allowedRegistries:
```
danpawlik, line 187: it will not raise any issue if they would be repeating? If not, it's fine.
Maybe setting fact  then `new_registry_var | unique`  should be helpful there.

jamepark4: Ack, will do.

danpawlik (approved):
thanks

frenzyfriday (approved):

PR #2913, opened 2025-04-21
Add scenario unidelta_ipv6 adoption

fyanac:
It's empty patch, the config will be defined in the config job.
but this file is required for the adoption_osp_deploy
role to run.


fyanac:
recheck

PR #2914, opened 2025-04-21
Use python from user's environment

dasm:
To allow for virtualenv usage, force the script to search for python set in an environment variable. By avoiding hard-coding, we don't need to modify system-wide requirements.

### DIFF
--- a/scripts/create_role_molecule.py
+++ b/scripts/create_role_molecule.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/env python3
 
 # Copyright Red Hat, Inc.
 # All Rights Reserved.


danpawlik (approved):

danpawlik (changes requested):
Could we apply for all files? After check `find . -type f -name '*.py' | xargs grep 'python'`  it seems many files could have such improvement.

dasm:
@danpawlik I tried to do so in my [prior commit](https://github.com/openstack-k8s-operators/ci-framework/commit/e09591140f4f0c6105a9be15ac94a5a3ecf48074). Unfortunately `ansible-test` doesn't allow to use `#!/usr/bin/env python` in modules.

danpawlik (approved):

udis (approved):

PR #2917, opened 2025-04-22
WIP Move common variable to group_vars

danpawlik:
There is no need to parse path, or call filters etc. to get the same value.


danpawlik:
We can do IMHO a big step, if we move some variables in ci-framework to be global. For sure, it would looks much better. According to the doc - https://zuul-ci.org/docs/zuul/latest/config/job.html#attr-job.include-vars.name - zuul will read the variables located in global dir, parse them and would be available later (that's the theory). For getting that, we need at least Zuul 11.2.0 (currently we have 11.0.0). I was doing tricks to avoid getting such feature - https://github.com/openstack-k8s-operators/ci-framework/pull/2917 - but... nested Ansible does not help at all.

danpawlik:
recheck

PR #2918, opened 2025-04-23
crawl_n_mask: Ensure keys of other datatypes are also handled

amartyasinha:
In crawl_n_mask, it was assumed that the keys in yaml files will be string. But, it is also possible that we might have keys as bool or int. It's better to typecast the keys to string during regex findall.


evallesp (commented):
LGTM, thanks for adding the test case. WDYT about adding a int and/or float.

evallesp (approved):
lgtm

dasm (approved):
LGTM

PR #2919, opened 2025-04-23
Reuse OCP cluster

eshulman2:
This patch is made to allow re-deploying openstack without having to re-deploy OCP every time.  This patch will allow optimizing CI jobs run time for NFV jobs.

### DIFF
--- a/deploy-edpm-reuse.yaml
+++ b/deploy-edpm-reuse.yaml
@@ -1,0 +1,102 @@
+---
+- name: Manage unique ID
+  ansible.builtin.import_playbook: playbooks/unique-id.yml
+
+- name: Reproducer prepare play
+  hosts: "{{ cifmw_target_host | default('localhost') }}"
+  gather_facts: true
+  pre_tasks:
+    - name: Prepare cleanup script
+      ansible.builtin.include_role:
+        name: reproducer
+        tasks_from: configure_cleanup.yaml
+
+    - name: Run Openstack cleanup
+      no_log: "{{ cifmw_nolog | default(true) | bool }}"
+      async: "{{ 7200 + cifmw_test_operator_timeout | default(3600) }}"  # 2h should be enough to deploy EDPM and rest for tests.
+      poll: 20
+      delegate_to: controller-0
+      ansible.builtin.command:
+        cmd: "/home/zuul/cleanup-architecture.sh"
+
+    - name: Inherit from parent scenarios if needed
+      ansible.builtin.include_tasks:
+        file: "ci/playbooks/tasks/inherit_parent_scenario.yml"
+
+    - name: Include common architecture parameter file
+      when:
+        - cifmw_architecture_scenario is defined
+        - cifmw_architecture_scenario | length > 0
+      ansible.builtin.include_vars:
+        file: "scenarios/reproducers/va-common.yml"
+
+    - name: Run reproducer validations
+      ansible.builtin.import_role:
+        name: reproducer
+        tasks_from: validations
+
+    - name: Gather OS facts
+      ansible.builtin.setup:
+        gather_subset:
+          - "!all"
+          - "!min"
+          - "distribution"
+
+    - name: Tweak dnf configuration
+      become: true
+      community.general.ini_file:
+        no_extra_spaces: true
+        option: "{{ config.option }}"
+        path: "/etc/dnf/dnf.conf"
+        section: "{{ config.section | default('main') }}"
+        state: "{{ config.state | default(omit) }}"
+        value: "{{ config.value | default(omit) }}"
+        mode: "0644"
+      loop: "{{ cifmw_reproducer_dnf_tweaks }}"
+      loop_control:
+        label: "{{ config.option }}"
+        loop_var: 'config'
+
+    - name: Install custom CA if needed
+      ansible.builtin.import_role:
+        name: install_ca
+
+    - name: Setup repositories via rhos-release if needed
+      tags:
+        - packages
+      when:
+        - ansible_facts['distribution'] == 'RedHat'
+        - cifmw_reproducer_hp_rhos_release | bool
+      vars:
+        cifmw_repo_setup_output: /etc/yum.repos.d
+        cifmw_repo_setup_rhos_release_args: "rhel"
+      ansible.builtin.import_role:
+        name: repo_setup
+        tasks_from: rhos_release
+
+  roles:
+    - role: ci_setup
+
+- name: Prepare switches
+  vars:
+    cifmw_configure_switches: "{{ 'switches' in groups }}"
+  ansible.builtin.import_playbook: playbooks/switches_config.yml
+
+- name: Reproducer reuse run
+  hosts: "{{ cifmw_target_host | default('localhost') }}"
+  gather_facts: false
+  tasks:
+    - name: Run reproducer reuse playbook
+      ansible.builtin.include_role:
+        name: reproducer
+        tasks_from: reuse_main
+
+    - name: Run deployment if instructed to
+      when:
+        - cifmw_deploy_architecture | default(false) | bool
+      no_log: "{{ cifmw_nolog | default(true) | bool }}"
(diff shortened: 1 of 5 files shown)


roles/reproducer/tasks/configure_cleanup.yaml
```diff
@@ -0,0 +1,56 @@
...
+            cifmw_reproducer_repositories |
+            selectattr('src', 'match', '^.*/ci-framework$') |
+            map(attribute='dest')
+          }}
+        _path: >-
+          {{
+            (custom_path | length > 0) |
+            ternary(custom_path | first, default_path)
+          }}
+      ansible.builtin.set_fact:
```
dasm, line 26: I'm not following what you're trying to do here. You're discovering `default_path`, then you're trying to fetch the `custom_path` which is only different by `_-`. If it's not found, you're reverting to the `default_path`. Is that correct?
Why is it like that?

roles/reproducer/tasks/reuse_main.yaml
```diff
@@ -0,0 +1,274 @@
...
+    - bootstrap_layout
+  when:
+    - cifmw_job_uri is defined
+  ansible.builtin.include_tasks:
+    file: ci_data.yml
+    apply:
+      tags:
+        - bootstrap_layout
+
+- name: Discover and expose CI Framework path on remote node
```
dasm, line 28: This is the same code, which you already provided in the `configure_cleanup.yaml`. Could it be unified?

roles/reproducer/tasks/reuse_main.yaml
```diff
@@ -0,0 +1,274 @@
...
+  tags:
+    - always
+  vars:
+    _use_crc: >-
+      {{
+        _cifmw_libvirt_manager_layout.vms.crc is defined and
+        (
+          (_cifmw_libvirt_manager_layout.vms.crc.amount is defined and
+           _cifmw_libvirt_manager_layout.vms.crc.amount|int > 0) or
+        _cifmw_libvirt_manager_layout.vms.crc.amount is undefined)
```
dasm, line 66: What happens if you end up with `undefined` here? How does it impact the line 75?

roles/reproducer/tasks/reuse_main.yaml
```diff
@@ -0,0 +1,274 @@
...
+          become: true
+          ansible.builtin.template:
+            dest: "/etc/motd.d/cifmw.motd"
+            src: "motd.j2"
+            mode: "0644"
+
+        - name: Rotate some logs
+          tags:
+            - always
+          ansible.builtin.include_tasks: rotate_log.yml
```
dasm, line 115: ditto

roles/reproducer/tasks/reuse_main.yaml
```diff
@@ -0,0 +1,274 @@
...
+            - "/home/zuul/ansible-bootstrap.log"
+
+        - name: Bootstrap environment on controller-0
+          environment:
+            ANSIBLE_LOG_PATH: "~/ansible-bootstrap.log"
+          no_log: "{{ cifmw_nolog | default(true) | bool }}"
+          ansible.builtin.command:
+            chdir: "{{ _cifmw_reproducer_framework_location }}"
+            cmd: >-
+              ansible-playbook -i ~/ci-framework-data/artifacts/zuul_inventory.yml
```
dasm, line 126: FYI, we're trying to get rid of such nested invocations of playbooks. We're not there yet, so we cannot do anything about that. However, we're gonna be slowly retiring such a dirty workarounds.

roles/reproducer/tasks/reuse_main.yaml
```diff
@@ -0,0 +1,274 @@
...
+            creates: "/home/zuul/ansible-bootstrap.log"
+
+        - name: Install dev tools from install_yamls on controller-0
+          environment:
+            ANSIBLE_LOG_PATH: "~/ansible-bootstrap.log"
+          vars:
+            _devsetup_path: >-
+              {{
+                (
+                  cifmw_install_yamls_repo | default('/home/zuul/src/github.com/openstack-k8s-operators/install_yamls'),
```
dasm, line 139: Please consider using `"{{ ansible_user_dir }}/src/github.com/openstack-k8s-operators/install_yamls/"`

dasm (changes requested):

eshulman2:
Hi @dasm this is mostly based on the structure of the current reproducer playbook and role are you planning a restructure or change of it any time soon? I believe this set of playbook can move with you and the changes you are doing in the reproducer and related roles but for now it is built in a mostly similar way. Can we have a quick meeting to discuss the changes you are planning? hit me up on slack when you can :)

roles/reproducer/tasks/configure_cleanup.yaml
eshulman2: This is similar to the configure architecture TBH didn't change this part of the logic but maybe we can reconsider it in both

roles/reproducer/tasks/reuse_main.yaml
eshulman2: As those are separate task files that might not always be called together I would like to keep them independent of each other to avoid issues in running only one of them

roles/reproducer/tasks/reuse_main.yaml
eshulman2: This would result in this value just being false as this var is equal to the evaluation result

roles/reproducer/tasks/reuse_main.yaml
eshulman2: Generally I'm against this as well I would follow you as you make those changes to remove them here as well

deploy-edpm-reuse.yaml
```diff
@@ -0,0 +1,102 @@
+---
+- name: Manage unique ID
+  ansible.builtin.import_playbook: playbooks/unique-id.yml
```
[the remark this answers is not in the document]
eshulman2: I would like to avoid this as well but it is not simply calling a role it is running several tasks, are you planning on moving this playbook content into a role?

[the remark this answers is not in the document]
dasm: That would be preferred to migrate it over to a role.

deploy-edpm-reuse.yaml
```diff
@@ -0,0 +1,102 @@
...
+
+- name: Reproducer prepare play
+  hosts: "{{ cifmw_target_host | default('localhost') }}"
+  gather_facts: true
+  pre_tasks:
+    - name: Prepare cleanup script
```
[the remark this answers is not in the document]
dasm: I'm sorry, I think I misunderstood the code. You're right, it's importing a role from the `reproducer`.
What I wanted to say I'd like to see more verbose explanation in the name.

deploy-edpm-reuse.yaml
```diff
@@ -0,0 +1,102 @@
...
+        name: repo_setup
+        tasks_from: rhos_release.yml
+
+  roles:
+    - role: ci_setup
+
+- name: Prepare switches
+  vars:
+    cifmw_configure_switches: "{{ 'switches' in groups }}"
+  ansible.builtin.import_playbook: playbooks/switches_config.yml
```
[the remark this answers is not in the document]
dasm: You answered my similar question already.

roles/reproducer/tasks/configure_cleanup.yaml
dasm: Ack. Let's keep the same logic. We can change that later.

dasm (dismissed):
@eshulman2 I've added CI-FMW and CI-Ops teams to the review.
I like where you're going with this change, even though it's not fully in-sync with other initiatives happening at the time. However, we can reconvene and change it, if it's needed.

eshulman2:
@dasm I changed something in the commit message mind re-adding your approval?

dasm (approved):

sdatko (approved):
I saw it in action on uni03gamma, it is fine for now, though some further work needs to be added for other uni jobs (as a separate changes preferably to not make this one even bigger)

PR #2921, opened 2025-04-23
Add OpenStack Operator Initialization

sathlan:
We have introduced new tasks for initializing the OpenStack operator. This is required for versions prior to v1.0.7, where deployment didn't include the initialization. This action only takes place if we are using an OLM-based update.

Additionally, we block the process until the OpenStack deployment is successful. This applies to all starting versions, as the operator should be ready before proceeding.

Closes: [OSPRH-15799](https://issues.redhat.com/browse/OSPRH-15799)

### DIFF
--- a/roles/update/tasks/main.yml
+++ b/roles/update/tasks/main.yml
@@ -44,6 +44,51 @@
     - cifmw_ci_gen_kustomize_values_installplan_approval is defined
     - cifmw_ci_gen_kustomize_values_installplan_approval | lower == 'manual'
 
+- name: Initialize the openstack operator if needed
+  kubernetes.core.k8s:
+    kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
+    api_key: "{{ cifmw_openshift_token | default(omit) }}"
+    context: "{{ cifmw_openshift_context | default(omit) }}"
+    definition: "{{ _openstack_init_resource }}"
+    state: present
+  vars:
+    _openstack_init_resource:
+      apiVersion: operator.openstack.org/v1beta1
+      kind: OpenStack
+      metadata:
+        name: openstack
+        namespace: openstack-operators
+  when:
+    - cifmw_ci_gen_kustomize_values_deployment_version is defined
+    - cifmw_ci_gen_kustomize_values_deployment_version is in ['v1.0.3', 'v1.0.6']
+
+- name: Ensure OpenStack deployment is successful and block until it is done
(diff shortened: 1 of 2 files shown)


sathlan:
Need to re-run all the tests with the latest version.

roles/update/tasks/main.yml
```diff
@@ -44,6 +44,51 @@
...
+      apiVersion: operator.openstack.org/v1beta1
+      kind: OpenStack
+      metadata:
+        name: openstack
+        namespace: openstack-operators
+  when:
+    - cifmw_ci_gen_kustomize_values_deployment_version is defined
+    - cifmw_ci_gen_kustomize_values_deployment_version is in ['v1.0.3', 'v1.0.6']
+
+- name: Ensure OpenStack deployment is successful and block until it is
```
holser, line 65: Could you update phrase to "Ensure OpenStack deployment is successful and block until it is >done<"? It looks like the sentence is not finished (at least I expect something after it is)

roles/update/tasks/main.yml
```diff
@@ -44,6 +44,51 @@
...
+  vars:
+    _openstack_init_resource:
+      apiVersion: operator.openstack.org/v1beta1
+      kind: OpenStack
+      metadata:
+        name: openstack
+        namespace: openstack-operators
+  when:
+    - cifmw_ci_gen_kustomize_values_deployment_version is defined
+    - cifmw_ci_gen_kustomize_values_deployment_version is in ['v1.0.3', 'v1.0.6']
```
dasm, line 63: If I understand it right, currently we do support only `v1.0.3` and `v1.0.6` [src](https://github.com/openstack-k8s-operators/ci-framework/blob/58bc72099dfcc8fb9440b3bd8c50e91c5d1b33fd/roles/ci_gen_kustomize_values/README.md) with a `v1.0.7` [like here](https://github.com/openstack-k8s-operators/architecture/blob/6980941078e83c1d6876a2e59981f811f1141b95/examples/common/olm-subscriptions/README.md) (but it currently does not exist).
Is this check required?

sathlan:
Tested on a manual non-trunk deployment using `v1.0.6` and also during a trunk `crc`  job (where it shouldn't have any impact) and both were successful.

dasm (approved):

ciecierski (approved):

danpawlik (approved):

PR #2923, opened 2025-04-24
Do not print oc completion

danpawlik:
When Ansible is executed in verbose mode, the oc completion output is not needed to print.


danpawlik:
recheck

dasm (approved):

PR #2927, opened 2025-04-25
Add conditional for use CentOS nodepool mirror

rlandy:


danpawlik:
Is it for review?

roles/repo_setup/tasks/ci_mirror.yml
```diff
@@ -12,6 +12,8 @@
     set -o pipefail
     source /etc/ci/mirror_info.sh
     sed -i -e "s|https://trunk.rdoproject.org|$NODEPOOL_RDO_PROXY|g" *.repo
+    {% if ansible_distribution | lower == 'centos' and ansible_distribution_major_version is not version('10', '==') %}
```
dasm, line 15: Can you add additional info about location of CentOS10 nodepool? If it doesn't exist, please create another `when` condition. Instead of nesting `if/else` here,  please modify the line 10, and extend it over there. For the Centos10 specific use-case, please add another task.
You might consider using nested conditions, something like:
```
- block
   - name: cs9
     when: major_version is < 10
   - name: cs10
     when: major_version is 10
   when:
     - mirror_path.stat.exists
```

roles/build_containers/templates/build_containers.sh.j2
```diff
@@ -49,7 +49,7 @@ openstack tcib container image build \
 {%   endfor %}
 {%   endif %}
      --tcib-extra tcib_release={{ ansible_distribution_major_version }} \
-     --tcib-extra tcib_python_version={{ (ansible_distribution_major_version is version('9', '<')) | ternary ('3.6', '3.9') }} \
+     --tcib-extra tcib_python_version={{ (ansible_distribution_major_version is version('10', '<')) | ternary ('3.9', '3.12') }} \
```
dasm, line 52: Tnis change is unrelated to your PR. Please remove it.

dasm (changes requested):

roles/build_containers/templates/build_containers.sh.j2
raukadah: Note this role is used in RHOSO-18 where python3.9 is used. I would drop 3.6 completely(No longer used by RHOSO anywhere) and keep 3.9 and 3.12 only.

dasm: OK, makes sense.

roles/repo_setup/tasks/ci_mirror.yml
```diff
@@ -8,10 +8,24 @@
...
+  block:
+    - name: Use RDO proxy mirrors
+      ansible.builtin.shell: |
+        set -o pipefail
+        source /etc/ci/mirror_info.sh
+        sed -i -e "s|https://trunk.rdoproject.org|$NODEPOOL_RDO_PROXY|g" *.repo
+      args:
+        chdir: "{{ cifmw_repo_setup_output }}"
+
+    # TODO(rlandy) remove when CentOS 10 mirrors exist on Nodepool Hosts
```
dasm, line 20: Thank you for the additional explanation. Please add this note to the `name`. It will be visible every time during the test run, so we won't loose track of it.

dasm (changes requested):

dasm (approved):

raukadah:
recheck

raukadah:
recheck

frenzyfriday (approved):

PR #2928, opened 2025-04-25
cd_dcn_site: Set the correct hostname

krcmarik:
Set the correct hostname (without the
extra prefix "edpm-"). The hostname of
the compute nodes is set to a name
without the "edpm-" prefix by ci-fmw
when the VMs are provisioned. Once the
dataplane is deployed with a different
hostname which would be configured in the
OCP DNS used by dataplane nodes (after
dataplane deployed) The original hostname
which is still set on the dataplane nodes
would not be recognized by the DNS and
the correct domain would not be used.
Instead the original domain which was set
by DHCP during the VM nodes provisioning
by ci-framework would be set until the
DHCP lease is lost and that eventually
leads to fqdn of dataplane nodes mismatch

### DIFF
--- a/roles/ci_dcn_site/tasks/az.yml
+++ b/roles/ci_dcn_site/tasks/az.yml
@@ -24,13 +24,11 @@
     command: >-
       openstack aggregate show {{ _az }} -c hosts -f value
 
-- name: Convert az_hosts string to list and remove extra text
+- name: Convert az_hosts string to list
   ansible.builtin.set_fact:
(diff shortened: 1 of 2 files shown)


krcmarik:
recheck

fultonj (approved):
/approve
/lgtm

fultonj (approved):
/approve
/lgtm

sbekkerm (approved):

PR #2929, opened 2025-04-28
DNM Use role instead of playbooks to flatten Ansible execution

danpawlik:
Most of the playbooks which are executed inside the nested Ansible (Ansible inside the Ansible) does not need to be in playbook, it can be in a simple role.
The previous CI execution for 01-bootstrap.yml file was:

    Zuul CI ->
        Ansible command ->
        Ansible playbook ->
        import playbook

With this change would be one layer less:

    Zuul CI ->
        Ansible command ->
        Ansible playbook ->
        cifmw_setup role


danpawlik:
recheck

danpawlik:
recheck

danpawlik:
recheck

dasm (approved):
Thanks @danpawlik !

danpawlik:
Just to remind https://github.com/openstack-k8s-operators/ci-framework/pull/2930/commits/e64c9f237745b91f755b2591a33ed011534c4e4d why we remove from playbook to roles: 
```
    Change the way how the hooks playbook is executed

    In many places, it was done that Ansible playbook is importing another
    ansible playbook, which is importing again another playbook and another.
    It is too complex and after doing simple improvement as it was done
    in this change by removing import_playbook: hooks.yml, we see
    that those tasks can be running in same play.
    With that approach we can:

    - easy understand playbook/role execution,
    - better control variables,
    - in some part faster execution due there is no need to make delegation
      or gather_facts,
    - easier to debug
```

PR #2930, opened 2025-04-28
Change the way how the hooks playbook is executed

danpawlik:
In many places, it was done that Ansible playbook is importing another ansible playbook, which is importing again another playbook and another. It is too complex and after doing simple improvement as it was done in this change by removing import_playbook: hooks.yml, we see that those tasks can be running in same play.
With that approach we can:

- easy understand playbook/role execution,
- better control variables,
- in some part faster execution due there is no need to make delegation or gather_facts,
- easier to debug


danpawlik:
recheck

evallesp (dismissed):
Great

evallesp (approved):

dasm (approved):
Great work @danpawlik !

PR #2931, opened 2025-04-28
Increase CEPH_TIMEOUT for component tests

fmount:


danpawlik:
NOTE:  There is many errors related to: `pull QPS exceeded` - https://access.redhat.com/solutions/4767061. 
Right now, we are against changing the default value for pulling images from 5 to 10 or more, due the hypervisors does not have set any IOPS/RW limitation and storing images to the disk is resource consuming. So if we change the value, it might raise additional complication or do a "noisy neighbor" issue and affect other jobs.

danpawlik (approved):

evallesp:
recheck

evallesp (approved):

dasm (approved):

PR #2932, opened 2025-04-28
Update adoption scenario uni06zeta

fyanac:
It's empty patch, the configuration is defined in the job
Similar empty patch to unialpha adoption: https://github.com/openstack-k8s-operators/ci-framework/commit/9526d90e250afaf743f57ce5375da1f12b34e4a3


dasm (changes requested):
Thank you @fyanac for bringing that up. I'm in full support to remove duplicated/outdated code. However I'm not sure what's happening here.
Would you be able to add extra explanation in your commit message (ideally with links to other changes) to show that it's indeed not required.

Please include also test project, which verifies that your changes do not affect `uni06zeta`.

fyanac:
> Thank you @fyanac for bringing that up. I'm in full support to remove duplicated/outdated code. However I'm not sure what's happening here. Would you be able to add extra explanation in your commit message (ideally with links to other changes) to show that it's indeed not required.
> 
> Please include also test project, which verifies that your changes do not affect `uni06zeta`.

Hi,
This change is for uni06zeta adoption for osp 17.1 initial deployment, so it doesn't affect to uni06zeta greenfield deployment. The change also is the same as unialpha adoption.

dasm (approved):

PR #2937, opened 2025-04-29
Convert pre-run/post-run to lists

dasm:
Zuul accept list or string as a value for pre-run and post-run. During a process of building jobs, Zuul combines pre/post-runs into lists. To avoid confusions, convert values into lists.

https://zuul-ci.org/docs/zuul/latest/config/job.html#attr-job.pre-run https://zuul-ci.org/docs/zuul/latest/config/job.html#attr-job.post-run

### DIFF
--- a/zuul.d/base.yaml
+++ b/zuul.d/base.yaml
@@ -15,7 +15,8 @@
     pre-run:
       - ci/playbooks/e2e-prepare.yml
       - ci/playbooks/dump_zuul_data.yml
-    post-run: ci/playbooks/collect-logs.yml
+    post-run:
+      - ci/playbooks/collect-logs.yml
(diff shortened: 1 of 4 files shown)


danpawlik (approved):

frenzyfriday (approved):
I didnt know we had the playbooks as strings like this 🤦🏽‍♀️

dasm:
> I didnt know we had the playbooks as strings like this 🤦🏽‍♀️

@frenzyfriday It's nothing wrong. Zuul allows to configure them as strings and as lists. Because `pre-run/post-run` are always combined, based on child/parent inheritance, I wanted to ensure that these are always represented in the same way.

PR #2938, opened 2025-04-30
Set jinja2_native in Ansible to preserve variable types during template operations

kstrenkova:
Currently, it is not possible to set concurrency using the 'stages' functionality. It is because the retyping to string is present and concurrency is supposed to be type integer. This patch fixes this bug and lets users to set the value of the parameter.

In other words, on reassignment variable, jinja2 is changing the type to string which raises an issue when running tempest. It raises an error:

```
    changed: false
    msg: 'Failed to create object: b''{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Tempest.test.openstack.org
      \\"tempest-tests-tempest\\" is invalid: spec.tempestRun.concurrency: Invalid value:
      \\"string\\": spec.tempestRun.concurrency in body must be of type integer: \\"string\\"","reason":"Invalid","details":{"name":"tempest-tests-tempest","group":"test.openstack.org","kind":"Tempest","causes":[{"reason":"FieldValueTypeInvalid","message":"Invalid
      value: \\"string\\": spec.tempestRun.concurrency in body must be of type integer:
      \\"string\\"","field":"spec.tempestRun.concurrency"}]},"code":422}\n'''
    reason: Unprocessable Entity
```

### DIFF
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -18,5 +18,6 @@ fact_caching_timeout = 0
 inventory = inventory.yml
 pipelining = True
 any_errors_fatal = True
+jinja2_native = True
 [ssh_connection]
 ssh_args = -o ControlMaster=auto -o ControlPersist=300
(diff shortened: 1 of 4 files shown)


danpawlik:
That's... not fine in Ansible.
After many tries in https://github.com/openstack-k8s-operators/ci-framework/pull/2949, I don't see other way to redefine it without big refactor.

danpawlik (dismissed):

evallesp (dismissed):
lgtm

danpawlik:
recheck

danpawlik:
recheck

danpawlik:
recheck

kstrenkova:
I changed the concurrency a tiny bit because we do not need to set the concurrency to int (it is already an integer). But the tasks are done in such way, that they take `cifmw_test_operator_tempest....` and concurrency does not have tempest in it's name, so it's not overwritten. That is why I still left the task there and it should work as it was intended in my ticket regarding this.

danpawlik (approved):

PR #2939, opened 2025-04-30
Retry OCP CSR approval

abays:
The `openshift_adm` role's `Check for pending certificate approval` can sometimes fail when the cluster API isn't fully stable yet.  Let's add a retry to avoid failing here unnecessarily.  I'm assuming the actual `Wait until the OpenShift cluster is stable` [1] has to come after CSR approval because, without the approval, the cluster cannot reach full stability.  If I'm wrong about that, perhaps we could just move the CSR task below [1].

[1] https://github.com/openstack-k8s-operators/ci-framework/blob/f8e17fe8789766d409037ec5cbf6e8cac77816e0/roles/openshift_adm/tasks/wait_for_cluster.yml#L63-L73

### DIFF
--- a/roles/openshift_adm/tasks/wait_for_cluster.yml
+++ b/roles/openshift_adm/tasks/wait_for_cluster.yml
@@ -49,8 +49,15 @@
 - name: Check for pending certificate approval.
   when:
     - _openshift_adm_check_cert_approve | default(false) | bool
+  register: _approve_csr
   approve_csr:
     k8s_config: "{{ cifmw_openshift_kubeconfig }}"
+  retries: 30
+  delay: 10
+  until:
+    - _approve_csr is defined
+    - _approve_csr.rc is defined
+    - _approve_csr.rc == 0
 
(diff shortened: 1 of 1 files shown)


abays:
> surprising, due it should not happening when etcd is on ramdisk. Does it happening when CI job has nested crc or it also happening when crc-cloud/crc-extracted?

It happens in my local env when I run the reproducer, every time on the first attempt.  Then if I run the reproducer again (without cleaning), it succeeds.  In my testing, adding the retries allows it to succeed on the first run.

abays:
> surprising, due it should not happening when etcd is on ramdisk. Does it happening when CI job has nested crc or it also happening when crc-cloud/crc-extracted?

Also, I am not using CRC with CIFMW.  This is the dev-scripts path.

leifmadsen:
I run into this issue every time I deploy with CIFMW. I have to wait for it to fail then re-run my deployment. It'd be nice to see something like this land so I don't have to double-run each deployment.

dasm (approved):

PR #2949, opened 2025-05-05
Set concurrency for tempest config without reference

danpawlik:
There was many tests done to set the integer value for concurrency
parameter in tempest config. Even forcing the value to integer
does not meet with conversion in the end.
We thought that it is k8s module issue, but it seems not. It migth be
related to jinja2 according to issue [1].
Until the value is not parsed properly, we need to set it without
reference.

[1] https://github.com/ansible/ansible/issues/18095


danpawlik:
recheck

danpawlik:
It requires to set `jinja2_native` in Ansible config: https://docs.ansible.com/ansible/latest/reference_appendices/config.html#default-jinja2-native

I don't see another way to set the value to integer without any code refactor like create a template, save to file then read in k8s module.

PR #2952, opened 2025-05-06
Simplify files to track by Zuul

danpawlik:
In many places it was not necessary to add a regex, due the whole role/playbook/module should be tracked.
In other words, we can live if README file would be changed and there would be a CI job running for veryfing that.

### DIFF
--- a/zuul.d/molecule.yaml
+++ b/zuul.d/molecule.yaml
@@ -2,7 +2,7 @@
     files:
     - ^common-requirements.txt
     - ^test-requirements.txt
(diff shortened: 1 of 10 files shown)


zuul.d/molecule.yaml
```diff
@@ -870,7 +870,7 @@
     files:
     - ^common-requirements.txt
     - ^test-requirements.txt
-    - ^roles/cifmw_external_dns/defaults|files|handlers|library|lookup_plugins|module_utils|molecule|tasks|templates|vars.*
```
danpawlik, line 873: NOTE: here was an error in regex.

dasm (approved):

PR #2954, opened 2025-05-07
dt-dcn: Make VM's names of DCN compute groups not to be substring

krcmarik:
The reason is to workaround a bug in neutron


fultonj (approved):
/approve
/lgtm

PR #2962, opened 2025-05-08
Fix docs ci job because of snowballstemmer

evallesp:
We need to pin snowballstemmer < 3 due to
https://github.com/sphinx-doc/sphinx/issues/13533

### DIFF
--- a/docs/doc-requirements.txt
+++ b/docs/doc-requirements.txt
@@ -7,3 +7,4 @@ sphinxemoji
 myst-parser[linkify]
 ansible-core
 ansible-doc-extractor
+snowballstemmer<3


evallesp:
recheck

frenzyfriday (approved):

danpawlik (approved):

danpawlik:
@dasm but the CI jobs fails if its not pinned.
We can merge and revert or just wait.

dasm:
> @dasm but the CI jobs fails if its not pinned. We can merge and revert or just wait.

I don't mind merging it and reverting it when it's needed. But the `v3.0.0` package [doesn't exist anymore](https://pypi.org/project/snowballstemmer/3.0.0/). The new one `v3.0.0.1` includes [the fix](https://github.com/snowballstem/snowball/issues/229)

We might be pulling the package from some local mirror then?

dasm:
Actually, I see the fix has been merged about 1h ago, so indeed it was not available earlier.

danpawlik:
cool, so we can close the PR.

PR #2963, opened 2025-05-08
Optionally configure glance with cinder_volume_type=multiattach

eduolivares:
With this patch, the hook that creates a multiattach volume type can also configure that volume type as the `cinder_volume_type`, if parameter `configure_cinder_volume_type` is set to true.

This is needed for octavia, because the amphora image that it creates needs to be based on a multiattach volume. Otherwise, it fails to create amphora VMs on different computes simultaneously, which is something that happens when octavia is configured with ACTIVE_STANDBY.

[OSPRH-16089](https://issues.redhat.com//browse/OSPRH-16089)
[OSPCIX-768](https://issues.redhat.com//browse/OSPCIX-768)

### DIFF
--- a/hooks/playbooks/cinder_multiattach_volume_type.yml
+++ b/hooks/playbooks/cinder_multiattach_volume_type.yml
@@ -24,3 +24,50 @@
         oc rsh openstackclient \
             openstack volume type set --property multiattach="<is> True" \
             {{ cifmw_volume_multiattach_type }}
+
+    # This block is needed for octavia because the Amphora image needs to be created on a multiattach volume
+    - name: Block to configure cinder_volume_type when needed
+      when: configure_cinder_volume_type | default('false') | bool
+      block:
+        - name: Create tempfile
+          ansible.builtin.tempfile:
+            state: file
+            prefix: glance_custom_service_config
+          register: _glance_custom_service_config_file
+
+        - name: Write current glance customServiceConfig to tempfile
+          ansible.builtin.shell: |
+            set -xe -o pipefail
+            crname=$(oc get openstackcontrolplane -o name -n {{ namespace }})
+            oc -n {{ namespace }} get ${crname} -o jsonpath={.spec.glance.template.customServiceConfig} > {{ _glance_custom_service_config_file.path }}
+          changed_when: false
+
+        - name: Ensure cinder_volume_type is configured with proper value in tempfile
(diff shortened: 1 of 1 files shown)


eduolivares:
recheck

hooks/playbooks/cinder_multiattach_volume_type.yml
```diff
@@ -24,3 +24,50 @@
...
+    - name: Block to configure cinder_volume_type when needed
+      when: configure_cinder_volume_type | default('false') | bool
+      block:
+        - name: Create tempfile
+          ansible.builtin.tempfile:
+            state: file
+            prefix: glance_custom_service_config
+          register: _glance_custom_service_config_file
+
+        - name: Write current glance customServiceConfig to tempfile
```
danpawlik, line 38: normally I will suggest to split into few tasks to avoid potential problems.

hooks/playbooks/cinder_multiattach_volume_type.yml
```diff
@@ -24,3 +24,50 @@
...
+            oc -n {{ namespace }} get ${crname} -o jsonpath={.spec.glance.template.customServiceConfig} > {{ _glance_custom_service_config_file.path }}
+          changed_when: false
+
+        - name: Ensure cinder_volume_type is configured with proper value in tempfile
+          community.general.ini_file:
+            path: "{{ _glance_custom_service_config_file.path }}"
+            section: "{{ default_backend | default('default_backend') }}"
+            option: cinder_volume_type
+            value: "{{ cifmw_volume_multiattach_type }}"
+            mode: "0644"
```
danpawlik, line 51: nit: don't think changing mode/ensuring that mode is set is needed.

danpawlik (approved):

fmount:
Is there a particular reason to not patch https://github.com/openstack-k8s-operators/architecture/blob/main/examples/dt/uni07eta/control-plane/service-values.yaml directly and avoid this patch entirely?

eduolivares:
recheck

fmount (changes requested):
I'm not sure this patch is required if we patch uni07eta glance config in advance.

fmount:
I'd like to clarify that aspect before moving forward with this.

fmount (approved):
/lgtm

bshewale (approved):

PR #2967, opened 2025-05-09
Drop PyYAML package for bindep 

raukadah:
PyYAML package is not available on RHEL distro and is replaced with
python-pyyaml, which is already there.

This pr will drops PyYAML from bindep.

Resolves: [OSPRH-16716](https://issues.redhat.com//browse/OSPRH-16716)

### DIFF
--- a/bindep.txt
+++ b/bindep.txt
@@ -22,7 +22,6 @@ podman            [platform:rpm]
 python3-devel     [platform:rpm !platform:rhel-7 !platform:centos-7]
 python3-libvirt   [platform:rpm]
 python3-lxml      [platform:rpm]
-PyYAML            [platform:rpm !platform:rhel-8 !platform:centos-8 !platform:rhel-9 !platform:centos-9 !platform:fedora]
 python3-pyyaml    [platform:rpm !platform:rhel-7 !platform:centos-7]
 python3-dnf       [platform:rpm !platform:rhel-7 !platform:centos-7]
 


amartyasinha (approved):

bindep.txt
```diff
@@ -22,7 +22,7 @@ podman            [platform:rpm]
 python3-devel     [platform:rpm !platform:rhel-7 !platform:centos-7]
 python3-libvirt   [platform:rpm]
 python3-lxml      [platform:rpm]
-PyYAML            [platform:rpm !platform:rhel-8 !platform:centos-8 !platform:rhel-9 !platform:centos-9 !platform:fedora]
+PyYAML            [platform:rpm !platform:rhel-8 !platform:centos-8 !platform:rhel-9 !platform:centos-9 !platform:rhel-10 !platform:centos-10 !platform:fedora]
```
SeanMooney, line 25: we are we continueign to do this if we are skipping all release

we should just remove this line

dasm: I agree with @SeanMooney 
What is the purpose of the `PyYAML` dependency, if we don't use it on any platform in recent history?
CentOS 7.9-2009[[3]](https://en.wikipedia.org/wiki/CentOS#cite_note-CentOS_7_Announcement-3) (12 November 2020) happened 4 years ago

danpawlik (changes requested):
Why you remove that instead of add new condition ?

EDIT:
I get the answer from comments, but not from commit message. 
Please update commit message and we can consider to merge.

bindep.txt
```diff
@@ -22,7 +22,6 @@ podman            [platform:rpm]
 python3-devel     [platform:rpm !platform:rhel-7 !platform:centos-7]
 python3-libvirt   [platform:rpm]
 python3-lxml      [platform:rpm]
-PyYAML            [platform:rpm !platform:rhel-8 !platform:centos-8 !platform:rhel-9 !platform:centos-9 !platform:fedora]
```
danpawlik, line 25: why not add `platform:centos-10` ?

```
podman run -it quay.io/centos/centos:stream10 bash 
yum install -y git vim python3-pip make sudo
git clone https://github.com/openstack-k8s-operators/ci-framework && cd ci-framework
./scripts/setup_env


[root@f210a56a991a ci-framework]# bindep --profiles
Platform profiles:
platform:base-py3
platform:centos
platform:centos-10
platform:centos-coughlan
platform:redhat
platform:rpm
```

SeanMooney: [platform:rpm !platform:rhel-8 !platform:centos-8 !platform:rhel-9 !platform:centos-9 !platform:fedora]

becaues the condtion say

install on rpm distros that are not one of rhel 8, rhel 9, centos 8 centos 9 or fedora

this repo does not support centos/rhel7  and the conditions have eliminated all but centos/rhel10  and that is what we would be adding.

so i suggested removing the package entrily since we have eliminated all distos it could be installed on with the conditions.

raukadah:
> Why you remove that instead of add new condition ?
> 
> EDIT: I get the answer from comments, but not from commit message. Please update commit message and we can consider to merge.
> 
> EDIT: So there are are two ways:
> 
>     * update commit message and remove pyyaml but dunno what would be a side effect
> 
>     * update bindep by adding: `!platform:centos-10` and leave it as it is.

Thank you @danpawlik for the review. I have updated the comment.

dasm (approved):

frenzyfriday (approved):

danpawlik (approved):

PR #2973, opened 2025-05-12
Change concurrency to tempest specific parameter

kstrenkova:
This patch introduces the cifmw_test_operator_tempest_concurrency parameter, as current concurrency parameter is used only in Tempest tests. It will also allow the stages loop to include concurrency, so now users can set different concurrency values for different tempest runs.


kstrenkova:
recheck

dasm (approved):
👍

danpawlik:
after merging recent changes, it does not work?

kstrenkova:
Hi :wave: No, some of the changes were reverted, so now it does not work. I think this approach is better than what we had in the previous patch also.

arxcruz (approved):

PR #2978, opened 2025-05-13
Add become for log delete

eshulman2:
In certain usecases some log files are generated by other tools and might have unexpected permissions set to them added become to avoid issues related in permissions

### DIFF
--- a/clean_openstack_deployment.yaml
+++ b/clean_openstack_deployment.yaml
@@ -15,3 +15,4 @@
       loop:
         - "/home/zuul/ci-framework-data/logs"
         - "/home/zuul/ci-framework-data/tests"
+      become: true


eshulman2:
for example after update this log is created and can't be deleted by zuul user
`[zuul@controller-0 ~]$ ll /home/zuul/ci-framework-data/logs/ceph-update/ceph_daemons.log
-rw-r--r--. 1 root root 7970 May 13 03:34 /home/zuul/ci-framework-data/logs/ceph-update/ceph_daemons.log`

clean_openstack_deployment.yaml
```diff
@@ -15,3 +15,4 @@
       loop:
         - "/home/zuul/ci-framework-data/logs"
         - "/home/zuul/ci-framework-data/tests"
+      become: true
```
danpawlik, line 18: nit: if you can add it into L12 would be better for the eye ;)

evallesp (approved):

clean_openstack_deployment.yaml
evallesp: +1

PR #2979, opened 2025-05-13
Reboot controller-0 to avoid running deployments

eshulman2:
Because deploy architecture runs detached from the zuul job and might run other playbooks in a detached way we need to make sure there are no leftover running deployments that might break the cleanup (trying to create additional resources while the cleanup is running). Added a reboot for controller-0 before cleanup to avoid such cases.

### DIFF
--- a/deploy-edpm-reuse.yaml
+++ b/deploy-edpm-reuse.yaml
@@ -2,6 +2,20 @@
 - name: Manage unique ID
   ansible.builtin.import_playbook: playbooks/unique-id.yml
 
+- name: Reboot controller-0 to make sure there are no running deployments
+  hosts: controller-0
+  gather_facts: false
+  tasks:
+    - name: Reboot controller-0
+      ansible.builtin.reboot:
+        reboot_timeout: 600
+      become: true
+
+    - name: Wait for controller-0 to come back online
+      ansible.builtin.wait_for_connection:
+        timeout: 600
+        delay: 10
+
 - name: Reproducer prepare play
   hosts: "{{ cifmw_target_host | default('localhost') }}"
   gather_facts: true


dasm (commented):
Thanks @eshulman2 !

dasm:
recheck

deploy-edpm-reuse.yaml
```diff
@@ -2,6 +2,22 @@
...
+      ansible.builtin.reboot:
+        reboot_timeout: 600
+      register: reboot_result
+      become: true
+
+    - name: Wait for controller-0 to come back online
+      ansible.builtin.wait_for_connection:
+        delay: 10
+        timeout: 600
+      when: reboot_result is changed
```
dasm, line 19: This line is triggering `ansible-lint` error. I'm not sure about a better way to implement that. We might consider disabling the linter on it.

deploy-edpm-reuse.yaml
```diff
@@ -2,6 +2,22 @@
...
   ansible.builtin.import_playbook: playbooks/unique-id.yml
 
+- name: Reboot controller-0 to make sure there are no running deployments
+  hosts: controller-0
+  gather_facts: false
+  tasks:
+    - name: Reboot controller-0
+      ansible.builtin.reboot:
+        reboot_timeout: 600
+      register: reboot_result
```
dasm, line 12: Maybe you could drop this line and replace it instead with [post_reboot_delay](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/reboot_module.html#parameter-post_reboot_delay) to allow for the delay mentioned [here](https://github.com/openstack-k8s-operators/ci-framework/pull/2979/files#diff-fd2cd96602531f6ce6919f24b71046575a74b6dfeddaabcc00ee246d7393fa2cR17).
This would allow to perform the reboot, wait for it and continue with the `Wait for controller`.

deploy-edpm-reuse.yaml
dasm: Scratch above. I think I suggested slightly better solution.

deploy-edpm-reuse.yaml
eshulman2: great idea, done

danpawlik:
I don't see related CI job that is verifying that change. Am I wrong?

deploy-edpm-reuse.yaml
```diff
@@ -2,6 +2,15 @@
 - name: Manage unique ID
   ansible.builtin.import_playbook: playbooks/unique-id.yml
 
+- name: Reboot controller-0 to make sure there are no running deployments
+  hosts: controller-0
+  gather_facts: false
+  tasks:
+    - name: Reboot controller-0
+      ansible.builtin.reboot:
+        post_reboot_delay: 600
```
dasm, line 11: Thanks @eshulman2 for incorporating my feedback. `post_reboot_delay` is a blocking activity, which will stop test execution for 10 minutes. Instead, you could use features from both `reboot` and `wait_for_connection`.
`reboot` can use `post_delay` while `wait_for_connection` will continue the process immediately

```
    ansible.builtin.reboot:
        reboot_timeout: 600
        post_reboot_delay: 60
      become: true

    - name: Wait for controller-0 to come back online
      ansible.builtin.wait_for_connection:
        timeout: 600
```

dasm: The delay might be adjusted

danpawlik (approved):

sdatko (approved):

PR #2983, opened 2025-05-14
Adjust Adoption OSP Deploy to use ipv4 or ipv6

ciecierski:
Currently ipv4 is the only supported protocol for initial OSP17 deployment. With this change based on cifmw_networking_env_definition file network version protocol is detected and used for generating network related files.


roles/adoption_osp_deploy/templates/os_net_config_undercloud.yml.j2
fyanac, line 1: Line13 - address for ipv6, it should be 128, no 32. It could be like this: https://gitlab.cee.redhat.com/eng/openstack/tripleo-ansible/-/blob/rhos-18.0-trunk-patches/tripleo_ansible/roles/tripleo_network_config/templates/undercloud.j2?ref_type=heads#L4-13

roles/adoption_osp_deploy/tasks/prepare_undercloud.yml
```diff
@@ -91,13 +91,13 @@
       vars:
         _undercloud_name: "{{ _vm_groups['osp-underclouds'] | first }}"
         _undercloud_net: "{{ cifmw_networking_env_definition.instances[_undercloud_name] }}"
-        _ctlplane_ip: "{{ _undercloud_net.networks.ctlplane.ip_v4 }}"
+        _ctlplane_ip: "{{ _undercloud_net.networks.ctlplane[ip_version|default('ip_v4')] }}"
         _ctlplane_vip: "{{ cifmw_adoption_osp_deploy_scenario.undercloud.ctlplane_vip }}"
```
fyanac, line 95: The _ctlplane_vip_ also needs to be adapted

roles/adoption_osp_deploy/templates/os_net_config_undercloud.yml.j2
```diff
@@ -38,8 +48,8 @@ network_config:
     mtu: {{ net.mtu }}
     vlan_id: {{ net.vlan_id }}
     addresses:
-    - ip_netmask: {{ net.ip_v4 }}/{{ net.prefix_length_v4 }}
-    - ip_netmask: {{ net.ip_v4.split('.')[:3] | join('.') }}.2/32
+    - ip_netmask: {{ net[ip_version|default('ip_v4')] }}/{{ net[prefix_length_version|default('prefix_length_v4')] }}
+    - ip_netmask: {{ net[ip_version|default('ip_v4')].split('.')[:3] | join('.') }}.2/32
```
fyanac, line 52: I think it has a mask /32 for ipv4

roles/adoption_osp_deploy/tasks/prepare_undercloud.yml
ciecierski: done

roles/adoption_osp_deploy/templates/os_net_config_undercloud.yml.j2
ciecierski: re-review the change after latest changes

roles/adoption_osp_deploy/tasks/config_files.yml
```diff
@@ -70,22 +70,31 @@
...
+        gw_version: gw_v6
+        prefix_length_version: prefix_length_v6
+      when: cifmw_networking_env_definition.networks.ctlplane.network_v6 is defined
+
     - name: Generate DeployedServerPortMap field
       vars:
         _node_instance_net: "{{ cifmw_networking_env_definition.instances[node] }}"
         _key_name: "{{ node }}-ctlplane"
-        _ctlplane_ip: "{{ _node_instance_net.networks.ctlplane.ip_v4 }}"
+        _ctlplane_ip: "{{ _node_instance_net.networks.ctlplane[ip_version|default('ip_v4')] }}"
```
hjensas, line 86: Would it read better if we remove the `| default()` and instead always `set_fact` above instead?

Something like:
```
ansible.builtin.set_fact:
  ip_version: >-
    {{ 
       "ip_v6" if cifmw_networking_env_definition.networks.ctlplane.network_v6 is defined 
       else "ip_v4"
    }}
```

Then everyhing below can be:
```
_ctlplane_ip: "{{ _node_instance_net.networks.ctlplane[ip_version] }}"
```
 ... etc.

hjensas (approved):
Just the nit on `|default()` usage - but ok. Can be ignored, or a follow up.

jistr (approved):

ciecierski:
@fyanac let us know if change fixed reported bug https://issues.redhat.com/browse/OSPRH-16510 and ipv6 deployment is successful

PR #2987, opened 2025-05-14
Add multus bridge attach to

evallesp:
We're adding a new variable attach at the networking_mapper to map its value
to the output variable multus_attach.
This is used then to generate NAD with information about where to attach
the multus type bridge, if in a interface or a linux-bridge.

### DIFF
--- a/roles/ci_multus/templates/nad.yml.j2
+++ b/roles/ci_multus/templates/nad.yml.j2
@@ -4,6 +4,11 @@
 {% else %}
 {% set multus_type = cifmw_ci_multus_default_nad_type %}
 {% endif %}
+{% if network_details.tools.get('multus', {}).get('multus_attach', None) %}
+{% set multus_attach = network_details.tools.multus.multus_attach %}
+{% else %}
+{% set multus_attach = cifmw_ci_multus_default_bridge_attach %}
+{% endif %}
 ---
 apiVersion: k8s.cni.cncf.io/v1
 kind: NetworkAttachmentDefinition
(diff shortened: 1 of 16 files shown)


evallesp:
recheck

evallesp:
recheck

roles/ci_multus/templates/nad.yml.j2
```diff
@@ -12,13 +22,16 @@ spec:
...
+      "type": "{{ multus_type }}",
+{% if multus_type == "macvlan" %}
       "master": "{{ network_details.interface_name }}",
 {% endif %}
-{% if cifmw_ci_multus_default_nad_type == "bridge" %}
-      "type": "bridge",
+{% if multus_type == "bridge" %}
+{% if multus_attach == "interface" %}
       "bridge": "{{ network_details.interface_name }}",
+{% elif multus_attach == "linux-bridge" %}
```
hjensas, line 32: nit, I guess the bridge name and network name does not always match. But I guess that is a constraint we can work with.

hjensas (commented):
Looks good to me.

roles/ci_multus/templates/nad.yml.j2
evallesp: I also think we can work on that if that contraint block us.

hjensas: Yes, I agree. Let's go with this as is.

evallesp:
recheck

evallesp:
recheck

danpawlik (dismissed):

evallesp:
recheck

amartyasinha (approved):

evallesp:
recheck

PR #2990, opened 2025-05-15
Update labels related to nodesets with rhel

danpawlik:
Some CI jobs when using multiple nodes, some VMs are running on RHEL 9.2, where it should been using newer version.


danpawlik:
recheck

danpawlik:
recheck

danpawlik:
We still need to use 9.2
Let's fix 9.2 image or investigate when it fails in the CI

PR #2991, opened 2025-05-15
Add a containers tag role

rlandy:


roles/build_containers/tasks/main.yml
```diff
@@ -90,6 +90,13 @@
     - cifmw_build_containers_buildah_push | default ('false') | bool
     - not cifmw_build_containers_push_containers | bool
 
+- name: Return containers built, tag, and artifacts location
```
raukadah, line 93: Can we also add a task here to call ta.yaml task file with different conditional like cifmw_build_containers_retag so that they can retag it with something different tag?

Reason: periodic-container-tcib-build-push-centos-10-master tags the container with hash, I want to reuse this playbook to retag it with current in the same.

danpawlik:
recheck

roles/build_containers/tasks/main.yml
```diff
@@ -90,6 +90,13 @@
     - cifmw_build_containers_buildah_push | default ('false') | bool
     - not cifmw_build_containers_push_containers | bool
 
+ - name: "Push retagged images - image: {{ item }}"
+  become: true
+  when: cifmw_build_containers_retag_images | default(false) | bool
+  ansible.builtin.command: >
+    buildah push {{ item }}:{{ cifmw_build_containers_retag_string }}
+  loop: "{{ built_images.stdout_lines }}"
```
raukadah, line 98:
```suggestion
 - name: "Push retagged images - image: {{ item }}"
   become: true
   when: cifmw_build_containers_retag_images | default(false) | bool
   ansible.builtin.command: >
       buildah push {{ item }}:{{ cifmw_build_containers_retag_string }}
    loop: "{{ built_images.stdout_lines }}"
```

frenzyfriday (commented):
I am missing some context here:
What is the difference between `cifmw_build_containers_tag_string`, `cifmw_build_containers_image_tag` and `cifmw_build_containers_retag_string`

I am wondering if we could have one list that will have as many tags and retags needed, and the tag and push tasks will iterate over that?

rlandy:
> I am missing some context here: What is the difference between `cifmw_build_containers_tag_string`, `cifmw_build_containers_image_tag` and `cifmw_build_containers_retag_string`
> 
> I am wondering if we could have one list that will have as many tags and retags needed, and the tag and push tasks will iterate over that?

We want to retag after running a test - so we are tagging at two diff times - if thst makes sense

roles/build_containers/tasks/main.yml
```diff
@@ -90,6 +90,14 @@
     - cifmw_build_containers_buildah_push | default ('false') | bool
     - not cifmw_build_containers_push_containers | bool
 
+- name: "Retag each image and push to registry: {{ item }}"
+  become: true
+  ansible.builtin.command: >
+    buildah tag {{ item }}:{{ cifmw_build_containers_image_tag }} {{ item }}:{{ cifmw_build_containers_retag_string }}
+    buildah push {{ item }}:{{ cifmw_build_containers_retag_string }} {{ item }}:{{ cifmw_build_containers_retag_string }}
```
raukadah, line 97: Based on my testing in different pr https://github.com/openstack-k8s-operators/ci-framework/pull/2996/files#diff-35c48fff63350c3d85b3c2c536183b5a12e5ad5d5858f74a5a9eecf3e2aa6d00R96
```suggestion
    buildah push --format v2s2 --all {{ item }}:{{ cifmw_build_containers_image_tag }} docker://{{ item }}:{{ cifmw_build_containers_retag_string }}
```
above command fixes my issue.
Below was the past issue 
Screenshot From 2025-05-20 09-26-42
and after running above command, it got addressed and now latest hash pointing to current .
Screenshot From 2025-05-20 10-59-35

danpawlik: nit: I suggest to add condintion:
```
when: cifmw_build_containers_tag_string != cifmw_build_containers_retag_string

**Comment by rlandy** [comment]
recheck

File: roles/build_containers/tasks/tag.yml [hunk 4]
```diff
@@ -0,0 +1,97 @@
...
+    mode: "0644"
+    force: true
+  register: result
+  until:
+    - result.status_code is defined
+    - result.status_code == 200
+  retries: 6
+  delay: 50
+
+- name: Get built_images from the log file # noqa risky-shell-pipe
```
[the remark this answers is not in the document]
rlandy: done

rlandy:
recheck

danpawlik:
recheck

bshewale:
lgtm

bshewale (approved):

danpawlik (approved):

PR #2992, opened 2025-05-16
Use modules for enabling rhel subscription

danpawlik:
We should start using Ansible modules instead of executing commands via command/shell module. Dedicated modules might handle in better way that operation.

### DIFF
--- a/roles/adoption_osp_deploy/tasks/login_registries.yml
+++ b/roles/adoption_osp_deploy/tasks/login_registries.yml
@@ -20,10 +20,12 @@
     - cifmw_adoption_osp_deploy_rhsm_key is defined
   become: true
   no_log: true
-  ansible.builtin.command: >-
-    subscription-manager register --force
-    --org "{{ cifmw_adoption_osp_deploy_rhsm_org }}"
-    --activationkey "{{ cifmw_adoption_osp_deploy_rhsm_key }}"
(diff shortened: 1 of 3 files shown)


danpawlik:
NIT: need to migrate rhel registration to config repo.

evallesp (dismissed):

amartyasinha (dismissed):

danpawlik:
recheck

zuul.d/adoption.yaml
```diff
@@ -127,6 +127,7 @@
       - ^roles/openshift_setup/.*
       - ^roles/repo_setup/.*
       - ^roles/cifmw_setup/.*
+      - ^roles/adoption_osp_deploy/.*
```
ciecierski, line 130: this role is not used by adoption-standalone job, so it doesn't make sense to include it in files

ciecierski (commented):
Change makes sense, but needs to verified in uni-gamma or alpha downstream adoption run.

danpawlik:
recheck

zuul.d/adoption.yaml
danpawlik: ah, right.

frenzyfriday (approved):

amartyasinha (approved):

PR #2993, opened 2025-05-16
Add variable to turn off artifact masking

lmiccini:
In our lab this takes 1h+ (va-pidone, 3 masters and 4 workers).
Let's add a variable to control w/e we want to mask the artifacts and give people a chance to disable it if required.


abays (dismissed):
/lgtm

abays:
recheck

amartyasinha (approved):

PR #2995, opened 2025-05-19
Use sync module instead of rsync command

eshulman2:
Change repo sync to use ansible.posix.synchronize to allow more flexibility

### DIFF
--- a/roles/reproducer/tasks/push_code.yml
+++ b/roles/reproducer/tasks/push_code.yml
@@ -124,8 +124,12 @@
       delegate_to: localhost
       when:
         - item.src is abs or item.src is not match('.*:.*')
-      ansible.builtin.command:  # noqa: command-instead-of-module
-        cmd: "rsync -ar {{ item.src }} zuul@controller-0:{{ item.dest }}"
+      ansible.posix.synchronize:
+        src: "{{ item.src }}"
+        dest: "zuul@controller-0:{{ item.dest }}"
+        archive: true
+        recursive: true
+        delete: true
       loop: "{{ _cifmw_reproducer_all_repositories }}"
       loop_control:
         label: "{{ item.src | basename }}"


roles/reproducer/tasks/push_code.yml
```diff
@@ -124,8 +124,13 @@
...
         - item.src is abs or item.src is not match('.*:.*')
-      ansible.builtin.command:  # noqa: command-instead-of-module
-        cmd: "rsync -ar {{ item.src }} zuul@controller-0:{{ item.dest }}"
+      ansible.posix.synchronize:
+        src: "{{ item.src }}"
+        dest: "zuul@controller-0:{{ item.dest }}"
+        archive: true
+        recursive: true
+        delete: true
+        use_ssh_args: true
```
cjeanner, line 133: I'd rather NOT use that `use_ssh_args`:
- controller-0 inventory entry is dynamically injected during the run
- the framework generates specific ssh_config options for all of the managed virtual machines, enabling proxyJump, user and some other things

If really you want to allow overriding the default behavior, you probably should use [rsync_opts](https://docs.ansible.com/ansible/latest/collections/ansible/posix/synchronize_module.html#parameter-rsync_opts) instead, with a default value set to `omit` so that it can be empty/omitted, matching the old behavior.

eshulman2: the thing is that without it it might not use certain configurations set to allow ssh connection. generally it is possible to use the rsync opts although I would like to avoid having to duplicate the already existing ssh config from the inventory into another var.

roles/reproducer/tasks/push_code.yml
```diff
@@ -125,7 +125,7 @@
       when:
         - item.src is abs or item.src is not match('.*:.*')
       ansible.builtin.command:  # noqa: command-instead-of-module
-        cmd: "rsync -ar {{ item.src }} zuul@controller-0:{{ item.dest }}"
+        cmd: "rsync {{ cifmw_push_code_extra_ssh_options | default('') }} -ar {{ item.src }} zuul@controller-0:{{ item.dest }}"
```
cjeanner, line 128: sounds safer than switching to the module + use _ssh_args.
Adding that option in the doc would be nice.
Also, I don't know if it's still the usage, but the parameter should be named differently, something like `cifmw_reproducer_push_code_ssh_opts` (referencing the role name as prefix).

You can also avoid the `|default('')` by adding it to the defaults/main.yml directly (with associated doc in the README).

eshulman2: will do after initial testing

roles/reproducer/tasks/push_code.yml
```diff
@@ -124,8 +124,12 @@
...
       when:
         - item.src is abs or item.src is not match('.*:.*')
-      ansible.builtin.command:  # noqa: command-instead-of-module
-        cmd: "rsync -ar {{ item.src }} zuul@controller-0:{{ item.dest }}"
+      ansible.posix.synchronize:
+        src: "{{ item.src }}"
+        dest: "zuul@controller-0:{{ item.dest }}"
+        archive: true
+        recursive: true
+        delete: true
```
evallesp, line 132: (non-blocking) question: Not saying it's bad. but do we want to add this behaviour?

eshulman2: yes, when using re-use I would like files to be removed to keep proper track of the state of the repo even for files that got removed.

evallesp (approved):

sdatko (approved):

PR #2998, opened 2025-05-20
ci_dcn_site: Set cross_az_attach at nova-api

krcmarik:
We set the cross_az_attach to False at the
nova compute level and we should set it at
the nova api level too.


fultonj (approved):
/approve
/lgtm

PR #2999, opened 2025-05-20
[BGP] Remove octbr from network-values

eduolivares:
With BGP, octavia provider network should only be deployed on networker nodes and to do this, it has to be removed from NNCP configuration.

OSPRH-10768


danpawlik (dismissed):
If you say that it is not needed :)

PR #3000, opened 2025-05-20
Move steps of update run from install_yaml to ci-framework

ciecierski:
Rewrite install_yaml update steps to ci-framework directly.
Steps in install_yamls are left for developers.

Fake update path is not used in CI so it is not migrated.

### DIFF
--- a/roles/update/tasks/update_run.yml
+++ b/roles/update/tasks/update_run.yml
@@ -1,0 +1,196 @@
+---
+- name: Wait for MinorUpdateAvailable condition available
+  kubernetes.core.k8s_info:
+    kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
+    api_key: "{{ cifmw_openshift_token | default(omit) }}"
+    context: "{{ cifmw_openshift_context | default(omit) }}"
+    api_version: core.openstack.org/v1beta1
+    kind: OpenStackVersion
+    namespace: "{{ cifmw_update_namespace }}"
+  register: _openstackversion_info
+  retries: 40
+  delay: 15
+  until:
+    - _openstackversion_info.resources[0].status.conditions is defined
+    - >-
+      _openstackversion_info.resources[0].status.conditions |
+      selectattr("type", "equalto", "MinorUpdateAvailable") |
+      selectattr("status", "equalto", "True") | length > 0
+
+# start update by changing targetVersion to new availableVersion
+- name: Change targetVersion to availableVersion
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  ansible.builtin.shell: >-
+    oc patch openstackversion
+    $(oc get openstackversion
+    -o=jsonpath='{.items[].metadata.name}')
+    --type merge --patch '{"spec":{"targetVersion":_cifmw_update_openstack_version}}'
+
+- name: Wait for ovn update completion on control plane
+  kubernetes.core.k8s_info:
+    kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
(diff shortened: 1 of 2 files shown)


roles/update/tasks/update_run.yml
```diff
@@ -0,0 +1,196 @@
...
+    PATH: "{{ cifmw_path | default(ansible_env.PATH) }}"
+  ansible.builtin.shell: >-
+    set -o pipefail;
+    oc -n {{ cifmw_update_namespace }}
+    get openstackdataplanenodeset -o name
+    | awk -F'/' '{print $2}'
+  register: cifmw_update_node_sets
+  changed_when: false
+
+- name: Create the OpenStackDataPlaneDeployment CR used for ovn udpate
```
pinikomarov, line 150: is this task specifically ovn related ? maybe the task name can be different , i,e, "general update trigger" ?

roles/update/tasks/update_run.yml
```diff
@@ -0,0 +1,196 @@
...
+
+- name: Create the OpenStackDataPlaneDeployment CR to trigger OpenStackDataPlaneDeployment update
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path | default(ansible_env.PATH) }}"
+  ansible.builtin.command: >-
+    oc -n {{ cifmw_update_namespace }}
+    create -f {{ cifmw_update_artifacts_basedir }}/edpm_update.yaml
+
+- name: Wait for update completion on data plane
```
pinikomarov, line 180: should there be a compute nodes reboot here ? is there a reboot handling of the computes nodes going on later in the existing workflow?

pinikomarov:
the overall update workflow looks looks , 
added some questions regarding implementation stuff

roles/update/tasks/update_run.yml
ciecierski: Reboot is tested later in existing workflow: https://github.com/openstack-k8s-operators/ci-framework/blob/b7e23c7bb7687c22120afd865dce420adc0d6e5b/roles/update/tasks/main.yml#L78

roles/update/tasks/update_run.yml
ciecierski: good catch, will fix it

roles/update/tasks/update_run.yml
```diff
@@ -0,0 +1,196 @@
...
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path | default(ansible_env.PATH) }}"
+  ansible.builtin.shell: >-
+    set -o pipefail;
+    nodes_with_ovn=()
+    # Get the names of all OpenStackDataPlaneNodeSet resources
+    openstackdataplanenodesets=$(oc get openstackdataplanenodeset -o custom-columns=NAME:.metadata.name,SERVICES:.spec.services --no-headers)
+    # Loop through each OpenStackDataPlaneNodeSet
+    while read -r node_name services; do
```
danpawlik, line 59: I understand - it would work, but... why? Why not doing that in 2 tasks and use ansible loop for that?

danpawlik:
I know that wild shell here is something normal, but maybe we should start do more things using ansible and relay on it? Otherwise we can really create a shell script and just execute it with `-x`.

jistr:
Hmm i wonder if we should reconsider such a split between what CI runs and what developers run. Aren't we worried this will be troublesome by creating too many issues of type "works in my env, doesn't work in CI" or vice versa?

Personally i would prefer if we can use a single test suite for both CI and dev envs. I don't have a strong opinion whether the test suite should live in install_yamls or in ci-framework, and whether it's bash scripts or Ansible, but i think we should have a single test suite that we can use from both places (like we did for Adoption or OS Migrate).

PR #3006, opened 2025-05-22
Reduce retry count for iscsi login

rajathere:
In a previous PR, we added custom login timeout values in the iscsid.conf file[1].
However, even with reduced values (from 120 seconds to 15 seconds), we still see failures in CI jobs due to timeout.

To further improve the situation, we can set the retry count to 1 making the wait time 5 seconds which should be sufficient to pass the tempest wait threshold (and we shouldn't really need to retry in jobs anyway).

CIX: https://issues.redhat.com/browse/OSPCIX-805

[1] https://github.com/openstack-k8s-operators/ci-framework/pull/2060

### DIFF
--- a/hooks/playbooks/compute-iscsi-config.yml
+++ b/hooks/playbooks/compute-iscsi-config.yml
@@ -9,7 +9,7 @@
         path: /etc/iscsi/iscsid.conf
         line: "{{ item }}"
       loop:
-        - 'node.session.initial_login_retry_max = 3'
+        - 'node.session.initial_login_retry_max = 1'
         - 'node.conn[0].timeo.login_timeout = 5'
 
     - name: Restart iscsid container to refresh /etcd/iscsid.conf


hooks/playbooks/compute-iscsi-config.yml
```diff
@@ -9,7 +9,7 @@
         path: /etc/iscsi/iscsid.conf
         line: "{{ item }}"
       loop:
-        - 'node.session.initial_login_retry_max = 3'
+        - 'node.session.initial_login_retry_max = 1'
```
sdatko, line 12: Let me just drop here a [comment](https://github.com/openstack-k8s-operators/ci-framework/pull/2060/files#r1673792367) from the original #2060 change:
> I think we can change this to 1, because it either connects or it doesn't, and these are for tests, not production environments...

sdatko (approved):

hooks/playbooks/compute-iscsi-config.yml
rajathere: That's true but even if you look at the successful login attempts, even 1 second is enough, we are being very generous to give 5

May 19 21:58:34 compute-y9ihgosq-1 nova_compute[73665]: 2025-05-19 16:28:34.166 134 DEBUG oslo_concurrency.processutils [-] CMD "iscsiadm -m node -T iqn.1992-08.com.netapp:sn.1b60aa2ba3ab11e8911600a098d00ab0:vs.5 -p 10.46.246.75:3260 --login" **returned: 0 in 0.030s** execute /usr/lib/python3.9/site-packages/oslo_concurrency/processutils.py:422

May 19 21:58:34 compute-y9ihgosq-1 nova_compute[73665]: 2025-05-19 16:28:34.168 134 DEBUG oslo_concurrency.processutils [-] CMD "iscsiadm -m node -T iqn.1992-08.com.netapp:sn.1b60aa2ba3ab11e8911600a098d00ab0:vs.5 -p 10.46.246.74:3260 --login" **returned: 0 in 0.033s** execute /usr/lib/python3.9/site-packages/oslo_concurrency/processutils.py:422

PR #3009, opened 2025-05-26
Stop test-operator pod when ImagePullError

kstrenkova:
When encountering ImagePullError in the test-operator role, the pod would be stuck in that state for a longer period of time. This patch aims to stop the pod, if the waiting reason is ImagePullError to report the issue faster.


kstrenkova:
I am closing this PR, as the patch could bring new errors (canceling jobs early) and would be useful only in one specific case - giving a wrong image name. That issue could be fixed just by giving the correct name in the next run and not adding any new risks. More is said in the ticket related to this issue.

PR #3010, opened 2025-05-26
Increase timeout for integration tests

danpawlik:
On some infras, sometimes there is a proxy error and the pods can not start properly because it can not pull images, so the CI job fails and in the log file, there is information:

    oc wait openstack/openstack -n openstack-operators --for condition=Ready --timeout=300s
    error: timed out waiting for the condition on openstacks/openstack
    make[1]: *** [Makefile:760: openstack_init] Error 1

Increase timeout twice to avoid potential errors and give some time for proxy and kubernetes to recover and retry pull image.


abays (approved):
/lgtm
/approve

bshewale (approved):

PR #3016, opened 2025-05-27
fix sushy_emulator: move register out of slurp module block

antonioromito:
This patch fixes a syntax error in the sushy_emulator role where the `register` keyword
was incorrectly placed inside the `slurp` module block. Ansible does not support that usage
and fails with:

  "Unsupported parameters for (ansible.builtin.slurp) module: register"

The fix moves `register` outside of the module block, restoring correct behavior.

Affected file:
- roles/sushy_emulator/tasks/collect_details.yml

Tested on a local deployment using ci-framework with an RHOSP17 undercloud.


antonioromito:
recheck

hjensas (approved):

PR #3022, opened 2025-05-30
Add retry when registering node in RH subscription or container registry

danpawlik:
Sometimes the CI jobs fails because there is too many request done for login to the container registry or to register system in Red Hat registration service.
Let's add delay and retry again to register the system or login to the container service.

### DIFF
--- a/roles/adoption_osp_deploy/tasks/login_registries.yml
+++ b/roles/adoption_osp_deploy/tasks/login_registries.yml
@@ -26,6 +26,10 @@
     release: "{{ ansible_distribution_version }}"
     force_register: true
     state: present
+  retries: 5
+  delay: 30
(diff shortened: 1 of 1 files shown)


odyssey4me (approved):

roles/adoption_osp_deploy/tasks/login_registries.yml
```diff
@@ -26,6 +26,10 @@
     release: "{{ ansible_distribution_version }}"
     force_register: true
     state: present
+  retries: 5
+  delay: 30
+  register: _rh_result
+  until: not _rh_result.failed
```
odyssey4me, line 32: Is 'not failed' the best option we have here?

frenzyfriday (approved):

PR #3024, opened 2025-05-30
Multinode EDPM CI job failing for nmstate provider

Jaganathancse:
This change is to update the incorrect network config template where used nic1 for dhcp and also used
br-ex controlplane config with default route.
so two default routes creating for nmstate provider scenario and this is not happened for ifcfg
provider and only one default route created for nic1. Proposed this PR is to fix incorrect network config.

### DIFF
--- a/hooks/playbooks/fetch_compute_facts.yml
+++ b/hooks/playbooks/fetch_compute_facts.yml
@@ -157,6 +157,12 @@
                     path: /spec/nodeTemplate/ansible/ansibleVars/neutron_public_interface_name
                     value: "{{ crc_ci_bootstrap_networks_out[_first_compute].default.iface | default('') }}"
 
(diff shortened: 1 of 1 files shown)


hooks/playbooks/fetch_compute_facts.yml
```diff
@@ -211,7 +211,7 @@
                         network_config:
                         - type: interface
                           name: nic1
-                          use_dhcp: true
+                          use_dhcp: false
```
karelyatin, line 214: nic1 is the primary nic which zuul connects to and disabling dhcp there would cause issues. so the failing job looks related to this.

karthiksundaravel (dismissed):
We need only one default route. Not sure if the nic1 needs DHCP enabled, if thats the necessity then this default route needs to be removed.

karelyatin (approved):

PR #3027, opened 2025-06-02
Use role instead of playbooks - 08-run-tests.yml

amartyasinha:
It is continuation of simplification job execution [1].

[1] https://github.com/openstack-k8s-operators/ci-framework/pull/2929

### DIFF
--- a/deploy-edpm.yml
+++ b/deploy-edpm.yml
@@ -85,13 +85,21 @@
   tags:
     - admin-setup
 
-- name: Import run test playbook
-  ansible.builtin.import_playbook: playbooks/08-run-tests.yml
-  vars:
-    pre_tests: "{{ (lookup('vars', 'pre_tempest', default=[])) }}"
-    post_tests: "{{ (lookup('vars', 'post_tempest', default=[])) }}"
-  tags:
-    - run-tests
+- name: Run cifmw_setup run_tests.yml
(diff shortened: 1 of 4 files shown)


amartyasinha:
recheck

eshulman2:
How should this role be used when debugging? should users create a playbook of their own to run this? I generally agree with this change but it would be nice to keep a version of this playbook for ease of use when debugging

amartyasinha:
> How should this role be used when debugging? should users create a playbook of their own to run this? I generally agree with this change but it would be nice to keep a version of this playbook for ease of use when debugging

As of now, we are not removing the playbooks. It may happen in the near future (currently not under the scope). For now, we are trying to remove nested import_playbook calls within the framework to reduce the complexity.

deploy-edpm.yml
```diff
@@ -85,13 +85,21 @@
...
+  gather_facts: false
+  tasks:
+    - name: Run Test
+      vars:
+        pre_tests: "{{ (lookup('vars', 'pre_tempest', default=[])) }}"
+        post_tests: "{{ (lookup('vars', 'post_tempest', default=[])) }}"
+      ansible.builtin.import_role:
+        name: cifmw_setup
+        tasks_from: run_tests.yml
+      when:
```
danpawlik, line 99: not needed condition comparing to old code.

PR #3032, opened 2025-06-03
Use role instead of playbooks - 99-logs.yml

bshewale:
It is continuation of simplification job execution [1].

[1]: https://github.com/openstack-k8s-operators/ci-framework/pull/2929

### DIFF
--- a/deploy-edpm.yml
+++ b/deploy-edpm.yml
@@ -117,6 +117,13 @@
         mode: "0644"
 
 - name: Run log related tasks
-  ansible.builtin.import_playbook: playbooks/99-logs.yml
-  tags:
-    - logs
+  hosts: "{{ cifmw_target_host | default('localhost') }}"
+  gather_facts: true
+  tasks:
+    - name: Run 99-logs.yml 
+      ansible.builtin.import_role:
+        name: os_must_gather
+        tasks_from: 99-logs.yml
+      tags:
+        - logs
+
--- a/roles/os_must_gather/tasks/99-logs.yml
+++ b/roles/os_must_gather/tasks/99-logs.yml
@@ -1,0 +1,107 @@
+- name: Early exit if no tests
+  when:
+    - zuul_log_collection | default('false') | bool
+  ansible.builtin.meta: end_play
+
+- name: Ensure cifmw_basedir param is set
+  when:
+    - cifmw_basedir is not defined
+  ansible.builtin.set_fact:
+    cifmw_basedir: "{{ ansible_user_dir }}/ci-framework-data"
+
+- name: Try to load parameters files
+  block:
+    - name: Check directory availabilty
+      register: param_dir
+      ansible.builtin.stat:
+        path: "{{ cifmw_basedir }}/artifacts/parameters"
+
+    - name: Load parameters files
+      when:
+        - param_dir.stat.exists | bool
+      ansible.builtin.include_vars:
+        dir: "{{ cifmw_basedir }}/artifacts/parameters"
+  always:
+    - name: Set custom cifmw PATH reusable fact
+      when:
+        - cifmw_path is not defined
+      ansible.builtin.set_fact:
+        cifmw_path: "{{ ansible_user_dir }}/.crc/bin:{{ ansible_user_dir }}/.crc/bin/oc:{{ ansible_user_dir }}/bin:{{ ansible_env.PATH }}"
+        cacheable: true
+
+- name: Set destination folder for the logs
+  ansible.builtin.set_fact:
+    logfiles_dest_dir: >-
+      {{
+        (
+          cifmw_basedir | default(ansible_user_dir ~ '/ci-framework-data'),
+          'logs/',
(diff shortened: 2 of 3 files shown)


deploy-edpm.yml
```diff
@@ -117,6 +117,13 @@
         mode: "0644"
 
 - name: Run log related tasks
```
danpawlik, line 119: it can be merged in previous play execution :) 
In other words, remove L119-L122

danpawlik:
> This seems more converging the entire `ci-framework` logging topic into the existing must-gather role. I see this as a wrong move because we risk to confuse the actions we execute in the role that shouldn't break the principle of managing a single thing (which is running openstack-must-gather). I'm not sure why removing a playbook and unpacking it into this role results in a simplification, but I see this as an attempt to rename must-gather into a logging role, while we should move the common code into a logging role.

If we are watching on that from this PR - no, that "simplification" is just an illusion, but if we check another playbooks what we were executing - then yes, it is simplification. Import playbook does not give transparent workflow, where moving that to roles:
- easy understand what is executed - no need to check which playbook was executing that role, because in the logs playbook name is not raised
- better control variables,
- in some part faster execution due there is no need to make delegation or gather_facts,
- easier to debug

About this PR: So if you think that [openstack-must-gather](https://github.com/openstack-k8s-operators/openstack-must-gather) is better for collecting logs, than let's remove the 99-logs play and use it directly. Is it ok @fmount ?

danpawlik:
@fmount `my point is simply related to the fact that I see tasks moved to the must-gather role that are not related to openstack-must-gather`  - not really. Tasks are comming from https://github.com/openstack-k8s-operators/ci-framework/blob/main/playbooks/99-logs.yml , copy-paste to new file which is located in roles. What @bshewale should mention in this PR - in the file `playbooks/99-logs.yml` there should be a comment: 

```
#
# NOTE: Playbook migrated to: cifmw_setup/tasks/SOMETHING.
# DO NOT EDIT THAT PLAYBOOK. IT WOULD BE REMOVED IN NEAR FUTURE.
#
```

At this point my suggestion is:
- move content from playbooks/99-logs.yml  to role
- add comment to the playbooks/99-logs.yml 
- when all playbooks would be migrated, drop playbooks
- migrate role execution related to 99-logs.yml and use openstack-must-gather

Alternative way:
- add comment to the playbooks/99-logs.yml
- in places, where we are calling 99-logs.yml  we should call openstack-must-gather

is it ok @fmount ?

fmount:
> * places, where we are calling 99-

The alternative way sounds good. You should include must-gather where you need, not the other way around.

bshewale:
recheck

ci/playbooks/e2e-collect-logs.yml
```diff
@@ -20,12 +20,18 @@
...
-      ansible.builtin.command:
-        chdir: "{{ ansible_user_dir }}/src/github.com/openstack-k8s-operators/ci-framework"
-        cmd: >-
-          ansible-playbook playbooks/99-logs.yml
-          -e @scenarios/centos-9/base.yml
+- name: Run log related tasks
+  hosts: "{{ cifmw_target_host | default('all') }}"
+  gather_facts: true
+  vars_files:
+    - @../../base.yml
```
amartyasinha, line 27: It should be:
```
  vars_files:
    - ../../scenarios/centos-9/base.yml
```

roles/ci_setup/tasks/run_logs.yml
```diff
@@ -0,0 +1,108 @@
+# end_play will end only current play, not the main edpm-deploy.yml
+- name: Early exit if no tests
+  when:
+    - zuul_log_collection | default('false') | bool
+  ansible.builtin.meta: end_play
```
amartyasinha, line 5: `end_play` should be replaced with `when` condition, as this task block is now in role instead of playbook. If `end_play` runs, it will end the main play instead of just ending logging.

Another way where `end_play` could be used can be referenced from [this](https://github.com/openstack-k8s-operators/ci-framework/blob/main/deploy-edpm.yml#L153-L184)

amartyasinha (changes requested):

amartyasinha (approved):

roles/cifmw_setup/tasks/run_logs.yml
```diff
@@ -0,0 +1,104 @@
+- name: Check if the logging requires
+  when: not zuul_log_collection | default('false') | bool
```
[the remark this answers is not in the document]
fmount: +1 to improve this on a new iteration.
Also, any reason why this condition couldn't be the other way around (it seems more intuitive) with a default defined here [1]?

[1] https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/cifmw_setup/defaults/main.yml

fmount (approved):
/lgtm

fmount:
I think this patch needs a follow up to run the must-gather role on the existing jobs :/

PR #3033, opened 2025-06-03
Add variable to overwrite NNCP_DNS_SERVER

danpawlik:
Earlier Ansible was relay on variable that was taken from /etc/ci/env/networking-info.yml which was generated during base job, but it does not include value with "ip4". This commit gives possibility to set custom NNCP_DNS_SERVER value.


danpawlik:
recheck

PR #3035, opened 2025-06-03
Drop -4 flag from curl command in download cirros-cloud image

ciecierski:
Do not force ipv4 in curl command only when resolving hostnames. In environment with ipv6 only let curl use ipv6 addresses when resolving hostnames.


ciecierski:
currently testing against both ipv6 and ipv4 environments

PR #3036, opened 2025-06-03
Separate deployment from testing

eshulman2:
Separate deployment and post-deployment tasks

This patch separates the deployment and post-deployment tasks
to allow for better reuse of the deployment/post-deployment tasks

- Separate the deployment and post-deployment tasks in the
  reproducer playbook
- Add a new playbook for post-deployment tasks (split deploy-edpm.yml)
- Update the deploy-edpm-reuse.yaml to include the new post-deployment playbook
- Update the roles/test_operator to include old run cleanup to allow re-use
- Update the roles/reproducer to include the new post-deployment tasks


arxcruz (dismissed):
/lgtm
/approve

deploy-edpm-reuse.yaml
```diff
@@ -114,3 +114,13 @@
       delegate_to: controller-0
       ansible.builtin.command:
         cmd: "/home/zuul/deploy-architecture.sh {{ cifmw_deploy_architecture_args | default('') }}"
+
+    - name: Run post deployment if instructed to
+      when:
+        - cifmw_post_deployment | default(true) | bool
+      no_log: "{{ cifmw_nolog | default(true) | bool }}"
+      async: "{{ 7200 + cifmw_test_operator_timeout | default(3600) }}"  # 2h should be enough to deploy EDPM and rest for tests.
```
danpawlik, line 122: default should not be 7200?

roles/reproducer/tasks/configure_post_deployment.yml
```diff
@@ -0,0 +1,27 @@
...
+- name: Prepare scripts on controller-0
+  delegate_to: controller-0
+  block:
+    - name: Push script
+      vars:
+        run_directory: "{{ _cifmw_reproducer_framework_location }}"
+        exports:
+          ANSIBLE_LOG_PATH: "~/ansible-post-deployment.log"
+        default_extravars:
+          - "@~/ci-framework-data/parameters/reproducer-variables.yml"
```
danpawlik, line 11: suggest to replace `~` with `{{ ansible_user_dir }}`

bshewale (dismissed):

roles/reproducer/tasks/configure_post_deployment.yml
```diff
@@ -0,0 +1,27 @@
+---
+- name: Prepare scripts on controller-0
+  delegate_to: controller-0
```
danpawlik, line 3: should not be "controllers"  group or we are sure that always would be controller-0 ?

eshulman2: Afaik should always be controller-0. This is also reflected in other playbook and parts of cifmw.

bshewale (approved):

arxcruz (approved):

deploy-edpm.yml
```diff
@@ -79,70 +79,3 @@
   ansible.builtin.import_playbook: playbooks/06-deploy-architecture.yml
   tags:
     - edpm
-
-- name: Run Post-deployment admin setup steps, test, and compliance scan
```
raukadah, line 83: @eshulman2 deploy-edpm.yml playbook is used in upstream edpm job. SInce we moved these tasks to post-deployment.yaml, tempest is no longer running in https://softwarefactory-project.io/zuul/t/rdoproject.org/builds?job_name=podified-multinode-edpm-deployment-crc job

PR #3037, opened 2025-06-03
Set default values to Resources parameter

kstrenkova:
There is a recent change that the Resources parameter will not have defaults for workflow, but it will be propagated from the main section of the CR. To make it work, it is important that there will be no empty value set, instead Resources parameter should be omitted.

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/3024


kstrenkova:
recheck

kstrenkova:
> @kstrenkova can you please create a PR that showcases that this PR will work hand in hand with the change from Alan? You can make a PR on the side of the test-operator with Alan's changes and make this one as Depends-On. I think we talked about this, if not apologies:). We can talk this through during next call.

Hi @lpiwowar, couldn't we just recheck Alan's patch as it already has the Depends-On? Or am I missing something?

kstrenkova:
The issue has been addressed and it seems to be working correctly :partying_face: I think we are ready to merge both patches, what do you think @lpiwowar?

adrianfusco (approved):

PR #3041, opened 2025-06-05
DNM: Decrease tempest concurrency

evallesp:
Once the tempest pod failed it gets the status OOMKilled, which means that it tried to get more memoery than the maximum.
Also those tempest execution has not the summary at the end of passing and failling tests.
We would need to increase the memory, but for testing purpose we went the way of reducing concurrency.

DNM


evallesp:
> Could the description/commit message be a bit more specific about this being applied just to a specific job? I was a bit concerned when I opened the pull request.

surething, I'm just checking. Updated.

PR #3047, opened 2025-06-06
Rename _network_data in create_cloud_init_iso to _network_data_config_drive

ciecierski:
There is conflict of variable names. _network_data is already defined and exposed using set_fact[1]. Variable for config_drive should be named diffrently to not generate wrong network-configs for vms.

[1] https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/libvirt_manager/tasks/create_networks.yml#L185

### DIFF
--- a/roles/libvirt_manager/tasks/create_cloud_init_iso.yml
+++ b/roles/libvirt_manager/tasks/create_cloud_init_iso.yml
@@ -50,12 +50,12 @@
       when:
         - vm_data.networkconfig | type_debug == "dict"
       ansible.builtin.set_fact:
-        _network_data: "{{ vm_data.networkconfig }}"
+        _libvirt_manager_network_data: "{{ vm_data.networkconfig }}"
     - name: "Define the network config for each vm"
       when:
         - vm_data.networkconfig | type_debug == "list"
       ansible.builtin.set_fact:
-        _network_data: "{{ vm_data.networkconfig[vm_idx] }}"
+        _libvirt_manager_network_data: "{{ vm_data.networkconfig[vm_idx] }}"
 
 - name: "Call the config_drive role"
   vars:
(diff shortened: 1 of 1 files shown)


roles/libvirt_manager/tasks/create_cloud_init_iso.yml
```diff
@@ -50,20 +50,20 @@
       when:
         - vm_data.networkconfig | type_debug == "dict"
       ansible.builtin.set_fact:
-        _network_data: "{{ vm_data.networkconfig }}"
+        _network_data_config_drive: "{{ vm_data.networkconfig }}"
```
evallesp, line 53: Let's add the role prefix name as suggestefd in ansible good practices: [3.1.4. Naming parameters](https://redhat-cop.github.io/automation-good-practices/#_naming_parameters)

roles/libvirt_manager/tasks/create_cloud_init_iso.yml
```diff
@@ -50,20 +50,20 @@
...
         - vm_data.networkconfig | type_debug == "dict"
       ansible.builtin.set_fact:
-        _network_data: "{{ vm_data.networkconfig }}"
+        _config_drive_network_data: "{{ vm_data.networkconfig }}"
     - name: "Define the network config for each vm"
       when:
         - vm_data.networkconfig | type_debug == "list"
       ansible.builtin.set_fact:
-        _network_data: "{{ vm_data.networkconfig[vm_idx] }}"
+        _config_drive_network_data: "{{ vm_data.networkconfig[vm_idx] }}"
```
holser, line 58: > All defaults and all arguments to a role should have a name that begins with the role name to help avoid collision with other names. Avoid names like packages in favor of a name like foo_packages.

Could you change it to libvirt_manager_network_data?

holser (dismissed):

roles/libvirt_manager/tasks/create_cloud_init_iso.yml
evallesp: yep please: Let's add the role prefix name as suggestefd in ansible good practices: [3.1.4. Naming parameters](https://redhat-cop.github.io/automation-good-practices/#_naming_parameters)

hjensas:
recheck

evallesp (approved):
Thanks!

odyssey4me:
Worth a good practise note from https://redhat-cop.github.io/automation-good-practices/#_ansible_guidelines

```
Use the smallest scope for variables. Facts are global for playbook run, so it is preferable to use other types of variables. Therefore limit (preferably avoid) the use of set_fact. Role variables are exposed to the whole play when the role is applied using roles: or import_role:. A more restricted scope such as task or block variables is preferred.
```

PR #3048, opened 2025-06-06
adoption_osp_deploy: allow to pass private overcloud args

tosky:
A new parameter allow users to specify an additional
overcloud deploy environment file (-e), which can be used
to pass private/restricted parameters.

cifmw_adoption_osp_deploy_overcloud_extra_args: points to a string
which contains an environment file.


tosky:
The change is ready for review

jistr (approved):

PR #3054, opened 2025-06-09
[DNM?] Revert "Use local copy of common-requirements.txt while installation"

sdatko:
The change https://github.com/openstack-k8s-operators/ci-framework/pull/3014 broke the deployments on baremetal environments and we are not sure when it will be fixed properly, yet we need to proceed with our work on some changes from `main` branch. This merge request will be used as depends-on in my testprojects.

This reverts commit 299a5d6035b6a26accd24a1b67572f9bef601b2e.


sdatko:
Yes, not needed anymore

PR #3055, opened 2025-06-09
Remove printing redhat-release in adoption_osp_deploy

danpawlik:
The task should iterate on list of hosts, which should be provided
later as "_vm" variable. That was missing, so the Ansible was
failing because of undefined variable.
Printing the redhat-release should be added into pre-run config after
registration, but in this kind of CI jobs we did not spotted any issues
now.

### DIFF
--- a/roles/adoption_osp_deploy/tasks/prepare_overcloud.yml
+++ b/roles/adoption_osp_deploy/tasks/prepare_overcloud.yml
@@ -121,15 +121,6 @@
         loop_var: _vm
         pause: 1
 
-    - name: Get current /etc/redhat-release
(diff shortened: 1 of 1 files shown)


ciecierski:
For the context: Printing content of /etc/redhat-release is both broken[1] and should
be implemented  in pre-run config playbook after registration tasks instead.   

[1]https://github.com/openstack-k8s-operators/ci-framework/pull/3040#discussion_r2134995734

danpawlik:
@tosky updated. Sorry for that, normally it should be a draft. After getting conversation with @ciecierski  it seems tasks does not need to be executed.

PR #3064, opened 2025-06-11
rsync only if source directory exists

fultonj:
In reproducer role's configure_controller.yml tasks file, stat the cifmw_reproducer_src_dir before attempting to rsync it and then only rsync it if it exists. This is a reasonable thing to do before calling rsync.

If the cifmw_reproducer_src_dir does not exist, then fall back to github for the Ansible dependencies. This will not break the current behavior for which the rsync task was added.

Jira: [OSPRH-17434](https://issues.redhat.com//browse/OSPRH-17434)

### DIFF
--- a/roles/reproducer/tasks/configure_controller.yml
+++ b/roles/reproducer/tasks/configure_controller.yml
@@ -350,7 +350,14 @@
             name: sushy_emulator
             tasks_from: verify.yml
 
-    - name: Sync local repositories to other hosts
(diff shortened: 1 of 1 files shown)


danpawlik (approved):

fultonj:
> @fultonj hey, thanks for that patch. I guess you are running reproducer locally, right? 

Yes, I am running it locally. 

It's very important for my work to be able to run it locally when creating new VAs or DTs.

PR #3076, opened 2025-06-18
Fix the default value of cifmw_nfs_target

danpawlik:
Change default variable value for cifmw_nfs role

The target var should be "computes", not "compute".
It raises an issue that later the host is skipped by the Ansible.

This commit fixes porting issue raised in Pull Request [1].

[1] https://github.com/openstack-k8s-operators/ci-framework/pull/3038


tosky:
Thank you!

PR #3077, opened 2025-06-18
fix upstream gate

eshulman2:
patch 3036 seperated testing from deployment but did not updated upstream job this patch should fix upstream gating and run post deployment as well as deployment playbook


raukadah:
recheck

PR #3078, opened 2025-06-18
Patch samples registry

sdatko:
Due to https://issues.redhat.com/browse/OCPBUGS-30313 we observe the following message when checking the cluster health: `clusteroperators/openshift-samples is progressing and degraded`. This change simply patches the samples registry so the problem is not occurring anymore.


danpawlik:
Looks promising

dasm (approved):

PR #3080, opened 2025-06-19
Set optimize.proxy_host_address value in tempest workflow

raukadah:
optimize.proxy_host_address needs controller_ip while running tempest in uni03gamma job. In order to achieve that, we need to set the value via ansible tasks.

This pr sets the same.

### DIFF
--- a/roles/test_operator/tasks/tempest-tests.yml
+++ b/roles/test_operator/tasks/tempest-tests.yml
@@ -159,7 +159,8 @@
               item |
               combine({'tempestconfRun': {'overrides':
                 item.tempestconfRun.overrides + ' ' +
-                'whitebox_neutron_plugin_options.proxy_host_address ' + controller_ip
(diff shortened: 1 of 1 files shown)


roles/test_operator/tasks/tempest-tests.yml
```diff
@@ -159,7 +159,8 @@
               item |
               combine({'tempestconfRun': {'overrides':
                 item.tempestconfRun.overrides + ' ' +
-                'whitebox_neutron_plugin_options.proxy_host_address ' + controller_ip
+                'whitebox_neutron_plugin_options.proxy_host_address ' + controller_ip + ' ' +
```
danpawlik, line 162: it's going to be more and more complex. Maybe add a comment in L153 for others with example how it should look like. WDYT?

raukadah: Thank you @danpawlik for the review. this pr is no longer needed.

PR #3081, opened 2025-06-19
Add cifmw_cleanup_architecture var to enable/disable cleanup

raukadah:
It will help to reuse the ocp environment when architecture deployment failed in pre step.


evallesp (approved):
LGTM

PR #3088, opened 2025-06-25
Set gather_facts to true for post deployment

eshulman2:
some issues are caused by missing facts after new post deployment seperation was introduced, setting gather_facts to true for post deployment to avoid those issues


eduolivares (approved):
this PR resolved an issue that affects integration jobs (at least)

evallesp (approved):

PR #3092, opened 2025-06-26
Allow setting test-operator namespace per stage

sdatko:
This will allow us to run the testing with test-operator in environments where multiple OpenStack namespaces exist on a single OpenShift cluster.


danpawlik (approved):

sdatko:
The CI passed here, so existing configurations of test-operator still work, but it also passed in my CI job with multiple namespaces – we are happy :heart:

PR #3119, opened 2025-07-10
Allow deploy OCP with single master node

danpawlik:
That commit allows to deploy OCP with single master node. It would be useful for environments with limited resources.


danpawlik:
taken from https://github.com/openshift-metal3/dev-scripts/blob/master/agent/docs/config-presets-and-options.md - that SNO is possible.

PR #3121, opened 2025-07-11
Set ovn-egress-iface=true for br-ex physical NIC

eduolivares:
In order to fix [OSPRH-17551](https://issues.redhat.com//browse/OSPRH-17551) and make BW limits properly applied to physical ports (ports from neutron VLAN and flat networks), the br-ex member interfaces need to be configured with ovn-egress-iface=true.

This is applies when os-net-config is used with the nmstate driver.


eduolivares:
this has been successfully tested downstream

evallesp (approved):

danpawlik (approved):

PR #3122, opened 2025-07-13
[DNM] Add compute-1 to cephadm deployment targets

lkuchlan:


playbooks/ceph.yml
```diff
@@ -261,7 +261,7 @@
 
 - name: Bootstrap Ceph and apply spec
   tags: cephadm
-  hosts: "{{ (groups[cifmw_ceph_target | default('computes')] | default([]))[:1] }}"
+  hosts: "{{ (groups[cifmw_ceph_target | default('computes')] | default([]))[:2] }}"
```
fmount, line 264: -1: this is the wrong way of doing that, because it might have side effects over the rest of the playbook where it is assumed that tasks are executed against `mon[0]`.
Is there a specific reason to install cephadm everywhere? we might want to eventually `extract` that task and and execute it outside of the boostrap context (e.g. before deploying ceph), to make sure all the prereq are met.
For example you can have a little playbook, executed against `cifmw_ceph_target | default(computes)` where you include:

```
- name: Install cephadm package
  ansible.builtin.include_tasks: install_cephadm.yml
```
and you can extend the cephadm deployment to the rest of the cluster. Note that in theory we only deploy in the first mon because we usually do not install it in CephStorage (OSD) nodes, but only in `_mon` or `_admin` cephadm nodes.

PR #3126, opened 2025-07-15
Remove deprecated playbooks - Part 1

bshewale:
As we we deprecated playbooks by moving the playbooks into the roles So this change is removing deprecated playbooks from the repo.


bshewale:
> This requires through testing downstream and it would be better to do it playbook-by-playbook. Please at least change it to draft.

I agree with this comment, and plan is same as you mentioned and also created jira task as per the plan. We are going to remove the playbooks one-by-one. I have created PR here https://github.com/openstack-k8s-operators/ci-framework/pull/3128  for one by one removal of playbook.
And also I was looking around that and was also checking the impact on both upstream and downstream. We are going to remove the playbooks after all testing and approvals from DFG's and CIOPS team.

PR #3130, opened 2025-07-15
Add cleanup run after tempest tests

dsariel:
Required by OSPNW-1148. Some tests have leftovers. This change cleans them and allows rerunning splitted test jobs.

OSPNW-1148.

### DIFF
--- a/roles/test_operator/files/openstack_cleanup.sh
+++ b/roles/test_operator/files/openstack_cleanup.sh
@@ -1,0 +1,134 @@
+#!/bin/bash
+set -euo pipefail
+
+log() {
+  echo "[+] $1"
+}
+
+require_command() {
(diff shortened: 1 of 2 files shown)


roles/test_operator/files/openstack_cleanup.sh
```diff
@@ -0,0 +1,134 @@
...
+
+log() {
+  echo "[+] $1"
+}
+
+require_command() {
+  command -v "$1" >/dev/null 2>&1 || { echo "Required command '$1' not found. Install it and retry."; exit 1; }
+}
+
+# require_command jq
```
danpawlik, line 12: leftover or wanna use 
```
if ! command -v jq; then exit 1; fi
```

PR #3131, opened 2025-07-15
(update): Simplify registry login for openstackclient

ciecierski:
This commit refactors the logic for creating the local openstackclient container.
The previous implementation relied on parsing OpenStackDataPlaneNodeSet
resources to determine registry credentials, which was complex and error-prone.

The new approach simplifies this by:
- Fetching the openstackclient image directly from its pod definition.
- Conditionally logging into `registry.redhat.io` only when the image originates from there.
- Using the `cifmw_registry_token` dictionary for credentials to logging into `registry.redhat.io`.

Fixes: [OSPCIX-969](https://issues.redhat.com//browse/OSPCIX-969)

### DIFF
--- a/roles/update/tasks/create_local_openstackclient.yml
+++ b/roles/update/tasks/create_local_openstackclient.yml
@@ -1,68 +1,4 @@
 ---
-- name: Gather NodeSet resource information
-  kubernetes.core.k8s_info:
-    kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
-    namespace: "openstack"
-    api_key: "{{ cifmw_openshift_token | default(omit) }}"
-    context: "{{ cifmw_openshift_context | default(omit) }}"
-    kind: "OpenStackDataPlaneNodeSet"
-    api_version: "dataplane.openstack.org/v1beta1"
-  register: _cifmw_update_osdpns_all_info
-
-- name: Fail if no OSDPNS resources are found
-  ansible.builtin.fail:
-    msg: "No OSDPNS resources found in the 'openstack' namespace!"
-  when: _cifmw_update_osdpns_all_info.resources | length == 0
-
-- name: Choose the first OSDPNS resource which has edpm_container_registry_logins defined
-  ansible.builtin.set_fact:
-    _cifmw_update_osdpns_info: >-
-      {{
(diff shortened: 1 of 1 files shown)


sathlan (commented):
LGTM already, running a testproject against an uni01alpha and non-trunk env.

roles/update/tasks/create_local_openstackclient.yml
```diff
@@ -80,7 +16,22 @@
...
-    openstackclient_image: "{{ _cifmw_update_openstackclient_pod.resources[0].spec.containers[0].image }}"
+    openstackclient_image: "{{ _cifmw_update_openstackclient_pod | community.general.json_query('resources[0].spec.containers[0].image') | default('') }}"
+
+- name: Login to registry.redhat.io if needed
+  when: "'registry.redhat.io' in openstackclient_image"
+  block:
+    - name: Fail if cifmw_registry_token.credentials is not defined
+      ansible.builtin.fail:
+        msg: "cifmw_registry_token.credentials is not defined, cannot login to registry.redhat.io"
+      when: (cifmw_registry_token | default({})).get('credentials') is not defined
```
danpawlik, line 27: maybe better:
```
when: "'credentials' not in  cifmw_registry_token | default({})"
```

roles/update/tasks/create_local_openstackclient.yml
```diff
@@ -80,7 +16,22 @@
 
 - name: Set the openstackclient image fact
   ansible.builtin.set_fact:
-    openstackclient_image: "{{ _cifmw_update_openstackclient_pod.resources[0].spec.containers[0].image }}"
+    openstackclient_image: "{{ _cifmw_update_openstackclient_pod | community.general.json_query('resources[0].spec.containers[0].image') | default('') }}"
```
danpawlik, line 19: nit: suggest to add one more task with `fail` that will verify if the openstackclient_image is not empty

ciecierski: There is Fail task if pod is not found before. If pod is found, then image should be there too

roles/update/tasks/create_local_openstackclient.yml
ciecierski: fixed, thank you

PR #3132, opened 2025-07-16
DNM: Ensure ci playbooks find action plugins when run in zuul

amartyasinha:
Since playbooks which are run directly by zuul does not uses ansible.cfg from ci-framework repo, playbooks are not able to find where custom action plugins are defined.

This commit tries to check whether a symlink to the action plugins help us or not.

### DIFF
--- a/ci/playbooks/action_plugins/ci_kustomize.py
+++ b/ci/playbooks/action_plugins/ci_kustomize.py
@@ -1,0 +1,1 @@
+../../../plugins/action/ci_kustomize.py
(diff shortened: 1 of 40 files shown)


amartyasinha:
Due to the limitation of legacy plugin system, it is not possible to use local module_utils in local action_plugins.
Ref: https://forum.ansible.com/t/import-of-local-module-utils-within-local-action-plugin-not-working/43965/2?u=amartyasinha

PR #3135, opened 2025-07-16
Adjust the MTU for tcib job

TristanCacqueray:
This change tries to fix an issue with container build timing out when the host has a large MTU.


danpawlik:
Let's postpone the PR for a while. After we switch network to other, the CI jobs seems to be working better (even with high MTU).

PR #3138, opened 2025-07-17
Removing pre_tempest and post_tempest is no longer in use

pojadhav:
Ticket: https://issues.redhat.com/browse/OSPRH-6826


bshewale:
I would suggest you can test this change downstream by running the job in testproject.

tosky:
Any results from the testing?

PR #3140, opened 2025-07-17
[debug] Print the rhos-release URL

sdatko:
We are investigating a case and it appeared that despite setting the variable to the desired value, in the end in the job it ends up requesting something else... This debug task would allow us inspecting what we really get.


sdatko:
recheck

PR #3142, opened 2025-07-18
Fix terminology: use "architecture-based" instead of "VA"

tosky:
The architecture repository contains severals DTs, some of which are VAs. At the beginning the architecture-runner code was developed with VA1 (later HCI VA) in mind, but we are really talking about architecture-based deployments, not specifically to VAs.

This change only affects the ansible output and it has no functional impact.

### DIFF
--- a/deploy-edpm.yml
+++ b/deploy-edpm.yml
@@ -150,7 +150,7 @@
       tags:
         - edpm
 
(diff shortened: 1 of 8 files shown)


danpawlik (approved):
Thank you

PR #3152, opened 2025-07-23
Trigger cifmw-pod-zuul-files on each change

danpawlik:
It is not a first time, that some molecule job is missing, because some role was added and cifmw-pod-zuul-files job was not triggered.

Add the CI job and run it for each change that was done.


sdatko (approved):

danpawlik:
recheck

PR #3153, opened 2025-07-23
Add KRB5_TRACE variable for trace logging output

adrianfusco:
Adding KRB5_TRACE to have more information about the authentication. [OSPCIX-936](https://issues.redhat.com//browse/OSPCIX-936)

### DIFF
--- a/roles/dlrn_report/tasks/dlrn_report_results.yml
+++ b/roles/dlrn_report/tasks/dlrn_report_results.yml
@@ -20,6 +20,8 @@
       kinit
       {{ cifmw_dlrn_report_krb_user_realm }}
       -k -t {{ cifmw_dlrn_report_keytab }}
+  environment:
+    KRB5_TRACE: /dev/stdout
   retries: 5
   delay: 60
   register: _kinit_status


sdatko (changes requested):
I like the idea of using environment, but why you are touching the cmd then? Please keep it as it was

roles/dlrn_report/tasks/dlrn_report_results.yml
```diff
@@ -16,10 +16,11 @@
 
 - name: Perform kinit for DLRN kerberos authentication
   ansible.builtin.command:
-    cmd: >-
-      kinit
-      {{ cifmw_dlrn_report_krb_user_realm }}
-      -k -t {{ cifmw_dlrn_report_keytab }}
+    cmd: >
+      kinit {{ cifmw_dlrn_report_krb_user_realm }}
+            -k -t {{ cifmw_dlrn_report_keytab }}
```
sdatko, line 21: I like the idea of using environment, but why you are touching the cmd then? Please keep it as it was

evallesp (approved):

PR #3158, opened 2025-07-24
Call tasks from roles if play host is same

danpawlik:
After we migrate some playbooks into roles, we can drop the convention how it was done earlier - few plays where each was done on same host and each was calling one (or few) task in role. It is not needed, we can call other tasks in role in one play, because all is done on same host.

### DIFF
--- a/deploy-edpm.yml
+++ b/deploy-edpm.yml
@@ -43,7 +43,7 @@
       tags:
         - infra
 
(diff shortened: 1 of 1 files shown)


amartyasinha (approved):

evallesp (approved):
LGTM!

PR #3162, opened 2025-07-25
Set bool value instead of string bool value

danpawlik:
There is no need to write bool values as string. They should be written directly as bool: true, false.

Example command how it was done:

  find ci docs hooks playbooks roles scenarios -type f -exec sed -i "s@('true')@(true)@g" {} \;
  find ci docs hooks playbooks roles scenarios -type f -exec sed -i "s@('false')@(false)@g" {} \;


amartyasinha (approved):

evallesp (approved):
LGTM!

PR #3163, opened 2025-07-25
Change reproducer ANSIBLE_LOG_PATH

evallesp:
With this we'll able to retrieve that log file

### DIFF
--- a/roles/reproducer/tasks/configure_architecture.yml
+++ b/roles/reproducer/tasks/configure_architecture.yml
@@ -6,7 +6,7 @@
       vars:
         run_directory: "{{ _cifmw_reproducer_framework_location }}"
         exports:
-          ANSIBLE_LOG_PATH: "~/ansible-deploy-architecture.log"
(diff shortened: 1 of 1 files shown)


roles/reproducer/tasks/configure_architecture.yml
```diff
@@ -6,7 +6,7 @@
       vars:
         run_directory: "{{ _cifmw_reproducer_framework_location }}"
         exports:
-          ANSIBLE_LOG_PATH: "~/ansible-deploy-architecture.log"
+          ANSIBLE_LOG_PATH: "~/ci-framework-data/logs/ansible-deploy-architecture.log"
```
danpawlik, line 9: there are no vars poiting to `ci-framework-data` or logs?

evallesp: No that I've found.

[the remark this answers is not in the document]
evallesp: I've found that {{ ansible_user_dir }} -> /home/zuul, so it should be:
{{ ansible_user_dir }}/ci-framework-data/logs/

PR #3156, opened 2025-07-24
libvirt multiple interfaces to the same network

eshulman2:
Allow multiple interfaces to be attached from the same network.

Jira: [OSPRH-18516](https://issues.redhat.com//browse/OSPRH-18516)

PR #3056, opened 2025-06-09
Make worker-Nodes take preference in nncp ci_kustomize

evallesp:
As for example in 3+3 scenarios, we want to run all openstack namespaces pods
in the worker nodes.

As we have different type of nodes: ocp, crc and ocp-worker; we want to
select the nodes by taking precedende to worker nodes if defined.
If not, we want to check if it's a SNO crc deployment, and lastly, we'll
go by typical scenarios where the control-plane, master and workers in the same
nodes.

PR #3170, opened 2025-07-30
[OSPRH-18411] Remove default bridged control plane network for HCI VA

abays:
Switches the HCI VA control plane network to use a single dedicated NIC instead of a bridged one.

Depends-on: https://github.com/openstack-k8s-operators/ci-framework/pull/3156
Depends-on: https://github.com/openstack-k8s-operators/architecture/pull/593

Jira: https://issues.redhat.com//browse/OSPRH-18411


abays:
recheck

abays:
> Would you please remove the two depends-on on in this PR?
> 
>     * architecture/pull/593 should depend on this patch (not the other way around)
> 
>     * [libvirt multiple interfaces to the same network #3156](https://github.com/openstack-k8s-operators/ci-framework/pull/3156) has merged

Doesn't this PR technically depend on https://github.com/openstack-k8s-operators/architecture/pull/593?  If this patch merged without that architecture PR having merged, then CIFMW would be broken for the `uni03gamma` job because it would be using [1] instead of still using [2] (so it would try to use `enp6s0` for the control plane network when it should still be using `ospbr` since https://github.com/openstack-k8s-operators/architecture/pull/593 didn't merge yet).

Maybe the PRs are co-dependent?

[1] https://github.com/abays/ci-framework/blob/4f32850efa1a202e5024b7603139b189d9c6d8f0/roles/ci_gen_kustomize_values/templates/hci/network-values/values.yaml.j2#L84-L86
[2] https://github.com/openstack-k8s-operators/ci-framework/blob/412a7b648027f8fda98cd8b332af57bbbf489708/roles/ci_gen_kustomize_values/templates/common/network-values/values.yaml.j2#L84-L85

abays:
Testing with uni03gamma (which uses HCI VA) completed successfully

fultonj:
> > Would you please remove the two depends-on on in this PR?
> > ```
> > * architecture/pull/593 should depend on this patch (not the other way around)
> > 
> > * [libvirt multiple interfaces to the same network #3156](https://github.com/openstack-k8s-operators/ci-framework/pull/3156) has merged
> > ```
> 
> Doesn't this PR technically depend on [openstack-k8s-operators/architecture#593](https://github.com/openstack-k8s-operators/architecture/pull/593)? If this patch merged without that architecture PR having merged, then CIFMW would be broken for the `uni03gamma` job because it would be using [1] instead of still using [2] (so it would try to use `enp6s0` for the control plane network when it should still be using `ospbr` since [openstack-k8s-operators/architecture#593](https://github.com/openstack-k8s-operators/architecture/pull/593) didn't merge yet).

I was thinking that if architecture/pull/593 merged without this patch that it wouldn't have what it needed. I guess it can work both ways.

> Maybe the PRs are co-dependent?
> 
> [1] https://github.com/abays/ci-framework/blob/4f32850efa1a202e5024b7603139b189d9c6d8f0/roles/ci_gen_kustomize_values/templates/hci/network-values/values.yaml.j2#L84-L86 [2]
> 
> https://github.com/openstack-k8s-operators/ci-framework/blob/412a7b648027f8fda98cd8b332af57bbbf489708/roles/ci_gen_kustomize_values/templates/common/network-values/values.yaml.j2#L84-L85

roles/ci_gen_kustomize_values/templates/hci/network-values/values.yaml.j2
```diff
@@ -0,0 +1,126 @@
+---
+# source: hci/network-values/values.yaml.j2
+{% set _ipv = cifmw_ci_gen_kustomize_values_ip_version_var_mapping %}
+{% set ns = namespace(interfaces={},
+                      ocp_index=0,
+                      lb_tools={}) %}
+data:
+{% for host in cifmw_networking_env_definition.instances.keys() -%}
+{%   if host is match('^(ocp|crc).*') %}
```
evallesp, line 9: In order to take worker nodes take preferences in master+worker nodes in 3+3 scenarios, we have an ongoing patch that do this briefly different: https://github.com/openstack-k8s-operators/ci-framework/pull/3056

I'm unsure if we want to update this after the other is (eventually) merge, or if we want directly go the same approach here.

evallesp (commented):
LGTM in general, a question about how to proceed.

evallesp (approved):

PR #3171, opened 2025-07-31
Revert "Trigger cifmw-pod-zuul-files on each change"

amartyasinha:
Reverts openstack-k8s-operators/ci-framework#3152

This patch is breaking other buildsets. Reverting it to unblock other buildsets.
Ref: https://softwarefactory-project.io/zuul/t/rdoproject.org/buildset/36068ef9763d4be0830a7348feaad7c7

### DIFF
--- a/zuul.d/project-templates.yaml
+++ b/zuul.d/project-templates.yaml
@@ -12,7 +12,6 @@
             dependencies:
               - openstack-k8s-operators-content-provider
         - cifmw-crc-podified-edpm-baremetal: *content_provider
-        - cifmw-pod-zuul-files
 
 - project-template:
     name: podified-multinode-edpm-pipeline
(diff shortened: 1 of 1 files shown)


bshewale (approved):

tosky:
> > there was a reason why @dpawlik enabled those jobs for verifying changes
> 
> Every change always has a reason, isn't it? Priority should be to unblock by reverting unless we can fix it quickly. How was #3152 tested before merging? AFAICT, it does not talk talk about running it in every repo.

But the fix was found and proposed, so priority was given to unblock.
Let's go with https://github.com/openstack-k8s-operators/ci-framework/pull/3173 ?

rabi:
> > > there was a reason why @dpawlik enabled those jobs for verifying changes
> > 
> > 
> > Every change always has a reason, isn't it? Priority should be to unblock by reverting unless we can fix it quickly. How was #3152 tested before merging? AFAICT, it does not talk talk about running it in every repo.
> 
> But the fix was found and proposed, so priority was given to unblock. Let's go with #3173 ?

Sure, if we can confirm that would fix the issue without rechecking PRs in other repos.

sdatko:
It did work properly in edpm-ansible – https://softwarefactory-project.io/zuul/t/rdoproject.org/build/73b2021521634d61b2e463314c612b33

Closing this merge request then :+1:

PR #3174, opened 2025-07-31
Fix cinder multiattach volume hook

maximsava12:


tosky:
If you are testing https://github.com/openstack-k8s-operators/ci-framework/pull/3169, you don't need this. 
You can just either:
- depend on the main branch of ci-framework instead 
- depend also on https://github.com/openstack-k8s-operators/ci-framework/pull/3167

tosky:
I would say we need this

sdatko (approved):
I missed it... Thanks!

evallesp (approved):

PR #3175, opened 2025-07-31
Fix a few more oc rsh openstacklient calls (add namespace)

tosky:
It seems that previous switch to the openstack project is not effective (maybe it takes more time than expected?) so let's specify the namespace explicitly in a few more places.


evallesp (approved):

tosky:
Probably not needed, it turned out that it was a deployment issue. Due to an hook, a certificate was added to the pods after the deployment, which caused the pods to restart. The subsequent hooks then didn't find the openstackclient pod as it was being removed and restarted.

PR #3178, opened 2025-08-06
Use delegate_to localhost when getting secret

fultonj:
Ensure that the secret file is read from the system that is running the playbook by using `delegate_to: localhost`.

This is a follow to PR 3116, which helps to run reproducer playbook locally and not on remote host directly. It is possible that the target host hypervisor may be different from the host that is running the playbook. Whatever host that is, we can access the secret file via delegation. If it is the hypervisor, the original case for PR 3116 is still covered since localhost will still refer to that host.

Jira: https://issues.redhat.com/browse/OSPRH-18840


fultonj:
Sorry I sent this PR in prematurely. I'm still testing. I'll remove the DNM label when it's ready.

fultonj:
root cause:
TASK `name: Check if pull secret src file exists` called `stat` which got `permission denied` so it failed and stopped the playbook.

fix:
Call stat with `ignore_errors: true` and when referencing the register variable `_ps_exists` use `| default(false) ` in case it had an error.

bshewale:
recheck

PR #3185, opened 2025-08-07
Update custom cert installation hook

sdatko:
According to Ansible module documentation [1] we should be fine to use `state=present` setting, as it will create resource if it does not exist and patch the existing one. After this change we should be able to effectively use this hook in early deployment stages and avoid having whole control plane being restarted when the patching occured for already deployed OpenStack cluster.

[1] https://docs.ansible.com/ansible/latest/collections/kubernetes/core/k8s_module.html#parameter-state


sdatko:
That was a bad idea (I would know if I knew OpenShift more; in test run it complains on a lot of values missing ~ seems like it actually tries to deploy whole control plane I guess, so not exactly as I intended)

PR #3187, opened 2025-08-11
Introduce the split update (alternative to monolithic)

jistr:
This commit refactors the update role to support two distinct update variants, controlled by the new `cifmw_update_variant` variable. The variants are:

- `monolithic`: (Default) Preserves the existing behavior by running the single `make openstack_update_run` target.

- `split`: Implements a new two-phase update by running the `make update_services` and `make update_system` targets.

To achieve this, the main task has been modified to dynamically include either `update_variant_monolithic.yml` or
`update_variant_split.yml` based on the selected variant.


jistr:
Thank you for the reviews, i believe we have quorum so i'll press the merge button.

PR #3188, opened 2025-08-11
Support podified-ci-testing-tcib repo in ebi role

sdatko:
This DRLN tag would be used just for building the containers, so the `podified-ci-testing` then will be only updated after we ensure we can properly build the containers. The reason for this is that we have a lot of jobs that expect to find container tag for `podified-ci-testing` already in repository, while the containers may never be pushed due to tcib issues. The intermediate tag would allow us to bypass this without redesigning all existing jobs.

### DIFF
--- a/roles/edpm_build_images/tasks/post.yaml
+++ b/roles/edpm_build_images/tasks/post.yaml
@@ -12,7 +12,8 @@
     - ironic-python-agent
 
 - name: Retag and push the images with podified-ci-testing tag
-  when: cifmw_repo_setup_promotion == "podified-ci-testing"
+  when: cifmw_repo_setup_promotion
+        in ("podified-ci-testing", "podified-ci-testing-tcib")
   block:
     - name: Retag the images with podified-ci-testing tag
       containers.podman.podman_tag:


roles/edpm_build_images/tasks/post.yaml
```diff
@@ -12,7 +12,8 @@
     - ironic-python-agent
 
 - name: Retag and push the images with podified-ci-testing tag
-  when: cifmw_repo_setup_promotion == "podified-ci-testing"
+  when: cifmw_repo_setup_promotion
```
danpawlik, line 15: ah, you decide to use "safe" way, ok :)
Normally I was expecting to use:

```
when: cifmw_repo_setup_promotion and cifmw_repo_setup_promotion.startswith('podified-ci-testing')
```

roles/edpm_build_images/tasks/post.yaml
```diff
@@ -12,7 +12,8 @@
...
+        in ("podified-ci-testing", "podified-ci-testing-tcib")
```
danpawlik, line 16: I don't know if tuple is ok, or it should not be a list...

danpawlik: ok, it seems jinja2 accept tuple

sdatko: It works in testproject, so Ansible seems fine with Tuples :wink:

roles/edpm_build_images/tasks/post.yaml
sdatko: That code is heavy-tied to what we have in repo-setup, so @evallesp's argument of being explicit convinced me here in the end

PR #3199, opened 2025-08-18
[cinder] Add Ceph client config for RBD mirroring

lkuchlan:
Ensure the primary cluster includes the required Ceph client configuration to access the secondary cluster. This is necessary for successful RBD mirroring operations.

### DIFF
--- a/hooks/playbooks/enable_rbd_mirror_replication.yml
+++ b/hooks/playbooks/enable_rbd_mirror_replication.yml
@@ -48,13 +48,14 @@
   hosts: ceph_replication_targets
   become: true
   vars:
-    # Host filesystem paths (what Ansible sees)
-    bootstrap_token_path_host: /tmp/bootstrap_token_site
-    token_tmp_path: /tmp/rbd_mirror_bootstrap_token
-    # Container filesystem paths (what cephadm container sees)
-    bootstrap_token_path_container: /rootfs/tmp/bootstrap_token_site
-    # Configurable pool name
+    # Pool configuration
     replication_pool: "{{ cifmw_replication_pool | default('volumes') }}"
+
+    # Derived paths using cifmw variables directly
+    primary_conf_path: "{{ cifmw_replication_ceph_conf_dir | default('/etc/ceph') }}/{{ cifmw_replication_primary_cluster }}.conf"
+    primary_keyring_path: "{{ cifmw_replication_ceph_conf_dir | default('/etc/ceph') }}/{{ cifmw_replication_primary_cluster }}.client.{{ cifmw_replication_client_name | default('openstack') }}.keyring"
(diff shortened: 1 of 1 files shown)


danpawlik (commented):
LGTM, but could you share link to job which is verifying it?

danpawlik (commented):
I think you should set this PR as draft, to avoid additional review. Thanks

fmount:
recheck

fmount (dismissed):
/lgtm

hooks/playbooks/enable_rbd_mirror_replication.yml
```diff
@@ -48,13 +48,17 @@
...
-    bootstrap_token_path_host: /tmp/bootstrap_token_site
-    token_tmp_path: /tmp/rbd_mirror_bootstrap_token
-    # Container filesystem paths (what cephadm container sees)
-    bootstrap_token_path_container: /rootfs/tmp/bootstrap_token_site
-    # Configurable pool name
+    # Pool and client configuration
     replication_pool: "{{ cifmw_replication_pool | default('volumes') }}"
+    client_name: "{{ cifmw_replication_client_name | default('openstack') }}"
+    primary_cluster_name: "{{ cifmw_replication_primary_cluster }}"
+    secondary_cluster_name: "{{ cifmw_replication_secondary_cluster }}"
```
fmount: @lkuchlan I'm wondering if we need to define `primary_cluster_name` and `secondary_cluster_name`. Can't be simply reusing `cifmw_replication_primary_cluster` and `cifmw_replication_secondary_cluster` where necessary? The same thing would apply to `client_name` since we rely on `cifmw_replication_client_name`.
We can avoid redefining these three variable by reusing existing `cifmw_` prefixed variables, while I agree that `conf` and `keyring` might be overridden.
Apart from this small change, I'm good with the patch.

lkuchlan: Thanks for the comment
Done

hooks/playbooks/enable_rbd_mirror_replication.yml
```diff
@@ -97,64 +98,121 @@
...
+      ansible.builtin.file:
+        path: "/tmp/{{ cifmw_replication_secondary_cluster }}.client.{{ cifmw_replication_client_name | default('openstack') }}.keyring"
+        state: absent
+      delegate_to: localhost
+      run_once: true
+
     - name: Clean up token file on remote hosts
       ansible.builtin.file:
-        path: "{{ bootstrap_token_path_host }}"
+        path: "/tmp/bootstrap_token_site"
```
danpawlik: nit: if something needs really to be removed after, better to use:
```
block:
(...)
always:
<remove files>

danpawlik (approved):
still don't get why so much variable, but I can sleep with that.
Assuming from the conversation that testproject was made for it and it pass (via DM messages).

evallesp (approved):
Approving, IMO those tmp locations should be defined somewhere. But LGTM.

fmount (approved):
/lgtm

PR #3203, opened 2025-08-19
Move BGP playbooks to ci-fmw

eduolivares:
These BGP playbooks were duplicated in both ci-fmw and ci-fmw-jobs (downstream project).
They had been recently updated in downstream only.

With this patch the playbooks are updated in upstream. They will be removed from the downstream repo too.

[OSPRH-19224](https://issues.redhat.com//browse/OSPRH-19224)


fultonj:
Looks reasonable to me. Want to rebase and take it out of draft?

PR #3204, opened 2025-08-20
Remove condition from run_logs tasks of cifmw_setup

cescgina:
After merging[1], jobs have stopped running must-gather to collect
service logs, e.g [2,3]. This seems to be caused by the change in [4].
Where before the 99-logs.yml playbook was run from an 'ansible-playbook'
invocation, without any inventory and thus the check for
'zuul_log_collection' was never triggered because the variable was never
set. Now, when importing the cifmw_setup role, the cifmw variables are
loaded and that variable which is in most jobs set to True prevents the
log collection, and most critically, must-gather collection.
This change removes the condition to preserve the previous behaviour.

[1] https://github.com/openstack-k8s-operators/ci-framework/pull/3032
[2] https://softwarefactory-project.io/zuul/t/rdoproject.org/build/63673310b4f94883a9aff2db719a2095
[3] https://softwarefactory-project.io/zuul/t/rdoproject.org/build/03e87102fbf04dc7b482ecdba7128085
[4] https://github.com/openstack-k8s-operators/ci-framework/commit/cc87f54332d129fc6af4e048870bfd293c1b7318#diff-691dae244b4fb4bfc1270fb5d3b708c6be6073b0e33d997db15850edfa3fd976L23

### DIFF
--- a/ci/playbooks/e2e-collect-logs.yml
+++ b/ci/playbooks/e2e-collect-logs.yml
@@ -21,11 +21,18 @@
       ansible.builtin.meta: end_host
 
 - name: Run log related tasks
-  hosts: "{{ cifmw_target_host | default('all') }}"
+  hosts: "{{ cifmw_zuul_target_host | default('all') }}"
   gather_facts: true
   vars_files:
     - ../../scenarios/centos-9/base.yml
   tasks:
+    - name: Filter out host if needed
+      when:
+        - cifmw_zuul_target_host is defined
+        - cifmw_zuul_target_host != 'all'
+        - inventory_hostname != cifmw_zuul_target_host
+      ansible.builtin.meta: end_host
+
     - name: Run logging
       ansible.builtin.import_role:
         name: cifmw_setup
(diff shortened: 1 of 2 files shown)


amartyasinha (approved):

danpawlik (dismissed):
Make sens. Thanks

evallesp (dismissed):

fmount:
thank you

cescgina:
I don't think this approach will work. There are calls to `include_vars` like in https://github.com/openstack-k8s-operators/ci-framework/pull/3032/files#diff-20b693089f9adf9a3c7c6053da094f591663657757d9c7c335f5eb85e79af43cR20 and other which will not work as intended now. They are executed in the controller (the zuul controller), but are passed paths of the `controller` cifmw vm. This was fine before because the playbook was called as a an `ansible-playbook` from the controller  vm, not the zuul-controller, but is now failing

cescgina:
Closing this in favor of https://github.com/openstack-k8s-operators/ci-framework/pull/3206

PR #3208, opened 2025-08-21
Run nested ansible command and call run_logs tasks to drop 99-logs play

danpawlik:
Earlier the 99-logs.yml playbook was executed as nested Ansible.
After we migrate that playbook to role, it can not be called directly,
because it contains: cimfw collection to be installed on Zuul executor
and include_vars can not read files, that are located on the controller,
so they needs to be synchronized with Zuul executor job dir.

### DIFF
--- a/ci/playbooks/e2e-collect-logs.yml
+++ b/ci/playbooks/e2e-collect-logs.yml
@@ -20,22 +20,28 @@
         - not cifmw_status.stat.exists
       ansible.builtin.meta: end_host
 
-    - name: Run log collection
-      ansible.builtin.command:
+- name: Run log collection when zuul_log_collection
+  hosts: "{{ cifmw_target_host | default(cifmw_zuul_target_host) | default('controller') }}"
+  gather_facts: true
+  tasks:
+    - name: Run run_logs tasks from cifmw_setup
+      ansible.builtin.command: >
+        ansible localhost
+        -m include_role
+        -a "name=cifmw_setup tasks_from=run_logs.yml"
+        -e "@scenarios/centos-9/base.yml"
+      args:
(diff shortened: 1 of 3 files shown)


cescgina:
this version is still missing the must-gather logs: https://logserver.rdoproject.org/dce/rdoproject.org/dce68e40e56e4814aae7623b7c2447ca/controller/ci-framework-data/logs/. Looking at the job output: https://softwarefactory-project.io/zuul/t/rdoproject.org/build/dce68e40e56e4814aae7623b7c2447ca/log/job-output.txt#7411 it looks like all the tasks from the `run_logs.yaml` (https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/cifmw_setup/tasks/run_logs.yml) are being skipped, I guess because of this conditional https://github.com/openstack-k8s-operators/ci-framework/blob/5703bdde587294121499365c3da7a82a063fdb90/roles/cifmw_setup/tasks/run_logs.yml#L2

danpawlik:
recheck

danpawlik:
exactly @cescgina . Earlier there was import_playbook which does not provide variables, because it is static. If someone change it to include playbook, would be aware that it does not work. In short, issue here is that the variable is defined and value is `true`, where earlier it was not defined so it pick default value which is `false`.

danpawlik:
confirmed with @cescgina  that https://github.com/openstack-k8s-operators/watcher-operator/pull/251 provides missing logs. Checking, if the 251 requires additional changes in "empty" watcher operator change https://github.com/openstack-k8s-operators/watcher-operator/pull/258

amartyasinha (approved):

PR #3209, opened 2025-08-21
Add generic must-gather fallback

evallesp:
Adding a generic must-gather fallback in case it fails fist, the second
one is a simpler one less likely to fail.

### DIFF
--- a/roles/os_must_gather/tasks/main.yml
+++ b/roles/os_must_gather/tasks/main.yml
@@ -99,27 +99,40 @@
           "{{ cifmw_os_must_gather_output_dir }}/logs/openstack-k8s-operators-openstack-must-gather"
 
   rescue:
-    - name: Create oc_inspect log directory
-      ansible.builtin.file:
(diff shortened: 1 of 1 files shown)


roles/os_must_gather/tasks/main.yml
```diff
@@ -99,27 +99,40 @@
...
-            )
-          ) | unique
-        }}
+        - name: Run fallback generic must-gather command
+          environment:
+            KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+            PATH: "{{ cifmw_path }}"
+          ansible.builtin.command:
+            cmd: oc adm must-gather --dest-dir /home/zuul/ci-framework-data/must-gather
+      rescue:
```
danpawlik, line 114: nit: maybe replace `rescue` with `always` ?

danpawlik (dismissed):

evallesp:
recheck

Valkyrie00 (approved):
lgtm

PR #3210, opened 2025-08-21
[nat64_appliance] Allow to specify a base image URL

eduolivares:
The nat64_appliance runs diskimage-builder to build the qcow image that
is used for the nat64 VM. The diskimage-builder by default downloads the
latest CentOS image from the provided CentOS version:
https://github.com/openstack/diskimage-builder/blob/75b99e1990a45d33b694740f6fa20bc691de9ef9/diskimage_builder/elements/centos/root.d/10-centos-cloud-image#L45

With this patch, instead of delegating the download of the CentOS image
to diskimage-builder, an image URL can be provided.
For example:
```
cifmw_nat64_appliance_image_url: https://cloud.centos.org/centos/9-stream/x86_64/images/CentOS-Stream-GenericCloud-9-20250818.1.x86_64.qcow2
```

### DIFF
--- a/roles/nat64_appliance/tasks/main.yml
+++ b/roles/nat64_appliance/tasks/main.yml
@@ -66,12 +66,33 @@
     dest: "{{ cifmw_nat64_appliance_workdir }}/edpm-image-builder"
     version: main
 
+- name: Download nat64 source image
+  when: cifmw_nat64_appliance_image_url | length > 0
+  block:
+    - name: Create nat64 appliance cache directory
+      ansible.builtin.file:
+        path: "{{ cifmw_nat64_appliance_workdir }}/cache"
+        state: directory
+        mode: "0755"
+
+    - name: Download the nat64 source image
+      ansible.builtin.get_url:
+        url: "{{ cifmw_nat64_appliance_image_url }}"
+        dest: "{{ cifmw_nat64_appliance_workdir }}/cache"
+        timeout: 20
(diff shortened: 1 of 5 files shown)


roles/nat64_appliance/tasks/main.yml
```diff
@@ -66,12 +66,31 @@
     dest: "{{ cifmw_nat64_appliance_workdir }}/edpm-image-builder"
     version: main
 
+- name: Find latest image name
+  when:
+    - cifmw_discovered_image_url is not defined
+    - not cifmw_edpm_build_images_dry_run
+  ansible.builtin.import_role:
+    name: discover_latest_image
```
hjensas, line 74: Does'nt this mean we would be using a RHEL image vs CentOS image - depending on what parameters are set up for `discover_latest_image`? Are we sure a RHEL image will work? 

I think for a RHEL image to to work we would have to update https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/nat64_appliance/files/nat64-appliance.yaml 

What issue are we trying to fix? 
There are bugs, and fixes proposed for diskimage-builder related to CentOS image changes - I am guessing, but if the reason for this change is due to those bugs we should instead focus on fixing that, and not change this role.

-  https://bugs.launchpad.net/ironic/+bug/2120974
- https://review.opendev.org/c/openstack/diskimage-builder/+/957983

PR #3218, opened 2025-08-23
Add roles for fixing Python encodings

sdatko:
On the freshly reprovisioned nodes in our CI, not very often, but still from time to time, we seem to lack the necessary encoding file to invoke Ansible modules properly. I suspect that all this happens due to some rogue cleaning script that removes too many files, such as the compiled Python codes (`*.pyc` files) and hence from time to time we hit this error (the distro images we use are being rebuild quite often).

By including the new role as first task in the CI jobs, we should never ever encounter this problem again.

### DIFF
--- a/roles/fix_python_encodings/tasks/main.yaml
+++ b/roles/fix_python_encodings/tasks/main.yaml
@@ -1,0 +1,71 @@
+---
+- name: Check if cp437 is available
+  ansible.builtin.raw: |-
+    python3 -c 'from encodings import cp437; print(cp437)'
+  register: _import_cp437
+  changed_when: false
+  ignore_errors: true
+
+- name: Fix missing cp437
+  when: _import_cp437 is not success
+  block:
+    - name: Install python3-libs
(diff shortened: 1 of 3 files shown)


roles/fix_python_encodings/tasks/main.yaml
```diff
@@ -0,0 +1,69 @@
...
+    python3 -c 'from encodings import cp437; print(cp437)'
+  register: _import_cp437
+  changed_when: false
+  ignore_errors: true
+
+- name: Fix missing cp437
+  when: _import_cp437 is not success
+  block:
+    - name: Install python3-libs
+      ansible.builtin.raw: |-
```
danpawlik, line 13: it really can not be command or shell? Almost sure that dnf module can handle that too

sdatko:
recheck

arxcruz (approved):

danpawlik (approved):

danpawlik:
Thats awesome. You rock @sdatko !

PR #3223, opened 2025-08-26
Fix LVMCluster until clause

evallesp:
We have faced some times a race error because of:
The error was: error while evaluating conditional (_cifmw_lvms_storage_cluster_lvmscluster_out.resources

| length == 1): ''dict object'' has no attribute ''resources''. ''dict object''

With this we should fix this and make the task more robust


abays (approved):
/lgtm
/approve

amartyasinha (approved):

PR #3224, opened 2025-08-26
Cleanup scripts and old ansible facts

eshulman2:
Cleanup scripts and old ansible facts leftovers from artifacts directory


roles/cleanup_openstack/tasks/main.yaml
```diff
@@ -99,6 +99,23 @@
...
+- name: Remove artifacts
+  ansible.builtin.file:
+    path: "{{ item }}"
+    state: absent
+  loop: "{{
+        artifacts_to_remove.files |
+        map(attribute='path') |
+        list
+    }}"
+  become: true
```
danpawlik, line 117: don't think become is needed here, but ok. Could you move it to L109?

amartyasinha (approved):

PR #3230, opened 2025-08-28
Add tobikoPatch parameter for custom patches

belolipa:
This pull request introduces a new parameter, tobikoPatch, to the Tobiko Custom Resource to allow for testing against specific. [OSPRH-11286]


roles/test_operator/defaults/main.yml
```diff
@@ -179,6 +179,7 @@ cifmw_test_operator_tobiko_image: "{{ stage_vars_dict.cifmw_test_operator_tobiko
 cifmw_test_operator_tobiko_image_tag: "{{ cifmw_test_operator_default_image_tag }}"
 cifmw_test_operator_tobiko_testenv: scenario
 cifmw_test_operator_tobiko_version: master
+cifmw_test_operator_tobiko_patch: null
```
kstrenkova, line 182: I don't think you should set the default to `null`. Looking at the code briefly I am not sure how tobiko values work with the `nulls`, but I think there is additional process that resolves them. You can try looking into it.

roles/test_operator/defaults/main.yml
kstrenkova, line 1: You will need to set value of the new patch parameter in tobiko_config parameter [1]. I think it will work like `resources` parameter [2], we want to omit it if it's not defined. If we use it, I think the translated config would look like this:
```
cifmw_test_operator_tobiko_patch:
  repository: <name>
  refspec: <change-num>
```

[1] https://github.com/openstack-k8s-operators/ci-framework/blob/82ae564d1fe504fd157c085cf9c40d0513f12231/roles/test_operator/defaults/main.yml#L195

[2] https://github.com/openstack-k8s-operators/ci-framework/blob/82ae564d1fe504fd157c085cf9c40d0513f12231/roles/test_operator/defaults/main.yml#L215C5-L215C92

roles/test_operator/README.md
```diff
@@ -135,6 +135,7 @@ Default value: {}
 * `cifmw_test_operator_tobiko_image_tag`: (String) Tag for the `cifmw_test_operator_tobiko_image`. Default value: `{{ cifmw_test_operator_default_image_tag }}`
 * `cifmw_test_operator_tobiko_testenv`: (String) Executed tobiko testenv. See tobiko `tox.ini` file for further details. Some allowed values: scenario, sanity, faults, neutron, octavia, py3, etc. Default value: `scenario`
 * `cifmw_test_operator_tobiko_version`: (String) Tobiko version to install. It could refer to a branch (master, osp-16.2), a tag (0.6.x, 0.7.x) or an sha-1. Default value: `master`
+* `cifmw_test_operator_tobiko_patch`: (String) A specific Git patch to apply to the Tobiko repository. This feature expects both `repository` and `refspec` to be defined.
```
kstrenkova, line 138: The patch parameter is not a string :sweat_smile: It's a dictionary. You should also add the default value which should be `{}`.

kstrenkova (changes requested):

roles/test_operator/README.md
belolipa: Sure. Also i added default value

kstrenkova (changes requested):
I have tested the change locally and the variable itself is working as expected :+1: My only concern is in the comment I left, I hope someone will take time to review it.

PR #3231, opened 2025-08-28
Drop pre-run and post-run for kuttl_multinode

danpawlik:
The playbook execution is duplicated and it raises complication in some cases.

### DIFF
--- a/zuul.d/kuttl_multinode.yaml
+++ b/zuul.d/kuttl_multinode.yaml
@@ -41,13 +41,9 @@
                 ip: 172.18.0.5
               tenant:
                 ip: 172.19.0.5
-    pre-run:
-      - ci/playbooks/e2e-prepare.yml
     run:
       - ci/playbooks/dump_zuul_data.yml
       - ci/playbooks/kuttl/run.yml
-    post-run:
-      - ci/playbooks/collect-logs.yml
     required-projects:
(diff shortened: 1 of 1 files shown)


cescgina (approved):
+1, these are duplicated from the parent job

dasm (approved):
Yes, the execution is duplicated:
[src](https://github.com/openstack-k8s-operators/ci-framework/blob/82ae564d1fe504fd157c085cf9c40d0513f12231/zuul.d/base.yaml#L143)

[Zuul docs](https://zuul-ci.org/docs/zuul/latest/config/job.html#job) describing the behavior:

> [...]
> In addition to the job’s main playbook, each job may specify one or more pre- and post-playbooks. These are run, in order, before and after (respectively) the main playbook. They may be used to set up and tear down resources needed by the main playbook. When combined with inheritance, they provide powerful tools for job construction. A job only has a single main playbook, and when inheriting from a parent, the child’s main playbook overrides (or replaces) the parent’s. However, the pre- and post-playbooks are appended and prepended in a nesting fashion. So if a parent job and child job both specified pre and post playbooks, the sequence of playbooks run would be:
> 
> * parent pre-run playbook
> * child pre-run playbook
> * child playbook
> * child post-run playbook
> * parent post-run playbook

PR #3235, opened 2025-08-28
Add cifmw_helpers role

danpawlik:
The new role would provide tasks that would be helpful to drop nested Ansible execution in the project.


evallesp (commented):
LGTM, will check properly when moved away from Draft.

danpawlik:
recheck

PR #3193, opened 2025-08-14
Drop unused playbooks except 08-run-tests.yml

danpawlik:
The playbooks were migrated into role cifmw_setup.
As it was mentioned in commit message [1], such approach
should be better for developers to maintain the code.
Also skip removing 08-run-tests.yml because the NFV team is using it
and it takes a time to verify that. The playbook would be removed in
next pull request.

[1] https://github.com/openstack-k8s-operators/ci-framework/pull/2930

Depends-On: https://review.rdoproject.org/r/c/rdo-jobs/+/58147
Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/3208

PR #3237, opened 2025-08-29
WIP: Move common variable to group_vars

amartyasinha:
There is no need to parse path, or call filters etc. to get the same value.


amartyasinha:
> @amartyasinha about `adoption-standalone-to-crc-ceph-provider` I think it might be better when #3193 would be working... OR https://review.rdoproject.org/r/c/rdo-jobs/+/58147

Should I add Depends-On and then recheck?

amartyasinha:
recheck

michburk:
recheck

amartyasinha:
recheck

amartyasinha:
recheck

amartyasinha:
recheck

amartyasinha:
recheck

amartyasinha:
recheck

amartyasinha:
recheck

PR #3240, opened 2025-09-01
Allow merging all stages result xml

eshulman2:
Polarion shows stages results as different test run which makes it
harder to manage. Allow merging all stages results if required

Solves: https://issues.redhat.com/browse/NFV-3317


tosky:
Just for the record, I think the real solution to this would be the implementation of "test phases", which would allow us to aggregate test results independently from how test operator handles the tests.
The idea would be to define test results as belonging to a certain test phase and aggregate based on that.

evallesp (approved):

PR #3246, opened 2025-09-03
Add logging to a file for os must gather task

danpawlik:
After we drop executing Ansible playbook and we move temporary to Ansible command directly, there is an issue that some tasks are not executed properly.
Add ANSIBLE_LOG_PATH for that task for now, to make clear view what is executed.


ci/playbooks/e2e-collect-logs.yml
amartyasinha, line 1: L28: Wouldn't it be better to use `ci_script` module to execute the nested ansible ad-hoc command so that we get live logs appended in the log file? AFAIK, command module provides log once the execution completes.

amartyasinha (approved):

cescgina (approved):

PR #3247, opened 2025-09-03
Dump openstack databases when must-gather is collected

fmount:
When we collect the `must-gather` info in CI, it would be useful to dump the `openstack` databases for troubleshooting purposes. In addition, this makes sure we properly test the `gather_db` code from `must-gather`.

This patch is required to test https://github.com/openstack-k8s-operators/openstack-must-gather/pull/102

Jira: https://issues.redhat.com/browse/OSPRH-19553

### DIFF
--- a/roles/os_must_gather/defaults/main.yml
+++ b/roles/os_must_gather/defaults/main.yml
@@ -35,3 +35,4 @@ cifmw_os_must_gather_namespaces:
   - metallb-system
   - crc-storage
 cifmw_os_must_gather_host_network: false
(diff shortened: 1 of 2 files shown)


fmount:
recheck

raukadah (approved):

fmount:
> LGTM Just a suggestion: Vars should be used to enable and disable desired features at runtime. `OPENSTACK_DATABASES: "{{ os_must_gather_dump_db }}"`

Done, updated the patch to consume it from an ansible variable, which is set to `ALL` by default.

raukadah:
recheck

PR #3248, opened 2025-09-03
Add feature to exclude updating certain packages

danpawlik:
In some CI jobs, especially where some systems are based on RHEL and there are additional repositories provided, where packages belongs to CentOS, it can be a situation where the SSHd service would raise an issue:

    kex_exchange_identification: read: Connection reset by peer

Where on the host, sshd service raises a message:

    OpenSSL version mismatch. Built against 30000070, you have 30500010

With that way, we are able to add into the Zuul job vars and exclude the packages like 'openssl-libs*' (in that case) for being updated.

### DIFF
--- a/ci/playbooks/multinode-customizations.yml
+++ b/ci/playbooks/multinode-customizations.yml
@@ -1,4 +1,23 @@
 ---
+- name: Exclude updating certain packages
+  hosts: all
+  tasks:
+    - name: Check if yum conf exists
+      become: true
+      ansible.builtin.stat:
+        path: "/etc/yum.conf"
+      register: _yum_conf
+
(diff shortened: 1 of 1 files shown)


danpawlik:
recheck

danpawlik:
In another PR I will add packages to exclude like: `openssl-libs*`. If I will add this one, feature would disappear when revert would be done.
Let's split it into two PRs.

danpawlik:
The affected CI jobs seems not to be important anymore.
Also there is a miss match between host (RHEL) and deployment (setup CentOS repos). The second part seems to be a workaround for something that was not removed after a while.

PR #3255, opened 2025-09-04
Simplify tasks on crc host for getting container logs

danpawlik:
There is no need to use root as a user to get the logs from the CRC host to Zuul.


danpawlik:
recheck

danpawlik:
recheck

PR #3257, opened 2025-09-05
Add post_ceph var for deploying Ceph in edpm ceph scenario

danpawlik:
The variable was missing, so the post_ceph hook was not executed.


amartyasinha (approved):

danpawlik:
it seems the hook is breaking the `cifmw-multinode-tempest` job. If now is working, let's don't touch it.

PR #3259, opened 2025-09-05
Add missing vars for update role

danpawlik:
There are some missing variables, so the Ansible is raising an error:

    task path: /home/zuul/src/github.com/openstack-k8s-operators/ci-framework/roles/update/tasks/init_monitoring.yml:29
        changed: false
        msg: 'AnsibleUndefinedVariable: ''cifmw_path'' is undefined. ''cifmw_path'' is undefined'

### DIFF
--- a/roles/update/defaults/main.yml
+++ b/roles/update/defaults/main.yml
@@ -67,3 +67,6 @@ cifmw_update_ctl_plane_max_tries: 84
 
 # Resource Monitoring during update
 cifmw_update_resources_monitoring_interval: 10 # in seconds.
+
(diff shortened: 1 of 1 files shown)


ciecierski:
@danpawlik I found the same in my recent test-project. What has changed so that we need those variables now? They were not required before.

danpawlik:
https://github.com/openstack-k8s-operators/ci-framework/commit/cc87f5433#diff-0a442c193c3751a2ceb52b2d447afc739bc5f159a36ff8b5c4e06fb520684525R64 here is the problematic part.

danpawlik:
https://github.com/openstack-k8s-operators/ci-framework/pull/3260

PR #3262, opened 2025-09-05
update ceph var for ipv6

katarimanojk:
cifmw filters vars only with prefix cifmw- so
updating the var so that it can be propogated
duirng job execution.


hooks/playbooks/ceph.yml
```diff
@@ -163,7 +163,7 @@
     - name: Set IPv4 facts
       when:
         - ansible_all_ipv4_addresses | length > 0
-        - not ceph_ipv6 | default(false)
+        - not cifmw_ceph_ipv6 | default(false)
```
fmount, line 166: +1

PR #3276, opened 2025-09-10
Include read_global_vars.yml in pre-run: zuul.d/molecule-base.yaml

amartyasinha:
The reason to make the change one file per commit is our flakey jobs. It is pain to get all jobs pass at once


amartyasinha:
recheck

amartyasinha:
recheck

PR #3285, opened 2025-09-11
Add hook_retry parameter for run_hook role

danpawlik:
Sometimes, some hooks fails like on downloading tools. Let's add a new parameter: hook_retry that will make a retry on those hooks, where the parameter `hook_retry` is set to true.


danpawlik:
Might be helpful to use it in "download-tools" which sometimes just fails.

PR #3288, opened 2025-09-12
libvirt_manager: allow users to specify per-VM devices

tosky:
Add a new parameter 'devices' to the definition of virtual machine
types in cifmw_libvirt_manager_configuration.
The parameter is optional. If set, it contains a dictionary
where the keys are the VMs of that type that needs devices
to be attached, and the values are lists of strings,
where each string must contain a valid <hostdev/> libvirt XML
element that will be passed to `virsh attach-device`.

This is useful to expose for example passthrough devices
to the virtual machines.


tosky:
recheck

tosky:
recheck

PR #3299, opened 2025-09-16
Fix: search openstack-operator-index in all namespaces

mnietoji:
Currently it is looking for it in openstack-operator-index, but it is configured by default in openshift-marketplace. Task will look for it in any namespace

### DIFF
--- a/roles/env_op_images/tasks/main.yml
+++ b/roles/env_op_images/tasks/main.yml
@@ -77,6 +77,24 @@
       ansible.builtin.set_fact:
         cifmw_openstack_service_images_content: "{{ _sa_images_content.stdout | from_json }}"
 
+    - name: Get all pods from all namespaces to find openstack-operator-index
+      kubernetes.core.k8s_info:
+        kind: Pod
+        api_version: v1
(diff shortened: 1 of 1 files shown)


roles/env_op_images/tasks/main.yml
```diff
@@ -96,7 +96,8 @@
 
     - name: Retrieve openstack-operator-index pod
       vars:
-        selected_pod: "{{ pod_list.resources| selectattr('metadata.generateName', 'equalto', 'openstack-operator-index-') | list | first }}"
+        all_pods: "{{ query('kubernetes.core.k8s', api_version='v1', kind='Pod', kubeconfig=cifmw_openshift_kubeconfig, field_selectors=['status.phase=Running']) }}"
```
evallesp, line 99: I think we need here to put k8s_info based on the ci jobs output.

danpawlik:
Could you share where can I find result of that change in logs? Thanks

mnietoji:
recheck

evallesp (approved):

evallesp (approved):

danpawlik:
I did not take care about commit message. It was bad, especially that the user was not Jenkins.

PR #3301, opened 2025-09-17
Update Cinder volume backend naming to use sequential identifiers

lkuchlan:
Change Cinder volume backend configuration from using cluster names to sequential naming (ceph1, ceph2, etc.) to ensure unique backend names across all deployments.


fmount:
+1 from me in general. As discussed, we need this to make sure we can properly distinguish between `cinder volume service` associated to different Ceph clusters.
With this change we can properly configure replication.
Let's see the testproject execution and then review the patch further.

lkuchlan:
We initially suspected that the "permission denied" issue when Cinder
tried to access the Ceph cluster was related to backend naming.
After further investigation, it turned out the issue was unrelated
to backend naming and instead seems to be an issue in Cinder itself.
Therefore, this patch is not needed.

PR #3310, opened 2025-09-18
Remove scenarios/reproducers/dz-storage.yml

fultonj:
When the dz-storage scenario (uni10kappa) is run by ci-framwork using our downstream environment, it needs other variable files found in ci-framework-jobs. Thus, there is little value in keeping this scenario file upstream in ci-framework, which is a tooling repo. This file will be moved to ci-framework-jobs too.

This is not a revert of PR 3217 because we still need roles/ci_gen_kustomize_values/templates/dz-storage (see also eb940a1e78b5601eccaa450c9be91519d21ac6d7)


evallesp (approved):
lGTM!"

sdatko (approved):
/lgtm
/approve

PR #3323, opened 2025-09-23
Use gzip with best parameter to make logs size smaller

danpawlik:
With the best parameter, the logs should be even smaller than before.


evallesp (approved):

fmount (approved):
/lgtm

PR #3326, opened 2025-09-23
Fix openshift_adm role context when kubeconfig has multiple clusters

fultonj:
Add configurable context parameter to openshift_adm role to explicitly target the deployed OpenShift cluster instead of relying on current-context.

- Add cifmw_openshift_adm_context variable defaulting to 'admin'

- Update all kubernetes.core tasks to use the specified context:
  - _get_nodes.yml: k8s_info task
  - wait_for_cluster.yml: k8s_drain task
  - api_cert.yml: k8s and k8s_info tasks
  - shutdown.yml: k8s_drain task

- Add context switching for custom modules that don't support context:
  - approve_csr module: switch context before CSR approval
  - openshift_auth module: switch context before authentication

- Replace static cifmw_openshift_api parameter with dynamic API server URL detection:
  - Add _get_api_server.yml task to retrieve URL from current context
  - Update URI check and authentication to use context-based URL

- Fixes issue where tasks fail with 403 Forbidden when current-context points to CI cluster where user lacks permissions

Jira: https://issues.redhat.com/browse/OSPRH-20252
Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/roles/openshift_adm/tasks/wait_for_cluster.yml
+++ b/roles/openshift_adm/tasks/wait_for_cluster.yml
@@ -18,16 +18,19 @@
 # We would wait till forbidden error is received. It indicates the endpoint
 # is reachable.
 
+- name: Get API server URL from current context
+  ansible.builtin.include_tasks: _get_api_server.yml
+
 - name: Wait until the OCP API endpoint is reachable.
   ansible.builtin.uri:
-    url: "{{ cifmw_openshift_api }}"
+    url: "{{ _current_api_server }}"
     return_content: true
     validate_certs: false
     status_code: 403
   register: ocp_api_result
   until: ocp_api_result.status == 403
   retries: "{{ cifmw_openshift_adm_retry_count }}"
-  delay: 5
+  delay: 30
 
 - name: Get nodes list
   ansible.builtin.import_tasks: _get_nodes.yml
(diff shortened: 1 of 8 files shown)


evallesp (approved):
LGTM, but pre-commit failed.

roles/openshift_adm/tasks/wait_for_cluster.yml
```diff
@@ -49,15 +53,21 @@
...
+    - name: Set current context to admin for CSR approval
+      ansible.builtin.shell: |
+        KUBECONFIG="{{ cifmw_openshift_kubeconfig }}" oc config use-context "{{ cifmw_openshift_adm_context }}"
+
+    - name: Approve pending certificate requests
+      register: _approve_csr
+      approve_csr:
+        k8s_config: "{{ cifmw_openshift_kubeconfig }}"
+      retries: 30
+      delay: 10
```
sdatko, line 66: lets do reverse maybe: 10 retries with 30 seconds delay

fultonj:
This worked in my env for adoption but we might want to test project for greenfield before merging

roles/openshift_adm/tasks/wait_for_cluster.yml
```diff
@@ -39,25 +42,32 @@
...
-  retries: 30
-  delay: 10
-  until:
-    - _approve_csr is defined
-    - _approve_csr.rc is defined
-    - _approve_csr.rc == 0
+  block:
+    - name: Set current context to admin for CSR approval
+      ansible.builtin.shell: |
+        KUBECONFIG="{{ cifmw_openshift_kubeconfig }}" oc config use-context "{{ cifmw_openshift_adm_context }}"
```
danpawlik, line 59: just nit: I guess here you forget to use `environment`.

PR #3327, opened 2025-09-24
Sync variables of test-operator role from defaults/main.yml in README.md automatically

sauragar:
Testing my script to update the README.md file to sync variables of test-operator role from defaults/main.yml in README.md automatically via script

Ticket: OSPRH-19423


tosky:
I would say the commit message should not be "Testing my script to update the README.md file to sync variables of tesat-operator role from defaults/main.yml" but closer to the message of the merge request

roles/test_operator/defaults/main.yml
```diff
@@ -132,6 +134,7 @@ cifmw_tempest_tempestconf_config_defaults:
 
 # Please refer to https://openstack-k8s-operators.github.io/test-operator/guide.html#executing-tempest-tests
 cifmw_test_operator_tempest_debug: false
+cifmw_test_operator_tempest_temp_var: "{{ stage_vars_dict.cifmw_test_operator_tempest_name }}-{{ _stage_vars.name }}"
```
kstrenkova: Not sure if this variable is needed, but if it's test please remove it later as well.

scripts/sync_test_operator_var_readme.py
```diff
@@ -0,0 +1,131 @@
...
+
+    for key, val in defaults.items():
+        cat = categorize_var(key)
+        categorized[cat].append((key, val))
+
+    md_sections = []
+    for cat, items in categorized.items():
+        if not items:
+            continue
+        md_sections.append(f"### {cat} Variables\n")
```
kstrenkova: Based on my comment above I would change this part where the markdown sections are created.

scripts/sync_test_operator_var_readme.py
kstrenkova: Please make sure the `Generic` section comes first. I would also not add the header if we know it comes first -> it will look the same as the current README.

kstrenkova (changes requested):
Thank you for the PR! In my opinion the main issue is that we lose parameter descriptions. Please focus on that part, other comments are just suggestions on improving the code.

roles/test_operator/README.md
```diff
@@ -92,188 +74,118 @@ cifmw_test_operator_stages:
...
+* `cifmw_test_operator_crs_path`: (String) The path into which the tests CRs file will be created in
+* `cifmw_test_operator_tempest_rerun_failed_tests`: (Default: `False`) — Activate tempest re-run feature. When activated, tempest will perform another run of the tests that failed during the first execution
+* `cifmw_test_operator_tempest_rerun_override_status`: (Default: `False`) — Allow override of exit status with the tempest re-run feature. When activated, the original return value of the tempest run will be overridden with a result of the tempest run on the set of failed tests
+* `cifmw_test_operator_tempest_tempestconf_config`: (Default: `{{ cifmw_tempest_tempestconf_config }}`) — This parameter can be used to customize the execution of the `discover-tempest-config` run. Please consult the test-operator documentation. For example, to pass a custom configuration for `tempest.conf`, use the `overrides` section:
+>>>>>>> 0daf6d11 (Testing my script to update the README.md file to sync variables of tesat-operator role from defaults/main.yml)
 ```
 cifmw_test_operator_tempest_tempestconf_config:
   overrides: |
     identity.v3_endpoint_type public
```
kstrenkova: You are missing the ending of the code section (``` symbol). That is why all the text after it is all blue.

roles/test_operator/README copy.md
kstrenkova: I am not sure that this file is for.

sauragar: I am just testing stuff, so this is just duplicate

roles/test_operator/README.md
```diff
@@ -2,52 +2,31 @@
...
-        command: ["sleep"]
-        args: ["infinity"]
-        volumeMounts: "{{ _test_operator_volume_mounts }}"
-    volumes: "{{ _test_operator_volumes }}"
-    tolerations: "{{ cifmw_test_operator_tolerations | default(omit) }}"
-```
+* `cifmw_test_operator_controller_ip`: (String) An ip address of the controller node. Default value: `ansible_default_ipv4.address` which defaults to (`""`).
+* `cifmw_test_operator_controller_namespace`: (String) Namespace inside which the test-operator-controller-manager is created. Default value: `openstack-operators`
+* `cifmw_test_operator_controller_priv_key_file_path`: (String) Specifies the path to the CIFMW private key file. Note: Please ensure this file is available in the environment where the ci-framework test-operator role is executed. Default value: `~/.ssh/id_cifw`
+* `cifmw_test_operator_crs_path`: Default value: See defaults in [main.yml](defaults/main.yml)
```
kstrenkova: Why does this parameter have the reference to defaults/main.yml and not just the value?

roles/test_operator/README.md
```diff
@@ -2,52 +2,31 @@
...
 * `cifmw_test_operator_default_groups`: (List) List of groups in the include list to search for tests to be executed. Default value: `[ 'default' ]`
+* `cifmw_test_operator_default_image_tag`: (String) Default tag for all the supported test frameworks (tempest, tobiko, horizontest and ansibletest) to pull their container images. It can be overridden at test framework level.  Defaults to `current-podified`
 * `cifmw_test_operator_default_jobs`: (List) List of jobs in the exclude list to search for tests to be excluded. Default value: `[ 'default' ]`
+* `cifmw_test_operator_default_namespace`: (String) Default registry's namespace for all the supported test frameworks (tempest, tobiko, horizontest and ansibletest) to pull their container images. It can be overridden at test framework level.  Defaults to `podified-antelope-centos9`
```
kstrenkova: This parameter has `Defaults to` instead of `Default value:`. Does this script take the original text and not look at the default values? I would imagine it takes the description from the README, but the `Default value:` is generated for every parameter automatically.

PR #3328, opened 2025-09-24
[test_operator] Allow to read files from remote hosts

eduolivares:
Several tasks from the test_operator role only worked when the ansible
playbook was run on localhost, because they used lookup file to read the
file content.
With this change, the file content is obtained with slurp, which allows
to execute the playbook from a different host.

[OSPRH-20268](https://issues.redhat.com//browse/OSPRH-20268)


eduolivares:
> nit: if the file is a yaml, you would be able to use https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/cifmw_helpers/tasks/var_file.yml

Thanks for sharing. I was not aware of this and it looks really useful. But it would not work for this case.

sdatko (approved):
/lgtm
/approve

PR #3330, opened 2025-09-24
Add a new hook for storing Tempest timing data

kstrenkova:
This new hook is enabling timing data to be stored into an internal repository. Later the timing data will be used to reduce test execution time.


hooks/playbooks/tempest_store_timing_data.yaml
```diff
@@ -0,0 +1,46 @@
+---
+- name: Store stestr timing data in internal repository
+  hosts: localhost
+  gather_facts: false
+  vars:
+    repo_path: "/tmp/timing-data"
+    tests_log_dir: "{{ ansible_user_dir }}/ci-framework-data/tests/test_operator"
+    timing_data_repo: "https://user:{{ git_token }}@gitlab.cee.redhat.com/ci-framework/timing-data"
```
sdatko: let's keep `https://repo.example.org/project/timing-data` here and the full desired value pass just as variable when the hook is called

hooks/playbooks/tempest_store_timing_data.yaml
```diff
@@ -0,0 +1,46 @@
+---
+- name: Store stestr timing data in repository
+  hosts: localhost
+  gather_facts: false
+  vars:
```
danpawlik: nit: I know that in all places we have done same way, because we relay on ansible var order, but normally and good habit would be to remove "vars" and move it to vars defaults, e.g. "{{ timing_data_repo | default('https://repo.example.org/project/timing-data') }}"

danpawlik (approved):

PR #3333, opened 2025-09-25
Ensure molecule finds group_vars

amartyasinha:
Molecule can also use group_vars. For that, we have to add the path to group_vars dir in molecule config. This commit adds the path. It should be noted that relative path from config file wont work. Since molecule will expand the relative path during the molecule test run, the path to group_vars dir should be relative with molecule jobs from where they are run. More details will be in the PR.

Molecule is loading and using the group_vars. One of the examples of passing job is here [1]. In this job, molecule used the config file, and also read the group_vars. Before adding the group_vars, the molecule test was failing with undefined var. But the moment it was added, it passed. There is one failure at the playbook executed by zuul-executor, but that failure is independent of this specific change, and that will be fixed in another patch.

[1]: https://softwarefactory-project.io/zuul/t/rdoproject.org/build/64dd407880d845969d94659a502b6751/log/job-output.txt


amartyasinha:
recheck

amartyasinha:
recheck

amartyasinha:
recheck

amartyasinha:
recheck

amartyasinha:
recheck

amartyasinha:
recheck

amartyasinha:
recheck

amartyasinha:
recheck

amartyasinha:
recheck

amartyasinha:
recheck

amartyasinha:
recheck

Valkyrie00 (approved):

PR #3336, opened 2025-09-25
Remove deprecated cifmw_tempest_tempestconf_config

kstrenkova:
This parameter has been deprecated for some time, so we should use the new variant cifmw_test_operator_tempest_tempestconf_config. All mentions of this parameter have been removed both from downstream and upstream, so this patch is the last step to removing the parameter.


sdatko (approved):
/lgtm
/approve

adrianfusco (approved):

PR #3343, opened 2025-09-26
Revert "Include read_global_vars.yml in pre-run: zuul.d/edpm_build_images.yam…"

amartyasinha:
Reverts openstack-k8s-operators/ci-framework#3279

The reason to remove the playbook from pre-run stage from Zuul is, Zuul is not storing the cached vars that we added in pre-run. Maybe it is possible by some other way, but lets remove this for now.


danpawlik:
suggest instead of making 5 revert, do 1 revert with all content and write revert in commit message.

PR #3345, opened 2025-09-26
Make a symlink of os_must_gather dir to static name dir

danpawlik:
Some teams are using static directory to grab latest CI job results. Let's help them and create a symlink to latest dir.

### DIFF
--- a/roles/os_must_gather/tasks/main.yml
+++ b/roles/os_must_gather/tasks/main.yml
@@ -16,7 +16,7 @@
 
 - name: Ensure directories are present
   ansible.builtin.file:
-    path: "{{ cifmw_os_must_gather_output_dir }}/logs/openstack-k8s-operators-openstack-must-gather"
+    path: "{{ cifmw_os_must_gather_output_log_dir }}"
     state: directory
     mode: "0755"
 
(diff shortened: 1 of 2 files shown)


sdatko (dismissed):
Seems legit
/lgtm
/approve

roles/os_must_gather/tasks/main.yml
```diff
@@ -63,12 +63,26 @@
...
+    - name: Find gather current dir
+      ansible.builtin.find:
+        paths: "{{ cifmw_os_must_gather_output_log_dir }}"
+        file_type: directory
+        depth: 1
+      register: _os_gather_latest_dir
+
+    - name: Make a symlink of os must gather log dir
+      ansible.builtin.file:
+        src: "{{ cifmw_os_must_gather_output_log_dir }}/{{ (_os_gather_latest_dir.files | sort(attribute='mtime', reverse=True) | first).path | basename }}"
```
sdatko, line 83: I think will be just `{{ (_os_gather_latest_dir.files | sort(attribute='mtime', reverse=True) | first).path | basename }}` ~ we would have to use relative path to the created `latest` symlink ~ otherwise on logs server the preserver absolute path that you set will be incorrect

sdatko (approved):
/lgtm
/approve

fmount (approved):
/lgtm

PR #3311, opened 2025-09-18
Add cifmw-crc-edpm-podified-baremetal-bootc job

slagle:
Add cifmw-crc-edpm-podified-baremetal-bootc job

The job uses bootc as the base OS for the edpm baremetal compute node.
Adds a new podified-multinode-edpm-baremetal-bootc-pipeline which will
run the new job.

Depends-On: https://github.com/openstack-k8s-operators/install_yamls/pull/1091
Signed-off-by: James Slagle <jslagle@redhat.com>

PR #3346, opened 2025-09-29
Fix zuul check jobs

amartyasinha:
This [1] commit broke our zuul check jobs. This commit fix the issue. We need to work on a better solution so that it does not happen again. It has occured multiple times.

[1]: https://github.com/openstack-k8s-operators/ci-framework/pull/3311/files


amartyasinha:
recheck

amartyasinha:
recheck

brjackma (approved):

roles/adoption_osp_deploy/tasks/gather_stack_nodes.yml
```diff
@@ -14,6 +14,10 @@
       }}
 
 - name: Gather other facts for stack {{ _overcloud_name }}"
+  when:
+    - "'hostname_groups_map' in cifmw_adoption_osp_deploy_scenario"
+    - "'roles_groups_map' in cifmw_adoption_osp_deploy_scenario"
+    - _network_data_file is defined and _network_data_file
```
jistr, line 20: Will the job actually be able to run if we don't execute the set_fact below? Or will we just defer the failure to a later point?

danpawlik: checking now by removing condition

Valkyrie00 (dismissed):

PR #3241, opened 2025-09-01
feat: replace hardcoded 'zuul' user with configurable variables

Valkyrie00:
This PR replaces hardcoded 'zuul' user and group references throughout the CI-Framework codebase with configurable variables to improve flexibility and follow the project's convention of using parameterized variables.

 ### 1. **libvirt_manager Role**
  - **Files modified:** `roles/libvirt_manager/tasks/manage_vms.yml`, `roles/libvirt_manager/defaults/main.yml`
  - **Changes:** 
    - Replaced hardcoded 'zuul' with `{{ cifmw_libvirt_manager_user }}`
    - Added configurable variable: `cifmw_libvirt_manager_user: "{{ ansible_user | default(lookup('env', 'USER')) }}"`

  ### 2. **Infrastructure Playbook**
  - **File modified:** `create-infra.yml`
  - **Changes:** 
    - Replaced hardcoded 'zuul' with `{{ ansible_user_id }}`

  ### 3. **Reproducer Role**
  - **Files modified:** `roles/reproducer/tasks/configure_controller.yml`, `roles/reproducer/tasks/ci_job.yml`, `roles/reproducer/defaults/main.yml`
  - **Changes:**
    - Replaced hardcoded 'zuul' with `{{ cifmw_reproducer_user }}`
    - Added configurable variable: `cifmw_reproducer_user: "{{ ansible_user | default('zuul') }}"`

  ### 4. **Hook Playbooks**
  - **File modified:** `hooks/playbooks/delete_all_pre_adoption_resources.yaml`
  - **Changes:**
    - Replaced hardcoded 'zuul' with `{{ ansible_user | default('zuul') }}`
  - **Impact:** Adoption cleanup operations work with different user contexts

  ## Backward Compatibility
  - **Variable precedence:** Uses appropriate Ansible facts (`ansible_user_id`, `ansible_user`) 
  based on context
  - **Backward compatibility:** Maintains 'zuul' as fallback where appropriate

PR #3347, opened 2025-09-29
[DNM] fix(libvirt_manager): use ansible_user_id instead of ansible_user in libvirt_manager_user

Valkyrie00:
## Problem
  When deploying locally (not through testproject automation), the `id_cifw` SSH key was being
  created with `root:root` ownership instead of the expected user ownership. This caused
  permission denied errors when the user tried to connect to compute nodes.

  The issue occurred because with `ansible_connection=local`, `ansible_user` can become
  undefined, causing the fallback `lookup('env', 'USER')` to return `root` instead of the
  intended user.

  ## Solution
  Switch to `ansible_user_id` which correctly detects the user even with local connections and
  resolves the file ownership problem.

  ## Test executed
  - [x] Validated locally that `ansible_user_id` works correctly
  - [x] Confirmed workaround using `cifmw_libvirt_manager_user: zuul` resolves the issue
  
Issue Ref  #3241


amartyasinha (approved):

danpawlik (changes requested):
please update commit message.

karelyatin (approved):

PR #3350, opened 2025-09-29
Refactor hard-coded `zuul` user in reproducer role

michburk:
The reproducer role used a hardcoded `zuul` user and hardcoded `/home/zuul` path in several places, particularly when configuring controller-0. This patch pulls out this hardcoded data into default vars in the reproducer.
This patch also fixes a few incorrect assumptions that a var like `ansible_user_id` can be used when delegating to or configuring controller-0. For example, the `ansible_user_id` will not resolve to controller-0's user, and instead resolve to the host that the task is being delegated from. If there is a mismatch between these two users, this causes problems. This has mostly not been an issue because our ci assumes that there is a `zuul` user everywhere, but people running local deployments have encountered issues when using the framework on a hypervisor with a non-`zuul` user.

### DIFF
--- a/roles/reproducer/tasks/configure_controller.yml
+++ b/roles/reproducer/tasks/configure_controller.yml
@@ -357,11 +357,10 @@
         - cifmw_reproducer_src_dir_stat.stat.exists
         - cifmw_reproducer_src_dir_stat.stat.isdir
 
-    - name: Check if local common-requirements.txt exists
-      delegate_to: localhost
+    - name: Check if common-requirements.txt exists on controller-0
       ansible.builtin.stat:
-        path: "{{ cifmw_project_dir_absolute }}/common-requirements.txt"
-      register: _local_common_requirements_check
+        path: "{{ cifmw_reproducer_controller_user_dir }}/src/github.com/openstack-k8s-operators/ci-framework/common-requirements.txt"
+      register: _controller_common_requirements_check
       run_once: true
       ignore_errors: true
 
@@ -370,10 +369,10 @@
       async: 600  # 10 minutes should be more than enough
       poll: 0
       ansible.builtin.pip:
-        requirements: "{{ have_local | ternary(local, remote) }}"
+        requirements: "{{ have_controller_reqs | ternary(controller_reqs, remote) }}"
       vars:
-        have_local: "{{ _local_common_requirements_check.stat is defined and _local_common_requirements_check.stat.exists }}"
-        local: "{{ cifmw_project_dir_absolute }}/common-requirements.txt"
+        have_controller_reqs: "{{ _controller_common_requirements_check.stat is defined and _controller_common_requirements_check.stat.exists }}"
+        controller_reqs: "{{ cifmw_reproducer_controller_user_dir }}/src/github.com/openstack-k8s-operators/ci-framework/common-requirements.txt"
         remote: https://raw.githubusercontent.com/openstack-k8s-operators/ci-framework/main/common-requirements.txt
 
     - name: Inject most of the cifmw_ parameters passed to the reproducer run


roles/reproducer/vars/main.yml
```diff
@@ -2,11 +2,11 @@
 # Default repositories we always want to have
 cifmw_reproducer_default_repositories:
   - src: "https://github.com/openstack-k8s-operators/ci-framework"
-    dest: "{{ cifmw_project_dir_absolute }}"
```
michburk: using `cifmw_project_dir_absolute` here does not always work as these `dest` paths are for controller-0. For example the task `reproducer : Clone repository if needed: ci-framework` will fail if `localhost` and `controller-0` have different users, as this task will try to create `/home/<localhost_user_here>/src/github.com/openstack-k8s-operators/ci-framework` on controller-0 rather than `/home/<controller_user_here>/src/github.com/openstack-k8s-operators/ci-framework`

roles/reproducer/defaults/main.yml
```diff
@@ -17,7 +17,8 @@
 
 # All variables intended for modification should be placed in this file.
 # All variables within this role should have a prefix of "cifmw_reproducer"
-cifmw_reproducer_user: "{{ ansible_user | default('zuul') }}"
+cifmw_reproducer_controller0_user: "{{ hostvars['controller-0']['ansible_ssh_user'] | default('zuul') }}"
+cifmw_reproducer_controller0_user_dir: "/home/{{ cifmw_reproducer_controller0_user }}"
```
michburk: Enrique brought up the possibility of using something like `$HOME` rather than using a long var like `cifmw_reproducer_controller0_user_dir` - this would work in many cases, but some tasks will fail with this. For example:
```
TASK [sushy_emulator : Create and start Sushy Emulator container
...                                                                                         
task path: /home/testuser/ci-framework/roles/sushy_emulator/tasks/create_container.yml:26                                                                                    
fatal: [localhost -> controller-0(controller-0.localhost)]: FAILED! =>                                                                                                       
    changed: false                                                                                                                                                           
    msg: Container cifmw-sushy_emulator exited with code 125 when runed                                                                                                      
    stderr: |                                                                                                                                                                
      Error: creating named volume "$HOME/ci-framework-data/artifacts/sushy_emulator/.htpasswd": running volume create option: names must match [a-zA-Z0-9][a-zA-Z0-9_.-]*: invalid argument
    stderr_lines:
    - 'Error: creating named volume "$HOME/ci-framework-data/artifacts/sushy_emulator/.htpasswd":
      running volume create option: names must match [a-zA-Z0-9][a-zA-Z0-9_.-]*: invalid
      argument'
```
It would be possible to default to using `$HOME` where possible, but we have to keep track of the controller0 user anyways and I figured it'd be good to be clear and consistent about using `controller0_user_dir` - open to other arguments though!

roles/reproducer/tasks/main.yml
```diff
@@ -281,13 +281,15 @@
...
+            - "{{ cifmw_reproducer_controller0_user_dir }}/ansible-bootstrap.log"
 
         - name: Bootstrap environment on controller-0
           vars:
             # NOTE: need to overwrite parent vars:
             # ./roles/reproducer/molecule/crc_layout/converge.yml
-            cifmw_basedir: "{{ ansible_user_dir ~ '/ci-framework-data' }}"
+            cifmw_basedir: "{{ cifmw_reproducer_controller0_basedir }}"
+            ansible_user_dir: "{{ cifmw_reproducer_controller0_user_dir }}"
+            ansible_user_id: "{{ cifmw_reproducer_controller0_user }}"
```
michburk: Looking for feedback on my decision to overwrite these vars here - This task delegates to controller-0 and ends up calling several other roles that configure things assuming they can use vars like `ansible_user_dir` and `ansible_user_id` , but this causes issues when there are different users between localhost and controller-0.

roles/reproducer/README.md
```diff
@@ -6,8 +6,10 @@ None
 
 ## Parameters
 
-* `cifmw_reproducer_user`: (String) User used for reproducer role. Defaults to `zuul`
 * `cifmw_reproducer_basedir`: (String) Base directory. Defaults to `cifmw_basedir`, which defaults to `~/ci-framework-data`.
+* `cifmw_reproducer_controller0_user`: (String) User on controller-0. Defaults to the `ansible_user_id` associated with controller-0 if available, otherwise defaults to `zuul`.
```
evallesp: (blocking) question: Should this be "Defaults to the `ansible_ssh_user`" ? Instead of ansible_user_id

roles/reproducer/defaults/main.yml
evallesp: For this task as it's using the module: podman_container it doesn't expand env variables, so I think you need to use the lookup in this way: {{ lookup('env', 'HOME') }}

But I agree with you of consistenly using the same solution.

roles/reproducer/tasks/main.yml
evallesp: (non-blocking) reflection: I'm not opposite to this solution

evallesp (commented):
LGTM in general, not approving to hold on until more people chime in

roles/reproducer/defaults/main.yml
michburk: Minor correction: `{{ lookup('env', 'HOME') }}` would not be looking up `HOME` in controller-0's env because this task is delegated, so this would lookup HOME on the host that is being delegated _from_. Something like `{{ hostvars['controller-0']['ansible_env']['HOME'] }}` works though.

danpawlik:
I will pick that change in next iteration in tesstproject. 
So far, it looks nice!

Valkyrie00 (approved):

PR #3351, opened 2025-09-29
Add Barbican adoption support with Proteccio HSM integration

mauricioharley:
Implements comprehensive Barbican service adoption from OSP 17.1 to RHOSO 18 while preserving Proteccio Hardware Security Module (HSM) integration. This extends the existing ci-framework adoption infrastructure rather than creating separate components.

Implements: [OSPRH-18874](https://issues.redhat.com//browse/OSPRH-18874)


vakwetu (changes requested):
Unclear yet as to which of these changes will be needed.  The playbook will definitely be needed, but not sure about the rest.

PR #3352, opened 2025-09-30
Improve stability of ci_multus molecule test

amartyasinha:
ci_multus molecule job is quite flaky. Resources from openshift are not ready when we need to create our resource dependending on them. Adding retries will help the openstack core resources to get ready.

As of now, we get the following error. Retries will ensure the annotation is made available through openstack core resources.
```
fatal: [instance]: FAILED! => changed=false
  msg: 'Failed to create object: b''{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"pods \\"pod-testnad\\" is forbidden: error fetching namespace \\"openstack\\": unable to find annotation openshift.io/sa.scc.uid-range","reason":"Forbidden","details":{"name":"pod-testnad","kind":"pods"},"code":403}\n'''
  reason: Forbidden
```


amartyasinha:
recheck

danpawlik (approved):

PR #3356, opened 2025-09-30
Add building Tempest timing data url

kstrenkova:
There is a relatively new parameter added to test-operator, called TimingDataUrl. We are trying to make the parameter work with jobs, but the issue is that there is no variable that stores the final Tempest pod name that will be created. Without it, we cannot build the timing data url, as we are saving the data under job/test-name folder.

This change adds a way to build the final url for both tests with and without workflow.

### DIFF
--- a/roles/test_operator/tasks/tempest-tests.yml
+++ b/roles/test_operator/tasks/tempest-tests.yml
@@ -210,3 +210,71 @@
               stage_vars_dict |
               combine({'cifmw_test_operator_tempest_workflow': no_resources_workflow})
           }}
+
+- name: Build tempest timing data URL (no workflow)
+  when:
+    - not cifmw_test_operator_dry_run | bool
+    - stage_vars_dict.cifmw_test_operator_tempest_timing_data_url is defined
+    - stage_vars_dict.cifmw_test_operator_tempest_timing_data_url
+    - stage_vars_dict.cifmw_test_operator_tempest_workflow | length == 0
+  vars:
+    final_timing_data_url: >-
+      {{
+        stage_vars_dict.cifmw_test_operator_tempest_timing_data_url +
+        test_operator_instance_name + '/stestr.tar.gz'
+      }}
+  ansible.builtin.set_fact:
+    test_operator_cr: >-
+      {{
+          test_operator_cr |
+          combine({'spec': {'timingDataUrl': final_timing_data_url}}, recursive=true)
+      }}
+    stage_vars_dict: >-
+      {{
+          stage_vars_dict |
+          combine({'cifmw_test_operator_tempest_timing_data_url': final_timing_data_url})
+      }}
+
+- name: Build tempest timing data URL (workflow)
+  when:
+    - not cifmw_test_operator_dry_run | bool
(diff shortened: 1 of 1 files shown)


roles/test_operator/tasks/tempest-tests.yml
```diff
@@ -210,3 +210,66 @@
               stage_vars_dict |
               combine({'cifmw_test_operator_tempest_workflow': no_resources_workflow})
           }}
+
+- name: Build tempest timing data URL (no workflow)
+  when:
+    - not cifmw_test_operator_dry_run | bool
+    - stage_vars_dict.cifmw_test_operator_tempest_timing_data_url is defined
+    - stage_vars_dict.cifmw_test_operator_tempest_timing_data_url != ""
+    - stage_vars_dict.cifmw_test_operator_tempest_workflow | list | length == 0
```
danpawlik, line 219: L218 and L219 they exclude each other.
Try just with simply way:
```
  when:
    - not cifmw_test_operator_dry_run | bool
    - stage_vars_dict.cifmw_test_operator_tempest_timing_data_url is defined
```

roles/test_operator/tasks/tempest-tests.yml
```diff
@@ -210,3 +210,66 @@
...
+  vars:
+    _base_url: "{{ stage_vars_dict.cifmw_test_operator_tempest_timing_data_url }}"
```
danpawlik, line 221: is it required to create var just for 1 task that uses it?

roles/test_operator/tasks/tempest-tests.yml
```diff
@@ -210,3 +210,66 @@
...
+  when:
+    - not cifmw_test_operator_dry_run | bool
+    - stage_vars_dict.cifmw_test_operator_tempest_workflow | list | length > 0
+  vars:
+    _base_url: "{{ stage_vars_dict.cifmw_test_operator_tempest_timing_data_url }}"
+  block:
+    - name: Add the full timing data url to workflow steps
+      when: item.timingDataUrl is defined
+      vars:
+        _final_timing_data_url: "{{ _base_url }}{{ test_operator_instance_name }}-s0{{ step_number }}-{{ item.stepName }}/stestr.tar.gz"
```
danpawlik, line 251: why don't you keep same name convention? Earlier it was `final_timing_data_url`, now `_final_timing_data_url`. Any reason?

kstrenkova: I changed the name because one is used as a local variable in the loop. But if it's better to have the same name, I have no problem with that. 

Also, I did change how the `final_timing_data_url` is build, but your recommended method created a space in the url link, so I used `{{ ..+ .. }}` construction instead.

kstrenkova:
@danpawlik Thank you for the review! It was very helpful, I even noticed one mistake I did in the logic. If you have any doubts please add comments again. It can be even nitpicks, as I am not very well versed in ansible yet I would like to learn from my mistakes :D.

sdatko (approved):
/lgtm
/approve

danpawlik (approved):

rebtoor (approved):

PR #3357, opened 2025-09-30
Fix output of must-gather log

sdatko:
Right now because of &> the output of ci_script task
in collected logs is empty. After the change we will
have the output along other calls of ci_script instead
of relying on the dedicated file.


fmount (dismissed):
/lgtm

danpawlik (approved):

PR #3359, opened 2025-10-01
fix(devscripts): add molecule group in roles/devscript/check_cluster_status/molecule.yml

amartyasinha:
PR [1] ensured our molecule jobs can use group_vars. During this, adding molecule group to one file was missed. This commit adds that.

[1] https://github.com/openstack-k8s-operators/ci-framework/pull/3333


amartyasinha:
recheck

PR #3364, opened 2025-10-03
Add parse ansible argument string to variables

danpawlik:
In some playbook, when nested Ansible is executed via shell/command module, there is a string which contains arguments to parse by the ansible-playbook binary. If nested Ansible can be removed, it would be required to parse such variables.
This helper task can help to parse properly variables, that later can be read properly by tasks without using nested Ansible execution.

### DIFF
--- a/roles/cifmw_helpers/README.md
+++ b/roles/cifmw_helpers/README.md
@@ -209,3 +209,117 @@ execute tasks on new host.
         name: cifmw_helpers
         tasks_from: inventory_file.yml
 ```
+
+#### Parse string of arguments and convert to list of variables or list of files
+
+In some playbook, when nested Ansible is executed via shell/command module,
+there is a string which contains arguments to parse by the ansible-playbook
+binary. If nested Ansible can be removed, it would be required to parse
+such variables. Below example how nested Ansible execution looks like,
+and how it could be replaced.
+
+NOTE: `test.yaml` is executed on `host-1`.
+
+Example:
+- all files are on same host which execute ansible-playbook
+
+```yaml
+- name: Nested Ansible execution
+  hosts: localhost
+  tasks:
+    - name: Run ansible-playbook
+      vars:
+        cmd_args: "-e@somefile.yml -e @/tmp/someotherfile.yml -e myvar=test"
+      ansible.builtin.command: |
+        ansible-playbook "{{ cmd_args }}" test.yaml
+```
+
+To:
+
+```yaml
+- name: Playbook that does not use nested Ansible - same host
+  hosts: localhost
+  vars:
+    cifmw_cmd_args: "-e@somefile.yml -e @/tmp/someotherfile.yml -e myvar=test"
+  tasks:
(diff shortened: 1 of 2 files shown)


danpawlik:
Definitely one day I need to add molecule test job for all cases. 
Good that README is clear (hope so)

roles/cifmw_helpers/tasks/parse_ansible_args_string.yml
```diff
@@ -0,0 +1,34 @@
...
+# to:
+#
+#     cmd_args_vars: [{'myvar': 'test'}]
+#     cmd_args_files: ['somefile.yml', '/tmp/someotherfile.yml']
+#
+
+- name: Split string of arguments into the lists of vars and files
+  when: cmd_args | length > 1
+  ansible.builtin.set_fact:
+    cmd_args_vars: "{{ cmd_args
```
evallesp, line 18: (non-blocking) suggestion: What do you think about prefix the name with cifmw_helpers ? I think we're following that rule for role variables, but I'm unsure if we're doing that for facts.

roles/cifmw_helpers/tasks/parse_ansible_args_string.yml
```diff
@@ -0,0 +1,34 @@
...
+         | split(' -e ')
+         | reject('search', '@')
+         | reject('equalto', '')
+         | map('regex_replace', '^(.*?)=(.*)$', '{\"\\1\": \"\\2\"}')
+         | map('from_yaml')
+         | list
+      }}"
+    cmd_args_files: "{{ cmd_args
```
evallesp, line 26: (non-blocking) same as above suggetion.

roles/cifmw_helpers/README.md
```diff
@@ -209,3 +209,109 @@ execute tasks on new host.
...
+
+    - name: Parse only variables from cmd_args_vars
+      when: cmd_args_vars is defined and cmd_args_vars | length > 0
+      vars:
+        various_vars: "{{ cmd_args_vars }}"
+      ansible.builtin.include_role:
+        name: cifmw_helpers
+        tasks_from: various_vars.yml
+
+    - name: Fetch cmd_args_files to executing host
```
evallesp, line 304: Great example!

roles/cifmw_helpers/README.md
```diff
@@ -209,3 +209,109 @@ execute tasks on new host.
...
+    - name: Read var files from cmd_args
+      when: cmd_args_files is defined and cmd_args_files | length > 0
+      ansible.builtin.include_vars:
+        file: "{{ item }}"
+      loop: "{{ cmd_args_files }}"
+```
+
+- files are located in remote host - controller
+
+In alternative version, variables are available on remote host (include_vars
```
evallesp, line 268: (non-blokcing) suggestion: To make easier (Actually just faster) to understand, I'd add here that you need to fetch the files

evallesp (dismissed):
Some suggestion but all of them are non-blocking. Approved! I'm failling in love this role.

amartyasinha (dismissed):

amartyasinha (approved):

PR #3367, opened 2025-10-03
Revert change to make preference nncp on worker nodes thatn master nodes

evallesp:
It seems these scenarios has master nodes and a single worker node which is tainted. This worker node is used for running tempest but nothing related with OCP.

Problematic scenarios are those which has a worker node tainted for tempest, but the workload is set in master nodes.

There's no reason to configure NNCP in master nodes when we have regular worker nodes, so we were assuming that in case of there's a worker node, then all NNCP goes to worker node, which is problematic when single worker node is tainted.

So we need to enhance the templates to:
* Check if there's a worker node, if so, check if those are tainted.
* If not tainted, worker takes preference.
* If tainted, master takes preference.
* If there's no worker node, then we go with master nodes.
* Same for SNO CRC scenarios.

### DIFF
--- a/roles/ci_gen_kustomize_values/templates/uni02beta/edpm-nodeset-values/values.yaml.j2
+++ b/roles/ci_gen_kustomize_values/templates/uni02beta/edpm-nodeset-values/values.yaml.j2
@@ -1,0 +1,77 @@
+---
+# source: uni02beta/edpm-nodeset-values/values.yaml.j2
+{% set _ipv = cifmw_ci_gen_kustomize_values_ip_version_var_mapping %}
+{% set instances_names = []                                                            %}
+{% set _original_nodeset = (original_content.data | default({})).nodeset | default({}) %}
+{% set _original_nodes = _original_nodeset.nodes | default({})                         %}
+{% set _original_services = _original_nodeset['services'] | default([])                %}
+{% set _vm_type = (_original_nodes.keys() | first).split('-')[1]                       %}
+{% for _inst in cifmw_networking_env_definition.instances.keys()                       %}
+{%   if _inst.startswith(_vm_type)                                                     %}
+{%     set _ = instances_names.append(_inst)                                           %}
+{%   endif                                                                             %}
+{% endfor                                                                              %}
+data:
+  ssh_keys:
+    authorized: {{ cifmw_ci_gen_kustomize_values_ssh_authorizedkeys | b64encode }}
+    private: {{ cifmw_ci_gen_kustomize_values_ssh_private_key | b64encode }}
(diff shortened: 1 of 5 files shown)


roles/ci_gen_kustomize_values/templates/uni02beta/edpm-nodeset-values/values.yaml.j2
```diff
@@ -0,0 +1,77 @@
+---
+# source: uni02beta/edpm-nodeset-values/values.yaml.j2
+{% set _ipv = cifmw_ci_gen_kustomize_values_ip_version_var_mapping %}
+{% set instances_names = []                                                            %}
+{% set _original_nodeset = (original_content.data | default({})).nodeset | default({}) %}
+{% set _original_nodes = _original_nodeset.nodes | default({})                         %}
+{% set _original_services = _original_nodeset['services'] | default([])                %}
```
danpawlik, line 7: nit: if everywhere you use dots to get the "key" why here `['services']` not `.services`

roles/ci_gen_kustomize_values/templates/uni02beta/edpm-nodeset-values/values.yaml.j2
sdatko, line 1: Waiting for testproject result (I will let you know of the status in ~3 hours)

roles/ci_gen_kustomize_values/templates/uni02beta/edpm-values/values.yaml.j2
```diff
@@ -0,0 +1,71 @@
+---
+# source: uni02beta/edpm-values/values.yaml.j2
+{% set _ipv = cifmw_ci_gen_kustomize_values_ip_version_var_mapping %}
+{% set instances_names = []                                                            %}
+{% set _original_nodeset = (original_content.data | default({})).nodeset | default({}) %}
+{% set _original_nodes = _original_nodeset.nodes | default({})                         %}
+{% set _original_services = _original_nodeset['services'] | default([])                %}
+{% set _vm_type = (_original_nodes.keys() | first).split('-')[1]                       %}
+{% for _inst in cifmw_networking_env_definition.instances.keys()                       %}
```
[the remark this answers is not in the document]
evallesp: Resolving, but notes are taking in consideration for refactor task

roles/ci_gen_kustomize_values/templates/uni02beta/edpm-nodeset-values/values.yaml.j2
evallesp: Resolving, but notes are taking in consideration for refactor task

PR #3381, opened 2025-10-08
DNM Improve cifmw-molecule-ci_nmstate job

danpawlik:
The CI job is randomly failing without any clear log. Let's take also logs from pods, events etc.


danpawlik:
recheck

PR #3528, opened 2025-12-03
Add new download cache validation

drosenfe:
This validation verifies that user may specify rpms and container images for download and caching on a compute node. After a deployment is executed it is verified that the proper rpms and container images are present in the compute node cache.

### DIFF
--- a/roles/validations/tasks/edpm/download-cache-service.yml
+++ b/roles/validations/tasks/edpm/download-cache-service.yml
@@ -1,0 +1,100 @@
+- name: Determine name of deployed NodeSet
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  cifmw.general.ci_script:
+    output_dir: "{{ cifmw_validations_basedir }}/artifacts"
+    script: >-
+      oc get -n {{ cifmw_validations_namespace }} osdpns --no-headers -o custom-columns=":metadata.name"
+  register: deployed_nodeset_name
+
+- name: Clean cache on a compute node
+  become: true
+  ansible.builtin.command:
+    cmd: dnf clean all
+  delegate_to: "{{ cifmw_validations_edpm_check_node }}"
+
+- name: Uninstall requested packages on a compute node
+  become: true
+  ansible.builtin.dnf:
+    name: "{{ item_cached_package }}"
+    state: absent
+  loop: "{{ cifmw_validations_cached_packages }}"
+  loop_control:
+    loop_var: item_cached_package
+  delegate_to: "{{ cifmw_validations_edpm_check_node }}"
+
+# Packages are removed from cache when they are installed. Create a custom service
+# that checks for cached packages. This service can be used to verify that packages
+# have been properly cached by executing before a service that loads packages.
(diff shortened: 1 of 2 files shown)


evallesp (dismissed):

roles/validations/tasks/edpm/download-cache-service.yml
```diff
@@ -0,0 +1,123 @@
...
+    output_dir: "{{ cifmw_validations_basedir }}/artifacts"
+    script: >-
+      oc get -n {{ cifmw_validations_namespace }} osdpns --no-headers -o custom-columns=":metadata.name"
+  register: deployed_nodeset_name
+
+- name: Clean cache on a compute node
+  become: true
+  ansible.builtin.shell:
+    cmd: >-
+      set -o pipefail && dnf clean all
```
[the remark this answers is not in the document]
drosenfe: It was copied from a previous validator without thinking about it very much. I can take it out.

roles/validations/tasks/edpm/download-cache-service.yml
```diff
@@ -0,0 +1,120 @@
...
+  delegate_to: "{{ cifmw_validations_edpm_check_node }}"
+
+- name: Patch nodeset with download cache configuration
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  cifmw.general.ci_script:
+    output_dir: "{{ cifmw_validations_basedir }}/artifacts"
+    script: >-
+      oc patch -n {{ cifmw_validations_namespace }} osdpns/"{{ deployed_nodeset_name.stdout | trim }}" --type json --patch '[{ "op": "add", "path": "/spec/nodeTemplate/ansible/ansibleVars/edpm_download_cache_running_services","value": {{ cifmw_validations_cache_run_services }} }]'
```
slagle: This var shouldn't need to be set. It's set by the edpm_download_cache role from either edpm_services_override, or edpm_services, both of which of those are set automatically by openstack-operator.

drosenfe: Yes, this was here because of: https://issues.redhat.com/browse/OSPRH-21737   Now that we know a fix is coming will remove this and let the role populate it automatically.

roles/validations/tasks/edpm/download-cache-service.yml
```diff
@@ -0,0 +1,120 @@
...
+      apiVersion: dataplane.openstack.org/v1beta1
+      kind: OpenStackDataPlaneDeployment
+      metadata:
+        name: download-cache-service
+        namespace: {{ cifmw_validations_namespace }}
+      spec:
+        nodeSets:
+          - "{{ deployed_nodeset_name.stdout | trim }}"
+        servicesOverride:
+          - download-cache
```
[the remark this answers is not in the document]
drosenfe: bootstrap added. Decided in slack to remove frr.

roles/validations/tasks/edpm/download-cache-service.yml
drosenfe: patch has been removed in latest version.

slagle (approved):

michburk (approved):

PR #3575, opened 2025-12-23
Add OIDC authentication flows tests for federation

xek:
This adds support for testing all OIDC authentication methods:
- v3oidcpassword (Resource Owner Password Credentials)
- v3oidcclientcredentials (Client Credentials)
- v3oidcaccesstoken (Access Token Reuse)
- v3oidcauthcode (Authorization Code)

New features:
- Templates for each OIDC auth method configuration
- Helper script to obtain tokens from Keycloak
- Keycloak client setup for Service Accounts and Device Auth
- Comprehensive test suite with assertions
- Updated README with full documentation

Note: v3oidcdeviceauthz requires Python 3.10+ and is not
available in OSP18 which ships with Python 3.9.

Jira: [OSPRH-22826](https://issues.redhat.com//browse/OSPRH-22826)


roles/federation/tasks/hook_post_deploy.yml
```diff
@@ -78,3 +78,26 @@
     - "{{ cifmw_federation_keycloak_testuser1_username }}"
     - "{{ cifmw_federation_keycloak_testuser2_username }}"
   when: not cifmw_federation_deploy_multirealm|bool
+
```
jagee: What do you think about us eliminating the above section 69-80? since the run_openstack_oidc_auth_tests.yml is covering the "[v3oidcpassword] Get token for test user 1" task we are doing some duplication here and it might make more since to just have one round of this test in the new yaml. the only thing we are not doing is testing user2 but i dont think that is really important as they are both in the same project.

jagee (commented):
Everything looks good to me, just wanted see if you though we should remove one duplicate auth test from the older code.

roles/federation/tasks/hook_post_deploy.yml
jagee: the second test user is in there more for the horizon gui testing in the other dfgs testing.

xek: Thanks for the review, I removed it.

jagee (dismissed):
looks good to me

danpawlik (dismissed):

michburk (dismissed):
/lgtm

jagee:
The new jq changes in the shell scripts looks good. yes we can remove the run_openstack_auth_test.yml, it was a left over from when i originaly wrote the role.  lgtm

evallesp (approved):
/lgtm

michburk (approved):
/approve

PR #3641, opened 2026-01-29
Fix task name truncation in ci_script plugin: use rstrip instead of rsplit

ciecierski:
Using rsplit("_", 1)[0] dropped the last segment when there was no trailing underscore (e.g. for short ansible task names like "Deploy ceph" became "deploy").

Use rstrip("_") so we only remove a trailing underscore at the end of string.

Update the script integration test expectations: ci_script_* artifact and log
paths now use the full normalized task-name slug (e.g. ..._failing_script,
..._debug_enabled, ..._chdir_option). Refresh SHA1 checksums for the generated
.sh wrappers and for the chdir task log; other log checksums are unchanged.

Signed-off-by: Mikołaj Ciecierski <mikolaj.ciecierski@redhat.com>

### DIFF
--- a/plugins/action/ci_script.py
+++ b/plugins/action/ci_script.py
@@ -192,9 +192,7 @@ class ActionModule(ActionBase):
         # truncate the length of the name to 20 chars at max
         # while trying to preserve entire words
         t_name = (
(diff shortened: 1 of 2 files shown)


brjackma (approved):
/lgtm

PR #3738, opened 2026-03-04
[ci_gen_kustomize_values] Co-locate provisionserver with metal3 to prevent DHCP failures

mnietoji:
…ures

When metal3-dnsmasq pod restarts during a node's DHCP lease renewal on the provisioning network (172.23.0.0/24), NetworkManager fails to renew and sets ipv4.method=disabled. NMState operator then preserves this disabled state, causing permanent loss of provisioning network connectivity on that node.

The issue occurs when OpenStackProvisionServer and metal3 pods run on different nodes. If metal3 restarts while a node is attempting DHCP renewal, the temporary unavailability of metal3-dnsmasq causes the renewal to fail.

Solution:
Automatically detect the node running metal3 pod (via k8s-app=metal3 label) and configure provisionServerNodeSelector in baremetalSetTemplate to schedule OpenStackProvisionServer on the same node. This ensures provisioning network connectivity is maintained because metal3-static-ip-manager maintains a static IP (172.23.0.3) on the metal3 node regardless of dnsmasq restarts.

### DIFF
--- a/roles/ci_gen_kustomize_values/templates/nfv-ovs-dpdk-sriov-hci/edpm-nodeset-values/values.yaml.j2
+++ b/roles/ci_gen_kustomize_values/templates/nfv-ovs-dpdk-sriov-hci/edpm-nodeset-values/values.yaml.j2
@@ -44,3 +44,9 @@ data:
       - "{{ svc }}"
 {%   endfor                                                                            %}
 {% endif                                                                               %}
+{% if cifmw_kustomize_deploy_metal3_node is defined                                    %}
+  baremetalSetTemplate:
+{{ original_content.data.baremetalSetTemplate | default({}) | to_nice_yaml(indent=2) | indent(4, first=true) }}
+    provisionServerNodeSelector:
+      kubernetes.io/hostname: "{{ cifmw_kustomize_deploy_metal3_node }}"
+{% endif                                                                               %}
--- a/roles/ci_gen_kustomize_values/templates/sriov/edpm-nodeset-values/values.yaml.j2
+++ b/roles/ci_gen_kustomize_values/templates/sriov/edpm-nodeset-values/values.yaml.j2
@@ -42,3 +42,9 @@ data:
       - "{{ svc }}"
 {%   endfor                                                                            %}
 {% endif                                                                               %}
+{% if cifmw_kustomize_deploy_metal3_node is defined                                    %}
+  baremetalSetTemplate:
+{{ original_content.data.baremetalSetTemplate | default({}) | to_nice_yaml(indent=2) | indent(4, first=true) }}
+    provisionServerNodeSelector:
+      kubernetes.io/hostname: "{{ cifmw_kustomize_deploy_metal3_node }}"
+{% endif                                                                               %}
--- a/docs/dictionary/en-custom.txt
+++ b/docs/dictionary/en-custom.txt
@@ -183,6 +183,7 @@ dlrnapi
 dlrnrepo
 dm
 dnf
+DPDK
 dns
 dnsdata
 dnsmasq
@@ -283,6 +284,7 @@ ipmi
 ips
 iptables
 ipv
+IOV
 iscsi
 isdir
 itldwuw
@@ -391,6 +393,7 @@ networker
 networkmanager
 networktype
 nfs
+NFV
 nftables
 nhc
 nic
@@ -461,6 +464,7 @@ ovirt
 ovirtmgmt
 ovn
 ovnkubernetes
+OVS
 pablintino
 param
 params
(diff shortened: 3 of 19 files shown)


playbooks/06-deploy-architecture.yml
```diff
@@ -163,6 +163,43 @@
...
+        - always
+      when:
+        - cifmw_sync_pullsecret_credentials | default(true) | bool
+      block:
+        - name: Extract credentials for each configured registry
+          loop: "{{ cifmw_reproducer_registry_list | default(['registry.stage.redhat.io']) }}"
+          loop_control:
+            loop_var: registry
+          ansible.builtin.include_role:
+            name: edpm_pullsecret_sync
```
Valkyrie00: I can't find the reference for this role. Is it being introduced in a separate PR?

roles/ci_gen_kustomize_values/templates/sriov/edpm-nodeset-values/values.yaml.j2
```diff
@@ -42,3 +43,11 @@ data:
       - "{{ svc }}"
 {%   endfor                                                                            %}
 {% endif                                                                               %}
+{% if cifmw_kustomize_deploy_metal3_node is defined                                    %}
+  baremetalSetTemplate:
+{%   for key, value in _original_baremetal_template.items()                            %}
+    {{ key }}: {{ value }}
+{%   endfor                                                                            %}
```
michburk: Would it be possible to use something like `| to_nice_yaml` for this instead of manually deconstructing and reconstructing the yaml `key: value`s?

roles/kustomize_deploy/tasks/execute_step.yml
```diff
@@ -126,6 +126,29 @@
       ansible.builtin.include_role:
         name: run_hook
 
+    - name: Detect metal3 pod node for baremetal nodeset provisioning
```
michburk: Would it make more sense to put these tasks somewhere other than `execute_step.yml`? Would these tasks be better-suited to living in some dedicated hook, rather than being part of this generic `execute_step.yml` file?

michburk (commented):
Has this been tested? If so, could you link a jira ticket with any relevant downstream links hidden behind comments or descriptions that are marked as 'Red Hat Employee'?

Thanks!

roles/kustomize_deploy/tasks/execute_step.yml
mnietoji: yes, you are right. moved the code to a different place

roles/ci_gen_kustomize_values/templates/sriov/edpm-nodeset-values/values.yaml.j2
mnietoji: yes, improved the code

roles/ci_gen_kustomize_values/templates/nfv-ovs-dpdk-sriov-hci/edpm-nodeset-values/values.yaml.j2
```diff
@@ -43,3 +43,9 @@ data:
       - "{{ svc }}"
 {%   endfor                                                                            %}
 {% endif                                                                               %}
+{% if cifmw_kustomize_deploy_metal3_node is defined                                    %}
+  baremetalSetTemplate:
+{{ ((original_content.data | default({})).baremetalSetTemplate | default({})) | to_nice_yaml(indent=2) | indent(4, first=true) }}
```
michburk: [non-blocking] nit: instead of
```
original_content.data | default({})).baremetalSetTemplate | default({})
```
could you do
```
original_content.data.baremetalSetTemplate | default({})
```

roles/ci_gen_kustomize_values/tasks/main.yml
```diff
@@ -14,6 +14,28 @@
...
+          - k8s-app=metal3
+      register: _cifmw_kustomize_deploy_metal3_pod_info
+
+    - name: Set metal3 node for provisionserver nodeSelector
+      ansible.builtin.set_fact:
+        cifmw_kustomize_deploy_metal3_node: "{{ _cifmw_kustomize_deploy_metal3_pod_info.resources[0].spec.nodeName }}"
+        cacheable: true
+      when:
+        - _cifmw_kustomize_deploy_metal3_pod_info.resources is defined
+        - _cifmw_kustomize_deploy_metal3_pod_info.resources | length > 0
```
michburk: [non-blocking] nit: Would something like
```
_cifmw_kustomize_deploy_metal3_pod_info.resources | default([]) | length > 0
```
achieve the same thing while being a little cleaner?

roles/ci_gen_kustomize_values/tasks/main.yml
```diff
@@ -14,6 +14,28 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+- name: Detect metal3 node for baremetal nodeset provisioning
```
michburk, line 17: suggestion: I can't track down the version of this that I reviewed before, and I can't remember exactly where these tasks were in that version, but I would prefer if these were in their own separate tasks file that could be imported here. Sorry if we end up going in a circle on where exactly these tasks end up located, but I feel that this introduces tasks that are specific to your use case to the main tasks file of this role.

Note that I'm not saying I'm 100% correct on this point, if you (or any others) disagree, I would be receptive to disagreement and would appreciate further explanation.

michburk (commented):
Sorry for the delay!
A couple of style nits, I'm still uncertain on the location of the two tasks you're introducting.

roles/ci_gen_kustomize_values/tasks/main.yml
mnietoji: ok, moved to another file

michburk:
changes are looking good, just wanted to double check: is this tested and working as expected?

mnietoji:
> changes are looking good, just wanted to double check: is this tested and working as expected?

Yes, it is tested and working

michburk (approved):

Valkyrie00 (approved):

PR #3739, opened 2026-03-04
[multiple]: Support agent-based BM SNO deployment

bogdando:
- Add options to deploy Single-Node OpenShift from the Zuul
    controller instead of the classic hybrid (dev-scripts + libvirt) job
    layout: treat the node that runs deploy architecture as the Ansible
    controller; skip libvirt/vmnet and do not provision VMs.
  - Replace dev-scripts with agent-based openshift-installer so the same
    path can grow into isolated L2 between Zuul, SNO bare metal, and EDPM
    bare metal (EDPM BM to follow).
  - On the SNO host, optionally enable USB boot and discover or validate
    the UEFI Virtual Media boot target so installs boot the ISO we build;
    avoid silent boots from the wrong device when disks are not wiped.
  - Support live debug on the agent appliance (e.g. autologin / debug
    shell) for discovery-phase console access.
  - Password handling for agent appliance and the eventual cluster:
    - After ISO build: emit MachineConfig (e.g. 99-core-password.yaml)
      so MCO applies the core user password post-bootstrap.
    - Before/around ISO use: coreos-installer iso ignition show, run
      patch_ignition.py to add core passwordHash and getty@tty1 autologin,
      then coreos-installer iso ignition embed -f to put ignition back in
      the agent ISO.
  - When reusing OCP on bare metal SNO (cifmw_bm_sno), skip controller
    reboot/wait in deploy-edpm-reuse, skip repo sync to the controller in
    push_code, and set reuse facts so OpenShift is assumed without CRC/OCP
    VM detection.
  - In deploy_architecture, derive controller network facts from the play
    host if no controller-* exists; only shrink OCP topology in
    architecture when the ocps group is present.
  - Let agent installs disable IPv4/IPv6 on extra NICs
    (cifmw_bm_agent_disabled_ifaces + networkConfig); install nmstate when
    that list is non-empty to satisfy agent-config validation.
  - Document bm_sno (incl. Zuul autohold iteration), reproducer scenarios
    vs baremetal extra-vars, and add spellcheck terms (NICs, autoheld,
    tty).

Jira: [OSPRH-26767](https://issues.redhat.com/browse/OSPRH-26767)
Generated-by: Cursor (claude-4.6-opus-high)
Signed-off-by: Bohdan Dobrelia <bdobreli@redhat.com>


danpawlik (dismissed):

bogdando:
This no longer depends on other PRs, and I tested it downstream

danpawlik (approved):
code and commit messages are fine. What's missing here is: SNO on baremetal host without using devscripts. I like that approach, especially that you find a reason how to use it.
I was hoping to use the solution in our CI but I realize, that we can not do any magic with provisioning VM and do an SNO on it without nested virtualization. We will do another solution, or even different one.

LGTM

bogdando:
recheck [adoption-standalone-to-crc-ceph-provider](https://softwarefactory-project.io/zuul/t/rdoproject.org/build/21565d3a47b140daaecad9506906b65e)

bogdando:
PTAL folks

brjackma (approved):
/LGTM

PR #3773, opened 2026-03-17
[multiple] nova05epsilon: adjust Ceph for BM SNO DCN cases

bogdando:
1.  Adjust ceph.yml post_stage_run hook for DCN conventions

    The ceph.yml post_stage_run hook (via cifmw_ceph_client role) writes Ceph config files to  
    cifmw_ceph_client_fetch_dir (default /tmp/). This template reads those files and provides them as base64-
    encoded values under data.ceph_conf (DCN convention).

2.  Allow overriding ssh and storage_mgmt
    
    To allow BM SNO with ceph using custom ceph CIDR values,
    make ssh_network_range and storage_mgmt_network_range overridable via
    cifmw_ceph_ssh_network_range and cifmw_ceph_storage_mgmt_network_range.
    Both are set in set_fact which clobbers extra vars, so we use the
    cifmw_ indirection with default() to preserve original defaults.
    
    NOTE: storage_network_range also needs this treatment.
    It use to be commented out in set_fact, and this change needs
    extra testing with Ceph ci jobs perhaps.

    Also gather network facts for IP-to-host mapping. 

3. Fix Swift by Ceph RGW on SNO setup                                                                                                                                                              

    On SNO with a single EDPM compute (single-host CephHCI), the Ceph
    ingress service (haproxy/keepalived) is not deployed because the
    ceph_rgw.yml.j2 spec template only creates it for multi-host clusters.

    Parameterize RGW port to correct the Keystone Swift endpoint for SNO.
    Change the VIP detection logic so that if cifmw_cephadm_rgw_vip
    is pre-set (e.g. to the host's storage IP for SNO cases) - it's
    preserved. Otherwise it falls back to cifmw_cephadm_vip
    (the ingress VIP) as before.

    Users will be able to chose from VIP:8080 vs host_ip:8082 accordingly.

Jira: [OSPRH-27641](https://redhat.atlassian.net/browse/OSPRH-27641)
Generated-by: claude-4.6-opus-high

### DIFF
--- a/docs/dictionary/en-custom.txt
+++ b/docs/dictionary/en-custom.txt
@@ -227,6 +227,7 @@ fips
 firewalld
 flbxutz
 fmw
+frontend
 fqdn
 freefonts
 frmo
--- a/hooks/playbooks/ceph.yml
+++ b/hooks/playbooks/ceph.yml
@@ -104,6 +104,10 @@
   gather_facts: false
   become: true
   tasks:
+    - name: Gather network facts for IP-to-host mapping
+      ansible.builtin.setup:
+        gather_subset:
+          - network
     # jq is normally installed by cifmw_block_device role, but when cifmw_ceph_spec_data_devices
     # is defined (indicating block devices are already present), the block device creation play
     # is skipped. Install jq explicitly here to ensure it's available for ceph operations.
@@ -186,9 +190,12 @@
       when:
         - not cifmw_ceph_ipv6 | default(false)
       ansible.builtin.set_fact:
-        ssh_network_range: 192.168.122.0/24
-        # storage_network_range: 172.18.0.0/24
-        storage_mgmt_network_range: 172.20.0.0/24
+        ssh_network_range: >-
+          {{ cifmw_ceph_ssh_network_range | default('192.168.122.0/24') }}
+        storage_network_range: >-
+          {{ cifmw_ceph_storage_network_range | default('172.18.0.0/24') }}
+        storage_mgmt_network_range: >-
+          {{ cifmw_ceph_storage_mgmt_network_range | default('172.20.0.0/24') }}
         all_addresses: ansible_all_ipv4_addresses
         ms_bind_ipv4: true
         ms_bind_ipv6: false
--- a/roles/ci_gen_kustomize_values/templates/nova05epsilon/edpm-nodeset-values-post-ceph/values.yaml.j2
+++ b/roles/ci_gen_kustomize_values/templates/nova05epsilon/edpm-nodeset-values-post-ceph/values.yaml.j2
@@ -1,0 +1,15 @@
+---
+# source: nova05epsilon/edpm-nodeset-values-post-ceph/values.yaml.j2
+# Auto-populates ceph_conf from files written by ceph.yml hook.
+# The ceph.yml post_stage_run hook (via cifmw_ceph_client role) writes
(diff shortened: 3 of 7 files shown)


bogdando:
recheck cifmw-crc-podified-edpm-baremetal

fultonj:
This looks like it should be fine to me. Let's confirm with a test project.

Nit: The commit message says something about "Add computes to zuul inventory" but I don't see that in the changed files.

fmount:
> This looks like it should be fine to me. Let's confirm with a test project.

+1 I agree, looks ok but testprojects w/ the Ceph related scenarios will confirm.

bogdando:
tested this downstream for BM SNO DCN job. I will test this also with an hci uni job for regressions - together with https://github.com/openstack-k8s-operators/ci-framework/pull/3972

bogdando:
recheck [cifmw-pod-ansible-test](https://gateway-cloud-softwarefactory.apps.ocp.cloud.ci.centos.org/zuul/t/rdoproject.org/build/c497f2ede8224bcfb3089c1cf599b556)

hooks/playbooks/fix_swift_endpoint.yml
```diff
@@ -0,0 +1,97 @@
+---
```
fmount: -1 on this: I think that instead of trying to workaround this way we should try to provide a real fix or support a single node deployment.

I have the following considerations so far:

1. [cifmw_cephadm_rgw_vip](https://github.com/openstack-k8s-operators/ci-framework/blob/main/hooks/playbooks/ceph.yml#L465C9-L465C30) is something we can override before calling the ceph deploy. If the RGW instance IP is predictable, we can keep that logic unchanged and pass it through an `extravar`

2. In my understanding if you have a single node `CephIngress` is not deployed at all (https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/cifmw_cephadm/templates/ceph_rgw.yml.j2#L28). 
This means that you might land to https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/cifmw_cephadm/tasks/configure_object.yml that requires a fix (expose the port as a parameter and `| default(8080)`). This way you can pass 8082 as an override and reach RGW directly.

@bogdando I might have missed pieces here for sure, but the main goal is to not create yet another playbook to workaround a scenario that requires a proper fix. Another option for that particular job is to not deploy `RGW` if it is not used by the job (`cifmw_ceph_daemons_layout.rgw_enabled: false`).

fmount (dismissed):

hooks/playbooks/fix_swift_endpoint.yml
bogdando: Thank you for reviewing! Yes, that makes sense, I will figure the solution instead of this w/a

roles/ci_local_storage/molecule/default/prepare.yml
```diff
@@ -17,6 +17,7 @@
 
 - name: Prepare
   hosts: all
+  become: true
```
bogdando: unrelated change, I will roll back it

bogdando:
downstream testproject passed tempest tests, the horizon test failures are unrelated

evallesp (approved):
/lgtm

bogdando:
please merge this one

fmount (approved):

fultonj (approved):
/approve
/lgtm

PR #3793, opened 2026-03-24
Adding loki and minio roles

lnatapov:
Adding loki and minio roles


roles/deploy_loki/defaults/main.yml
```diff
@@ -0,0 +1,56 @@
...
+
+cifmw_deploy_loki_openstack_namespace: openstack
+cifmw_deploy_loki_s3_secret_name: cloudkitty-loki-s3
+cifmw_deploy_loki_s3_bucket: loki
+
+# Referenced by the S3 secret manifest (must match your MinIO install). When
+# cifmw_deploy_loki_deploy_minio is true, defaults match roles/deploy_minio.
+cifmw_deploy_minio_namespace: minio-dev
+cifmw_deploy_minio_access_key: minio
+cifmw_deploy_minio_secret_key: minio123
```
michburk: These vars violate your own rule you set on L18 of this file:
`# All variables within this role should have a prefix of "cifmw_deploy_loki"`.

There are a couple ways you could remedy this,
for example, you could define these vars as `cifmw_deploy_loki_minio_*` instead and do something like:

```
- name: Deploy MinIO for Loki object storage
  vars:
    cifmw_deploy_minio_namespace: "{{ cifmw_deploy_loki_minio_namespace }}"
    cifmw_deploy_minio_access_key: "{{ cifmw_deploy_loki_minio_access_key }}"
    cifmw_deploy_minio_secret_key: "{{ cifmw_deploy_loki_minio_secret_key }}"
  ansible.builtin.include_role:
    name: deploy_minio
  when:
    - cifmw_deploy_loki | bool
    - cifmw_deploy_loki_deploy_minio | bool
```

This makes the data flow explicit and avoids defining two separate sources of truth for each variable.

michburk (commented):
Would it be possible to use `kubernetes.core.k8s` and `kubernetes.core.k8s_info` in place of running `oc` commands through `anisble.builtin.shell`/`ansible.builtin.command`?

Additionally, README files for each role would help explain the usage and purpose of each role. In particular, explaining how the `deploy_loki` role can deploy minio itself, and expectations around which vars are passed/set and where when using pre-deployed minio vs having the deploy_loki role deploy minio.

roles/deploy_minio/templates/minio_for_ck.yaml.j2
```diff
@@ -0,0 +1,100 @@
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: {{ cifmw_deploy_minio_namespace }}
+  labels:
+    name: {{ cifmw_deploy_minio_namespace }}
+---
+apiVersion: v1
+kind: Pod
```
stuggi: I think we should use a deployment so the pod gets re-created when terminate for some reason

roles/deploy_minio/templates/minio_for_ck.yaml.j2
```diff
@@ -0,0 +1,100 @@
...
+{% if cifmw_deploy_minio_route_console_host | length > 0 %}
+  host: {{ cifmw_deploy_minio_route_console_host }}
+{% endif %}
+  to:
+    kind: Service
+    name: minio
+    weight: 100
+  port:
+    targetPort: console
+  wildcardPolicy: None
```
stuggi: maybe add tls?
```
  tls:
    termination: edge
    insecureEdgeTerminationPolicy: Redirect
```

michburk:
Just wanted to follow up on this pr - I see that portions of these changes have been merged in a separate pr. Should we close this one @lnatapov  ?

PR #3799, opened 2026-03-25
Add default('') to discovered image vars

fultonj:
`cifmw_discovered_image_url` and `cifmw_discovered_hash` are set dynamically by the `discover_latest_image` role. When `networking_mapper` serializes hostvars early in the play, these variables are not yet defined, causing a fatal undefined variable error. Add `| default('')` so serialization succeeds; the real values are populated later when `discover_latest_image` runs.

### DIFF
--- a/scenarios/reproducers/va-hci.yml
+++ b/scenarios/reproducers/va-hci.yml
@@ -70,8 +70,8 @@ cifmw_libvirt_manager_configuration:
       uefi: "{{ cifmw_use_uefi }}"
       root_part_id: "{{ cifmw_root_partition_id }}"
       amount: "{{ [cifmw_libvirt_manager_compute_amount|int, 3] | max }}"
-      image_url: "{{ cifmw_discovered_image_url }}"
-      sha256_image_name: "{{ cifmw_discovered_hash }}"
+      image_url: "{{ cifmw_discovered_image_url | default('') }}"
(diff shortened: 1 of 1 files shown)


danpawlik (approved):

michburk (commented):
I understand the motivation for the changes and they lgtm, but is there a reason to only apply these changes to `va-hci.yml`? If this is an issue for other scenarios, we should consider a more widespread fix.

fultonj:
recheck

fultonj:
> I understand the motivation for the changes and they lgtm, but is there a reason to only apply these changes to `va-hci.yml`? If this is an issue for other scenarios, we should consider a more widespread fix.

I was trying to reproduce `va-hci.yml` in my environment and encountered the problem this patch fixes. 

I didn't try to fix it for other scenarios because I was not testing them.

evallesp (approved):

evallesp:
> `cifmw_discovered_image_url` and `cifmw_discovered_hash` are set dynamically by the `discover_latest_image` role. When `networking_mapper` serializes hostvars early in the play, these variables are not yet defined, causing a fatal undefined variable error. Add `| default('')` so serialization succeeds; the real values are populated later when `discover_latest_image` runs.

Created: https://redhat.atlassian.net/browse/OSPCIFMW-105
Let's merge this as it-is now.

evallesp (approved):
/lgtm
/approve

PR #3814, opened 2026-03-30
[disconnected] Create hook to configure disconnected cluster

drosenfe:
Create a hook to configure openshift cluster for disconnected deployment. The hook should execute after the cluster is deployed but before openstack is deployed.

jira: https://redhat.atlassian.net/browse/OSPRH-21316

Signed-off-by: David Rosenfeld drosenfe@redhat.com

### DIFF
--- a/hooks/playbooks/config_cluster_for_disconnected_deployment.yml
+++ b/hooks/playbooks/config_cluster_for_disconnected_deployment.yml
@@ -1,0 +1,310 @@
+---
+# This is a pre_infra ci-framework hook that will configure the cluster for
+# disconnected deployment. The variable: cifmw_ci_gen_kustomize_values_ooi_image
+# must be specified. These examples for that variable are supported:
+#
+# registry-proxy.engineering.redhat.com/rh-osbs/iib:1125611
+# registry.redhat.io/redhat/redhat-operator-index:v4.18
+#
+# Due to being in deprecated sqlite format this is unsupported:
+# images.paas.redhat.com/podified-main-rhos-18-rhel-9/openstack-operator-index:trunk-patches-latest
+#
+# sqlite requires deprecated v1 oc-mirror workflow instead of the supported v2
+# oc-mirror workflow
+#
+#
+- name: Update cluster for disconnected deployment
+  hosts: "{{ cifmw_target_host | default('localhost') }}"
+  vars:
+    oc_mirror_download_url: "{{ cifmw_disconnected_mirror_url | default('https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/latest/oc-mirror.rhel9.tar.gz') }}"
+    mirror_registry_url: "{{ cifmw_disconnected_registry_url | default('https://mirror.openshift.com/pub/cgw/mirror-registry/latest/mirror-registry-amd64.tar.gz') }}"
+    disconnect_working_dir: "{{ cifmw_disconnected_working_dir | default('/home/zuul/disconnect_working_dir') }}"
+    mirror_location: "{{ disconnect_working_dir }}/mirror_location"
+    local_registry: "{{ disconnect_working_dir }}/local_registry"
+    oc_mirror_cert_manager_catalog_url: "{{ cifmw_cert_manager_catalog_url | default('registry.redhat.io/redhat/redhat-operator-index:v4.18') }}"
+  tasks:
+    - name: Create disconnected working directories
+      ansible.builtin.file:
+        path: "{{ item }}"
+        state: directory
+        mode: '0777'
+      loop:
+        - "{{ disconnect_working_dir }}"
+        - "{{ mirror_location }}"
+        - "{{ local_registry }}"
+
+    - name: Download oc mirror image to controller
+      ansible.builtin.get_url:
+        url: "{{ oc_mirror_download_url }}"
+        dest: "{{ disconnect_working_dir }}/oc-mirror.rhel9.tar.gz"
+        mode: '0644'
+
+    - name: Extract downloaded oc mirror archive
+      ansible.builtin.unarchive:
+        src: "{{ disconnect_working_dir }}/oc-mirror.rhel9.tar.gz"
+        dest: "{{ disconnect_working_dir }}"
+        remote_src: true
+
+    - name: Make oc mirror executable
+      ansible.builtin.command:
+        cmd: chmod +x {{ disconnect_working_dir }}/oc-mirror
+
+    - name: Move oc mirror to bin directory
+      become: true
+      ansible.builtin.command:
+        cmd: mv {{ disconnect_working_dir }}/oc-mirror /usr/local/bin/.
+
+    - name: Get host FQDN
+      ansible.builtin.command:
+        cmd: hostname -f
+      register: host_fqdn
+
+    - name: Create mirror location file
+      become: true
+      ansible.builtin.copy:
+        dest: "/etc/containers/registries.conf.d/010-stage.conf"
+        mode: '0644'
+        content: |
+          unqualified-search-registries = ["registry.access.redhat.com", "docker.io"]
+          short-name-mode = ""
+
+          [[registry]]
+          prefix = ""
+          location = "registry.redhat.io"
+
+          [[registry.mirror]]
+          location = "registry.stage.redhat.io"
+          pull-from-mirror = "digest-only"
+
+    - name: Create update service namespace
+      kubernetes.core.k8s:
+        state: present
+        definition:
+          apiVersion: v1
+          kind: Namespace
+          metadata:
+            name: openshift-update-service
+            annotations:
+              openshift.io/node-selector: ""
+            labels:
+              openshift.io/cluster-monitoring: "true"
+
+    - name: Create update service operator group
+      kubernetes.core.k8s:
+        state: present
+        definition:
+          apiVersion: operators.coreos.com/v1
+          kind: OperatorGroup
+          metadata:
+            name: update-service-operator-group
+            namespace: openshift-update-service
+          spec:
+            targetNamespaces:
+              - openshift-update-service
+
+    - name: Create subscription service
+      kubernetes.core.k8s:
+        state: present
+        definition:
+          apiVersion: operators.coreos.com/v1alpha1
+          kind: Subscription
+          metadata:
+            name: update-service-subscription
+            namespace: openshift-update-service
+          spec:
+            channel: v1
+            installPlanApproval: "Automatic"
+            source: "redhat-operators"
+            sourceNamespace: "openshift-marketplace"
+            name: "cincinnati-operator"
+
+    - name: Wait for update service operator to be installed
+      ansible.builtin.shell: |
+        set -o pipefail
+        oc get crd | grep -i updateservice.operator.openshift.io
+      register: crd_out
+      until: "'updateservice.operator.openshift.io' in crd_out.stdout"
+      retries: 10
+      delay: 30
+
+    - name: Create Image Set yaml when two catalogs are required
+      ansible.builtin.copy:
+        dest: "{{ disconnect_working_dir }}/imageset-config-v2.yaml"
+        mode: '0644'
+        content: |
+          kind: ImageSetConfiguration
+          apiVersion: mirror.openshift.io/v2alpha1
+          mirror:
+            operators:
+              - catalog: {{ cifmw_ci_gen_kustomize_values_ooi_image }}
+                packages:
+                 - name: openstack-operator
+                 - name: local-storage-operator
+              - catalog: {{ oc_mirror_cert_manager_catalog_url }}
+                packages:
+                 - name: kubernetes-nmstate-operator
+                 - name: openshift-cert-manager-operator
+                 - name: metallb-operator
+                 - name: lvms-operator
+                 - name: cluster-observability-operator
+            additionalImages:
+             - name: registry.redhat.io/ubi8/ubi:latest
+             - name: registry.redhat.io/ubi9/ubi@sha256:20f695d2a91352d4eaa25107535126727b5945bff38ed36a3e59590f495046f0
+      when: cifmw_ci_gen_kustomize_values_ooi_image != oc_mirror_cert_manager_catalog_url
+
+    - name: Create Image Set yaml when only one catalog is required
+      ansible.builtin.copy:
+        dest: "{{ disconnect_working_dir }}/imageset-config-v2.yaml"
+        mode: '0644'
+        content: |
+          kind: ImageSetConfiguration
+          apiVersion: mirror.openshift.io/v2alpha1
+          mirror:
+            operators:
+              - catalog: {{ cifmw_ci_gen_kustomize_values_ooi_image }}
+                packages:
+                 - name: openstack-operator
+                 - name: local-storage-operator
+                 - name: kubernetes-nmstate-operator
+                 - name: openshift-cert-manager-operator
+                 - name: metallb-operator
+                 - name: lvms-operator
+                 - name: cluster-observability-operator
+            additionalImages:
+             - name: registry.redhat.io/ubi8/ubi:latest
+             - name: registry.redhat.io/ubi9/ubi@sha256:20f695d2a91352d4eaa25107535126727b5945bff38ed36a3e59590f495046f0
+      when: cifmw_ci_gen_kustomize_values_ooi_image == oc_mirror_cert_manager_catalog_url
+
+    - name: Get registry.redhat.io username and password from pull secret
+      ansible.builtin.shell: |
+        set -o pipefail
+        oc get -n openshift-config secret pull-secret -o json | jq '.data[".dockerconfigjson"]' -r | base64 -d | jq '.auths["registry.redhat.io"].auth' -r | base64 -d
+      register: pull_secret_user_pass
+
+    - name: Login to registry.redhat.io
+      containers.podman.podman_login:
+        username: "{{ pull_secret_user_pass.stdout.split(':')[0] }}"
+        password: "{{ pull_secret_user_pass.stdout.split(':')[1] }}"
+        registry: "registry.redhat.io"
+
+    - name: Login to registry.stage.redhat.io
+      containers.podman.podman_login:
+        username: "{{ cifmw_registry_token.credentials.username }}"
+        password: "{{ cifmw_registry_token.credentials.password }}"
+        registry: "registry.stage.redhat.io"
+
+    - name: Mirror specified image set configuration to disk
+      ansible.builtin.shell: |
+        oc mirror --v2 --config {{ disconnect_working_dir }}/imageset-config-v2.yaml file://{{ mirror_location }} >>{{ disconnect_working_dir }}/mirror_images.log
+      register: mirror_image_result
+      until: mirror_image_result is not failed
+      retries: 1
+
+    - name: Download mirror registry to controller
+      ansible.builtin.get_url:
+        url: "{{ mirror_registry_url }}"
+        dest: "{{ disconnect_working_dir }}/mirror-registry-amd64.tar.gz"
+        mode: '0644'
+
+    - name: Extract downloaded mirror registry archive
+      ansible.builtin.unarchive:
+        src: "{{ disconnect_working_dir }}/mirror-registry-amd64.tar.gz"
+        dest: "{{ disconnect_working_dir }}"
+        remote_src: true
+
+    - name: Install mirror registry
+      ansible.builtin.shell: |
+        {{ disconnect_working_dir }}/mirror-registry install --quayHostname {{ host_fqdn.stdout }} --quayRoot \
+        {{ local_registry }} --initPassword {{ cifmw_mirror_registry_init_password }} >{{ disconnect_working_dir }}/registry_install.log
+
+    - name: Increase gunicorn-web timeout in quay-app container
+      ansible.builtin.command:
+        cmd: >
+          podman exec quay-app {% raw %}sed -i '/command=gunicorn
+          -c %(ENV_QUAYCONF)s\/gunicorn_web.py web:application/c\command=gunicorn
+          --timeout 900 -c %(ENV_QUAYCONF)s\/gunicorn_web.py web:application'
+          /quay-registry/conf/supervisord.conf{% endraw %}
+
+    - name: Reread supervisord.conf in quay-app container
+      ansible.builtin.command:
+        cmd: podman exec quay-app supervisorctl -c /quay-registry/conf/supervisord.conf reread
+
+    - name: Restart gunicorn-web quay-app container with new timeout value
+      ansible.builtin.command:
+        cmd: podman exec quay-app supervisorctl -c /quay-registry/conf/supervisord.conf restart gunicorn-web
+
+    - name: Copy local registry root certificate to system anchors directory
+      become: true
+      ansible.builtin.copy:
+        src: "{{ local_registry }}/quay-rootCA/rootCA.pem"
+        dest: /etc/pki/ca-trust/source/anchors/
+        mode: '0644'
+        remote_src: true
+
+    - name: Configure system to trust mirror registry root ca
+      become: true
+      ansible.builtin.command:
+        cmd: update-ca-trust extract
+
+    - name: Login to  mirror registry
+      containers.podman.podman_login:
+        username: "init"
+        password: "{{ cifmw_mirror_registry_init_password }}"
+        registry: "{{ host_fqdn.stdout }}:8443"
+
+    - name: Create mirror registry root ca configmap
+      ansible.builtin.command:
+        cmd: oc create configmap registry-cas -n openshift-config --from-file={{ host_fqdn.stdout }}..8443={{ local_registry }}/quay-rootCA/rootCA.pem
+
+    - name: Patch cluster with mirror registry root ca
+      ansible.builtin.command:
+        cmd: oc patch image.config.openshift.io/cluster --patch '{"spec":{"additionalTrustedCA":{"name":"registry-cas"}}}' --type=merge
+
+    - name: Get cluster's current pull secret
+      ansible.builtin.shell: |
+        oc get secret {% raw %}pull-secret -n openshift-config -o template='{{index .data ".dockerconfigjson" | base64decode}}'{% endraw %} > {{ disconnect_working_dir }}/pull-secret.json
+
+    - name: Configure cluster to use pull secret from mirror registry
+      ansible.builtin.shell: |
+        set -eux
+        oc registry login --registry {{ host_fqdn.stdout }}:8443 --auth-basic=init:{{ cifmw_mirror_registry_init_password }} --to={{ disconnect_working_dir }}/pull-secret.json
+        oc set data secret/pull-secret -n openshift-config --from-file=.dockerconfigjson={{ disconnect_working_dir }}/pull-secret.json
+
+    - name: Mirror contents of generated image set to target mirror registry
+      ansible.builtin.shell: |
+        oc mirror --v2 --config {{ disconnect_working_dir }}/imageset-config-v2.yaml --from file://{{ mirror_location }} docker://{{ host_fqdn.stdout }}:8443 >>{{ disconnect_working_dir }}/mirror_contents.log
+      register: mirror_contents_result
+      until: mirror_contents_result is not failed
+      retries: 1
+
+    - name: Disable catalog source
+      ansible.builtin.shell: |
+        oc patch OperatorHub cluster --type json -p '[{"op": "add", "path": "/spec/disableAllDefaultSources", "value": true}]'
+
+    - name: Extract image name and tag from catalog source
+      ansible.builtin.set_fact:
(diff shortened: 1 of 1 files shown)


hooks/playbooks/config_cluster_for_disconnected_deployment.yml
```diff
@@ -0,0 +1,226 @@
...
+        cat <<EOF >{{ disconnect_working_dir }}/imageset-config-v2.yaml
+        kind: ImageSetConfiguration
+        apiVersion: mirror.openshift.io/v2alpha1
+        mirror:
+          operators:
+            - catalog: {{ oc_mirror_catalog_url }}
+              packages:
+               - name: openstack-operator
+               - name: kubernetes-nmstate-operator
+               - name: openshift-cert-manager-operator
```
slagle: cert-manager-operator needs to be listed with a different catalog like so:

```
kind: ImageSetConfiguration
apiVersion: mirror.openshift.io/v2alpha1
mirror:
  operators:
    - catalog: {{ oc_mirror_catalog_url }}
      packages:
       - name: openstack-operator
       - name: kubernetes-nmstate-operator
       - name: openshift-cert-manager-operator
       - name: metallb-operator
       - name: local-storage-operator
       - name: lvms-operator
       - name: cluster-observability-operator
    - catalog: {{ oc_mirror_cert_manager_catalog_url }}
      packages:
       - name: openshift-cert-manager-operator
  additionalImages:
   - name: registry.redhat.io/ubi8/ubi:latest
   - name: registry.redhat.io/ubi9/ubi@sha256:20f695d2a91352d4eaa25107535126727b5945bff38ed36a3e59590f495046f0
```

I've used a separate var for the catalog source for openshift-cert-manager-operator, `{{ oc_mirror_cert_manager_catalog_url }}`, which should default to `registry.redhat.io/redhat/redhat-operator-index:v4.18`
This is because the cert-manager operator is not included in our index image we use in CI.

After using this ImageSetConfiguration you'll end up with CatalogSources generated for both sources:

```
[stack@host05 disconnected]$ ls -l mirror_location/working-dir/cluster-resources/
total 24
-rw-r--r--. 1 stack stack  417 Apr  9 21:11 cc-iib-1125611.yaml
-rw-r--r--. 1 stack stack  448 Apr  9 21:11 cc-redhat-operator-index-v4-18.yaml
-rw-r--r--. 1 stack stack  421 Apr 10 06:54 cs-iib-1125611.yaml
-rw-r--r--. 1 stack stack  436 Apr 10 13:34 cs-redhat-operator-index-v4-18.yaml
-rw-r--r--. 1 stack stack 1909 Apr  9 21:11 idms-oc-mirror.yaml
-rw-r--r--. 1 stack stack 1091 Apr  9 21:11 itms-oc-mirror.yaml
```

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
```diff
@@ -0,0 +1,226 @@
...
+      ansible.builtin.shell: |
+        oc patch OperatorHub cluster --type json -p '[{"op": "add", "path": "/spec/disableAllDefaultSources", "value": true}]'
+
+    - name: Extract image name and tag from catalog source
+      ansible.builtin.set_fact:
+        index_image_name_tag: "{{ oc_mirror_catalog_url.split('/') | last | replace(':', '-') | replace('.', '-') }}"
+
+    - name: Prepare catalog source for environment
+      ansible.builtin.shell: |
+        sed -i 's/cs-{{ index_image_name_tag | quote }}/redhat-operators/g' {{ mirror_location }}/working-dir/cluster-resources/cs-{{ index_image_name_tag }}.yaml
```
slagle: you only want this to apply to cs-redhat-operator-index-v4-18.yaml

cifmw will create it's own CatalogSource called openstack-operator-index that will be configured to use `cifmw_ci_gen_kustomize_values_ooi_image`, and then the mirror configuration will make sure the mirror gets used.

slagle (commented):
you'll also need another task to login the dataplane nodes to the mirror registry. you can add this to edpm_container_registry_logins somewhere.

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
drosenfe: In the example in oc_mirror_catalog_url section the openshift-cert-manager-operator line should be removed?

slagle: yes, i think it should probably be removed. I didn't notice that when I teseted it though, and it seemed to be ok. let's try removing it though and see what happens.

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
drosenfe: I believe this will only apply to: cs-redhat-operator-index-v4-18.yaml. The previous set fact statement uses variable: oc_mirror_cert_manager_catalog_url to extract: redhat-operator-index-v4-18. I could hardcode the file name, but think this gives a little more flexibility.

TristanCacqueray:
recheck

drosenfe:
recheck

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
```diff
@@ -0,0 +1,296 @@
...
+      ansible.builtin.unarchive:
+        src: "{{ disconnect_working_dir }}/oc-mirror.rhel9.tar.gz"
+        dest: "{{ disconnect_working_dir }}"
+        remote_src: true
+
+    - name: Install oc mirror
+      ansible.builtin.shell: |
+        set -eux
+        chmod +x {{disconnect_working_dir}}/oc-mirror &&
+        sudo mv {{disconnect_working_dir}}/oc-mirror /usr/local/bin/.
```
michburk: Here and in a a couple other places, add spaces around the expression within the `{{ }}`. E.g.
`{{disconnect_working_dir}}` -> `{{ disconnect_working_dir }}`

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
```diff
@@ -0,0 +1,296 @@
...
+        {{ local_registry }} --initPassword {{ cifmw_mirror_registry_init_password }} >{{disconnect_working_dir}}/registry_install.log
+
+    - name: Increase gunicorn-web timeout in quay-app container
+      ansible.builtin.command: >-
+        podman exec -it quay-app {% raw %}sed -i '/command=gunicorn
+        -c %(ENV_QUAYCONF)s\/gunicorn_web.py web:application/c\command=gunicorn
+        --timeout 900 -c %(ENV_QUAYCONF)s\/gunicorn_web.py web:application'
+        /quay-registry/conf/supervisord.conf{% endraw %}
+
+    - name: Reread supervisord.conf in quap-app container
```
michburk: Typo: quap -> quay

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
```diff
@@ -0,0 +1,296 @@
...
+
+    - name: Configure system to trust mirror registry root ca
+      become: true
+      ansible.builtin.shell: |
+        cp {{ local_registry }}/quay-rootCA/rootCA.pem /etc/pki/ca-trust/source/anchors/
+        update-ca-trust extract
+
+    - name: Login to  mirror registry
+      ansible.builtin.shell: |
+        podman login -u init -p {{ cifmw_mirror_registry_init_password }} {{ host_fqdn.stdout }}:8443
```
michburk: Could you use `containers.podman.podman_login:` here like you do above?

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
```diff
@@ -0,0 +1,296 @@
...
+      ansible.builtin.shell: |
+        set -o pipefail
+        oc get crd | grep -i updateservice.operator.openshift.io
+      register: crd_out
+      until: "'updateservice.operator.openshift.io' in crd_out.stdout"
+      retries: 10
+      delay: 30
+
+    - name: Create Image Set yaml when two catalogs are required
+      ansible.builtin.shell: |
```
michburk: Could you use `ansible.builtin.copy`/`ansible.builtin.template` here instead of `ansible.builtin.shell` + `cat` + heredoc?

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
```diff
@@ -0,0 +1,296 @@
...
+
+    - name: Apply yaml files from results directory to cluster
+      ansible.builtin.shell: |
+        oc apply -f {{ mirror_location }}/working-dir/cluster-resources
+
+    - name: Wait for mirrored operators to be available
+      ansible.builtin.shell: |
+        oc get packagemanifests.packages.operators.coreos.com
+      register: packagemanifest_out
+      until: "'openstack-operator' and 'kubernetes-nmstate-operator' in packagemanifest_out.stdout"
```
michburk: Is this condition meant to be checking if both of these strings are in `packagemanifest_out.stdout`? If so I think this condition should be more like
```
until:
  - 'openstack-operator' in packagemanifest_out.stdout
  - 'kubernetes-nmstate-operator' in packagemanifest_out.stdout
```

If I'm not mistaken, as it is written now, the task is checking if
1. the expression `'openstack-operator'` is true
and
2. `'kubernetes-nmstate-operator' in packagemanifest_out.stdout` is true

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
```diff
@@ -0,0 +1,296 @@
...
+        chmod +x {{disconnect_working_dir}}/oc-mirror &&
+        sudo mv {{disconnect_working_dir}}/oc-mirror /usr/local/bin/.
+
+    - name: Get host FQDN
+      ansible.builtin.command: hostname -f
+      register: host_fqdn
+
+    - name: Create mirror location file
+      become: true
+      ansible.builtin.shell: |
```
michburk: Rather than using `ansible.builtin.shell` with `cat` and a heredoc, could you use a module like `ansbile.buitlin.copy` instead?

michburk (commented):
There are a couple of places where you do straightforward operations with the `oc` command, is it possible to use modules like `kubernetes.core.k8s`/`kubernetes.core.k8s_info` instead? I know that not all instances of `ansible.builtin.shell` + `oc ...` can be translated as such, but if I'm not mistaken, some of the simpler `oc apply/patch/create` commands could.

Additionally, do you have a jira you could link here? And if you have run this in a testproject, we would greatly appreciate if that were attached to the jira ticket / in the comments of that jira ticket.

Thanks!

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
drosenfe: Believe ansible.builtin.template is for copying an existing file to a new location and modifying it in the process. This file and other files in the review are new files that don't exist in any form in the server. Does the comment apply in that case? If so can it be clarified some more about what is required?

michburk: Maybe my suggestion to use `ansible.builtin.template` was a little misplaced. This snippet isn't huge, there's no complicated jinja going on here, and this is a hook playbook so finding a place for the `.j2` file might be awkward. Refactoring to `ansible.builtin.copy` would be good IMO. e.g.
```
- name: Create Image Set yaml when two catalogs are required
  ansible.builtin.copy:
    dest: "{{ disconnect_working_dir }}/imageset-config-v2.yaml"
    content: |
      <all of the same content you put in the heredoc>
  when: cifmw_ci_gen_kustomize_values_ooi_image != oc_mirror_cert_manager_catalog_url
```
and if you want to control the `mode` as well you could supply `mode: '0644'` (or whatever might be appropriate) to the copy module.

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
drosenfe: Updated in latest version

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
drosenfe: fixed in latest version

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
drosenfe: fixed in latest version

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
drosenfe: fixed in latest version

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
drosenfe: fixed in latest version

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
drosenfe: fixed in latest version

drosenfe:
> There are a couple of places where you do straightforward operations with the `oc` command, is it possible to use modules like `kubernetes.core.k8s`/`kubernetes.core.k8s_info` instead? I know that not all instances of `ansible.builtin.shell` + `oc ...` can be translated as such, but if I'm not mistaken, some of the simpler `oc apply/patch/create` commands could.
> 
> Additionally, do you have a jira you could link here? And if you have run this in a testproject, we would greatly appreciate if that were attached to the jira ticket / in the comments of that jira ticket.
> 
> Thanks!

@michburk looked at using module: kubernetes.core.k8s. The only line that was potentially supported was: oc apply -f {{ mirror_location }}/working-dir/cluster-resources. However, that is applying a directory rather than a file which the module kubernetes.core.k8s did not support. I believe all other comments were addressed. Are you OK with review now?

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
```diff
@@ -0,0 +1,303 @@
...
+# sqlite requires deprecated v1 oc-mirror workflow instead of the supported v2
+# oc-mirror workflow
+#
+#
+- name: Update cluster for disconnected deployment
+  hosts: "{{ cifmw_target_host | default('localhost') }}"
+  vars:
+    oc_mirror_download_url: "{{ cifmw_disconnected_mirror_url | default('https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/latest/oc-mirror.rhel9.tar.gz') }}"
+    mirror_registry_url: "{{ cifmw_disconnected_registry_url | default('https://mirror.openshift.com/pub/cgw/mirror-registry/latest/mirror-registry-amd64.tar.gz') }}"
+    openstack_namespace: "{{ cifmw_openstack_namespace | default('openstack') }}"
```
michburk: var is defined but unused

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
```diff
@@ -0,0 +1,303 @@
...
+        remote_src: true
+
+    - name: Install mirror registry
+      ansible.builtin.shell: |
+        {{ disconnect_working_dir }}/mirror-registry install --quayHostname {{ host_fqdn.stdout }} --quayRoot \
+        {{ local_registry }} --initPassword {{ cifmw_mirror_registry_init_password }} >{{ disconnect_working_dir }}/registry_install.log
+
+    - name: Increase gunicorn-web timeout in quay-app container
+      ansible.builtin.command: >-
+        podman exec -it quay-app {% raw %}sed -i '/command=gunicorn
```
michburk: Just checking for my own understanding: are the `-it` flags on each of these `podman exec` commands necessary?

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
```diff
@@ -0,0 +1,303 @@
...
+        podman exec -it quay-app supervisorctl -c /quay-registry/conf/supervisord.conf help reread
+
+    - name: Restart gunicorn-web quay-app container with new timeout value
+      ansible.builtin.command:
+        podman exec -it quay-app supervisorctl -c /quay-registry/conf/supervisord.conf restart gunicorn-web
+
+    - name: Configure system to trust mirror registry root ca
+      become: true
+      ansible.builtin.shell: |
+        cp {{ local_registry }}/quay-rootCA/rootCA.pem /etc/pki/ca-trust/source/anchors/
```
michburk: Could be split into two tasks, `ansible.builtin.copy`  this file and a `ansible.builtin.command` task for `update-ca-trust extract`.

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
```diff
@@ -0,0 +1,303 @@
...
+
+    - name: Login to  mirror registry
+      containers.podman.podman_login:
+        username: "init"
+        password: "{{ cifmw_mirror_registry_init_password }}"
+        registry: "{{ host_fqdn.stdout }}:8443"
+
+    - name: Configure cluster to trust mirror registry root ca
+      ansible.builtin.shell: |
+        set -eux
```
michburk: could be split into two separate `ansible.builtin.command` tasks

michburk (commented):
> > There are a couple of places where you do straightforward operations with the `oc` command, is it possible to use modules like `kubernetes.core.k8s`/`kubernetes.core.k8s_info` instead? I know that not all instances of `ansible.builtin.shell` + `oc ...` can be translated as such, but if I'm not mistaken, some of the simpler `oc apply/patch/create` commands could.
> > Additionally, do you have a jira you could link here? And if you have run this in a testproject, we would greatly appreciate if that were attached to the jira ticket / in the comments of that jira ticket.
> > Thanks!
> 
> @michburk looked at using module: kubernetes.core.k8s. The only line that was potentially supported was: oc apply -f {{ mirror_location }}/working-dir/cluster-resources. However, that is applying a directory rather than a file which the module kubernetes.core.k8s did not support. I believe all other comments were addressed. Are you OK with review now?

A couple more things, very sorry for not catching them on the first pass. It's generally considered best practice to avoid `shell` where possible and prefer more specific modules or `ansible.builtin.command` if we have to use shell-like commands and don't need pipes/&&/||/etc.

I do want to follow-up and ask, is it the case that the tasks
`Create update service namespace`,  `Create update service operator group`, and `Create subscription service` for example can't be converted to use `kubernetes.core.k8s`? Very sorry if I'm missing something obvious here.

I'll keep an eye on this pr to try to be more responsive, sorry for not seeing your comment yesterday.

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
```diff
@@ -0,0 +1,310 @@
...
+    - name: Restart gunicorn-web quay-app container with new timeout value
+      ansible.builtin.command:
+        podman exec -it quay-app supervisorctl -c /quay-registry/conf/supervisord.conf restart gunicorn-web
+
+    - name: Copy local registry root certificate to system anchors directory
+      become: true
+      ansible.builtin.copy:
+        src: "{{ local_registry }}/quay-rootCA/rootCA.pem"
+        dest: /etc/pki/ca-trust/source/anchors/
+        remote_src: true
```
michburk, line 242: the pre-commit check is failing with
```
WARNING  Listing 1 violation(s) that are fatal
hooks/playbooks/config_cluster_for_disconnected_deployment.yml:237: risky-file-permissions: File permissions unset or incorrect.
```
providing an explicit, appropriate `mode:` should fix it

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
drosenfe: openstack_namespace removed in latestet version.

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
drosenfe: I removed the -it options in latest version and it still worked.

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
drosenfe: Fixed in latest version.

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
drosenfe: Fixed in latest version.

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
drosenfe: Fixed in latest version.

drosenfe:
> > > There are a couple of places where you do straightforward operations with the `oc` command, is it possible to use modules like `kubernetes.core.k8s`/`kubernetes.core.k8s_info` instead? I know that not all instances of `ansible.builtin.shell` + `oc ...` can be translated as such, but if I'm not mistaken, some of the simpler `oc apply/patch/create` commands could.
> > > Additionally, do you have a jira you could link here? And if you have run this in a testproject, we would greatly appreciate if that were attached to the jira ticket / in the comments of that jira ticket.
> > > Thanks!
> > 
> > 
> > @michburk looked at using module: kubernetes.core.k8s. The only line that was potentially supported was: oc apply -f {{ mirror_location }}/working-dir/cluster-resources. However, that is applying a directory rather than a file which the module kubernetes.core.k8s did not support. I believe all other comments were addressed. Are you OK with review now?
> 
> A couple more things, very sorry for not catching them on the first pass. It's generally considered best practice to avoid `shell` where possible and prefer more specific modules or `ansible.builtin.command` if we have to use shell-like commands and don't need pipes/&&/||/etc.
> 
> I do want to follow-up and ask, is it the case that the tasks `Create update service namespace`, `Create update service operator group`, and `Create subscription service` for example can't be converted to use `kubernetes.core.k8s`? Very sorry if I'm missing something obvious here.
> 
> I'll keep an eye on this pr to try to be more responsive, sorry for not seeing your comment yesterday.

 `Create update service namespace`, `Create update service operator group`, and `Create subscription service` changed to use kubernetes.core.k8s in latest version.

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
```diff
@@ -0,0 +1,305 @@
...
+      ansible.builtin.unarchive:
+        src: "{{ disconnect_working_dir }}/oc-mirror.rhel9.tar.gz"
+        dest: "{{ disconnect_working_dir }}"
+        remote_src: true
+
+    - name: Install oc mirror
+      ansible.builtin.shell: |
+        set -eux
+        chmod +x {{ disconnect_working_dir }}/oc-mirror &&
+        sudo mv {{ disconnect_working_dir }}/oc-mirror /usr/local/bin/.
```
michburk: Should probably be two tasks, where the `mv ...` command should use `become: true` instead of `sudo`

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
```diff
@@ -0,0 +1,305 @@
...
+
+    - name: Wait for mirrored operators to be available
+      ansible.builtin.shell: |
+        oc get packagemanifests.packages.operators.coreos.com
+      register: packagemanifest_out
+      until:
+        - "'openstack-operator' in packagemanifest_out.stdout"
+        - "'kubernetes-nmstate-operator' in packagemanifest_out.stdout"
+      retries: 10
+      delay: 30
```
michburk, line 306: these could be `ansible.builtin.command` instead of `ansible.builtin.shell`

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
```diff
@@ -0,0 +1,305 @@
...
+        username: "{{ cifmw_registry_token.credentials.username }}"
+        password: "{{ cifmw_registry_token.credentials.password }}"
+        registry: "registry.stage.redhat.io"
+
+    - name: Mirror specified image set configuration to disk
+      ansible.builtin.shell: |
+        oc mirror --v2 --config {{ disconnect_working_dir }}/imageset-config-v2.yaml file://{{ mirror_location }} >>{{ disconnect_working_dir }}/mirror_images.log
+      register: mirror_image_result
+      until: mirror_image_result is not failed
+      retries: 1
```
michburk, line 201: not blocking or anything and sorry if I'm missing something, but is 1 retry with no delay intentional here? Just want to double check

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
```diff
@@ -0,0 +1,305 @@
...
+        set -eux
+        oc registry login --registry {{ host_fqdn.stdout }}:8443 --auth-basic=init:{{ cifmw_mirror_registry_init_password }} --to={{ disconnect_working_dir }}/pull-secret.json
+        oc set data secret/pull-secret -n openshift-config --from-file=.dockerconfigjson={{ disconnect_working_dir }}/pull-secret.json
+
+    - name: Mirror contents of generated image set to target mirror registry
+      ansible.builtin.shell: |
+        oc mirror --v2 --config {{ disconnect_working_dir }}/imageset-config-v2.yaml --from file://{{ mirror_location }} docker://{{ host_fqdn.stdout }}:8443 >>{{ disconnect_working_dir }}/mirror_contents.log
+      register: mirror_contents_result
+      until: mirror_contents_result is not failed
+      retries: 1
```
michburk, line 278: same comment about 1 retry with no delay

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
```diff
@@ -0,0 +1,305 @@
...
+
+    - name: Disable catalog source
+      ansible.builtin.shell: |
+        oc patch OperatorHub cluster --type json -p '[{"op": "add", "path": "/spec/disableAllDefaultSources", "value": true}]'
```
michburk, line 282: this could be `ansible.builtin.command` instead of `ansible.builtin.shell`

michburk (commented):
There are a number of places where you use `ansible.builtin.command` as with the command string on a new line without a `cmd` parameter (picking a random example):
```
    - name: Wait until the OpenShift cluster is stable
      ansible.builtin.command:
        oc adm wait-for-stable-cluster --minimum-stable-period=5s --timeout=30m
```
where the following is preferred:
```
    - name: Wait until the OpenShift cluster is stable
      ansible.builtin.command:
        cmd: oc adm wait-for-stable-cluster --minimum-stable-period=5s --timeout=30m
```
Sorry for not catching this earlier, I suppose the first method has worked in your testing, but style-wise the command should either be on the same line as `ansible.builtin.command` i.e.
```
ansible.builtin.command: echo hi
```
or preceded by `cmd: `, i.e.
```
ansible.builtin.command:
  cmd: echo hi
```

really sorry if I'm dragging this along and not catching everything in earlier reviews, patch is generally looking good to me and I keep catching style nits. Sorry about that.

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
drosenfe: This task downloads around 180 container images taking over 20 minutes. When it fails its usually one image that didn't download properly. The retry goes quickly because it doesn't download the images that worked the first time. - just the failed image gets corrected. This task has been working well without a delay.

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
drosenfe: Same as previous comment. Around 180 images are mirrored. A failure is usually result of just one image not being mirrored correctly. The one retry without delay has been successful in recovering from a failure.

drosenfe:
> There are a number of places where you use `ansible.builtin.command` as with the command string on a new line without a `cmd` parameter (picking a random example):
> 
> ```
>     - name: Wait until the OpenShift cluster is stable
>       ansible.builtin.command:
>         oc adm wait-for-stable-cluster --minimum-stable-period=5s --timeout=30m
> ```
> 
> where the following is preferred:
> 
> ```
>     - name: Wait until the OpenShift cluster is stable
>       ansible.builtin.command:
>         cmd: oc adm wait-for-stable-cluster --minimum-stable-period=5s --timeout=30m
> ```
> 
> Sorry for not catching this earlier, I suppose the first method has worked in your testing, but style-wise the command should either be on the same line as `ansible.builtin.command` i.e.
> 
> ```
> ansible.builtin.command: echo hi
> ```
> 
> or preceded by `cmd: `, i.e.
> 
> ```
> ansible.builtin.command:
>   cmd: echo hi
> ```
> 
> really sorry if I'm dragging this along and not catching everything in earlier reviews, patch is generally looking good to me and I keep catching style nits. Sorry about that.

cmd: added to all ansible.builtin.command: tasks.

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
drosenfe: Tried with ansible.builtin.command. It gave error:  Syntax Error while loading YAML. mapping values are not allowed in this context   ansible.builtin.command doesn't like the :'s within the command. I put it back to the original format.

michburk (approved):
/lgtm

slagle (approved):

PR #3820, opened 2026-03-31
[env_op_images] Add pulled images report to env_op_images role

nemarjan:
Builds a pulled_images_report which was created by cross-referencing pod images against ICSP/IDMS mirror rules to report which images have a mirror configured and which pull directly from the original registry. As this is a what "should happen" report we need to verify via CRIO logs on node.

To verify a updated report is created which cross-references the pulled-images report with CRI-O journal logs
from cluster nodes to confirm which images were actually pulled by
the container runtime

### DIFF
--- a/ci/playbooks/collect-logs.yml
+++ b/ci/playbooks/collect-logs.yml
@@ -1,4 +1,15 @@
 ---
+- name: Ensure ci-framework-data base directories exist on all nodes
+  hosts: all
+  gather_facts: false
+  tasks:
+    - name: Create ci-framework-data/logs directory if missing
+      ansible.builtin.file:
+        path: "{{ ansible_user_dir }}/ci-framework-data/logs"
+        state: directory
+        mode: "0755"
+      ignore_errors: true  # noqa: ignore-errors
+
 - name: "Run ci/playbooks/collect-logs.yml"
   hosts: "{{ cifmw_zuul_target_host | default('all') }}"
   gather_facts: true
--- a/roles/env_op_images/README.md
+++ b/roles/env_op_images/README.md
@@ -4,6 +4,9 @@ A role to gather the container images used in the openstack deployment with spec
 ## Parameters
 * `cifmw_env_op_images_dir`: (String) Directory where the operator_images.yaml will be stored. Defaults to `~/ci-framework-data/artifacts`
 * `cifmw_env_op_images_file`: (String) Name of the file storing the operator images and tags. Defaults to `operator_images.yaml`
+* `cifmw_env_op_images_pulled_report_path`: Pulled-images policy report (ICSP/IDMS + pod image refs).
+* `cifmw_env_op_images_verified_report_path`: Output path for the CRI-O-enriched report. After the pulled report runs, fetches `oc adm node-logs NODE -u crio` per node, then writes this file with digest-level CRI-O fields (`node_verified_image_origin`, `log_evidence_uri`, `log_evidence_node`).
+* `cifmw_env_op_images_crio_logs_dir`: Directory for per-node `*.crio.log` files used during verification.
 
 ## Examples
 ```YAML
--- a/roles/env_op_images/defaults/main.yml
+++ b/roles/env_op_images/defaults/main.yml
@@ -21,3 +21,8 @@
 cifmw_env_op_images_dir: "{{ cifmw_basedir }}"
 cifmw_env_op_images_file: operator_images.yaml
 cifmw_env_op_images_dryrun: false
+
+cifmw_env_op_images_pulled_report_file: pulled_images_report.yaml
+cifmw_env_op_images_pulled_report_namespaces:
+  - "{{ cifmw_openstack_namespace | default('openstack') }}"
+  - "{{ operator_namespace | default('openstack-operators') }}"
--- a/roles/env_op_images/defaults/main.yml
+++ b/roles/env_op_images/defaults/main.yml
@@ -22,7 +22,20 @@ cifmw_env_op_images_dir: "{{ cifmw_basedir }}"
 cifmw_env_op_images_file: operator_images.yaml
 cifmw_env_op_images_dryrun: false
 
-cifmw_env_op_images_pulled_report_file: pulled_images_report.yaml
+cifmw_env_op_images_pulled_report_path: >-
+  {{
+    (cifmw_env_op_images_dir, 'artifacts', 'pulled_images_report.yaml')
+    | path_join
+  }}
+
+cifmw_env_op_images_verified_report_path: >-
+  {{
+    (cifmw_env_op_images_dir, 'artifacts', 'pulled_images_report_verified.yaml')
+    | path_join
+  }}
+cifmw_env_op_images_crio_logs_dir: >-
+  {{ (cifmw_env_op_images_dir, 'artifacts', 'crio_logs') | path_join }}
+
 cifmw_env_op_images_pulled_report_namespaces:
   - "{{ cifmw_openstack_namespace | default('openstack') }}"
   - "{{ operator_namespace | default('openstack-operators') }}"
--- a/roles/env_op_images/library/verify_pulled_report_crio.py
+++ b/roles/env_op_images/library/verify_pulled_report_crio.py
@@ -1,0 +1,297 @@
+#!/usr/bin/python
+
+# Copyright Red Hat, Inc.
+# All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+DOCUMENTATION = r"""
+---
+module: verify_pulled_report_crio
+
+short_description: Enrich pulled_images_report with CRI-O pull evidence
+
+description:
+  - Reads the YAML produced by the env_op_images pulled-images report role task.
+  - "Parses CRI-O journal lines for C(msg=\"Pulled image: ...@sha256:...\")."
+  - Adds per-row verification fields using trusted mirror domains from
+    C(summary.mirror_rules).
+  - When images carry a C(node) field, evidence is matched against the
+    specific node's CRI-O log first.  If the digest is only found on a
+    different node the entry is counted as a cross-node mismatch.
+  - Log files are expected to follow the C(<node-name>.crio.log) naming
+    convention produced by the role task.
+
+options:
+  report_path:
+    description: Path to C(pulled_images_report.yaml) (input).
+    required: true
+    type: str
+  output_path:
+    description: Path for the enriched YAML report (output).
+    required: true
+    type: str
+  log_paths:
+    description:
+      - Explicit list of log files to parse (e.g. per-node CRI-O logs).
+      - Combined with files found under I(log_dir) when set.
+    required: false
+    type: list
+    elements: str
+    default: []
+  log_dir:
+    description:
+      - Directory containing CRI-O log files matching I(log_glob).
+    required: false
+    type: str
+  log_glob:
+    description: Glob under I(log_dir). Used only when I(log_dir) is set.
+    required: false
+    default: "*.crio.log"
+    type: str
+
+author:
+  - Nemanja Marjanovic (@nemarjan)
+
+notes:
+  - Requires PyYAML on the controller (same as other cifmw.general modules).
+"""
+
+EXAMPLES = r"""
+- name: Enrich pulled report using fetched node logs
+  cifmw.general.verify_pulled_report_crio:
+    report_path: "{{ cifmw_env_op_images_pulled_report_path }}"
+    log_dir: "{{ cifmw_env_op_images_crio_logs_dir }}"
+    output_path: "{{ cifmw_env_op_images_verified_report_path }}"
+"""
+
+RETURN = r"""
+changed:
+  description: Whether the output file was written.
+  type: bool
+  returned: always
+trusted_mirrors:
+  description: Hostnames extracted from mirror rules in the report summary.
+  type: list
+  elements: str
+  returned: always
+log_files:
+  description: Number of log files read.
+  type: int
+  returned: always
+entries_with_digest:
+  description: Image rows that had a sha256 digest in C(image_id).
+  type: int
+  returned: always
+cross_node_entries:
+  description: >-
+    Image rows where evidence was found only on a different node
+    than where the pod ran.
+  type: int
+  returned: always
+nodes_with_evidence:
+  description: >-
+    Node names that had at least one C(Pulled image) log entry.
+  type: list
+  elements: str
+  returned: always
+"""
+
+import glob
+import os
+import re
+
+import yaml
+from ansible.module_utils.basic import AnsibleModule
+
+LOG_PATTERN = re.compile(
+    r'msg="Pulled image: (?P<actual_uri>[^@\s]+)@(?P<id>sha256:[a-f0-9]+)"'
+)
+SHA256_PATTERN = re.compile(r"sha256:[a-f0-9]+")
+
+
+def _node_from_path(path):
+    """Derive node name from the ``<node>.crio.log`` naming convention."""
+    basename = os.path.basename(path)
+    suffix_pos = basename.find(".crio.log")
+    if suffix_pos > 0:
+        return basename[:suffix_pos]
+    return os.path.splitext(basename)[0]
+
+
+def _domain_from_uri(uri):
+    """Return the registry host (+ optional port) from an image URI."""
+    return uri.split("/")[0].strip()
+
+
+def _apply_evidence(img, actual_uri, evidence_node, trusted_mirrors):
+    """Set common verification fields on an image row that has log evidence."""
+    actual_domain = _domain_from_uri(actual_uri)
+    img["node_verified_image_origin"] = (
+        "mirror" if actual_domain in trusted_mirrors else "source"
+    )
+    img["log_evidence_uri"] = actual_uri
+    img["log_evidence_node"] = evidence_node
+    return actual_domain
+
+
+def _collect_log_evidence(paths, module):
+    """Parse CRI-O logs into per-node and global evidence dicts.
+
+    Returns:
+        per_node: ``{node_name: {sha256_digest: pull_uri}}``
+        global_evidence: ``{sha256_digest: (pull_uri, node_name)}``
+            (last writer wins across nodes for the global dict)
+    """
+    per_node = {}
+    global_evidence = {}
+    for path in paths:
+        node = _node_from_path(path)
+        node_ev = per_node.setdefault(node, {})
+        try:
+            with open(path, "r") as f:
+                for line in f:
+                    match = LOG_PATTERN.search(line)
+                    if match:
+                        digest = match.group("id")
+                        uri = match.group("actual_uri")
+                        node_ev[digest] = uri
+                        global_evidence[digest] = (uri, node)
+        except IOError as exc:
+            module.fail_json(
+                msg="Cannot read CRI-O log file {0}: {1}".format(path, str(exc))
+            )
+    return per_node, global_evidence
+
+
+def run_module():
+    module_args = dict(
+        report_path=dict(type="str", required=True),
+        output_path=dict(type="str", required=True),
+        log_paths=dict(type="list", required=False, elements="str", default=[]),
+        log_dir=dict(type="str", required=False),
+        log_glob=dict(type="str", required=False, default="*.crio.log"),
+    )
+
+    module = AnsibleModule(argument_spec=module_args, supports_check_mode=True)
+
+    report_path = module.params["report_path"]
+    output_path = module.params["output_path"]
+    log_paths = module.params["log_paths"] or []
+    log_dir = module.params["log_dir"]
+    log_glob = module.params["log_glob"]
+
+    paths = list(log_paths)
+    if log_dir:
+        paths.extend(sorted(glob.glob(os.path.join(log_dir, log_glob))))
+
+    if not paths:
+        module.fail_json(
+            msg="No CRI-O log files: set log_paths and/or log_dir with matching files."
+        )
+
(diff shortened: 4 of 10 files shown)


roles/env_op_images/tasks/verify_pulled_report_crio.yml
```diff
@@ -0,0 +1,114 @@
...
+          }}
+
+    - name: Fail when oc get nodes did not succeed
+      when: _verify_crio_nodes_json.rc != 0
+      ansible.builtin.fail:
+        msg: >-
+          oc get nodes failed (rc={{ _verify_crio_nodes_json.rc }}); cannot fetch CRI-O logs.
+
+    # Filename is sanitised to avoid path-traversal with unusual node names.
+    - name: Fetch CRI-O unit logs per node
```
evallesp: (blocking) suggestion: let's move this task.

One per take the result and another to write in the logs.

Use for the first one: ansible.builtin.command -> cmd: oc adm node-logs "{{ item }}" -u crio --since=-2h

(From previous command, we should think if 2h is enough.

Then following task should be using ansible.builtin.copy where dest is the log path, and origin is stdout from previous task registered variable.

roles/env_op_images/library/verify_pulled_report_crio.py
```diff
@@ -0,0 +1,298 @@
+#!/usr/bin/python
```
evallesp, line 1: (non-blocking) requirement: We need to add tests for this script.
You can check crawl_n_mask tests at: tests/unit/modules

I'm Ok if we prepare following work to address this (reason why it's marked as non-blocking) but probably we can check the output of a AI model.

roles/env_op_images/tasks/pulled_images_report.yml
```diff
@@ -0,0 +1,227 @@
...
+        kind: Pod
+        namespace: "{{ item }}"
+      register: _pulled_report_pods
+      loop: "{{ cifmw_env_op_images_pulled_report_namespaces | unique }}"
+      loop_control:
+        label: "{{ item }}"
+      failed_when: false
+
+    # Flatten all loop results, then one report row per container + initContainer.
+    - name: Build per-pod pulled images report
```
evallesp, line 114: (blocking) suggestion: This should go in a jinja2 temlate file under template folder in the role.

roles/env_op_images/tasks/pulled_images_report.yml
```diff
@@ -0,0 +1,227 @@
...
+          {# Default: host = first path segment of image ref; basis = source #}
+          {%     set match = namespace(host=image.split('/')[0], expected_pull_basis='source') %}
+          {# First mapping where image startswith source: use mirror registry host #}
+          {%     for m in _pulled_report_mirror_mappings if m.source is defined and m.source and image.startswith(m.source) %}
+          {%       if loop.first %}
+          {%         set match.host = m.mirror.split('/')[0] %}
+          {%         set match.expected_pull_basis = 'mirror' %}
+          {%       endif %}
+          {%     endfor %}
+          {%     set _ = entries.append({
```
evallesp: (non-blocking) suggestion: I'm unsure if we want these all values.

roles/env_op_images/library/verify_pulled_report_crio.py
```diff
@@ -0,0 +1 @@
+../../../plugins/modules/verify_pulled_report_crio.py
```
evallesp: this requires new line.

ci/playbooks/collect-logs.yml
```diff
@@ -44,7 +55,7 @@
                 chdir: "{{ ansible_user_dir }}/zuul-output/logs/ci-framework-data"
                 cmd: |
                   cp -ra {{ ansible_user_dir }}/ci-framework-data/logs . ;
-                  cp -ra {{ ansible_user_dir }}/ci-framework-data/artifacts . ;
+                  cp -ra {{ ansible_user_dir }}/ci-framework-data/artifacts . || true ;
```
evallesp: (blocking) question: Is this required?

roles/env_op_images/README.md
```diff
@@ -4,6 +4,9 @@ A role to gather the container images used in the openstack deployment with spec
 ## Parameters
 * `cifmw_env_op_images_dir`: (String) Directory where the operator_images.yaml will be stored. Defaults to `~/ci-framework-data/artifacts`
 * `cifmw_env_op_images_file`: (String) Name of the file storing the operator images and tags. Defaults to `operator_images.yaml`
```
evallesp, line 6: (blocking) suggestion: I think we should add proper doc about this new feature.

roles/env_op_images/tasks/pulled_images_report.yml
```diff
@@ -0,0 +1,227 @@
...
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  block:
+    - name: Ensure artifacts directory exists
+      ansible.builtin.file:
+        path: "{{ cifmw_env_op_images_dir }}/artifacts"
+        state: directory
+        mode: "0755"
+
+    # Legacy OpenShift mirror CRD; empty or error is OK (parsed as no items).
```
evallesp: (non-blocking) suggestion: I'd remove some of these comments. Just leave where tasks have some lofic like thsoe jinja templates.
Also I'd reword them to feel more natural.

roles/env_op_images/tasks/pulled_images_report.yml
```diff
@@ -0,0 +1,227 @@
...
+        _full_report:
+          summary: "{{ _pulled_report_summary }}"
+          images: "{{ _pulled_images_report }}"
+      ansible.builtin.copy:
+        dest: "{{ cifmw_env_op_images_pulled_report_path }}"
+        content: "{{ _full_report | to_nice_yaml }}"
+        mode: "0644"
+
+    # Console visibility: split rows by how ICSP/IDMS classification turned out.
+    - name: Images with expected_pull_basis source
```
evallesp: (non-blocking) question: Do we need these debugs statements?

roles/env_op_images/tasks/verify_pulled_report_crio.yml
```diff
@@ -0,0 +1,114 @@
...
+    # Non-fatal: some nodes may be unreachable (e.g. NotReady).
+    - name: Warn when node log fetch failed for a node
+      when: item.rc | default(0) != 0
+      ansible.builtin.debug:
+        msg: "oc adm node-logs failed for node (rc={{ item.rc | default('n/a') }}): {{ item.item | default('unknown') }}"
+      loop: "{{ _verify_crio_fetch.results | default([]) }}"
+      loop_control:
+        label: "{{ item.item | default('') }}"
+
+    - name: Find fetched CRI-O log files
```
evallesp: (non-blockign) suggestion: If we split the task into two, one for executing oc adm node-logs other for retrieving the logs, this might not be necessary anymore.

roles/env_op_images/tasks/pulled_images_report.yml
```diff
@@ -0,0 +1,227 @@
...
+    # Preferred / current mirror CRD alongside ICSP.
+    - name: Get IDMS mirror rules
+      ansible.builtin.command:
+        cmd: oc get imagedigestmirrorset -o json
+      register: _pulled_report_idms
+      failed_when: false
+
+    # One flat list of rules for templating: each mirror list entry becomes
+    # its own row so prefix matching can use the same loop for all mirrors.
+    - name: Build source-to-mirror mapping from ICSP/IDMS
```
evallesp, line 61: (blocking) suggestion: I think this comment has room for improvement. Maybe an example of the output?

evallesp (commented):
I like the way is getting. And functionally now is working as expected.
Check my suggestions. Feel free to close them specially those marked as "non-blocking"

roles/env_op_images/tasks/verify_pulled_report_crio.yml
nemarjan: Separated tasks and increased limit CRI-O logs to last 24 hours , i think it should be enough to demonstrate our logging.

roles/env_op_images/library/verify_pulled_report_crio.py
nemarjan: Great idea Enrique. I created a test script but i will add this after this is merged as I dont want to introduce more files as patch is quiet large as is.

roles/env_op_images/tasks/pulled_images_report.yml
nemarjan: Created a Template call -- the inline Jinja is now a one-liner lookup('template', 'pulled_images_report.j2').

ci/playbooks/collect-logs.yml
nemarjan: Not required , was something that was left from fixing bugs. Now removed.

roles/env_op_images/tasks/pulled_images_report.yml
nemarjan: Not required and now moved.

roles/env_op_images/tasks/verify_pulled_report_crio.yml
nemarjan: The find is still needed -- it gives us the file-size check for the empty-log warning and a single .matched gate for downstream tasks, which the copy register alone doesn't provide.

roles/env_op_images/tasks/pulled_images_report.yml
nemarjan: I have reduced the list of information we re providing: 
- container: barbican-api
  expected_pull_basis: source
  expected_pull_location: quay.io
  image: quay.io/podified-antelope-centos9/openstack-barbican-api:current-podified
  image_id: quay.io/podified-antelope-centos9/openstack-barbican-api@sha256:26f90c...
  namespace: openstack
  node: crc
  pod: barbican-api-5d47bf4d-86rwj
  node_verified_image_origin: quay.io
  log_evidence_uri: quay.io/podified-antelope-centos9/openstack-barbican-api@sha256:26f90c...
  log_evidence_node: crc

roles/env_op_images/library/verify_pulled_report_crio.py
nemarjan: Added a new line to address issue.

nemarjan:
recheck

nemarjan:
uni02beta-base test project completed successfully: https://sf.apps.int.gpc.ocp-hub.prod.psi.redhat.com/zuul/t/components-integration/buildset/03bf01aee72c4265b6cac510c7b83638

Logs consist off two files stored in ci-framework-data/artifacts:
- pulled_images_report.yaml ( Expected Mirror rules on cluster)
- pulled_images_report_verified.yaml ( Verified by crio logs)

Example of verified logs:

container: barbican-api  
      expected_pull_basis: source  
      expected_pull_location: images.paas.redhat.com  
      image: images.paas.redhat.com/podified-rhos18-rhel9/openstack-barbican-  api:c9e20839baeb524ec5fd4883df7e6fda  
      image_id: images.paas.redhat.com/podified-rhos18-rhel9/openstack-barbican-api@sha256:b81bee9dfec5701498f28451b44bed755e077dc3af28fcf179770eabab1e0279  
      namespace: openstack  
      node: master-2  
      pod: barbican-api-85f46556b4-d2rmr  
      node_verified_image_origin: source  
      log_evidence_uri: images.paas.redhat.com/podified-rhos18-rhel9/openstack-barbican-api  
      log_evidence_node: master-2

container: bootstrap-edpm-deployment-openstack-edpm
  expected_pull_basis: mirror
  expected_pull_location: registry.stage.redhat.io
  image: registry.redhat.io/rhoso-operators/ee-openstack-ansible-ee-rhel9@sha256:700547e85197e2cfab167dfbd60be6082487eaf8f571ab7aa94588c1074842c9
  image_id: registry.redhat.io/rhoso-operators/ee-openstack-ansible-ee-rhel9@sha256:700547e85197e2cfab167dfbd60be6082487eaf8f571ab7aa94588c1074842c9
  namespace: openstack
  node: master-0
  pod: bootstrap-edpm-deployment-openstack-edpm-8xq4n
  node_verified_image_origin: source
  log_evidence_uri: registry.redhat.io/rhoso-operators/ee-openstack-ansible-ee-rhel9
  log_evidence_node: master-0

nemarjan:
recheck

roles/env_op_images/tasks/pulled_images_report.yml
```diff
@@ -0,0 +1,155 @@
...
+    # mirror in one loop.
+    # Example output (_pulled_report_mirror_mappings):
+    #   - source: registry.redhat.io/openstack-k8s-operators
+    #     mirror: quay.example.com/rh-osbs/openstack-k8s-operators
+    #   - source: registry.redhat.io/openstack-k8s-operators
+    #     mirror: internal-registry.corp.net/openstack-k8s-operators
+    - name: Build source-to-mirror mapping from ICSP/IDMS
+      vars:
+        _icsp_items: >-
+          {{ (_pulled_report_icsp.stdout | default('{}') | from_json).get('items', []) }}
```
Valkyrie00: Question, Does the `from_json` crash if the ICSP or IDMS CRD doesn't exist on the cluster?
Investigate if necessary replace with something like this: ```{{ (_pulled_report_icsp.stdout | default('{}', true) | from_json).get('items', []) }}```

roles/env_op_images/tasks/verify_pulled_report_crio.yml
```diff
@@ -0,0 +1,112 @@
...
+                  {{
+                    (_verify_crio_nodes_json.stdout | from_json).get('items', [])
+                    | map(attribute='metadata.name') | list
+                  }}
+
+            - name: Fetch CRI-O unit logs per node
+              ansible.builtin.command:
+                cmd: >-
+                  oc adm node-logs "{{ item }}" -u crio
+                  --since=-4h
```
Valkyrie00: Is it enough?

roles/env_op_images/templates/pulled_images_report.j2
```diff
@@ -0,0 +1,40 @@
...
+{%   endfor %}
+{%   for cs in pod.status.initContainerStatuses | default([]) %}
+{%     set _ = image_statuses.append({'cs': cs, 'type': 'init_container'}) %}
+{%   endfor %}
+{%   for entry in image_statuses %}
+{%     set cs = entry.cs %}
+{%     set image = cs.image | default('unknown') %}
+{%     set image_id = cs.imageID | default('') %}
+{%     set match = namespace(host=image.split('/')[0], expected_pull_basis='source') %}
+{%     for m in _pulled_report_mirror_mappings if m.source is defined and m.source and image.startswith(m.source) %}
```
Valkyrie00, line 22: nit question, I see the matching logic uses `if loop.first` to pick the first rule where `image.startswith(m.source)` is true.

What happens if there are overlapping source prefixes in the ICSP/IDMS rules? For example:
```
- source: registry.redhat.io -> mirror: mirror-generic.com
- source: registry.redhat.io/openstack  -> mirror: mirror-openstack.com
```

For an image like `registry.redhat.io/openstack/nova:latest`, both rules would match via startswith, but the result depends on which one appears first in the flattened list. CRI-O uses longest-prefix-match for this.
Is this a scenario that can actually occur in our clusters? I also don't know if this is possibile.

Valkyrie00:
First look: apart from a few minor comments, the implementation looks great! As we discussed privately, I’m not entirely agree with the effort-benefit trade-off, but that’s just my non-technical take. Your code is correct. GJ @nemarjan ! Once you fix the checks and comments, I'll approve it. 👍

roles/env_op_images/defaults/main.yml
```diff
@@ -21,3 +21,23 @@
 cifmw_env_op_images_dir: "{{ cifmw_basedir }}"
 cifmw_env_op_images_file: operator_images.yaml
 cifmw_env_op_images_dryrun: false
+
+cifmw_env_op_images_pulled_report_file: pulled_images_report.yaml
```
danpawlik: TBH adding dedicated var just for static file name... Maybe we can live without var and just merge into the path.

roles/env_op_images/defaults/main.yml
```diff
@@ -21,3 +21,23 @@
...
+cifmw_env_op_images_pulled_report_path: >-
+  {{
+    (cifmw_env_op_images_dir, 'artifacts', cifmw_env_op_images_pulled_report_file)
+    | path_join
+  }}
+
+cifmw_env_op_images_verified_report_file: pulled_images_report_verified.yaml
```
danpawlik: similar here

roles/env_op_images/library/verify_pulled_report_crio.py
```diff
@@ -0,0 +1,285 @@
+#!/usr/bin/python
+
+# Copyright: (c) 2026, Nemanja Marjanovic <nemarjan@redhat.com>
```
danpawlik: If you add copyright header, maybe worth to add also Red Hat header. Thanks

roles/env_op_images/library/verify_pulled_report_crio.py
```diff
@@ -0,0 +1,285 @@
...
+
+    if not isinstance(data, dict):
+        module.fail_json(msg="Report root must be a mapping (dict).")
+
+    trusted_mirrors = set()
+    summary_section = data.get("summary") or {}
+    for rule in summary_section.get("mirror_rules") or []:
+        if not isinstance(rule, dict):
+            continue
+        mirror_url = rule.get("mirror") or ""
```
danpawlik: nit: `or ""` is not needed. If `rule.get("mirror")` does not exists, `mirror_url` would be `None` so no need to set "".  Later condition will work with both: `""` and `None`.

roles/env_op_images/tasks/pulled_images_report.yml
danpawlik: nit: single quotes are not needed here `default('{}')`. Don't think jinja2 expect string `{}`

roles/env_op_images/tasks/pulled_images_report.yml
```diff
@@ -0,0 +1,155 @@
...
+      register: _pulled_report_pods
+      loop: "{{ cifmw_env_op_images_pulled_report_namespaces | unique }}"
+      loop_control:
+        label: "{{ item }}"
+      failed_when: false
+
+    - name: Build per-pod pulled images report
+      ansible.builtin.set_fact:
+        _pulled_images_report: >-
+          {{ lookup('template', 'pulled_images_report.j2') | trim | from_yaml }}
```
danpawlik, line 117: nit: construction is fine, but more safe would be make a template, then read it and set as fact. Just better for future debugging.

roles/env_op_images/tasks/verify_pulled_report_crio.yml
```diff
@@ -0,0 +1,112 @@
...
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  block:
+    - name: Check pulled images report exists
+      ansible.builtin.stat:
+        path: "{{ cifmw_env_op_images_pulled_report_path }}"
+      register: _verify_crio_pulled_stat
+
+    - name: Collect CRI-O logs and verify
+      when: _verify_crio_pulled_stat.stat.exists | bool
```
danpawlik: nit: no need to dump into bool, because the stat module needs to return into bool.

danpawlik (dismissed):
LGTM

roles/env_op_images/templates/pulled_images_report.j2
nemarjan: Good observation Vito. This template only produces a policy-level estimate (expected_pull_basis) — so yes, if two rules overlap, startswith will pick the first match. But in practice ICSP/IDMS rules use full repository paths (e.g. registry.redhat.io/openstack-k8s-operators/heat-api-rhel9), so overlapping prefixes don't really occur. Either way, the actual verification happens downstream in verify_pulled_report_crio, which checks CRI-O logs directly and doesn't depend on prefix ordering at all.

roles/env_op_images/tasks/verify_pulled_report_crio.yml
nemarjan: I updated to 24h, i was just thinking that as the report is created on setup we wouldnt need to go back a significant amount of time to catch our logs for crio.

roles/env_op_images/tasks/pulled_images_report.yml
nemarjan: Great catch Vito, will implement a fix for this. I will add  true to the default filters so from_json won't crash on an empty stdout when the ICSP or IDMS CRD doesn't exist on the cluster.

roles/env_op_images/README.md
nemarjan: Where do you think this doc should go?

roles/env_op_images/defaults/main.yml
nemarjan: Thanks Daniel, I agree, will implement.

roles/env_op_images/library/verify_pulled_report_crio.py
nemarjan: Done. Thanks.

roles/env_op_images/library/verify_pulled_report_crio.py
nemarjan: Makes sense, good catch, changing to: mirror_url = rule.get("mirror")

roles/env_op_images/tasks/pulled_images_report.yml
nemarjan: I think the from_json filter expects a string input to parse. If we pass a bare {} (a Jinja2 dict literal), from_json would receive an actual dict object rather than a JSON string, and it would fail. The '{}' provides a valid JSON string that from_json can parse.

roles/env_op_images/tasks/verify_pulled_report_crio.yml
nemarjan: Removed. Thanks

nemarjan:
recheck

danpawlik (approved):

Valkyrie00 (approved):

roles/env_op_images/tasks/pulled_images_report.yml
nemarjan: Reduced number of comments. Thanks for suggestion.

roles/env_op_images/README.md
evallesp: (non-blocking) I'd say here in this same file. (We can do in following MR)

evallesp:
LGTM

evallesp (commented):
/lgtm

evallesp (approved):
/lgtm

evallesp (commented):
/approve

PR #3831, opened 2026-04-06
[build_openstack_packages] Removing whitebox-n-t-p and tempest from branchless dict

evallesp:
So we want them to be actually checkout to antelope code instead
of master


evallesp:
recheck

danpawlik (approved):

bshewale (approved):

Valkyrie00 (approved):

karelyatin:
tempest and whitebox both are branchless projects so those needs to be kept in branchless dict. This needs to be reverted

PR #3838, opened 2026-04-08
[multi-namespace] Skip container image update when primary NS has no overrides

lmiccini:
The ns2_update_containers playbook unconditionally applies custom container images to the openstack2 namespace using CI parameter variables (registry, org, tag). However, when the primary openstack namespace does not have customContainerImages set on its OpenStackVersion CR, those CI parameters produce incorrect images that override the correct operator defaults.

Add a pre-check that reads the primary namespace's OpenStackVersion customContainerImages. If empty, skip the update_containers role so that openstack2 uses the same operator defaults as openstack.


abays (approved):
/lgtm

michburk (approved):
/lgtm
/approve

PR #3843, opened 2026-04-12
[SKMO] Grant swiftoperator role to cinder for backup

lkuchlan:
Add swiftoperator role assignment for cinder service user to enable Cinder Backup with Swift backend in SKMO deployments.

The role is granted in the prepare-leaf hook after Keystone and openstackclient are ready, ensuring the cinder service can create backup containers in Swift.


evallesp:
Hi, Could you push a single commit for this? AFAIK the relevant content is in the last one, and seems this is getting conflicts with CI.

PR #3846, opened 2026-04-13
[ci] Add AGENTS.md and CLAUDE.md for AI agent guidance

rebtoor:
Add structured documentation for AI coding agents working on this
repository. AGENTS.md provides repo-wide context: variable naming
rules, generated file warnings, testing commands, commit conventions,
and repository layout. CLAUDE.md adds Claude-specific behavioral
guidance and references AGENTS.md as the primary source of truth.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>


evallesp (dismissed):
/lgtm

michburk:
After doing a bit of testing, it looks like claude code doesn't read AGENTS.md files at the start of a new session. I know quite a few people use claude code, so should we include a CLAUDE.md that is a symlink to AGENTS.md? Or maybe in CLAUDE.md just write something like "if you haven't already, read AGENTS.md" in case some tools load both AGENTS.md and CLAUDE.md? I'm not really sure if there is already some existing best approach for dealing with this.

Aside from that, a file like this is very useful, so thanks for making this!

rebtoor:
> After doing a bit of testing, it looks like claude code doesn't read AGENTS.md files at the start of a new session. I know quite a few people use claude code, so should we include a CLAUDE.md that is a symlink to AGENTS.md? Or maybe in CLAUDE.md just write something like "if you haven't already, read AGENTS.md" in case some tools load both AGENTS.md and CLAUDE.md? I'm not really sure if there is already some existing best approach for dealing with this.
> 
> Aside from that, a file like this is very useful, so thanks for making this!

It's a great idea! Let me address this issue then.

AGENTS.md
```diff
@@ -0,0 +1,221 @@
...
+`hooks/playbooks/`, `ci/`, `scripts/`, `docs/`) are safe to edit following the
+conventions in this file.
+
+### Debugging patterns
+
+Use `block`/`rescue` for complex task sequences. Dump relevant variables in
+the `rescue` block, then `ansible.builtin.fail` to stop execution. This
+makes CI failures much easier to diagnose.
+
+## Playbook numbering
```
danpawlik: is it not playbooks that would be removed? Was it generated by AI?
If those file you want to add, its better to say "do not watch on playbooks" than "Top-level playbooks"

danpawlik:
was it done by AI? I think that should be mentioned.

rebtoor:
> was it done by AI? I think that should be mentioned.

Partly, but yes, i missed the "co-authored-by". Fixing rn.

danpawlik (dismissed):

CLAUDE.md
```diff
@@ -0,0 +1,25 @@
...
+2. `TEST_SINGLE_ROLE=<role> make molecule` — targeted Molecule test.
+3. `make ansible_test` — plugin unit/sanity/integration tests (only if plugins changed).
+4. `make docs` — only if documentation was modified.
+
+### macOS limitations
+
+The Makefile test targets (`make tests`, `make molecule`, `make ansible_test`,
+`make setup_tests`, `make setup_molecule`, and their `_nodeps` variants) **do not
+work on macOS**. The underlying scripts use `readlink -f`, which is not available
+on macOS. These targets are designed for Linux CI environments only.
```
michburk: [non-blocking] question: Is this info claude-code specific? Or could it also live in the agents.md file?

michburk:
The `@AGENTS.md` syntax seems to work nicely by the way, thanks!

CLAUDE.md
rebtoor: It something that I need to think about. 

I mean, the obvious answer is: "it belongs to AGENTS.md since it's the main file for agentic guideline" but i want to test how other agents react to those instructions.

michburk:
recheck

michburk (dismissed):
/lgtm

AGENTS.md
```diff
@@ -0,0 +1,269 @@
...
+- **ansible-test** for unit, sanity, and integration tests on plugins.
+- **Sphinx** for documentation (hosted on ReadTheDocs).
+- **Zuul**, **GitHub Actions**, and **Prow** for CI.
+
+## Repository layout
+
+| Path | Description |
+|---|---|
+| `roles/` | Ansible roles. Each has `defaults/`, `tasks/`, `molecule/`, `README.md`. |
+| `playbooks/` | End-to-end flows. Numbered stages (`01-bootstrap.yml` ... `99-logs.yml`), plus subdirectories for adoption, ceph, etc. |
```
danpawlik: I guess that part can be updated and remove `01-bootstrap.yml` ... `99-logs.yml`

AGENTS.md
```diff
@@ -0,0 +1,269 @@
...
+| `containerfiles/` | Podman images for CI (`Containerfile.ci`, `Containerfile.tests`). |
+| `group_vars/` | Shared group variables (e.g., `all.yml`). Changes here affect every playbook run. |
+| `zuul.d/` | Zuul job and project definitions. **Some files are generated -- see below.** |
+| `_skeleton_role_/` | Template used by `ansible-galaxy role init` when creating new roles. |
+
+## Critical rules
+
+### Variable naming
+
+All Ansible role variables **must** match the pattern `^cifmw_[a-z_][a-z0-9_]*$`.
```
danpawlik: nit: 
```
All Ansible role variables **must** match the pattern `^cifmw_[a-z_][a-z0-9_]*$` where after `cifmw_` prefix, there is a role name then variable name
```

AGENTS.md
```diff
@@ -0,0 +1,269 @@
...
+| `custom/` | Gitignored. Local-only overrides, never committed. |
+| `hooks/playbooks/roles/` | Excluded from ansible-lint. Owned by hook authors. |
+
+All other paths (`roles/`, `playbooks/`, `plugins/`, `group_vars/`, `scenarios/`,
+`hooks/playbooks/`, `ci/`, `scripts/`, `docs/`) are safe to edit following the
+conventions in this file.
+
+### Debugging patterns
+
+Use `block`/`rescue` for complex task sequences. Dump relevant variables in
```
danpawlik: nit:
```
Use `block`/`rescue`/`all`
```

AGENTS.md
```diff
@@ -0,0 +1,269 @@
...
+the `rescue` block, then `ansible.builtin.fail` to stop execution. This
+makes CI failures much easier to diagnose.
+
```
danpawlik, line 104: I suggest to add rule like:
```
### Don't be too verbose

There is no need to provide explanation for each variable in each task file. It is enough when the variable description is available in role README file.
Add comment to the complex code.
```

another:
```
### Do not create too much variables

Balance amount of the variables: there is no need to create additional variables especially for static files that later would be used by other module or role. In that case, less variables means more
```

and:
```
### Make tasks easier to debug in potential issue

The tasks does not need to make many things in single step, that in the future on failure would be difficult to debug. In some cases, adding more tasks that are fast to execute and would be easier to debug would be better.
Of course don't add too much debug messages - add it just in very complex places
```

AGENTS.md
```diff
@@ -0,0 +1,269 @@
...
+### Excluded from linting
+
+ansible-lint skips: `.github/`, `scripts/`, `docs/`, `containerfiles/`, `ci/`,
+and the generated Zuul files (`zuul.d/projects.yaml`, `zuul.d/molecule.yaml`).
+
+## Commit conventions
+
+- **Title**: Must begin with the role name in brackets or parentheses:
+  `[my_role] Add feature X` or `(my_role) Fix bug Y`.
+  For cross-cutting changes use a category: `[ci]`, `[docs]`, `[Feature]`.
```
danpawlik, line 201: that would require to improve our script for parsing commit message... Maybe let's just keep: 
```
*Title**: Must begin with the role name in brackets or parentheses:
  `[my_role] Add feature X` or `(my_role) Fix bug Y`. If changes were done in multiple roles, the title should start with `[multiple]` or `(multiple)`
```

AGENTS.md
```diff
@@ -0,0 +1,269 @@
...
+To keep a clean git history, prefer a single commit per feature or fix:
+
+1. Create the initial commit normally.
+2. For subsequent changes on the same branch, amend the existing commit
+   (`git commit --amend`) instead of creating new ones.
+3. After amending, use `git push --force` to update the remote branch.
+
+Never push directly to `main` — it is a protected branch. Always work on
+a feature branch. Force pushing is only appropriate for **solo feature
+branches**, never for `main` or shared branches.
```
danpawlik, line 227: nit: suggest to add also a rule:
```
If change is not directly related to the main goal of the pull request, but without it, the pull request would not work, consider to add it as a separate commit.
Remember, that if you need to change something that is in the commit that belongs to the pull request, edit that commit instead of create new commit. Be careful to edit just the commits that belongs to the same pull request.
```

AGENTS.md
rebtoor: done!

AGENTS.md
rebtoor: done!

AGENTS.md
rebtoor: done!

AGENTS.md
rebtoor: done!

AGENTS.md
rebtoor: done!

AGENTS.md
rebtoor: done!

michburk (approved):
/lgtm

michburk:
recheck

michburk:
recheck

rebtoor:
recheck

rebtoor:
Folks, I'm skipping the github checks since we're adding 2 markdown files here :)

PR #3845, opened 2026-04-13
ipa: fix kube-rbac-proxy image pull failure

xek:
## Summary

Fixes the LDAP CI job failure in the keystone component pipeline.

- **Job:** `component-keystone-edpm-rhel9-rhoso18.0-crc-ldap`
- **Error:** `ErrImagePull` / `ImagePullBackOff` for `gcr.io/kubebuilder/kube-rbac-proxy:v0.5.0`
- **Root cause:** The `freeipa-operator` Deployment includes a `kube-rbac-proxy` sidecar referencing `gcr.io/kubebuilder/kube-rbac-proxy:v0.5.0` which is no longer available (Google Container Registry deprecation)
- **Fix:** Override the sidecar image to `registry.k8s.io/kubebuilder/kube-rbac-proxy:v0.16.0` via `kustomize edit set image` during IPA operator installation

## Test plan

- [x] `component-keystone-edpm-rhel9-rhoso18.0-crc-ldap` passes (verified 2x on [testproject MR !2210](https://gitlab.cee.redhat.com/ci-framework/ci-framework-testproject/-/merge_requests/2210))

## Related

- [OSPCIX-1321](https://redhat.atlassian.net/browse/OSPCIX-1321)
- [OSPRH-28983](https://redhat.atlassian.net/browse/OSPRH-28983)
- Federation fix (separate): #3847

Assisted-by: Claude Opus 4 (Anthropic)

PR #3847, opened 2026-04-13
federation: split controlplane hook into pre- and post-deploy

vakwetu:
Recent changes were made to hook_controlplane_config.yml to accomodate adding federation to the SKMO deployment.  This deployment differs though, in that the control plane already exists in the SKMO case.  The changes broke the existing component job.

The cases are sufficiently different that its just cleaner to just use two different playbooks.  That way, we do not have to worry about getting different conditionals right.  This PR just does this - reverting the existing playbook to its original state (for the component job) and adding a new one for the SKMO case.

split into two task files:
  - hook_controlplane_config.yml  (pre-deploy / component pipeline, unchanged)
  - hook_controlplane_config_postdeploy.yml  (post-deploy / SKMO pipeline)

Add a new playbook federation-controlplane-config-postdeploy.yml that wraps the new post-deploy task file.

The architecture/automation/vars/multi-namespace-skmo.yaml is updated separately to call the new post-deploy playbook.

Fix variable names in both task files to use the required cifmw_ prefix so they pass the var-naming[pattern] rule enforced by ansible-lint.

Related-To: [OSPCIX-1321](https://redhat.atlassian.net/browse/OSPCIX-1321)

### DIFF
--- a/roles/federation/tasks/hook_controlplane_config.yml
+++ b/roles/federation/tasks/hook_controlplane_config.yml
@@ -13,121 +13,26 @@
 # WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
 # License for the specific language governing permissions and limitations
 # under the License.
-
-# ---------------------------------------------------------------------------
-# Step 1 - read the Keycloak CA cert written by federation-pre-deploy
-# ---------------------------------------------------------------------------
-- name: Get ingress operator CA cert
-  ansible.builtin.slurp:
-    src: "{{ [ansible_user_dir, 'ci-framework-data', 'tmp', 'ingress-operator-ca.crt'] | path_join }}"
-  register: federation_sso_ca
-
-# ---------------------------------------------------------------------------
-# Step 2 - read the live OSCP to determine where the CA bundle lives.
 #
-# Priority for the secret name:
-#   1. spec.tls.caBundleSecretName already set on the OSCP  (use it as-is).
-#   2. cifmw_custom_ca_certs_secret_name variable (if set by caller).
-#   3. Hard default: "custom-ca-certs".
+# Pre-deploy hook for the component / CRC pipeline.
 #
-# This makes the hook self-healing: it does not rely on the kustomize having
-# correctly propagated caBundleSecretName, and it works on fresh installs
-# where the secret does not yet exist.
-# ---------------------------------------------------------------------------
-- name: Read current OpenStackControlPlane state
-  kubernetes.core.k8s_info:
-    kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
-    api_version: core.openstack.org/v1beta1
-    kind: OpenStackControlPlane
-    name: controlplane
-    namespace: "{{ cifmw_federation_run_osp_cmd_namespace }}"
-  register: _federation_oscp_info
+# Called before kustomize_deploy brings up the OpenStackControlPlane.
+# Creates the required secrets and writes a kustomization file that
+# kustomize_deploy will apply when it deploys the control plane.
+# The OSCP does not exist yet at this point; no direct OSCP patching
+# is performed here.
+#
+# For the post-deploy (architecture / integration / SKMO) pipeline where
+# the OSCP is already running, use hook_controlplane_config_postdeploy.yml
+# instead.
 
-- name: Resolve CA bundle secret name and check if OSCP already references one
-  ansible.builtin.set_fact:
-    _federation_ca_bundle_secret_name: >-
-      {{
-        ((_federation_oscp_info.resources | first).spec.tls | default({})).caBundleSecretName
-        | default(cifmw_custom_ca_certs_secret_name, true)
-        | default('custom-ca-certs', true)
-      }}
-    _federation_oscp_has_ca_bundle: >-
-      {{
-        (
-          ((_federation_oscp_info.resources | first).spec.tls | default({})).caBundleSecretName
-          | default('')
-        ) | length > 0
-      }}
-
-# ---------------------------------------------------------------------------
-# Step 3 - preserve any keys already in the target secret
-# ---------------------------------------------------------------------------
-- name: Look up existing CA bundle secret
-  kubernetes.core.k8s_info:
-    kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
-    api_version: v1
-    kind: Secret
-    name: "{{ _federation_ca_bundle_secret_name }}"
-    namespace: "{{ cifmw_federation_run_osp_cmd_namespace }}"
-  register: _federation_existing_ca_bundle
-
-- name: Capture existing CA bundle secret data
-  ansible.builtin.set_fact:
-    _federation_ca_bundle_existing_data: >-
-      {{
-        (_federation_existing_ca_bundle.resources | first).data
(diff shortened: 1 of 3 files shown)


xek (commented):
Nice split — separating the pre-deploy and post-deploy cases makes this much cleaner.

One concern: `hook_controlplane_config_postdeploy.yml` Step 5 still uses `kubernetes.core.k8s state: patched` for the `OpenStackControlPlane` CR. This is the same pattern that was hitting the `"Failed to find exact match for core.openstack.org/v1beta1.OpenStackControlPlane"` bug in the component pipeline. The SKMO/architecture pipeline may run a different kubernetes Python client version so it might work there, but it's the same underlying resource-discovery issue ([kubernetes-client/python#1536](https://github.com/kubernetes-client/python/issues/1536), [kubernetes.core#553](https://github.com/ansible-collections/kubernetes.core/issues/553) — both wontfix upstream).

If the SKMO pipeline is hitting the same error, the fix would be to replace Step 5 with `oc patch` using the **plural** resource name:

```yaml
- name: Patch OpenStackControlPlane to set caBundleSecretName (when unset)
  when: not cifmw_federation_oscp_has_ca_bundle | bool
  ansible.builtin.command:
    cmd: >-
      oc patch openstackcontrolplanes.core.openstack.org controlplane
      -n {{ cifmw_federation_run_osp_cmd_namespace }}
      --type merge
      -p '{"spec":{"tls":{"caBundleSecretName":"{{ cifmw_federation_ca_bundle_secret_name }}"}}}'
  environment:
    PATH: "{{ cifmw_path }}"
  changed_when: true
```

Key detail: `oc patch` needs the **plural** name `openstackcontrolplanes` (not singular `openstackcontrolplane`). The CRD defines `plural: openstackcontrolplanes`, `singular: openstackcontrolplane`, `shortNames: [osctlplane, oscp]`.

Also — the postdeploy file is missing a Step 7 that patches the OSCP with the Keystone federation config (httpdCustomization, customServiceConfig). The original had this as Step 8. Without it, the SKMO pipeline won't actually configure Keystone for OIDC.

For the pre-deploy file: this looks correct — the component pipeline doesn't need to patch the OSCP directly since kustomize_deploy handles it. 

The pre-deploy fix should unblock OSPCIX-1321 for the component pipeline. Our PR #3845 has the same fix for the pre-deploy path (plus the IPA kube-rbac-proxy image fix for the LDAP job) — happy to close ours in favor of this one if you want, or we can keep the IPA fix separate.

xek (commented):
Correction on my earlier comment — you're right that the root cause is the execution flow, not the kubernetes.core.k8s resource discovery or plural/singular naming.

In the component pipeline the hook runs at `pre_deploy` stage. Before 398d55b0, this only wrote kustomization files and secrets — `kustomize_deploy` would apply them later when creating the controlplane. After 398d55b0, Steps 5 and 8 tried to directly patch the OSCP, but the CRD isn't even registered with the API server at `pre_deploy` time. The `k8s_info` returns `ok` with an empty resource list, `_federation_oscp_has_ca_bundle` resolves to `False`, and the patch task fires against a non-existent resource type.

The split into pre-deploy (kustomization-only) and post-deploy (direct patching for SKMO) is the right fix. `state: patched` in the postdeploy file should work fine since in the SKMO pipeline the OSCP will actually be running by then.

LGTM for the pre-deploy revert. Our PR #3845 can be narrowed down to just the IPA kube-rbac-proxy image fix (separate issue from the federation breakage).

vakwetu:
tested - this woks both for SKMO and the keystone component jobs.

roles/federation/tasks/hook_controlplane_config.yml
```diff
@@ -13,121 +13,20 @@
...
-          caBundleSecretName: "{{ _federation_ca_bundle_secret_name }}"
-
-# ---------------------------------------------------------------------------
-# Step 6 - kustomization for CRC/devscripts flow (not consumed by kustomize_deploy)
-# ---------------------------------------------------------------------------
-- name: Ensure kustomization controlplane directory exists
-  ansible.builtin.file:
-    path: "{{ cifmw_manifests_dir }}/kustomizations/controlplane"
-    state: directory
-    mode: "0755"
```
abays, line 128: Without this, could the new L31 possibly fail?  i.e. if something else didn't create the `{{ cifmw_manifests_dir }}/kustomizations/controlplane` directory first?

roles/federation/tasks/hook_controlplane_config.yml
```diff
@@ -141,71 +40,58 @@
...
   kubernetes.core.k8s:
     kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
     state: present
     definition:
       apiVersion: v1
       kind: Secret
+      type: Opaque
       metadata:
-        name: keystone-httpd-override
+        name: keycloakca
```
evallesp, line 86: (non-blocking) question: Should this match keycloak-ca.crt ?
Here it creates: data: {KeyCloakCA: <cert>}
And in the post it merges at:
  data: >-                                                                                                                                                                                                          
    {{                                                                                                                                                                                                              
      cifmw_federation_ca_bundle_existing_data |
      combine({'keycloak-ca.crt': cifmw_federation_sso_ca.content})
    }}

 so: data: {KeyCloakCA: <cert>, keycloak-ca.crt: <cert>} ?

evallesp (commented):
I have some problems reviewing this.

I see that we have mismatch cert name inside the secret.
ALso the way the secret name in the post is calculated, is make me hard to find if it boths run in the same pipeline the post adapts to the name created in the pre (by reading OSCP).
But if both are running in different pipelines, then it has different naming but we don't mind.


I think the certificate name inside the secret should be the same to avoid duplication, but I might be wrong.

roles/federation/tasks/hook_controlplane_config.yml
vakwetu: No, it doesn't have to match.  The way this works is you define a secret that contains all your custom ca certs.  In this case, that secret would be keycloakca. 

libcommon will read that secret and iterate over all the fields and extract the contents and add those to a chain of ca certs.  The field names do not matter at all.

roles/federation/tasks/hook_controlplane_config.yml
vakwetu: Possibly.  I just basically reverted to what was there before, so that I was sure it would work (and it does).  But I can add this stanza to make things more robust.

roles/federation/tasks/hook_controlplane_config.yml
vakwetu: @evallesp  - the two playbooks run in different pipelines.  One runs in the keystone federation component job and the other runs in the SKMO uni-job.  There would be no reason to run this twice.

The pre-config job  just reverts to what was there before so that we can be sure that the pre-existing component job just runs as before.

The post-config job is the one with newer, more robust code - as it needs to be because the control plane already exists for this case.

roles/federation/tasks/hook_controlplane_config.yml
vakwetu: fixed in latest patch @abays

danpawlik (approved):

michburk (approved):
/lgtm
/approve

PR #3850, opened 2026-04-14
[dlrn_report] Set crypto policies when FIPS is enabled

danpawlik:
Since some CI jobs are running using the PreMetal tool, and we move the job execution from controller node to hypervisor directly, the crypto policies are different.
For example, when the task was executed on controller:

    update-crypto-policies --show
    DEFAULT

Where, in some case, where uni job require FIPS, output is:

    update-crypto-policies --show
    FIPS

Set the AD-SUPPORT policy when just FIPS is enabled.


morucci (approved):

evallesp (approved):
/approve

PR #3852, opened 2026-04-14
Drop labels dedicated to cloud provider

danpawlik:
When IBM private cloud would be removed, there is no need to keep
nodesets that would force spawning CI jobs on specific cloud
provider.

After merging: https://github.com/openstack-k8s-operators/ci-framework/pull/3851

Depends-On: https://github.com/openstack-k8s-operators/watcher-operator/pull/363


danpawlik:
recheck

Valkyrie00 (approved):

PR #3698, opened 2026-02-19
[test_operator] Expose ansibleGitBranch for ansibletest via cifmw variable

gmarcian:
This change adds the ability to set ansibleGitBranch in the test_operator config via the cifmw_test_operator_ansibletest_ansible_git_branch variable.

Signed-off-by: Julia Marciano jmarcian@redhat.com

PR #3855, opened 2026-04-15
[test_operator] Fix sync README script

kstrenkova:
Empty lists/dicts now display inline (e.g., []) instead of redirecting to defaults file. Removed line number tracking and anchors - links now point to main.yml file without #L123 fragments for simpler navigation. This change was also done, so that when adding a new parameter, users don't have to run the sync script first to apply changes to the line numbers.


kstrenkova:
Hello @michburk :wave: I have created a change that removes the line numbers from the link and hopefully also removes all the dead code created by such change. Additionally, with my new parameter I noticed that even if the default is an empty array `[]`, the script would put a link there instead. So I fixed that as well, as you can see in the new README that was generated by running the script.

michburk:
Looks good @kstrenkova thank you! Before we merge this, do you think we should merge https://github.com/openstack-k8s-operators/ci-framework/pull/3698, then rebase this on top + re-run the updated sync script in this pr? 3689 has been open for a good while and I don't want to delay it further with another snag because of more readme sync changes :sweat_smile:

kstrenkova:
@michburk I approved the other MR in the morning, rebased this one atop it and I think this one is now ready to get merged :D

michburk (approved):
/lgtm

danpawlik (approved):

PR #3857, opened 2026-04-15
[os_must_gather] Add exit rc to os-must-gather main task

evallesp:
With this we allow to run the rescue block. Previously any timeout
in the main task would return the code from the echo which was
always success.

Also, we're removing from the rescue block the always block.
And we've create a always section for finding existing os-must-gather
directories and the symlink creation.

Also we've changed the dest-dir for the generic fallback command
to match the same folder of the symlink.

Removed oc inspect as we're not getting so many errors from oc adm must-gather
so probably this wouldn't be usefull.

Finally, we've parametriced SOS_EDPM as cifmw_os_must_gather_sos_edpm
and we've added default value to "all"

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>


evallesp:
Reopened from: https://github.com/openstack-k8s-operators/ci-framework/pull/3643

evallesp:
Not tested yet timeout behaviour.
No regression found.

evallesp:
Log information at: https://redhat.atlassian.net/browse/OSPCIFMW-95
We've ran a mocked task to raise a timeout. It excecuted the code properly.

TristanCacqueray:
recheck

evallesp:
recheck

evallesp:
recheck

evallesp:
recheck

evallesp:
recheck

evallesp:
recheck

evallesp:
recheck

evallesp:
recheck

evallesp:
recheck

evallesp:
recheck

evallesp:
recheck

roles/os_must_gather/tasks/main.yml
```diff
@@ -95,52 +121,11 @@
...
-
-        - name: Inspect the cluster after must-gather failure
-          ignore_errors: true # noqa: ignore-errors
-          environment:
-            KUBECONFIG: "{{ cifmw_openshift_kubeconfig | default(cifmw_os_must_gather_kubeconfig) }}"
-            PATH: "{{ cifmw_path }}"
-          cifmw.general.ci_script:
-            output_dir: "{{ cifmw_os_must_gather_output_dir }}/artifacts"
-            script: |
-              oc adm inspect namespace/{{ item }} --dest-dir={{ cifmw_os_must_gather_output_dir }}/logs/oc_inspect
```
Valkyrie00, line 137: The old `rescue -> always` block ran oc adm inspect `namespace/{{ item }}` over all configured namespaces as a last-resort diagnostic collection. This has been entirely removed in the new code. Was this intentional? I'm just asking out of curiosity.

roles/os_must_gather/tasks/main.yml
```diff
@@ -79,14 +79,40 @@
...
+        - _must_gather_result.rc is defined
+        - _must_gather_result.rc != 0
+      ansible.builtin.debug:
+        msg: "OpenStack must-gather failed with rc={{ _must_gather_result.rc }}"
+
+    - name: Run fallback generic must-gather command without SOS report when timed out
+      when:
+        - _must_gather_result is defined
+        - _must_gather_result.rc is defined
+        - _must_gather_result.rc == 124
```
Valkyrie00, line 105: What happen If must-gather fails for other reasons? I mean something different from 124.
Is there no recovery attempt anymore?

roles/os_must_gather/tasks/main.yml
```diff
@@ -79,14 +79,40 @@
           SOS_EDPM=$SOS_EDPM
           OMC=$OMC
           SOS_DECOMPRESS=$SOS_DECOMPRESS
-          gather
-          2>&1 || {
```
Valkyrie00, line 83: nit: With it removed, are we sure stderr still ends up in the log output?

Valkyrie00:
LGTM! I just added a few comments, but only to be on the safe side. Once I've checked it, I'll approve it 🙂

roles/os_must_gather/tasks/main.yml
evallesp: I'm unsure if makes sense to remove that check.
We're running must-gather in a simpler way, to make it worked within the timeout error.

If there's any other error, I'm unsure if it makes sense to again timeout long. So my decition was to just log it.
If failed, then we don't have logs in the job output.

I'm good to remove it, just that I'm unsure it might make sense.
Check: https://github.com/openstack-k8s-operators/ci-framework/pull/3643/changes#r2745723552

roles/os_must_gather/tasks/main.yml
evallesp: We got some input in that moment about this: https://github.com/openstack-k8s-operators/ci-framework/pull/3643#discussion_r2812091872

Now we have get_top file but we've lost a bit info with this. Taking that back! thanks for opening the question again.

roles/os_must_gather/tasks/main.yml
evallesp: Good catch but we're doing that directly in ci_script: http://github.com/openstack-k8s-operators/ci-framework/blob/main/plugins/action/ci_script.py#L117

roles/os_must_gather/tasks/main.yml
evallesp: I've make back "oc adm inspect namespace" task when RC != timeout.
Also I've added a new molecule scenario.

roles/os_must_gather/tasks/main.yml
evallesp: I've make back "oc adm inspect namespace" task when RC != timeout.
Also I've added a new molecule scenario.

roles/os_must_gather/tasks/main.yml
evallesp: This is tested in new scenario task: " - name: Verify must-gather failed with non-timeout error"

roles/os_must_gather/tasks/main.yml
Valkyrie00: Awesome!

roles/os_must_gather/tasks/main.yml
Valkyrie00: Awesome! Thanks for clarifying that

Valkyrie00 (approved):

roles/os_must_gather/tasks/main.yml
```diff
@@ -79,14 +79,75 @@
...
+      when:
+        - _must_gather_result is defined
+        - _must_gather_result.rc is defined
+        - _must_gather_result.rc == 124
+      environment:
+        KUBECONFIG: "{{ cifmw_openshift_kubeconfig | default(cifmw_os_must_gather_kubeconfig) }}"
+        PATH: "{{ cifmw_path }}"
+      ansible.builtin.command:
+        cmd: >-
+          timeout {{ (cifmw_os_must_gather_timeout | community.general.to_seconds) + 120 }}
```
michburk, line 111: non-blocking question: If this also times out, do we want to handle it any particular way?

evallesp: In that case I'd say behaviour depends on value of ignore_errors when importing the job.
But if second fail imo it should make the playbook fail as well.

michburk (commented):
Changes LGTM generally, but I just want to generally check to make sure I understand correctly: Currently the framework is tolerant of timeouts during must-gather and doesn't cause a post-failure or anything if it does time out. With this merged, jobs that might have passed before might now time out and get something like a post_failure because of a timed-out must-gather. Should we check how common it is for a build to succeed but then time out during must-gather before merging this?

evallesp:
> Changes LGTM generally, but I just want to generally check to make sure I understand correctly: Currently the framework is tolerant of timeouts during must-gather and doesn't cause a post-failure or anything if it does time out. With this merged, jobs that might have passed before might now time out and get something like a post_failure because of a timed-out must-gather. Should we check how common it is for a build to succeed but then time out during must-gather before merging this?

I don't think it's necessary.
This role is executed from: https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/artifacts/tasks/main.yml#L65 with ignore_errors true.

Here we're:
* Fixing the RC no matter if we want to ignore that or not.
* Before making that as a success (previously, always success), we're trying to actual have some data and not blank logs folders. And also we can check that didn't work properly.

The thing of not using "rescue" is because if you push ignore_errors as true, the rescue is not ran. So here we're checking the RC from the normal path in the always block to check if we want to try with less resources process.

michburk (approved):
/approve

PR #3863, opened 2026-04-17
Os must gather mock return error

evallesp:
DNM.


evallesp:
Mocked works fine more info at: https://redhat.atlassian.net/browse/OSPCIFMW-95

PR #3866, opened 2026-04-20
Fix namespace stuck on cleanup by separating deletion phases

mnietoji:
[cleanup_openstack] Fix namespace stuck on cleanup by separating deletion phases
Split CRs deletion into two phases to prevent namespace getting stuck
during cleanup:
- Phase 1: Delete application resources (DataPlane, ControlPlane, BMH, etc.)
- Phase 2: Delete operators (NMState, MetalLB, OpenStack operators, OLM)
- Wait between phases to ensure Phase 1 resources are fully deleted

Each resource type (DataPlaneDeployment, DataPlaneNodeSet, ControlPlane,
RabbitmqCluster) is waited on individually in dependency order. If a
resource gets stuck, its finalizers are force-removed before moving to
the next type. This replaces the previous RabbitmqCluster-specific
workaround with a generic mechanism.

This prevents race condition where operator controller is deleted before
it can process finalizers from application resources, which causes the
namespace to hang indefinitely.

Signed-off-by: Miguel Angel Nieto Jimenez <mnietoji@redhat.com>
Co-Authored-By: Claude Sonnet 4.5 <noreply@anthropic.com>
Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/roles/cleanup_openstack/tasks/main.yaml
+++ b/roles/cleanup_openstack/tasks/main.yaml
@@ -78,7 +78,7 @@
             value: []
       when: rabbitmq_cell1_info.resources | default([]) | length > 0
 
-- name: Delete deployment CRs
+- name: Delete deployment CRs (Phase 1 - Application Resources)
   vars:
     _stages_crs: >-
       {{
@@ -99,11 +99,6 @@
     _external_dns_crs:
       - "{{ cifmw_manifests_dir }}/cifmw_external_dns/ceph-local-dns.yml"
       - "{{ cifmw_manifests_dir }}/cifmw_external_dns/ceph-local-cert.yml"
-    _operators_crs:
-      - "{{ cifmw_kustomize_deploy_nmstate_dest_file }}"
-      - "{{ cifmw_kustomize_deploy_metallb_dest_file }}"
-      - "{{ cifmw_kustomize_deploy_kustomizations_dest_dir }}/openstack.yaml"
-      - "{{ cifmw_kustomize_deploy_olm_dest_file }}"
     _bmh_crs: >-
       {{
         bmh_crs.files |
@@ -121,9 +116,41 @@
         _external_dns_crs +
         _stages_crs_path +
         _bmh_crs +
-        _bmh_secrets_crs +
-        _operators_crs
+        _bmh_secrets_crs
       }}
+  ansible.builtin.import_tasks: cleanup_crs.yaml
+
+- name: Wait for Phase 1 resources to be deleted
+  kubernetes.core.k8s_info:
+    kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
+    api_key: "{{ cifmw_openshift_token | default(omit) }}"
+    context: "{{ cifmw_openshift_context | default(omit) }}"
+    namespace: "{{ cifmw_openstack_namespace }}"
+    kind: "{{ item }}"
+    api_version: "{{ api_versions[item] }}"
+  register: _phase1_resources
+  until: _phase1_resources.resources | default([]) | length == 0
+  retries: 60
+  delay: 10
+  loop:
+    - OpenStackControlPlane
+    - OpenStackDataPlaneNodeSet
+    - OpenStackDataPlaneDeployment
+  vars:
+    api_versions:
+      OpenStackControlPlane: core.openstack.org/v1beta1
+      OpenStackDataPlaneNodeSet: dataplane.openstack.org/v1beta1
+      OpenStackDataPlaneDeployment: dataplane.openstack.org/v1beta1
+  failed_when: false
+
+- name: Delete deployment CRs (Phase 2 - Operators)
+  vars:
+    _operators_crs:
+      - "{{ cifmw_kustomize_deploy_nmstate_dest_file }}"
+      - "{{ cifmw_kustomize_deploy_metallb_dest_file }}"
+      - "{{ cifmw_kustomize_deploy_kustomizations_dest_dir }}/openstack.yaml"
+      - "{{ cifmw_kustomize_deploy_olm_dest_file }}"
+    _crs_to_delete: "{{ _operators_crs }}"
   ansible.builtin.import_tasks: cleanup_crs.yaml
 
 - name: Get artifacts scripts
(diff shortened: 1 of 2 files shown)


roles/cleanup_openstack/tasks/main.yaml
```diff
@@ -121,11 +143,20 @@
         _external_dns_crs +
         _stages_crs_path +
         _bmh_crs +
-        _bmh_secrets_crs +
-        _operators_crs
+        _bmh_secrets_crs
       }}
   ansible.builtin.import_tasks: cleanup_crs.yaml
 
+- name: Delete deployment CRs (Phase 2 - Operators)
```
evallesp, line 113: (non-blocking) suggestion: Might be a good addition checking if previous phase has ended before running phase 2 ?

roles/cleanup_openstack/tasks/main.yaml
```diff
@@ -121,11 +116,43 @@
...
+  loop:
+    - OpenStackControlPlane
+    - OpenStackDataPlaneNodeSet
+    - OpenStackDataPlaneDeployment
+  vars:
+    api_versions:
+      OpenStackControlPlane: core.openstack.org/v1beta1
+      OpenStackDataPlaneNodeSet: dataplane.openstack.org/v1beta1
+      OpenStackDataPlaneDeployment: dataplane.openstack.org/v1beta1
+  failed_when: false
```
michburk: This isn't a blocking criticism, and sorry if this is a stupid question, but why `failed_when: false` here?

roles/cleanup_openstack/tasks/main.yaml
mnietoji: done

roles/cleanup_openstack/tasks/main.yaml
```diff
@@ -121,11 +116,75 @@
...
+        - OpenStackControlPlane
+        - OpenStackDataPlaneNodeSet
+        - OpenStackDataPlaneDeployment
+      vars:
+        _phase1_api_versions:
+          OpenStackControlPlane: core.openstack.org/v1beta1
+          OpenStackDataPlaneNodeSet: dataplane.openstack.org/v1beta1
+          OpenStackDataPlaneDeployment: dataplane.openstack.org/v1beta1
+
+  rescue:
```
Valkyrie00: If the until wait fails on the first loop item (e.g. OpenStackControlPlane), Ansible skips the remaining items (OpenStackDataPlaneNodeSet, OpenStackDataPlaneDeployment) and jumps straight to rescue.

At that point `_phase1_resources` only contains the result for the item that failed, not all three. The rescue strips finalizers on that one type, then Phase 2 deletes the operators, but the two unchecked types might still have active finalizers that need the operator alive to process, potentially reintroducing the stuck namespace scenario. 

Is my assumption correct?

roles/cleanup_openstack/tasks/main.yaml
mnietoji: updated and removed it

roles/cleanup_openstack/tasks/main.yaml
mnietoji: yes, you are right, fixing it

roles/cleanup_openstack/tasks/main.yaml
```diff
@@ -121,11 +89,37 @@
...
         _stages_crs_path +
         _bmh_crs +
-        _bmh_secrets_crs +
-        _operators_crs
+        _bmh_secrets_crs
       }}
   ansible.builtin.import_tasks: cleanup_crs.yaml
 
+- name: Ensure Phase 1 resources are removed before deleting operators
+  ansible.builtin.include_tasks: wait_andcifmw_cleanup_resource.yaml
```
Valkyrie00: missing underscore between and and cleanup, with a stray `cifmw_`

roles/cleanup_openstack/tasks/main.yaml
```diff
@@ -121,11 +89,37 @@
...
+      api_version: dataplane.openstack.org/v1beta1
+    - kind: OpenStackControlPlane
+      api_version: core.openstack.org/v1beta1
+    - kind: RabbitmqCluster
+      api_version: rabbitmq.com/v1beta1
+  loop_control:
+    loop_var: cifmw_cleanup_resource
+  vars:
+    cifmwcifmw_cleanup_resource_kind: "{{ cifmw_cleanup_resource.kind }}"
+    cifmwcifmw_cleanup_resource_api_version: "{{ cifmw_cleanup_resource.api_version }}"
```
Valkyrie00: There are too many `cifmw`s here

roles/cleanup_openstack/tasks/main.yaml
mnietoji: fixed

roles/cleanup_openstack/tasks/main.yaml
mnietoji: fixed

Valkyrie00 (approved):
/lgtm

roles/cleanup_openstack/tasks/main.yaml
```diff
@@ -51,34 +51,7 @@
...
-        api_key: "{{ cifmw_openshift_token | default(omit) }}"
-        context: "{{ cifmw_openshift_context | default(omit) }}"
-        namespace: "{{ cifmw_openstack_namespace }}"
-        kind: RabbitmqCluster
-        name: rabbitmq-cell1
-        patch:
-          - op: replace
-            path: /metadata/finalizers
-            value: []
-      when: rabbitmq_cell1_info.resources | default([]) | length > 0
```
michburk, line 79: I see you're now including the removal of these two tasks in this patch, but the commit title/message and the pr message don't mention this change or explain why you are doing this. Would the deletion of these two tasks be a better fit for a separate PR? Or at least a separate commit in this same pr?

Thanks!

mnietoji: No, this is all related. It is not 2 different changes in the same commit, it is all the same issue. I am removing 2 taks which  are not working and do it in a different way.

Dividing cleaning in 2 stages is a way to ensure that all crs are removed. Rabbitmq should be removed now and those 2 tasks are not neccesary any more. 

In any case there is a rescue part in which we force the removal if something is not removed. It is something more general in which it can be included other resources if we would find issues in the future. We include here the rabbitmq that is what is causing all the issues now.

Everything is explained in the commit message, but not in the PR, copying it to the PR.

michburk: Thanks for the explanation!

michburk (approved):
/approve

PR #3870, opened 2026-04-20
Fix multi-cell edpm_nodes generation in adoption_vars template

ashu-011:
[adoption_osp_deploy] Fix multi-cell edpm_nodes generation in adoption_vars template

The template logic was broken for multi-cell scenarios since commit 78bf394d,
which replaced the multi-cell-aware loop with single-cell-only logic that only
checked for 'osp-computes' group.

This fix restores the original _stack_names-driven approach that:
- Loops through stacks (overcloud, cell1, cell2, etc.)
- Uses appropriate prefix for each stack ('' for overcloud, 'cell1-' for cell1)
- Generates edpm_nodes with correct structure based on deployment type
- Works for both single-cell and multi-cell scenarios

Backward compatibility:
- Single-cell deployments (only overcloud): edpm_nodes.<node-name>
- Multi-cell deployments: edpm_nodes.<cell-name>.<node-name>
- Preserves existing adoption job behavior for single-cell scenarios

Also adds molecule test for multi-cell fixtures to 
verify the logic and prevent future breakages.

Resolves : https://redhat.atlassian.net/browse/OSPCIX-1325
Resolves: https://redhat.atlassian.net/browse/OSPRH-28982

### DIFF
--- a/roles/adoption_osp_deploy/templates/adoption_vars.yaml.j2
+++ b/roles/adoption_osp_deploy/templates/adoption_vars.yaml.j2
@@ -76,28 +76,41 @@ source_galera_members: |
   {% endfor %}
 {%+ endif +%}
 
-{% if _vm_groups['osp-computes'] | default([]) | length > 0 %}
+# Generate edpm_nodes using _stack_names-driven approach
+# - Single-cell: _stack_names=['overcloud'] → looks for 'osp-computes' → creates edpm_nodes.<default_cell_name>
+# - Multi-cell: _stack_names=['overcloud','cell1','cell2'] → looks for 'cell1-osp-computes', 'cell2-osp-computes' → creates edpm_nodes.cell1, edpm_nodes.cell2
 edpm_nodes:
-  {% for compute in _vm_groups['osp-computes'] %}
-  {% set node_nets = cifmw_networking_env_definition.instances[compute] %}
-  {{ compute }}:
-    hostName: {{ compute }}.{{ cifmw_adoption_osp_deploy_scenario.cloud_domain }}
-    {% for net in node_nets.networks.keys() if cifmw_adoption_osp_deploy_scenario.network_tripleo_network_map[net]|default(net) == "ctlplane" %}
-    ansible:
-      ansibleHost: {{ node_nets.networks[net][ip_version|default('ip_v4')] }}
-    {% endfor %}
-    networks:
-    {% for net in node_nets.networks.keys() if net not in cifmw_adoption_osp_deploy_adoption_vars_exclude_nets %}
+{% for stack in _stack_names %}
+  {% if stack == 'overcloud' %}
+    {% set cell = _default_cell_name %}
+    {% set prefix = '' %}
+  {% else %}
+    {% set cell = stack %}
+    {% set prefix = stack ~ '-' %}
+  {% endif %}
+  {% if _vm_groups[prefix ~ 'osp-computes'] | default([]) | length > 0 %}
+  {{ cell }}:
+    {% for compute in _vm_groups[prefix ~ 'osp-computes'] %}
+    {% set node_nets = cifmw_networking_env_definition.instances[compute] %}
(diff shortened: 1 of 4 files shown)


roles/adoption_osp_deploy/templates/adoption_vars.yaml.j2
```diff
@@ -76,7 +76,52 @@ source_galera_members: |
   {% endfor %}
 {%+ endif +%}
 
-{% if _vm_groups['osp-computes'] | default([]) | length > 0 %}
+{%- set _cell_compute_groups = [] -%}
```
bogdando: could you please add molecule test with a multi-cell fixtures to verify this logic and prevent it from future breakages?

roles/adoption_osp_deploy/templates/adoption_vars.yaml.j2
```diff
@@ -76,7 +76,52 @@ source_galera_members: |
...
+  {%- if 'cell' in group or (group != 'osp-computes' and 'osp-compute' in group) -%}
+    {%- set _has_cell_specific_groups = true -%}
+  {%- endif -%}
+{%- endfor -%}
+{%+ if multi_cell|default(false) and _has_cell_specific_groups +%}
+# Multi-cell scenario: auto-detected from VM group names
+# Generates edpm_nodes with cell structure based on existing groups
+edpm_nodes:
+{% for group_name in _cell_compute_groups | sort %}
+  {%- if group_name != 'osp-computes' -%}
```
bogdando: this drops nodes when osp-computes coexists with cellN-osp-computes. Should be fine if we don't test this in CI (the product doesn't support placing computes in the control plane stack AFAIK)

roles/adoption_osp_deploy/templates/adoption_vars.yaml.j2
```diff
@@ -76,7 +76,52 @@ source_galera_members: |
...
+{%- for group in _cell_compute_groups -%}
+  {%- if 'cell' in group or (group != 'osp-computes' and 'osp-compute' in group) -%}
+    {%- set _has_cell_specific_groups = true -%}
+  {%- endif -%}
+{%- endfor -%}
+{%+ if multi_cell|default(false) and _has_cell_specific_groups +%}
+# Multi-cell scenario: auto-detected from VM group names
+# Generates edpm_nodes with cell structure based on existing groups
+edpm_nodes:
+{% for group_name in _cell_compute_groups | sort %}
```
bogdando: nit: structural inconsistency with source_mariadb_ip / source_galera_members — Those existing multi-cell blocks are driven by _stack_names and derive prefixes (overcloud -> '', else stack ~ '-'). The new edpm_nodes logic instead auto-detects from *-osp-computes keys in _vm_groups. Stacks without compute groups, or extra groups not aligned with stacks, can produce edpm_nodes that don't mirror  the MariaDB/Galera topology. Using a _stack_names-driven approach would be more consistent and less fragile. But I would not expect those misalignments to ever happen in our existing CI jobs, I am also not sure the original implementation was free of that issue.

bogdando (commented):
please see inline

roles/adoption_osp_deploy/templates/adoption_vars.yaml.j2
ashu-011: Sure, make sense

roles/adoption_osp_deploy/templates/adoption_vars.yaml.j2
ashu-011: Added the molecule tests

bogdando (dismissed):
Good, the test project has passed. LGTM

ashu-011:
test-project added : https://redhat.atlassian.net/browse/OSPRH-28982?focusedCommentId=16801550

bogdando (approved):
/LGTM

brjackma (approved):

PR #3873, opened 2026-04-21
[edpm_deploy_baremetal] Allow overriding edpm_bootstrap_command

Valkyrie00:
The inline kustomization hardcodes edpm_bootstrap_command to "sudo dnf -y update", preventing downstream jobs from customizing the bootstrap command via file-based kustomizations (which are applied first and get overwritten).
Use a configurable variable with the same default so existing jobs are unaffected, while jobs that need a custom bootstrap command can set cifmw_edpm_deploy_baremetal_bootstrap_command.

### DIFF
--- a/roles/edpm_deploy_baremetal/README.md
+++ b/roles/edpm_deploy_baremetal/README.md
@@ -19,6 +19,8 @@ This role doesn't need privilege escalation.
 * `cifmw_edpm_deploy_baremetal_repo_setup_override`: (Boolean) Override the repo-setup service in OpenStackDataPlane with repo-setup-downstream. Default: `false`
 * `cifmw_edpm_deploy_baremetal_create_vms`: (Boolean) If enabled, compute nodes are pre-provisioned using Ironic else OpenStackProvisioner. Default: `true`
 * `cifmw_edpm_deploy_baremetal_nova_compute_extra_config`: (String) Oslo config snippet defining extra configuration for the nova-compute services. Defaults to an empty string.
+* `cifmw_edpm_deploy_baremetal_custom_bootstrap`: (Boolean) When true, skips the inline `edpm_bootstrap_command` kustomization patch so that a file-based kustomization can provide a custom bootstrap command instead. Default: `false`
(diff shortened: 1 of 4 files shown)


roles/edpm_deploy_baremetal/README.md
```diff
@@ -19,6 +19,8 @@ This role doesn't need privilege escalation.
 * `cifmw_edpm_deploy_baremetal_repo_setup_override`: (Boolean) Override the repo-setup service in OpenStackDataPlane with repo-setup-downstream. Default: `false`
 * `cifmw_edpm_deploy_baremetal_create_vms`: (Boolean) If enabled, compute nodes are pre-provisioned using Ironic else OpenStackProvisioner. Default: `true`
 * `cifmw_edpm_deploy_baremetal_nova_compute_extra_config`: (String) Oslo config snippet defining extra configuration for the nova-compute services. Defaults to an empty string.
+* `cifmw_edpm_deploy_baremetal_custom_bootstrap`: (Boolean) When true, skips the inline `edpm_bootstrap_command` kustomization patch so that a file-based kustomization can provide a custom bootstrap command instead. Default: `false`
```
karelyatin, line 22: but file based override should already work if those are named correctly, or it's not working even with that? what about non baremetal job, how it works there?

arxcruz (approved):
/lgtm
/approve

belolipa (approved):
/lgtm
/approve

danpawlik (approved):

roles/edpm_deploy_baremetal/README.md
Valkyrie00: Hey @karelyatin, during my analysis I discovered the file-based kustomization is applied before the inline one
(`kustomization_files_goes_first` defaults to true in ci_kustomize), so the inline patch in edpm_deploy_baremetal always overwrites edpm_bootstrap_command with `sudo dnf -y update` regardless of the file name.

Valkyrie00: Also checked others, this only affects baremetal jobs since non-baremetal jobs don't use this role. The flag lets the job opt out of the inline patch so the file-based one takes effect.

karelyatin (approved):

PR #3882, opened 2026-04-23
[cifmw_setup] Use cifmw_helpers role to read variable files

danpawlik:
In the CI job, where the nested Ansible execution was removed, there is an issue to read the variables using include_vars module, because that module is searching the files on the host where the Ansible was executed. It means that if the nested Ansible was dropped, the Ansible bin was called from the Zuul executor, on which the /home/zuul does not exists.


danpawlik:
recheck

Valkyrie00 (approved):
LGTM

michburk (approved):

PR #3891, opened 2026-04-24
[libvirt_manager] Add predictable network interface names to EDPM

holser:
Enable systemd predictable network interface naming inside guest VMs
by removing `net.ifnames=0` from kernel args via virt-customize. This
gives guests consistent PCI-topology-based names (enp1s0, enp2s0, etc.)
instead of legacy ethN naming. Predictable network interfaces are
requirement for testing Leapp upgrade functionality.

Controlled by cifmw_libvirt_manager_predictable_nic_names (defaults
to false).

Jira: [OSPRH-29381](https://redhat.atlassian.net/browse/OSPRH-29381)

### DIFF
--- a/roles/libvirt_manager/defaults/main.yml
+++ b/roles/libvirt_manager/defaults/main.yml
@@ -84,6 +84,7 @@ cifmw_libvirt_manager_firewalld_zone_libvirt_forward: true
 cifmw_libvirt_manager_firewalld_default_zone: public
 cifmw_libvirt_manager_firewalld_default_zone_masquerade: true
 cifmw_libvirt_manager_attach_dummy_interface_on_bridges: true
+cifmw_libvirt_manager_predictable_nic_names: false
 cifmw_libvirt_manager_extra_network_configuration: {}
 
 cifmw_libvirt_manager_vm_users: []
--- a/roles/libvirt_manager/tasks/create_vms.yml
+++ b/roles/libvirt_manager/tasks/create_vms.yml
@@ -61,6 +61,14 @@
         xml: "{{ lookup('template', cifmw_libvirt_manager_vm_template) }}"
         uri: "qemu:///system"
 
+    - name: "Disable net.ifnames=0 for {{ vm }}"
+      when:
+        - vm_data.disk_file_name != 'blank'
+        - cifmw_libvirt_manager_predictable_nic_names | default(false) | bool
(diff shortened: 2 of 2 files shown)


evallesp:
LGTM, asked for a TP.

roles/libvirt_manager/tasks/attach_interface.yml
```diff
@@ -80,6 +80,14 @@
       }}
     _type: "{{ cifmw_libvirt_manager_network_interface_types[network.name] | default('bridge') }}"
     _clean_vm: "{{ vm_name | replace('cifmw-', '') }}"
+    _vm_short: >-
+      {{
+        (vm_item | default(_clean_vm)) | replace('-', '')
+      }}
+    _target_dev: >-
```
nemarjan: (praise) Great improvement adding --target for predictable NIC naming.
(Question) : _target_dev is currently computed as (_vm_short ~ 'p' ~ (net_idx|int))[:15]. Because this relies on truncation to satisfy the 15‑char limit, there’s a chance of collisions for VMs with similar prefixes. Just another thing to mention, once net_idx goes to 2+ digits (p9 → p10), the suffix length changes and the truncation point shifts, which increases the chance of two different inputs collapsing to the same 15‑char string.

Would you be up for considering a slightly more collision‑resistant naming scheme that still stays within 15 characters (for example a fixed layout like <vm_prefix><short_hash_of _clean_vm><idx>, using only letters and numbers)? Alternatively, a small guard/assert to catch duplicate targets early could also work nicely.

holser: I have removed this part as it requires unique naming for all interfaces. PR update relies on virt-customize only and VM OS ability to set up networking interfaces based on pci lanes

michburk:
Changes lgtm, especially seeing that they are opt-in, but @holser do you have a testproject that uses these changes that you could link in the comments of that jira story? I see Enrique requested one here and in the jira comments.

holser:
> Changes lgtm, especially seeing that they are opt-in, but @holser do you have a testproject that uses these changes that you could link in the comments of that jira story? I see Enrique requested one here and in the jira comments.

Sorry. I was PTO for a while. I have created test project to test the changes.

evallesp:
Got TP. merging

evallesp (approved):
/lgtm
/approve

PR #3901, opened 2026-05-06
[kustomize_deploy] Fix MetalLB CR creation race with webhook readiness

rebtoor:
The MetalLB validating webhook can reject CR creation with "no endpoints available" when the operator's webhook services aren't fully ready yet. The existing deployment availability checks pass before the backing Endpoints objects have addresses populated, creating a window where CR creation fails.

Add an explicit wait for all four MetalLB webhook endpoint services (metallb-operator-controller-manager-service,
metallb-operator-webhook-server-service,
metallb-operator-webhook-service, webhook-service) to have at least one ready address before applying MetalLB CRs. Also add retry logic (retries: 10, delay: 15) to the CR apply step itself as a safety net for any remaining transient webhook failures.

### DIFF
--- a/roles/kustomize_deploy/tasks/install_operators.yml
+++ b/roles/kustomize_deploy/tasks/install_operators.yml
@@ -325,6 +325,28 @@
       register: _webhook_server_deployment
       until: _webhook_server_deployment is success
 
+    - name: Wait for MetalLB webhook endpoints to have addresses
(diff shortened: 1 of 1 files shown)


evallesp (approved):
/lgtm

michburk (approved):
/approve

PR #3905, opened 2026-05-06
Add retry logic to ceph node package installation

rebtoor:
The dnf task installing os-net-config and openvswitch on ceph nodes had no retry, causing transient repo failures to fail the entire job. Added retries: 3 with delay: 10 to both setup_cephnodes.yaml and setup_cephnodes_ipv6.yaml, consistent with other dnf tasks in the playbooks.

### DIFF
--- a/hooks/playbooks/setup_cephnodes.yaml
+++ b/hooks/playbooks/setup_cephnodes.yaml
@@ -52,6 +52,10 @@
           - os-net-config
           - openvswitch
         state: present
+      retries: 5
+      delay: 30
+      register: _ceph_pkgs_install
+      until: _ceph_pkgs_install is success
 
     - name: Generate os-net-config YAML
       ansible.builtin.copy:
--- a/hooks/playbooks/setup_cephnodes_ipv6.yaml
+++ b/hooks/playbooks/setup_cephnodes_ipv6.yaml
@@ -52,6 +52,10 @@
           - os-net-config
           - openvswitch
         state: present
+      retries: 5
+      delay: 30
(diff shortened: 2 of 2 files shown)


hooks/playbooks/setup_cephnodes.yaml
```diff
@@ -52,6 +52,10 @@
           - os-net-config
           - openvswitch
         state: present
+      retries: 5
+      delay: 30
+      register: _ceph_pkgs_install
```
fmount, line 57: I'm ok to have a `retry/delay` mechanism here, but I don't understand the reason of this transient error before the config provided by `os-net-config` is applied.
maybe I need more context but +1 from me if this solves an issue in CI

hooks/playbooks/setup_cephnodes.yaml
```diff
@@ -52,6 +52,10 @@
...
+      until: _ceph_pkgs_install is success
```
fmount, line 58: maybe this could be named `_net_pkgs_install` because we're installing `os-net-config` and `openvswitch`

fmount (commented):
just made a first pass.
In general I'm ok to have a retry mechanism even though I'm not sure what generates this transient error.

fmount:
recheck

hooks/playbooks/setup_cephnodes.yaml
rebtoor: definitely.

evallesp (approved):
/approve

PR #3906, opened 2026-05-07
[cifmw_cephadm] Add cifmw_cephadm_auth_allowed_ciphers parameter

fultonj:
Add a new optional string parameter cifmw_cephadm_auth_allowed_ciphers to the cifmw_cephadm role. When set, the role runs the fullowing during cluster configuration (cephadm_config_set.yml)

  ceph mon set auth_allowed_ciphers <value>

The parameter defaults to "" and is a no-op when unset, so existing deployments are unaffected. Document the parameter and its use cases in the role README.

Jira: [OSPRH-29668](https://redhat.atlassian.net/browse/OSPRH-29668)


fmount (approved):
/lgtm

evallesp (approved):
/approve

PR #3908, opened 2026-05-07
Wait for stable cluster after openshift_setup configuration changes

mnietoji:
The openshift_setup role modifies cluster-wide resources (Image/cluster additionalTrustedCA, registrySources, ImageContentSourcePolicy, Network operator config) that trigger rolling updates of the API server and other control plane components. Without waiting for these rollouts to complete, subsequent tasks (e.g. kustomize_deploy) fail with 401 Unauthorized because the API server pods are being recycled.

### DIFF
--- a/roles/cifmw_setup/tasks/infra.yml
+++ b/roles/cifmw_setup/tasks/infra.yml
@@ -46,6 +46,15 @@
   ansible.builtin.import_role:
     name: openshift_setup
 
+- name: Wait for cluster to stabilize after openshift_setup changes
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  ansible.builtin.command:
+    cmd: >-
+      oc adm wait-for-stable-cluster --minimum-stable-period=5s --timeout=30m
+  changed_when: false
+
 - name: Deploy Observability operator.
   when:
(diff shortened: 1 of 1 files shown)


roles/cifmw_setup/tasks/infra.yml
```diff
@@ -46,6 +46,14 @@
   ansible.builtin.import_role:
     name: openshift_setup
 
+- name: Wait for cluster to stabilize after openshift_setup changes
+  ansible.builtin.command:
```
ciecierski, line 53:
```suggestion
  environment:
    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
    PATH: "{{ cifmw_path }}"
  ansible.builtin.command:
    cmd: >-
      oc adm wait-for-stable-cluster --minimum-stable-period=5s --timeout=30m
```

This is common pattern in ci-framework. KUBECONFIG as env and with timeout

ciecierski (changes requested):

brjackma:
recheck

michburk:
Hey I see you've marked this as ready for review, but `rdoproject.org/github-check` hasn't passed, so we wouldn't merge even if we're happy with the changes. Can you make sure the CI is green?

If you suspect there is some issue with these jobs unrelated to your changes, let us know.

Thanks!

PR #3909, opened 2026-05-07
[reproducer] Wait for controller-0 SSH before configuring

mnietoji:
After controller-0 reboots, SSH may become transiently unreachable while background system initialization completes. The configure_controller block delegates tasks to controller-0 without verifying connectivity first, causing intermittent UNREACHABLE failures on the install_ca task.

Add wait_for_connection at the start of the delegate_to block to ensure controller-0 is fully accessible before proceeding.

### DIFF
--- a/roles/reproducer/tasks/configure_controller.yml
+++ b/roles/reproducer/tasks/configure_controller.yml
@@ -29,6 +29,11 @@
         default('zuul')
       }}
   block:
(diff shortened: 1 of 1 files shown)


evallesp (approved):
/lgtm

michburk (approved):
/approve

PR #3914, opened 2026-05-08
[env_op_images] Add unit tests for verify_pulled_report_crio

nemarjan:
Add focused unit coverage for CRI-O pulled report verification, including successful enrichment, cross-node evidence accounting, and failure when no logs are provided. Reuse shared test utilities with a local fallback so tests run in both collection-style and local environments.

Assisted-By: Cursor Code


nemarjan:
recheck

nemarjan:
recheck

tests/unit/modules/test_verify_pulled_report_crio.py
```diff
@@ -0,0 +1,190 @@
...
+        "roles",
+        "env_op_images",
+        "library",
+        "verify_pulled_report_crio.py",
+    )
+)
+
+
+class TestVerifyPulledReportCrio(ModuleBaseTestCase):
+    def test_enriches_report_and_counts_cross_node(self):
```
evallesp, line 27: (non-blocking) suggestion: let's add docstrings per each of the tests. I'm setting this as non-blocking as in other places we don't have docstrings. But definetely would be great to have.

It can be something simple or a more elaborated:
    """
    GIVEN ... 
    WHEN ...
    THEN ...
    """

tests/unit/modules/test_verify_pulled_report_crio.py
```diff
@@ -0,0 +1,190 @@
...
+            "images": [
+                {
+                    "image_id": "quay.io/demo/app@sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
+                    "node": "node-a",
+                },
+                {
+                    "image_id": "quay.io/demo/other@sha256:bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb",
+                    "node": "node-b",
+                },
+                {"image_id": "no-digest-here", "node": "node-a"},
```
evallesp, line 57: (non-blocking) question: Do we need this? I'm unsure if we're using somewhere.

tests/unit/modules/test_verify_pulled_report_crio.py
```diff
@@ -0,0 +1,190 @@
...
+
+    def test_fails_when_no_log_files(self):
+        set_module_args(
+            dict(report_path="/tmp/in.yaml", output_path="/tmp/out.yaml", log_paths=[])
+        )
+
+        with self.assertRaises(AnsibleFailJson) as rst:
+            VERIFY_MODULE.run_module()
+
+        self.assertIn("No CRI-O log files", rst.exception.args[0]["msg"])
```
evallesp, line 202: (blocking) question:I guess we can have more tests case for the exception at L211, L215, L175

evallesp (commented):
LGTM in general, some questions.

nemarjan:
recheck

tests/unit/modules/test_verify_pulled_report_crio.py
nemarjan: Make sense, added docstrings as suggested.

tests/unit/modules/test_verify_pulled_report_crio.py
nemarjan: Yes, this entry is needed as it exercises the skip branch in run_module() where images without a sha256: digest are ignored. In practice, digest-less images can appear when containers are referenced by tag only (e.g. quay.io/demo/app:latest) and the report captures the image_id before CRI-O resolves it, or when images are locally built/imported without a registry digest. The module can't match these against CRI-O pull logs, so it skips them. This test entry ensures the skip works correctly and that entries_with_digest only counts images that actually carry a digest (2, not 3).

tests/unit/modules/test_verify_pulled_report_crio.py
nemarjan: Addressed below , each of these exception paths now has a dedicated test:

L175 (IOError in log reading) → test_fails_when_log_file_unreadable
L211 (IOError in report reading) → test_fails_when_report_unreadable
L215 (YAMLError in parsing) → test_fails_when_report_has_invalid_yaml"

evallesp (approved):
/lgtm

michburk (commented):
Changes lgtm, but I do have one critique that isn't necessarily blocking:

It looks like this pr is making two changes:
1. Add the unit tests
2. Move the verify_pulled_report_crio.py file and use the fcqn for `verify_pulled_report_crio`

This isn't a huge set of changes, but I think generally we should keep commits focused on one set of changes at a time. For example, in the event we need to revert something, we can create one revert commit that can specifically target the exact broken changes, rather than trying to de-tangle a smaller subset of changes from a larger commit.

nemarjan:
> Changes lgtm, but I do have one critique that isn't necessarily blocking:
> 
> It looks like this pr is making two changes:
> 
> 1. Add the unit tests
> 2. Move the verify_pulled_report_crio.py file and use the fcqn for `verify_pulled_report_crio`
> 
> This isn't a huge set of changes, but I think generally we should keep commits focused on one set of changes at a time. For example, in the event we need to revert something, we can create one revert commit that can specifically target the exact broken changes, rather than trying to de-tangle a smaller subset of changes from a larger commit.

Thanks for the feedback Michael, yes I agree with you also, when working with creation of unit test i realized the python script was better placed in modules as tests were not running in gate stage in the previous location so that was reason for movement , noted on best practices and will look to incorporate like that in future. Do you think we need take the approach you mentioned for this MR?

michburk (approved):
/lgtm
/approve

PR #3919, opened 2026-05-08
[reproducer] Use include_role for install_ca in delegated block

mnietoji:
When import_role is used inside a delegate_to block, Ansible triggers Python interpreter discovery on the original host (hypervisor) rather than the delegate target (controller-0). If the SSH connection to the hypervisor is unavailable at that point, the task fails with UNREACHABLE.

Switch to include_role which loads tasks dynamically within the delegated execution context, avoiding the interpreter discovery on the original host.


evallesp (approved):
/lgtm

PR #3921, opened 2026-05-09
added dependecy of openstack cloud

hgoelredhat:


rebtoor (changes requested):
```
❯ rg -l 'openstack\.cloud'

❯
```

The collection isn't used in the ci-framework, so it's pointless to add this collection as dependency.

PR #3922, opened 2026-05-10
[shiftstack] Use GitHub for shiftstack-qa clone and fix false-positive test results

imatza-rh:
## Summary

Three fixes for the shiftstack CI role:

### 1. Clone shiftstack-qa from GitHub instead of GerritHub

shiftstack-qa moved code review to GitHub PRs ([shiftstack/shiftstack-qa#2](https://github.com/shiftstack/shiftstack-qa/commit/aaf4e79565372b8db0141a647ff8cde2b0607837)). The role default still clones from GerritHub, which is no longer maintained. GerritHub main is 3+ commits behind GitHub main - missing the `4-stable` OCP channel switch, `lb_tests` disable, and `.gitreview` removal.

- Change `cifmw_shiftstack_qa_repo` default to `https://github.com/shiftstack/shiftstack-qa`
- Rename `cifmw_shiftstack_qa_gerrithub_change` to `cifmw_shiftstack_qa_change_ref` - supports both GitHub PR refs and GerritHub change refs

### 2. Fix false-positive and connection-drop bugs in exec_command_in_pod.yml

oc exec/rsh connections to the shiftstackclient pod drop after ~38 minutes during long-running tests. The previous fix (marker file + polling) didn't solve it because the marker file write was chained to the same bash process that dies when the connection drops.

Root cause verified from tp!2297 build d349fb73 inner test log: the test ran for 1h 37m inside the pod (Manila StorageClass failure at 14:20 UTC), but oc exec returned after only 38m 50s. The marker file was never created - 720 retries over 2h all got No such file or directory.

Fix: writes the command to a script file, copies it into the pod, then runs it detached via nohup. The process survives oc exec connection drops. Also kills orphaned background processes from previous invocations to prevent marker file poisoning.

Default polling timeout set to 8h (5760 retries x 5s) to cover the full test suite duration. Configurable via `cifmw_shiftstack_exec_retries` and `cifmw_shiftstack_exec_delay`, documented in README.md.

Verified on the live shiftstackclient pod (serval70) - 6-case Ansible test suite covering success, exit codes 1/42, dollar-sign commands, orphan kill, retry timeout, and nohup survival after oc exec termination.

### 3. Increase compute VM memory for osasinfra DT scenario

Increase `cifmw_libvirt_manager_compute_memory` from 50 to 60 GB in `dt-osasinfra.yml`. With 24 GB master flavors and `ram_allocation_ratio=1.0`, two master-sized OCP VMs cannot share a 46 GB compute node (2x24576=49152 MB > 45957 MB schedulable). 60 GB gives ~51 GB usable, (6 GB margin). serval70 has 176 GB free host RAM - the increase adds only 28 GB across 3 computes.

## Test plan

- [x] GitHub clone validated in tp!2297 build 70b1f455 (5h 8m) - OCP installed, cluster checks passed
- [x] 6-case Ansible test suite on live pod (serval70)
- [x] nohup survival test: 30s command, oc exec killed at 3s, marker appeared at t+30s
- [x] YAML syntax + ansible-lint clean
- [x] Compute memory verified via live Placement API on serval70 (2026-08-06)
- [x] Testproject re-validation ([tp!2499](https://gitlab.cee.redhat.com/ci-framework/ci-framework-testproject/-/merge_requests/2499) - SUCCESS [42cbd430](https://sf.apps.int.gpc.ocp-hub.prod.psi.redhat.com/zuul/t/components-integration/build/42cbd43055db4b6c9d58a5eab04dfae9), Aug 12, 6h 57m, 1122 tests)


imatza-rh:
recheck

ekuris-redhat (commented):
Review of the exec_command_in_pod.yml rewrite — 7 findings, focused on the nohup+polling approach. The GitHub→GerritHub migration and variable rename look clean.

roles/shiftstack/tasks/exec_command_in_pod.yml
```diff
@@ -14,20 +14,101 @@
...
+      environment:
+        PATH: "{{ cifmw_path }}"
+        KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+      ansible.builtin.command:
+        cmd: >-
+          oc exec -n {{ namespace }} {{ pod_name }} --
+          bash -c 'cat /tmp/cifmw_cmd_rc'
+      register: pod_exit_code
+      changed_when: false
+      until: pod_exit_code.rc == 0
```
ekuris-redhat: The polling condition `pod_exit_code.rc == 0` treats every non-zero `oc exec` exit code as "command still running." But if the pod is evicted, OOM-killed, or the API server becomes unreachable, `oc exec` also returns non-zero , indistinguishable from "file not yet created." This means a dead pod causes 5760 retries (8 hours) of silent failure before the task finally times out.

Consider adding a pod-liveness check inside the polling loop so it fails fast when the pod is gone, e.g. wrapping the cat in a compound check:
```bash
bash -c 'if cat /tmp/cifmw_cmd_rc 2>/dev/null; then exit 0; elif oc get pod -n NS POD -o jsonpath="{.status.phase}" | grep -q Running; then exit 1; else echo "Pod is no longer running" >&2; exit 2; fi'
```
Or add a separate task after the loop that checks pod phase when `pod_exit_code.rc != 0`.

roles/shiftstack/tasks/exec_command_in_pod.yml
```diff
@@ -14,20 +14,101 @@
...
+          exit code {{ pod_exit_code.stdout | trim }}.
+      when: (pod_exit_code.stdout | trim) != '0'
+
   rescue:
     - name: Fail when the command module fails
       ansible.builtin.fail:
-        msg: '"{{ command }}" command execution failed with exit code "{{ pod_command_result.rc }}" and error: "{{ pod_command_result.stderr }}".'
+        msg: >-
+          "{{ command }}" failed.
+          oc exit code: {{ pod_command_result.rc | default('N/A') }},
```
ekuris-redhat: With the nohup approach, `pod_command_result.stderr` is from the backgrounding step (empty), and `pod_exit_code.stderr` is from `cat /tmp/cifmw_cmd_rc` (empty or "No such file"). The actual command output sits in `/tmp/cifmw_cmd_output.log` inside the pod but is never retrieved.

This is a regression from the old synchronous `oc rsh` which surfaced stderr directly in `pod_command_result.stderr`.

Consider fetching the output log before reporting the failure:
```yaml
rescue:
  - name: Retrieve command output from pod
    ansible.builtin.command:
      cmd: >-
        oc exec -n {{ namespace }} {{ pod_name }} --
        bash -c 'tail -c 8192 /tmp/cifmw_cmd_output.log 2>/dev/null || echo "(output not available)"'
    register: pod_output_log
    failed_when: false
    changed_when: false

  - name: Fail when the command module fails
    ansible.builtin.fail:
      msg: >-
        "{{ command }}" failed.
        pod exit code: {{ pod_exit_code.stdout | default('N/A') | trim }},
        output: {{ pod_output_log.stdout | default('(not retrieved)') }}
```

roles/shiftstack/tasks/exec_command_in_pod.yml
```diff
@@ -14,20 +14,101 @@
...
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+  ansible.builtin.command:
+    cmd: >-
+      oc exec -n {{ namespace }} {{ pod_name }} --
+      bash -c 'rm -f /tmp/cifmw_cmd_rc /tmp/cifmw_cmd_pid
+      /tmp/cifmw_cmd_script.sh /tmp/cifmw_cmd_output.log'
+  changed_when: false
+  failed_when: false
+
+- name: Write the command to a local script file
```
ekuris-redhat: The five setup tasks (lines 17–68: kill previous, clear markers, write script, copy to pod, chmod) run _before_ the `block:` at line 69. If any of them fails, e.g. a transient API-server error on "Copy the script into the pod"  the `rescue:` diagnostic message and the `always:` log-saving block are both skipped entirely. The caller gets a raw Ansible error with no structured diagnostics and no log file written.

Consider moving these setup tasks inside the `block:` so that failures are caught by the existing rescue/always handlers.

roles/shiftstack/tasks/exec_command_in_pod.yml
```diff
@@ -45,6 +126,7 @@
               pod_name: "{{ pod_name }}"
               command: "{{ command }}"
               pod_command_result: "{{ pod_command_result }}"
```
ekuris-redhat: With the nohup approach, `pod_command_result` now contains only the background PID (e.g. `{stdout: "1234", rc: 0}`), not the actual command output. The saved log file loses its diagnostic value compared to the old synchronous approach.

Consider fetching `/tmp/cifmw_cmd_output.log` from the pod in the `always:` block and including it in the saved command info.

roles/shiftstack/tasks/exec_command_in_pod.yml
```diff
@@ -14,20 +14,101 @@
...
 
+- name: Kill any previous background command
+  environment:
+    PATH: "{{ cifmw_path }}"
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+  ansible.builtin.command:
+    cmd: >-
+      oc exec -n {{ namespace }} {{ pod_name }} --
+      bash -c 'if [ -f /tmp/cifmw_cmd_pid ]; then
+      kill $(cat /tmp/cifmw_cmd_pid) 2>/dev/null; fi'
```
ekuris-redhat: The kill and clear-markers run as two separate `oc exec` round-trips (lines 17–39). Between them there's a window where the killed process could still write `/tmp/cifmw_cmd_rc` if the script was already finishing, or where PID reuse could cause the kill to target an unrelated process in the pod.

Consider combining into a single atomic operation:
```yaml
- name: Kill any previous background command and clear markers
  ansible.builtin.command:
    cmd: >-
      oc exec -n {{ namespace }} {{ pod_name }} --
      bash -c 'if [ -f /tmp/cifmw_cmd_pid ]; then
      kill $(cat /tmp/cifmw_cmd_pid) 2>/dev/null;
      wait $(cat /tmp/cifmw_cmd_pid) 2>/dev/null; fi;
      rm -f /tmp/cifmw_cmd_rc /tmp/cifmw_cmd_pid
      /tmp/cifmw_cmd_script.sh /tmp/cifmw_cmd_output.log'
  changed_when: false
  failed_when: false
```

roles/shiftstack/tasks/exec_command_in_pod.yml
```diff
@@ -14,20 +14,101 @@
...
+      /tmp/cifmw_cmd_script.sh /tmp/cifmw_cmd_output.log'
+  changed_when: false
+  failed_when: false
+
+- name: Write the command to a local script file
+  ansible.builtin.copy:
+    content: |
+      #!/bin/bash
+      {{ command }}
+    dest: /tmp/cifmw_cmd_script.sh
```
ekuris-redhat: The script is written to a hardcoded path `/tmp/cifmw_cmd_script.sh` on the Ansible controller. If two CI jobs run this role concurrently on the same executor host, they'd overwrite each other's scripts. Consider incorporating the pod name into the path to avoid collisions:
```yaml
dest: "/tmp/cifmw_cmd_script_{{ pod_name }}.sh"
```

roles/shiftstack/defaults/main.yml
```diff
@@ -33,8 +33,8 @@ cifmw_shiftstack_hypervisor: "{{ hostvars[hostvars['controller-0']['cifmw_hyperv
 cifmw_shiftstack_installation_dir: "{{ cifmw_shiftstack_basedir }}/installation"
 cifmw_shiftstack_manifests_dir: "{{ cifmw_shiftstack_basedir }}/manifests"
 cifmw_shiftstack_project_name: "shiftstack"
-cifmw_shiftstack_qa_gerrithub_change: ""
-cifmw_shiftstack_qa_repo: "https://review.gerrithub.io/shiftstack/shiftstack-qa"
+cifmw_shiftstack_qa_change_ref: ""
```
ekuris-redhat, line 38: `cifmw_shiftstack_exec_delay` and `cifmw_shiftstack_exec_retries` are documented in the README with concrete defaults (5 and 5760), but they're not declared here. Every other role variable with a concrete default is defined in `defaults/main.yml` , the file header at line 18 states "All variables intended for modification should be placed in this file."

The inline `| default()` fallbacks work functionally, but adding them here makes them discoverable and overridable through the standard Ansible variable precedence:
```yaml
cifmw_shiftstack_exec_delay: 5
cifmw_shiftstack_exec_retries: 5760
```

imatza-rh:
recheck

imatza-rh:
recheck

roles/shiftstack/tasks/exec_command_in_pod.yml
imatza-rh: Fixed. Changed the script's "still running" exit code from 1 to 10 (a code oc exec never returns on its own). The polling until is now pod_exit_code.rc != 10 - any unexpected rc (1, 137, etc.) from an oc exec connection failure breaks the loop immediately instead of retrying for 8 hours. A new "Fail if the pod is no longer reachable" task handles these cases with a diagnostic message including the oc exec stderr.

Additionally, the launch command now clears stale markers (rm -f /tmp/cifmw_cmd_rc /tmp/cifmw_cmd_pid) before starting nohup, so even if the cleanup task (which has failed_when: false) fails silently, a leftover marker from a previous run can't trick the polling into reporting stale results.

roles/shiftstack/tasks/exec_command_in_pod.yml
imatza-rh: Addressed - the rescue block now fetches tail -c 8192 /tmp/cifmw_cmd_output.log from the pod and includes it in the failure message. The always block also retrieves the last 16KB for the saved log file.

roles/shiftstack/tasks/exec_command_in_pod.yml
imatza-rh: Addressed - all tasks (kill, copy, execute, poll, fail checks) are now inside the block, so the rescue/always handlers catch any setup failure.

roles/shiftstack/tasks/exec_command_in_pod.yml
imatza-rh: Addressed - the always block now retrieves tail -c 16384 /tmp/cifmw_cmd_output.log from the pod and includes it in the saved command info YAML alongside the other fields.

roles/shiftstack/tasks/exec_command_in_pod.yml
imatza-rh: Addressed - combined into a single bash -c that kills, waits, and removes all markers atomically in one oc exec round-trip.

roles/shiftstack/tasks/exec_command_in_pod.yml
imatza-rh: Addressed differently - switched to kubernetes.core.k8s_cp with inline content parameter, which copies directly to the pod without writing a local file. Each pod has its own /tmp filesystem, so concurrent jobs on different pods don't conflict.

roles/shiftstack/defaults/main.yml
imatza-rh: Addressed - both cifmw_shiftstack_exec_delay (5) and cifmw_shiftstack_exec_retries (5760) are now declared in defaults/main.yml.

imatza-rh:
recheck

imatza-rh:
recheck

imatza-rh:
recheck

imatza-rh:
recheck

imatza-rh:
recheck

ekuris-redhat (approved):
/lgtm

Valkyrie00:
recheck

roles/shiftstack/tasks/exec_command_in_pod.yml
```diff
@@ -16,23 +16,142 @@
...
         KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
       ansible.builtin.command:
-        cmd: "oc rsh -n {{ namespace }} {{ pod_name }} bash -c '{{ command }}'"
-      register: pod_command_result
+        cmd: >-
+          oc exec -n {{ namespace }} {{ pod_name }} --
+          bash -c 'if cat /tmp/cifmw_cmd_rc 2>/dev/null; then exit 0;
+          elif [ -f /tmp/cifmw_cmd_pid ] && kill -0 $(cat /tmp/cifmw_cmd_pid) 2>/dev/null; then exit 10;
+          else exit 2; fi'
+      register: pod_exit_code
```
Valkyrie00, line 67: (nit) Hey @imatza-rh, the new register variables (pod_exit_code, pod_output_log, pod_output_for_log) are file-local but don't use the _ prefix convention. Not blocking, just a style alignment if you're touching those lines.

roles/shiftstack/tasks/exec_command_in_pod.yml
```diff
@@ -16,23 +16,142 @@
...
+      when:
+        - pod_exit_code.rc != 0
+        - pod_exit_code.rc != 2
+        - pod_exit_code.rc != 10
+
+    - name: Fail if the background process died without writing the marker
+      ansible.builtin.fail:
+        msg: >-
+          The background process in pod {{ pod_name }} exited without writing a result.
+          The command may have been killed by OOM or a signal.
```
Valkyrie00, line 89: (nit) Hey @imatza-rh, when pod_exit_code.rc == 2, the error message says the process may have been OOM-killed, but rc=2 could also be an oc exec error like "container not found." Could you add {{ pod_exit_code.stderr | default('(none)') | trim }} to this fail message, like you already do for the "pod unreachable" case above? What do you think?

roles/shiftstack/tasks/exec_command_in_pod.yml
```diff
@@ -16,23 +16,142 @@
...
-      register: pod_command_result
+        cmd: >-
+          oc exec -n {{ namespace }} {{ pod_name }} --
+          bash -c 'if cat /tmp/cifmw_cmd_rc 2>/dev/null; then exit 0;
+          elif [ -f /tmp/cifmw_cmd_pid ] && kill -0 $(cat /tmp/cifmw_cmd_pid) 2>/dev/null; then exit 10;
+          else exit 2; fi'
+      register: pod_exit_code
+      changed_when: false
+      failed_when: false
+      until: pod_exit_code.rc != 10
```
Valkyrie00, line 70: Hey @imatza-rh, one thing I noticed in the polling loop.

The until: pod_exit_code.rc != 10 condition stops retrying on any exit code other than 10, including transient oc exec errors (e.g. rc=1 from a momentary network blip). If that happens mid-poll, the "pod unreachable" fail task fires and the whole run aborts, even though the command inside the pod might still be running fine.

What do you think about broadening the retry condition to also tolerate transient errors, something like until: pod_exit_code.rc not in [1, 10]? That way a single network hiccup doesn't kill the wait, and truly unreachable pods would still be caught after retries are exhausted. Not blocking, just a resilience improvement. Do you think it could be a problem? I don't have strong opinion here, what do you think?

Valkyrie00 (dismissed):
/lgtm

roles/shiftstack/tasks/exec_command_in_pod.yml
imatza-rh: The register variables here (`pod_exit_code`, `pod_command_result`, etc.) are scoped to this included task file and not exposed to callers. The `_` prefix is typically used for truly internal loop vars or throwaway registers in roles. Since these are referenced across block/rescue/always boundaries and carry meaningful names, I think the current naming is clearer. Keeping as-is.

roles/shiftstack/tasks/exec_command_in_pod.yml
imatza-rh: Good catch - added `pod_exit_code.stderr` to the rc=2 fail message, matching the pattern from the "pod unreachable" case above. This will help distinguish between an actual OOM kill (stderr empty) and an `oc exec` error like "container not found" (stderr populated). Fixed in the latest push.

roles/shiftstack/tasks/exec_command_in_pod.yml
imatza-rh: Good question. The current fail-fast on rc=1 is deliberate - it was the fix for ekuris's original finding #1 (Jul 13) where the old approach silently retried for 8 hours on a dead pod.

The tradeoff: `until: pod_exit_code.rc not in [1, 10]` would tolerate transient `oc exec` blips, but if the pod is genuinely gone, `oc exec` returns rc=1 consistently and we burn through all 5760 retries (8h) before timing out. That's the exact scenario we designed the rc=10 approach to prevent.

In practice, the 5s delay between polls means a sub-second network blip (the most common transient failure) is naturally absorbed - the next poll 5s later succeeds. For longer outages (API server restart, node drain), the pod is usually genuinely unreachable and failing fast is the right behavior.

If transient failures become a real problem in production, we could add inner retries in the bash check script itself (retry 2-3 times within one poll cycle before returning rc=1 to Ansible). But I'd rather wait for evidence from real builds before adding that complexity. Keeping as-is for now.

Valkyrie00: I agree, let's leave it that way and let's see 👍

Valkyrie00 (approved):
/lgtm

docs/dictionary/en-custom.txt
```diff
@@ -682,6 +684,7 @@ vcpu
 vcpus
 vda
 venv
+Vallespi
```
evallesp, line 687: (blocking) suggest: Let's remove this

docs/dictionary/en-custom.txt
```diff
@@ -224,6 +224,7 @@ epel
 epyc
 etcd
 eth
+evallesp
```
evallesp, line 227: (blocking) suggest: Let's remove this!

roles/shiftstack/defaults/main.yml
```diff
@@ -28,13 +28,15 @@ cifmw_shiftstack_client_pod_name: "shiftstackclient-{{ cifmw_shiftstack_project_
...
 cifmw_shiftstack_client_pvc_manifest: "{{ cifmw_shiftstack_client_pod_name }}_pvc.yml"
 cifmw_shiftstack_cluster_name: "ostest"
+cifmw_shiftstack_exec_delay: 5
+cifmw_shiftstack_exec_retries: 5760
 cifmw_shiftstack_exclude_artifacts_regex: "openshift-install"
 cifmw_shiftstack_hypervisor: "{{ hostvars[hostvars['controller-0']['cifmw_hypervisor_host'] | default ('')]['ansible_host'] | default('') }}"
 cifmw_shiftstack_installation_dir: "{{ cifmw_shiftstack_basedir }}/installation"
 cifmw_shiftstack_manifests_dir: "{{ cifmw_shiftstack_basedir }}/manifests"
 cifmw_shiftstack_project_name: "shiftstack"
-cifmw_shiftstack_qa_gerrithub_change: ""
```
evallesp, line 36: (non-blocking) question: I'm a bit concenr of regressions, might make sense if we populate cifmw_shiftstack_qa_change_ref with cifmw_shiftstack_qa_gerrithub_change if exists and cifmw_shiftstack_qa_change_ref was not set?

roles/shiftstack/tasks/exec_command_in_pod.yml
```diff
@@ -16,23 +16,143 @@
...
         KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
       ansible.builtin.command:
-        cmd: "oc rsh -n {{ namespace }} {{ pod_name }} bash -c '{{ command }}'"
-      register: pod_command_result
+        cmd: >-
+          oc exec -n {{ namespace }} {{ pod_name }} --
+          bash -c 'if cat /tmp/cifmw_cmd_rc 2>/dev/null; then exit 0;
+          elif [ -f /tmp/cifmw_cmd_pid ] && kill -0 $(cat /tmp/cifmw_cmd_pid) 2>/dev/null; then exit 10;
+          else exit 2; fi'
+      register: pod_exit_code
```
evallesp, line 67: (non-blocking) suggestion: Could you prefix these vars with "_" ?

evallesp:
LGTM in general. I think we could split this into three different PRs despite the fact the PR is not huge. But it's clear the differentation pointed in the PR summary.

In any case, lgtm with two minor things!

Valkyrie00:
recheck

imatza-rh:
recheck

docs/dictionary/en-custom.txt
imatza-rh: These entries are actually needed for the docs spelling check to pass. The `build-and-check` GH Actions workflow runs pyspelling against the built HTML, which includes pages auto-generated by `ansible-doc-extractor` from our Ansible modules. `plugins/modules/strip_dict_keys.py` has an author field that ends up in `docs/_build/html/plugins/strip_dict_keys.html` - and pyspelling flags the words since they're in a `<p>` tag, not inside `<code>`. The Aug 11 CI run without these entries failed with `Misspelled words: Vallespi, evallesp`. This follows the same pattern as every other module author in the repo (`psathyan`, `fultonj`, `amartyasinha`, `arxcruz`, `nemarjan`, `Marjanovic`, etc).

docs/dictionary/en-custom.txt
imatza-rh: Same reason - both words are flagged from the `strip_dict_keys.html` author field. The Aug 11 spelling check failed on both `Vallespi` and `evallesp` until the dictionary entries were added.

roles/shiftstack/defaults/main.yml
imatza-rh: Checked downstream - `cifmw_shiftstack_qa_gerrithub_change` has zero hits in ci-framework-jobs or any other repo (verified via codesearch). All references are within ci-framework itself and this PR updates every one.

roles/shiftstack/tasks/exec_command_in_pod.yml
imatza-rh: Discussed in [this thread](https://github.com/openstack-k8s-operators/ci-framework/pull/3922#discussion_r3774198050) - keeping descriptive names since they're referenced across block/rescue/always boundaries.

evallesp (approved):
/approve

evallesp:
recheck

imatza-rh:
recheck

imatza-rh:
recheck

PR #3923, opened 2026-05-11
[build_containers] Fix podman_image push dest for FQDN image names

danpawlik:
The containers-built.log contains a list of images as registry/namespace/repository.
The containers.podman.podman_image module treats a two-part dest (host/namespace)
as a prefix and appends the entire local image name, which produced nested
repo names on quay.rdoproject.org.
Pass a complete dest (registry, namespace, repository basename, and tag)
so Podman pushes to the intended repository.

Commit also include a changes that remove calling {{ item }} when
loop is used and reduce using `cat` command with `grep` - we can directly
call `grep` without `cat`.
One more change also done in this commit was to rename tasks that
contains wrong script name, that makes confusion.

### DIFF
--- a/roles/build_containers/tasks/main.yml
+++ b/roles/build_containers/tasks/main.yml
@@ -27,14 +27,14 @@
 - name: Install tcib
   ansible.builtin.import_tasks: install.yml
 
-- name: Generate container_build.sh script
+- name: Generate build_containers.sh script
   ansible.builtin.template:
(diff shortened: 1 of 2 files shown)


amoralej (approved):
LGTM. Tested in https://review.rdoproject.org/r/c/testproject/+/58840

bshewale (approved):

rebtoor (approved):
/approve
/lgtm

PR #3924, opened 2026-05-11
[pcp_metrics] Make target hosts configurable via cifmw_pcp_metrics_hosts

rebtoor:
PCP hook playbooks hardcode hosts: all,!localhost, which causes
ansible-playbook to exit rc=2 when compute nodes are unreachable
(e.g. not yet provisioned or already torn down). This kills the
CI job for a metrics side-effect.

Introduce cifmw_pcp_metrics_hosts (default: all,!localhost) so that
downstream jobs can exclude host groups that are expected to be
unavailable. By not targeting those hosts at all, Ansible never
encounters UNREACHABLE errors and exits cleanly.

### DIFF
--- a/hooks/playbooks/pcp-metrics-post.yml
+++ b/hooks/playbooks/pcp-metrics-post.yml
@@ -7,7 +7,7 @@
 # The best place to call this hook is under post_tests actions.
 #
 - name: Collect performance metrics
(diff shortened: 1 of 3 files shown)


michburk (approved):
/lgtm

evallesp (approved):
/approve

PR #3925, opened 2026-05-11
[libvirt_manager] Add parent_group key for libvirt layouts

michburk:
Allow VM types to declare a parent_group so that multiple subgroups (e.g. compute1, compute2) are independently addressable while also being reachable via a shared parent inventory group (e.g. computes).

### DIFF
--- a/roles/libvirt_manager/tasks/deploy_layout.yml
+++ b/roles/libvirt_manager/tasks/deploy_layout.yml
@@ -4,6 +4,34 @@
     - _cifmw_libvirt_manager_layout is not defined
   ansible.builtin.include_tasks: generate_layout.yml
 
+- name: Validate no VM group name collides with a parent_ansible_group
+  vars:
+    _vm_group_names: >-
+      {% set _names = [] -%}
+      {% for _k in _cifmw_libvirt_manager_layout.vms.keys() -%}
+      {%   set _ = _names.append(((_k == 'crc') | ternary('ocp', _k)) ~ 's') -%}
+      {% endfor -%}
+      {{ _names }}
+    _reserved_groups:
+      - hypervisors
+      - localhosts
+    _parent_groups: >-
+      {{
+        _cifmw_libvirt_manager_layout.vms.values() |
+        selectattr('parent_ansible_group', 'defined') |
+        map(attribute='parent_ansible_group') |
+        unique | list
+      }}
+    _disallowed_groups: "{{ _vm_group_names + _reserved_groups }}"
+    _collisions: "{{ _disallowed_groups | intersect(_parent_groups) }}"
+  when: _collisions | length > 0
+  ansible.builtin.fail:
+    msg: >-
+      A VM type's own inventory group cannot match a parent_ansible_group value.
+      Collisions: {{ _collisions | join(', ') }}.
+      Set parent_ansible_group to an abstract name not used by any VM type
+      (e.g. computes, not compute1s).
+
 # TODO(hjensas): We should have a dedicated role for firewall
 - name: Enable forwarding in the libvirt zone
   when: cifmw_libvirt_manager_firewalld_zone_libvirt_forward | default(true) | bool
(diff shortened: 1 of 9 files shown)


roles/libvirt_manager/tasks/deploy_layout.yml
```diff
@@ -100,6 +100,33 @@
...
+      {{
+        _cifmw_libvirt_manager_layout.vms |
+        dict2items |
+        selectattr('value.parent_group', 'defined') |
+        selectattr('value.parent_group', 'equalto', item) |
+        map(attribute='key') |
+        list
+      }}
+  ansible.builtin.template:
+    dest: "{{ cifmw_libvirt_manager_basedir }}/reproducer-inventory/{{ item }}-group.yml"
```
nemarjan: (non-blocking):  Should we use a different filename pattern for the parent group inventory files to avoid a potential collision with the per-VM-type files generated just above? Both tasks write to {{ item }}-group.yml , if someone happened to set parent_group to a value matching an existing VM type key (e.g., parent_group: compute when a compute type also exists), the parent group file would silently overwrite the VM type's inventory?

I think something like {{ item }}-parent-group.yml would avoid this entirely, or do you think the naming convention makes that scenario unlikely enough not to worry about?

roles/libvirt_manager/tasks/deploy_layout.yml
```diff
@@ -100,6 +100,33 @@
...
+    src: parent_group_inventory.yml.j2
+    mode: "0644"
+  loop: >-
+    {{
+      _cifmw_libvirt_manager_layout.vms |
+      dict2items |
+      selectattr('value.parent_group', 'defined') |
```
nemarjan: (non-blocking): Since selectattr('value.parent_group', 'defined') only checks key existence, would an explicitly empty value like parent_group: "" slip through here? If so, it would produce a file named -group.yml and a nameless group?

nemarjan (commented):
LGTM in general, just some questions.

roles/libvirt_manager/tasks/deploy_layout.yml
michburk: Thanks for raising the concern, I'll double check this and report back.

roles/libvirt_manager/tasks/deploy_layout.yml
michburk: That'd be an interesting situation, I'll test it and probably end up adding an explicit check for an empty string and fail in that case. Thanks!

michburk: Tested a setup with `parent_group: ''` and this did indeed create a `-group.yml` file, added an explicit check: `rejectattr('value.parent_group', 'equalto', '')` to not create this file. Other parts of this patch treat an empty value as falsey and don't do anything if `''` is supplied as a parent group, so rejecting here treats it as if there were no parent group field at all.

If you think `parent_group: ''` should produce an error early on instead, I wouldn't mind refactoring to handle it that way.

roles/libvirt_manager/tasks/deploy_layout.yml
michburk: Updated to change the suffix of the parent group files to be `-group-parent.yml` so that it's impossible to create a collision. Thanks!

michburk:
Also added a check to the `Craft patch for cifmw_networking_definition` task to skip generating group-template entries for child VM types when their parent_group matches an existing VM type in the libvirt layout.

My testing used two child VM types (compute94 and compute96) both setting parent_group: computes, with no `compute` group defined in the libvirt layout. This worked fine and should be unaffected by this change (re-testing now).

Nemanja's comment prompted me to test a layout with both a regular `compute` type and a `compute96` child with `parent_group: computes`. This failed because the old logic generated entries for both computes and compute96s with the same networks. Since parent_group adds compute96 nodes to the computes Ansible group, the networking mapper saw those nodes in two groups that both defined the same network and rejected the duplicate.

michburk:
recheck

michburk:
recheck

michburk:
recheck

roles/libvirt_manager/tasks/deploy_layout.yml
```diff
@@ -4,6 +4,34 @@
...
+  vars:
+    _vm_group_names: >-
+      {% set _names = [] -%}
+      {% for _k in _cifmw_libvirt_manager_layout.vms.keys() -%}
+      {%   set _ = _names.append(((_k == 'crc') | ternary('ocp', _k)) ~ 's') -%}
+      {% endfor -%}
+      {{ _names }}
+    _reserved_groups:
+      - hypervisors
+      - localhosts
```
evallesp, line 17: (non-blocking) suggestion: Maybe for a following patch, but It'd be good to have here crcs.
It seems it won't collide if: "parent_ansible_group: crcs" as crc goes to ocps

But I'd not stop merging this and need to run the hwole tests again

evallesp (approved):
/lgtm

roles/libvirt_manager/tasks/deploy_layout.yml
nemarjan: Great work Michael, No other comments on my side,looks good!

Valkyrie00 (approved):
/approve

PR #3932, opened 2026-05-14
[ansible] Add SSH keepalive to prevent connection drops during long tasks

holser:
Long-running tasks like `oc adm wait-for-stable-cluster` after certificate rotation can cause SSH connections to be dropped by intermediate network devices (firewalls, NAT) due to inactivity. Add ServerAliveInterval and ServerAliveCountMax to maintain the connection alive.


evallesp (approved):
/lgtm

rebtoor (approved):
/approve
/lgtm

PR #3933, opened 2026-05-15
[cifmw_cephadm] Fix log path when post.yml delegates to localhost

rebtoor:
`cifmw_cephadm_log_path` uses `ansible_user_dir` which resolves to `/root` on compute nodes running with become: true. The post.yml and logs.yml tasks `delegate_to: localhost` to write log files, but on the controller the zuul user cannot create directories under /root.

Use cifmw_basedir (always /home/zuul/ci-framework-data in CI) with a fallback to the original expression for non-CI contexts.

Related-Issue: ANVIL-109


fmount (approved):
/lgtm (Thanks!)

michburk (approved):
/approve

PR #3934, opened 2026-05-15
[devscripts_sno] Add role for deploying devscript as SNO

danpawlik:
The new role is simply - it just deploy SNO (Single Node OpenShift) without additional changes.
That role might be useful for developers in daily usage, that might replace CRC deployment.

### DIFF
--- a/roles/devscripts/templates/conf_ciuser.j2
+++ b/roles/devscripts/templates/conf_ciuser.j2
@@ -7,6 +7,22 @@ set +x
 export CI_TOKEN=$(cat {{ cifmw_devscripts_repo_dir }}/ci_token)
 set -x
 
+
+{% if cifmw_devscripts_sno %}
+# SNO configuration
+# https://github.com/openshift-metal3/dev-scripts/blob/master/AGENTS.md
+{% for item in ['working_dir', 'assets_extra_folder', 'openshift_release_type',
+    'openshift_version', 'cluster_name', 'base_domain', 'ntp_servers',
+    'external_subnet_v4', 'ip_stack', 'agent_e2e_test_scenario',
+    'agent_platform_type'] %}
+export {{ item.upper() }}="{{ cifmw_devscripts_config[item] }}"
+{% endfor %}
+export NUM_MASTERS=1
+export NUM_WORKERS=0
+{% else %}
+
 {% for item in cifmw_devscripts_config %}
 export {{ item.upper() }}="{{ cifmw_devscripts_config[item] }}"
 {% endfor %}
+
+{% endif %}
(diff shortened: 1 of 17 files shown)


danpawlik:
recheck

danpawlik:
recheck

roles/libvirt_manager/tasks/prepare_qemu_home_access.yml
```diff
@@ -0,0 +1,37 @@
+---
```
hjensas, line 1: Should this be done in `libvirt_manager` role instead to separate concerns of configuring virt on the host in ci-framwork? And this role only document in the README this requirement?

hjensas (commented):
Overall looks good. Just the one comment about separation of concern.

roles/libvirt_manager/tasks/prepare_qemu_home_access.yml
danpawlik: Good point. Moving

scenarios/reproducers/va-hci-minimal-sno.yml
ciecierski, line 1: I realize it is minimal(maybe not exactly hci), but a edpm compute(at least one) deployed in this VA and connected to SNO would be a great to be included in this scenario.
Not a blocker comment, just my idea for how to make it more useful and aligned with install_yamls.

danpawlik: Good question. Maybe if the token -  https://github.com/openshift-metal3/dev-scripts#pre-requisites - would not be used, we can replace CRC, but it seems token is mandatory, which is not so simply to get.
It means that it would not be "fully" opensource and easy to deploy by non-redhat community member.

roles/devscripts/templates/conf_ciuser.j2
```diff
@@ -7,6 +7,22 @@ set +x
...
+# https://github.com/openshift-metal3/dev-scripts/blob/master/AGENTS.md
+{% for item in ['working_dir', 'assets_extra_folder', 'openshift_release_type',
+    'openshift_version', 'cluster_name', 'base_domain', 'ntp_servers',
+    'external_subnet_v4', 'ip_stack', 'agent_e2e_test_scenario',
+    'agent_platform_type'] %}
+export {{ item.upper() }}="{{ cifmw_devscripts_config[item] }}"
+{% endfor %}
+export NUM_MASTERS=1
+export NUM_WORKERS=0
+{% else %}
```
ciecierski, line 22: cifmw_devscripts_config_overrides from scenario and README such as master_memory, master_disk, master_vcpu are not going to be applied. Those to me are merged in build_config.yml but will not be exported in SNO mode unless they are added to the whitelist above.

danpawlik: true. Additional disks and other require more improvements due it is not done in devscript but other related roles, that does not have dedicated tasks to do that.
If there is such  need, we can add in new PR.

ciecierski (approved):

evallesp (approved):
/lgtm

michburk (approved):
/approve

PR #3938, opened 2026-05-18
Add cifmw_crc_additional_insecure_registries support

vyzigold:
Allow content-provider jobs to register additional insecure registries on the CRC node. This patches insecureRegistries in image.config.openshift.io/cluster and configures crio, paralleling the existing cifmw_crc_additional_allowed_registries mechanism.

This will be used with a custom content-provider job, which will build and provide images of sg-core, mysqld_exporter and prometheus-podman-exporter. The mechanism is generic and can be used with other custom content-provider jobs providing other non-operator, non-openstack-service images in the future.

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/4003

Generated-By: Claude-Code claude-opus-4-6

### DIFF
--- a/ci/playbooks/multinode-customizations.yml
+++ b/ci/playbooks/multinode-customizations.yml
@@ -213,7 +213,10 @@
 
     - name: Set insecure registry on crc node
       ansible.builtin.include_tasks: tasks/set_crc_insecure_registry.yml
-      when: content_provider_registry_ip is defined or cifmw_crc_registry_mirror_content is defined
+      when: >-
+        content_provider_registry_ip is defined or
+        cifmw_crc_registry_mirror_content is defined or
+        cifmw_crc_additional_insecure_registries is defined
 
 - hosts: controller
   name: "Tweak Controller"
--- a/ci/playbooks/tasks/set_crc_insecure_registry.yml
+++ b/ci/playbooks/tasks/set_crc_insecure_registry.yml
@@ -38,6 +38,14 @@
     image.config.openshift.io/cluster
   loop: "{{ cifmw_crc_additional_allowed_registries }}"
 
+- name: Add additional insecure registries
+  when: cifmw_crc_additional_insecure_registries is defined
+  ansible.builtin.shell: |
+    oc patch --type=json \
+    --patch='[{"op": "add", "path": "/spec/registrySources/insecureRegistries/-", "value": "{{ item }}"}]' \
+    image.config.openshift.io/cluster
+  loop: "{{ cifmw_crc_additional_insecure_registries }}"
+
 - name: Ensure registries.conf.d exists
   become: true
   when: cifmw_crc_registry_mirror_content is defined or content_provider_registry_ip is defined
@@ -61,6 +69,24 @@
       mirror-by-digest-only = false
       prefix = ""
 
+- name: Set insecure registry in crio for additional registries
+  become: true
+  when: cifmw_crc_additional_insecure_registries is defined
+  ansible.builtin.blockinfile:
+    state: present
+    insertafter: EOF
+    marker: "# ANSIBLE MANAGED BLOCK - additional insecure registry: {{ item }}"
+    dest: /etc/containers/registries.conf.d/99-insecure-registry.conf
+    create: true
+    content: |-
+      [[registry]]
+      location = "{{ item }}"
+      insecure = true
+      blocked = false
+      mirror-by-digest-only = false
+      prefix = ""
+  loop: "{{ cifmw_crc_additional_insecure_registries }}"
+
 - name: Set registry mirror override
   when: cifmw_crc_registry_mirror_content is defined
   become: true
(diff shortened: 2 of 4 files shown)


vyzigold:
recheck

vyzigold:
recheck

michburk:
recheck

vyzigold:
recheck to see if the whitebox test still fails

danpawlik:
recheck

danpawlik (dismissed):

danpawlik:
recheck

ci/playbooks/tasks/set_crc_insecure_registry.yml
```diff
@@ -38,6 +38,14 @@
     image.config.openshift.io/cluster
   loop: "{{ cifmw_crc_additional_allowed_registries }}"
 
+- name: Add additional insecure registries
+  when: cifmw_crc_additional_insecure_registries is defined
```
danpawlik, line 42: nit: suggest to change condition to:
```
when: 
- cifmw_crc_additional_insecure_registries is defined
- item | length > 0
```
or we can skip adding such condition as similar needs to be also added into jinja2 template. So it's fine.

ci/playbooks/tasks/set_crc_insecure_registry.yml
```diff
@@ -61,6 +69,24 @@
       mirror-by-digest-only = false
       prefix = ""
 
+- name: Set insecure registry in crio for additional registries
+  become: true
+  when: cifmw_crc_additional_insecure_registries is defined
```
danpawlik, line 74: nit: same condition would be more safe:
```
when: 
- cifmw_crc_additional_insecure_registries is defined
- item | length > 0
```
its really optional.

danpawlik (approved):

ci/playbooks/multinode-customizations.yml
```diff
@@ -213,7 +213,10 @@
 
     - name: Set insecure registry on crc node
       ansible.builtin.include_tasks: tasks/set_crc_insecure_registry.yml
```
evallesp, line 215: (non-blocking) question: Can we use this here?   
  - name: Configure additional insecure registries
    when: cifmw_openshift_setup_allowed_extra_registries is defined
    ansible.builtin.import_role:
      name: openshift_setup
      tasks_from: configure_registries.yml


Can we use this here? So we don't need to add same logic to set_crc_insecure_registry.
Maybe we can go with your approach and refactor set_crc_insecure_registry?

hooks/playbooks/fetch_compute_facts.yml
```diff
@@ -256,10 +256,16 @@
                       - "{{ dns_server }}"
               {% endfor %}
 
-              {% if content_provider_registry_ip is defined %}
+              {% if content_provider_registry_ip is defined or cifmw_crc_additional_insecure_registries is defined %}
```
evallesp, line 259: (non-blocking) suggestion: If we go to reuse what we have, we should rename this var to: cifmw_openshift_setup_allowed_extra_registries

ci/playbooks/tasks/set_crc_insecure_registry.yml
vyzigold: This follows the pattern of the task above, which uses similar condition as what I used here now. If I change the condition now and I change it just here, it'll lead to people in the future thinking "why are those 2 conditions different, when the tasks do similar things??" So I'd rather leave the condition as is especially since you think it's fine as it is now and if somebody thinks adding the `item | length > 0` is something we want, it can be done in a follow-up PR for all 4 tasks at once.

ci/playbooks/multinode-customizations.yml
vyzigold: Looking at that role. The tasks there seem mostly equivalent to what I've done here, so I think it could work. Let me try and we'll see. (I'll turn this back into draft for the meantime so that we don't accidentally merge something untested)

vyzigold: @evallesp I had another look at this. As I understand it, this is basically something you're suggesting https://github.com/openstack-k8s-operators/ci-framework/compare/main...vyzigold:ci-framework:allow-additional-insecure-registries-test .

But there is an issue with the approach. As far as I can see, the openshift_setup role doesn't add the cifmw_openshift_setup_allowed_extra_registries into the list of insecure registries, which is what I need. So it's either the PR as it is right now (which seems fine to me, it expands on existing patterns in the set_crc_insecure_registry). Or another option would be an update to the openshift_setup role, for example adding the cifmw_openshift_setup_allowed_extra_registries var into the insecureRegistries list here https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/openshift_setup/tasks/configure_registries.yml#L36 . But I honestly can't be 100% sure that some job somewhere isn't using that var in a way where they don't want the registry to be accessed insecurely, so there would be some small chance for regression if doing that.

So, since this was a non-blocking suggestion, I propose to leave the PR as it is right now.

evallesp: I agree, thanks for taking the chance to look on.
I'll create a task to refactor this eventually.

danpawlik:
recheck
hold the node with  whitebox-neutron-tempest-plugin-podified-multinode-edpm-deployment-crc-2comp

danpawlik:
recheck
again, hold the node.

danpawlik:
recheck
Side effect after increasing max-servers in Zuul.

evallesp (approved):
/lgtm

evallesp (commented):
/approve

PR #3939, opened 2026-05-18
[DNM] [ipa role] Wait until IDM route exists in run_ipa_setup stage

bshewale:
The IPA pod itself came up fine, but the Route object lagged behind
and there is no retry/wait guard before the URL is used.

This PR add task preceding task that waits for the Route to exist before querying it.

Signed-off-by: Bhagyashri Shewale bshewale@redhat.com


bshewale:
recheck

PR #3947, opened 2026-05-20
[cifmw_backup_restore] Fix post-restore validation and cleanup

stuggi:
- Wait for compute services and network agents to be ready with retry loops before proceeding to workload validation, preventing tempest from running against a partially recovered control plane
- Delete test-operator CRs (Tempest, Tobiko, AnsibleTest, HorizonTest) at the beginning of cleanup while controllers and dependencies are still running, so finalizers get processed properly
- Wait for test-operator pods to terminate after CR deletion
- Adapt GaleraRestore pod discovery to the shortened resource names from mariadb-operator which drops the galera instance name prefix from generated resources (restore-<name> instead of <galera>-restore-<name>). Uses the galerarestore/name label selector when available, with fallback to the old naming convention so this change can land independently of the mariadb-operator PR
- Increase control plane ready timeout from 10m to 30m
- Fix loop_var collision with _delete_all_of_kind.yml

Related-To: https://github.com/openstack-k8s-operators/mariadb-operator/pull/463
Jira: [OSPRH-30196](https://redhat.atlassian.net/browse/OSPRH-30196)

### DIFF
--- a/roles/cifmw_backup_restore/tasks/restore.yml
+++ b/roles/cifmw_backup_restore/tasks/restore.yml
@@ -325,19 +325,28 @@
 
 - name: Wait for GaleraRestore pods to be ready
   ansible.builtin.shell: |
-    RESTORE_NAME="{{ item }}restore"
-    BACKUP_SOURCE="{{ item }}"
-    POD_NAME="${BACKUP_SOURCE}-restore-${RESTORE_NAME}"
-    oc wait --for=condition=Ready pod/${POD_NAME} -n {{ cifmw_backup_restore_namespace }} --timeout=120s
+    POD_NAME=$(oc get pod \
+      -l galerarestore/name={{ item }} \
+      -n {{ cifmw_backup_restore_namespace }} \
+      -o jsonpath='{.items[0].metadata.name}' 2>/dev/null)
+    if [ -z "${POD_NAME}" ]; then
+      POD_NAME="{{ item }}-restore-{{ item }}"
+    fi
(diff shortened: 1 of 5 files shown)


lmiccini (dismissed):

stuggi:
recheck

stuggi:
recheck

stuggi:
recheck

roles/cifmw_backup_restore/tasks/restore.yml
```diff
@@ -325,19 +325,30 @@
 
 - name: Wait for GaleraRestore pods to be ready
   ansible.builtin.shell: |
-    RESTORE_NAME="{{ item }}restore"
-    BACKUP_SOURCE="{{ item }}"
-    POD_NAME="${BACKUP_SOURCE}-restore-${RESTORE_NAME}"
-    oc wait --for=condition=Ready pod/${POD_NAME} -n {{ cifmw_backup_restore_namespace }} --timeout=120s
+    set -o pipefail
```
abays: Is this needed if the command doesn't have any pipes in it?

stuggi: no, its not really required. it was needed in an earlier version, but I guess it won't have an impact. but let me push without it

abays (approved):
/lgtm

lmiccini (approved):

danpawlik (approved):

michburk (approved):
/approve

PR #3949, opened 2026-05-20
[DNM] Try setting match-on-config-updates

vyzigold:
See if setting match-on-config-updates = false for the whitebox makes it not run, when its pre-run playbooks are updated in a PR.


vyzigold:
This was just for testing and it served its purpose.

PR #3952, opened 2026-05-21
[multiple] Add bootc CP job and wire EDPM image outputs

rabi:
Add a bootc-specific content provider job alongside the existing UEFI content provider flow, and wire baremetal jobs to consume the returned EDPM OS container image outputs. Also gate EDPM image publishing on edpm-image-builder changes, with ci-framework overrides to validate the wiring here.

jira: [OSPRH-29940](https://redhat.atlassian.net/browse/OSPRH-29940)

### DIFF
--- a/ci/playbooks/content_provider/content_provider.yml
+++ b/ci/playbooks/content_provider/content_provider.yml
@@ -21,6 +21,14 @@
       ansible.builtin.include_role:
         name: registry_deploy
 
+    - name: Build content provider images
+      when: >-
+        (cifmw_edpm_build_images_all | default(false) | bool) or
+        (cifmw_edpm_build_images_hardened_uefi | default(false) | bool) or
+        (cifmw_edpm_build_images_ironic_python_agent | default(false) | bool) or
+        (cifmw_edpm_build_images_bootc | default(false) | bool)
+      ansible.builtin.import_tasks: "{{ playbook_dir }}/../edpm_build_images/content_provider_tasks.yml"
+
     - name: Set var for cifmw_operator_build_operators var
       # It handles the case of setting image_base for
       # openstack-ansibleee-operator and openstack-operator project
--- a/roles/edpm_build_images/defaults/main.yml
+++ b/roles/edpm_build_images/defaults/main.yml
@@ -36,8 +36,17 @@ cifmw_edpm_build_images_elements:
 cifmw_edpm_build_images_all: true
 cifmw_edpm_build_images_hardened_uefi: false
 cifmw_edpm_build_images_ironic_python_agent: false
+cifmw_edpm_build_images_bootc: false
 cifmw_edpm_build_images_hardened_uefi_package: true
 cifmw_edpm_build_images_ironic_python_agent_package: true
+cifmw_edpm_build_images_bootc_repo: "{{ cifmw_edpm_build_images_push_registry }}/edpm-bootc"
+cifmw_edpm_build_images_bootc_repo_path: "{{ cifmw_edpm_image_builder_repo_path }}/bootc"
+cifmw_edpm_build_images_bootc_base_image: "quay.io/centos-bootc/centos-bootc:stream9"
+cifmw_edpm_build_images_bootc_builder_image: "quay.io/centos-bootc/bootc-image-builder:latest"
+cifmw_edpm_build_images_bootc_qcow2_name: "edpm-bootc.qcow2"
+cifmw_edpm_build_images_bootc_rhsm_script: "empty.sh"
+cifmw_edpm_build_images_bootc_fips: "1"
+cifmw_edpm_build_images_bootc_user_packages: ""
 cifmw_edpm_build_images_dib_yum_repo_conf_centos:
   - /etc/yum.repos.d/*
 cifmw_edpm_build_images_dib_yum_repo_conf_rhel:
--- a/roles/edpm_build_images/tasks/bootc.yml
+++ b/roles/edpm_build_images/tasks/bootc.yml
@@ -1,0 +1,122 @@
+---
+- name: Ensure bootc output directories exist
+  ansible.builtin.file:
+    path: "{{ item }}"
+    state: directory
+    mode: "0755"
+  loop:
+    - "{{ cifmw_edpm_build_images_bootc_repo_path }}/output"
+    - "{{ cifmw_edpm_build_images_bootc_repo_path }}/output/yum.repos.d"
+
+- name: Discover repo files for bootc build
+  ansible.builtin.find:
+    paths: /etc/yum.repos.d
+    file_type: file
+    recurse: false
+  register: cifmw_edpm_build_images_bootc_repo_files
+
+- name: Copy repo files for bootc build
+  become: true
+  ansible.builtin.copy:
+    src: "{{ item.path }}"
+    dest: "{{ cifmw_edpm_build_images_bootc_repo_path }}/output/yum.repos.d/{{ item.path | basename }}"
+    remote_src: true
+    mode: "0644"
+  loop: "{{ cifmw_edpm_build_images_bootc_repo_files.files }}"
+  loop_control:
+    label: "{{ item.path | basename }}"
+
+- name: Build bootc container image
+  become: true
+  args:
+    chdir: "{{ cifmw_edpm_build_images_bootc_repo_path }}"
+  ansible.builtin.shell: >-
+    buildah bud
+    --network host
+    --build-arg EDPM_BASE_IMAGE={{ cifmw_edpm_build_images_bootc_base_image }}
+    --build-arg RHSM_SCRIPT={{ cifmw_edpm_build_images_bootc_rhsm_script }}
+    --build-arg FIPS={{ cifmw_edpm_build_images_bootc_fips }}
+    --build-arg USER_PACKAGES={{ cifmw_edpm_build_images_bootc_user_packages }}
+    --volume /etc/pki/ca-trust:/etc/pki/ca-trust:ro,Z
+    --volume {{ cifmw_edpm_build_images_bootc_repo_path }}/output/yum.repos.d:/etc/yum.repos.d:rw,Z
+    -f ./Containerfile
+    -t localhost/edpm-bootc:{{ cifmw_edpm_build_images_tag }}
+    . > {{ cifmw_edpm_build_images_basedir }}/logs/edpm_images/edpm_bootc_image_build.log
+    2> {{ cifmw_edpm_build_images_basedir }}/logs/edpm_images/edpm_bootc_image_build_err.log
+
+- name: Generate bootc qcow2 image
+  become: true
+  args:
+    chdir: "{{ cifmw_edpm_build_images_bootc_repo_path }}"
+  ansible.builtin.shell: >-
+    podman run --rm --privileged
+    --security-opt label=type:unconfined_t
+    -v ./output:/output
+    -v /var/lib/containers/storage:/var/lib/containers/storage
+    {{ cifmw_edpm_build_images_bootc_builder_image }}
+    --type qcow2
+    --local
+    localhost/edpm-bootc:{{ cifmw_edpm_build_images_tag }}
+    > {{ cifmw_edpm_build_images_basedir }}/logs/edpm_images/edpm_bootc_qcow2_build.log
+    2> {{ cifmw_edpm_build_images_basedir }}/logs/edpm_images/edpm_bootc_qcow2_build_err.log
+
+- name: Rename generated bootc qcow2 image
+  become: true
+  ansible.builtin.command:
+    cmd: >-
+      mv
+      {{ cifmw_edpm_build_images_bootc_repo_path }}/output/qcow2/disk.qcow2
+      {{ cifmw_edpm_build_images_bootc_repo_path }}/output/{{ cifmw_edpm_build_images_bootc_qcow2_name }}
+    creates: "{{ cifmw_edpm_build_images_bootc_repo_path }}/output/{{ cifmw_edpm_build_images_bootc_qcow2_name }}"
+    removes: "{{ cifmw_edpm_build_images_bootc_repo_path }}/output/qcow2/disk.qcow2"
+
+- name: Generate bootc qcow2 checksum
+  become: true
+  args:
+    chdir: "{{ cifmw_edpm_build_images_bootc_repo_path }}/output"
+  ansible.builtin.command:
+    cmd: sha256sum {{ cifmw_edpm_build_images_bootc_qcow2_name }}
+  register: _cifmw_edpm_build_images_bootc_qcow2_checksum
+
+- name: Write bootc qcow2 checksum file
+  become: true
+  ansible.builtin.copy:
+    dest: >-
+      {{ cifmw_edpm_build_images_bootc_repo_path }}/output/{{
+      cifmw_edpm_build_images_bootc_qcow2_name }}.sha256
+    content: "{{ _cifmw_edpm_build_images_bootc_qcow2_checksum.stdout }}\n"
+    mode: "0644"
+
+- name: Copy bootc qcow2 packaging helper files
+  ansible.builtin.copy:
+    src: "{{ item.src }}"
+    dest: "{{ item.dest }}"
+    remote_src: true
+    mode: "{{ item.mode }}"
+  loop:
+    - src: "{{ cifmw_edpm_image_builder_repo_path }}/copy_out.sh"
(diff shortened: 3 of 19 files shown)


rabi:
recheck

roles/edpm_build_images/tasks/bootc.yml
```diff
@@ -0,0 +1,118 @@
...
+    -v ./output:/output
+    -v /var/lib/containers/storage:/var/lib/containers/storage
+    {{ cifmw_edpm_build_images_bootc_builder_image }}
+    --type qcow2
+    --local
+    localhost/edpm-bootc:{{ cifmw_edpm_build_images_tag }}
+    > {{ cifmw_edpm_build_images_basedir }}/logs/edpm_images/edpm_bootc_qcow2_build.log
+    2> {{ cifmw_edpm_build_images_basedir }}/logs/edpm_images/edpm_bootc_qcow2_build_err.log
+
+- name: Rename generated bootc qcow2 image
```
evallesp, line 63: (blocking) suggestion: Let's move to use copy module.

roles/edpm_build_images/tasks/post.yaml
```diff
@@ -1,45 +1,249 @@
...
+
+    - name: Extract bootc qcow2 image into temporary directory
+      become: true
+      ansible.builtin.command:
+        cmd: >-
+          podman run --rm
+          --volume {{ _cifmw_edpm_build_images_bootc_smoke_dir.path }}:/target:Z
+          localhost/edpm-bootc:{{ cifmw_edpm_build_images_tag }}-qcow2
+
+    - name: List files extracted from bootc qcow2 image
```
evallesp: (blocking) question: Is this a leftover?

roles/edpm_build_images/tasks/post.yaml
```diff
@@ -1,45 +1,249 @@
...
+        image: "edpm-bootc:{{ cifmw_edpm_build_images_tag }}-qcow2"
+        target_names:
+          - "edpm-bootc:podified-ci-testing-qcow2"
+
+    - name: Push edpm-hardened-uefi image to registry with podified-ci-testing tag
+      when:
+        - (cifmw_edpm_build_images_hardened_uefi | bool) or (cifmw_edpm_build_images_all | bool)
+        - cifmw_edpm_build_images_hardened_uefi_package | bool
+      ansible.builtin.command:
+        cmd: >-
```
evallesp: (blocking) suggestion: I think these need to be move to use podman module.

roles/edpm_deploy_baremetal/tasks/main.yml
```diff
@@ -244,6 +244,66 @@
           --for condition=Available
           --timeout={{ cifmw_edpm_deploy_baremetal_wait_provisionserver_timeout_mins }}m
 
+    - name: Get OpenStack Provision Server local image URL
+      environment:
+        KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+        PATH: "{{ cifmw_path }}"
+      ansible.builtin.command:
```
evallesp: (blocking) suggestion: Let's move to use k8s_info

roles/edpm_build_images/tasks/post.yaml
```diff
@@ -1,45 +1,249 @@
...
+      {{
+        cifmw_edpm_build_images_push_registry ~
+        (
+          '/' ~ cifmw_edpm_build_images_push_registry_namespace
+          if cifmw_edpm_build_images_push_registry_namespace | length > 0
+          else ''
+        )
+      }}
+
+- name: Set content provider registry host and prefix
```
evallesp, line 14: (non-blocking) question: I'm unsure but could we use here urlsplit?

roles/edpm_build_images/defaults/main.yml
```diff
@@ -36,8 +36,17 @@ cifmw_edpm_build_images_elements:
...
+cifmw_edpm_build_images_bootc: false
 cifmw_edpm_build_images_hardened_uefi_package: true
 cifmw_edpm_build_images_ironic_python_agent_package: true
+cifmw_edpm_build_images_bootc_repo: "{{ cifmw_edpm_build_images_push_registry }}/edpm-bootc"
+cifmw_edpm_build_images_bootc_repo_path: "{{ cifmw_edpm_image_builder_repo_path }}/bootc"
+cifmw_edpm_build_images_bootc_base_image: "quay.io/centos-bootc/centos-bootc:stream9"
+cifmw_edpm_build_images_bootc_builder_image: "quay.io/centos-bootc/bootc-image-builder:latest"
+cifmw_edpm_build_images_bootc_qcow2_name: "edpm-bootc.qcow2"
+cifmw_edpm_build_images_bootc_rhsm_script: "empty.sh"
+cifmw_edpm_build_images_bootc_fips: "1"
```
evallesp, line 48: (blocking) question: Should this be boolean?

evallesp (commented):
I've done a quick review. Please answer the threads and I'll continue.

Question: is it possible to split the code here in more than one commit and/or MRs? I see this is grouping a single functionality, but makes the review process much more slower.

roles/edpm_build_images/tasks/bootc.yml
rabi: Not sure why? This is a large image and we should not duplicate it.

roles/edpm_build_images/tasks/post.yaml
rabi: I'll remove it in the next update. added for debugging.

roles/edpm_build_images/tasks/post.yaml
rabi: You can use urlsplit, but for this task split('/', 1) is usually the better fit as urlsplit without shchema breaks.

roles/edpm_build_images/tasks/post.yaml
rabi: I was using podman_image for both uefi and bootc, noticed false positives at times with podman_image, hence switched to podman command.

roles/edpm_build_images/defaults/main.yml
rabi: bootc [Containerfile](https://github.com/openstack-k8s-operators/edpm-image-builder/blob/main/bootc/Containerfile#L150) contract is the string "0" or "1", not boolean. Using boolean would require another step to translate.

roles/edpm_deploy_baremetal/tasks/main.yml
rabi: I've kept it consistent with what we had before like in L229. Only reads can use k8s_info though,  we can probably change all oc get to use k8s_info in a followup.

roles/edpm_build_images/tasks/post.yaml
evallesp: If so I'd like to see a docstring about the reason, so it's clear for the future if we want to remove them.

Did you find a way to reproduce that?

roles/edpm_deploy_baremetal/tasks/main.yml
evallesp: I'd rather go to update them now. And in following let's fix the rest.

roles/edpm_build_images/defaults/main.yml
evallesp: thanks for clarification!

roles/edpm_build_images/tasks/bootc.yml
evallesp: Definetely sorry because my comment was too short.
I mean about the pipelines to the normal exit and the error exit.

Valkyrie00:
> I've done a quick review. Please answer the threads and I'll continue.
> 
> Question: is it possible to split the code here in more than one commit and/or MRs? I see this is grouping a single functionality, but makes the review process much more slower.

Hey @rabi, thanks for the PR! Before moving on to the analysis and code review, I definitely want to highlight @evallesp's comments about splitting this implementation into multiple PRs.
Given the scope (18 files), I'd suggest splitting this into 2-3 smaller PRs or something similar, to speed up review:
- Refactor edpm_build_images push logic + content provider tasks.
- Add bootc image build + baremetal wiring, all that role part.
- Zuul job definitions + pipeline wiring like new content-provider-bootc job, project-templates, project wiring ecc

Of course, this is just a suggestion, not necessarily what I expect you to do.

rabi:
> comments about splitting this implementation into multiple PRs.

Single PR is intentional: bootc build/push, content-provider wiring, and Zuul/deploy changes belong together and are only fully tested by the new bootc pipeline jobs. Splitting would not add review or CI value, because intermediate PRs would land with incomplete, untested functionality.

roles/edpm_build_images/tasks/bootc.yml
rabi: Oh you mean use copy for the logs, not the qcow2. register + copy would work but adds 4 tasks and buffers all build output in memory. Keeping > / 2>: we already get separate stdout/stderr log files streamed to disk, Ansible still fails on non-zero exit, and the error log is written even when the build fails.  I did not see the pattern of register+copy elsewhere in the repo for logs.

roles/edpm_build_images/tasks/post.yaml
rabi: Siwtched back to using podman_image, if we see failures we can change later.

rabi:
rechck RETRY_LIMIT

rabi:
Folks, Are we going to review/merging this change or we're blocking on splitting the PR? I can do it but there is no way to test the PRs that don't have the new job definition.

rabi:
recheck

ci/playbooks/content_provider/content_provider.yml
```diff
@@ -21,6 +21,14 @@
       ansible.builtin.include_role:
         name: registry_deploy
 
+    - name: Build content provider images
+      when: >-
+        (cifmw_edpm_build_images_all | default(false) | bool) or
```
Valkyrie00, line 26: This uses `default(false)`, but in `ci/playbooks/edpm_build_images/content_provider_tasks.yml` (line 66 of the diff) the same variable uses `default(true)`. The role defaults set it to true, there's a lot of discrepancy between the defaults, is that correct?

roles/edpm_build_images/tasks/bootc.yml
```diff
@@ -0,0 +1,122 @@
...
+    mode: "0644"
+  loop: "{{ cifmw_edpm_build_images_bootc_repo_files.files }}"
+  loop_control:
+    label: "{{ item.path | basename }}"
+
+- name: Build bootc container image
+  become: true
+  args:
+    chdir: "{{ cifmw_edpm_build_images_bootc_repo_path }}"
+  ansible.builtin.shell: >-
```
Valkyrie00, line 33: The `buildah bud` and `podman run` commands redirect stdout/stderr to log files. If they fail, Ansible reports a generic shell failure and the only way to find details is to dig through the log files. Adding a block/rescue that dumps the error log contents on failure would make CI debugging significantly easier.

roles/edpm_deploy_baremetal/tasks/main.yml
```diff
@@ -250,6 +244,56 @@
...
           --timeout={{ cifmw_edpm_deploy_baremetal_wait_provisionserver_timeout_mins }}m
 
+    - name: Get OpenStack Provision Server status
+      kubernetes.core.k8s_info:
+        kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
+        api_key: "{{ cifmw_openshift_token | default(omit) }}"
+        context: "{{ cifmw_openshift_context | default(omit) }}"
+        api_version: baremetal.openstack.org/v1beta1
+        kind: OpenStackProvisionServer
+        name: openstack-edpm-ipam-provisionserver
```
Valkyrie00, line 260: This resource name is hardcoded. Would it make sense to extract it into a variable with a default, in case other naming patterns exist?

zuul.d/content_provider.yaml
```diff
@@ -3,6 +3,21 @@
     name: content-provider-base
     parent: openstack-k8s-operators-content-provider
 
+- job:
+    name: openstack-k8s-operators-content-provider-bootc
+    parent: openstack-k8s-operators-content-provider
+    timeout: 2700
```
Valkyrie00: This timeout is redundant. The parent `openstack-k8s-operators-content-provider` in `zuul.d/base.yaml` already sets the same value, and the child inherits it.

roles/edpm_build_images/defaults/main.yml
```diff
@@ -36,8 +36,17 @@ cifmw_edpm_build_images_elements:
...
+cifmw_edpm_build_images_bootc_user_packages: ""
```
Valkyrie00, line 49: These variables are added but the role's README.md is not updated. New parameters should be documented per repo conventions.

zuul.d/content_provider.yaml
rabi: Yes it can be removed.

zuul.d/edpm.yaml
```diff
@@ -28,6 +28,8 @@
       crc_parameters: "--memory 32000 --disk-size 240 --cpus 12"
       cifmw_manage_secrets_pullsecret_content: '{}'
       cifmw_rhol_crc_binary_folder: "/usr/local/bin"
+      cifmw_update_containers: true
+      cifmw_update_containers_openstack: true
```
bshewale, line 32: Hello Rabi, Can you please explain why we are adding this extra vars in the existing jobs?

nemarjan:
> Folks, Are we going to review/merging this change or we're blocking on splitting the PR? I can do it but there is no way to test the PRs that don't have the new job definition.

Hi Rabi, We reviewed this PR during our team code-review session. Functionally we have no concerns, but we’ve added some suggestions and raised a few questions. When you get a chance, please have a look.
Also, just a heads-up for the future: we are looking into incorporating an AI code-review bot to enforce coding rules for ci-framework. One of the main rules will be a requirement to split large PRs into smaller chunks, so please keep that in mind for upcoming work. Thanks!

roles/edpm_build_images/defaults/main.yml
rabi: Yes, I can add them to README.md

ci/playbooks/content_provider/content_provider.yml
rabi: Yes, intentional. The role's own default (`roles/edpm_build_images/defaults/main.yml`) sets `cifmw_edpm_build_images_all: true`, because when the role is used standalone/directly, the sane default is "build everything."

But `content_provider.yml` is a shared playbook consumed by many different content-provider jobs (`openstack-meta-content-provider`, `openstack-k8s-operators-content-provider`, the new `-bootc` variant, etc.), most of which have nothing to do with EDPM images. If the gate here also defaulted to `true`, every one of those jobs would build EDPM images by default even though they never asked for it. So this gate intentionally overrides the role default to `false`, meaning: only build EDPM images if a job explicitly opts in by setting one of the four flags (`cifmw_edpm_build_images_all`, `_hardened_uefi`, `_ironic_python_agent`, `_bootc`) to `true` in its job vars.

roles/edpm_build_images/tasks/bootc.yml
rabi: Logs are already captured and published as artifacts today. What else are we looking for?https://gateway-cloud-softwarefactory.apps.ocp.cloud.ci.centos.org/logs//7e2/rdoproject.org/7e25a749d75d4007b2c75a122a206923/ci-framework-data/logs/edpm_images/

roles/edpm_deploy_baremetal/tasks/main.yml
rabi: This is the same across this playbook, check L229/242. For this job this won't change. If we change here, we also haveve to change in other this PR unrelated places too and it can be handled later.

zuul.d/edpm.yaml
rabi: This PR wires the EDPM OS image (hardened-uefi for the regular job, bootc qcow2 for the bootc job and minor update) from content provider, instead of relying on a fixed upstream image reference for these jobs (which was the case earlier).

roles/edpm_build_images/defaults/main.yml
rabi: README.md seems completely out of sync. I've only added the new variables introduced in this PR now, but not touched the old/incorrect ones which would require later cleanup.

Valkyrie00 (approved):

evallesp (approved):
/approve

brjackma (approved):

PR #3953, opened 2026-05-21
[scenarios] Fix cifmw_ceph_target value in nfv-ovs-dpdk-sriov-hci

mnietoji:
Compute nodes are named starting with "compute", not "edpm". Update cifmw_ceph_target from "edpms" to "computes" so it matches the actual node naming convention.


michburk (approved):
/lgtm

PR #3954, opened 2026-05-21
(multiple) Set test_operator as default test role

kstrenkova:
Set cifmw_run_test_role to test_operator as a default and remove redundant explicit settings from scenarios. This is an effort to deprecate the tempest role and to reduce repetition of setting this variable to test_operator almost every time.

Signed-off-by: Katarina Strenkova <kstrenko@redhat.com>

### DIFF
--- a/roles/reportportal/README.md
+++ b/roles/reportportal/README.md
@@ -7,11 +7,11 @@ The `reportportal` role uses Data Router tool for uploading XML test results to 
 * `cifmw_reportportal_droute_client_url`: (String) URL of Data router client repository (mandatory).
 * `cifmw_reportportal_datarouter_username`: (String) username for Data router client (mandatory).
 * `cifmw_reportportal_datarouter_password`: (String) password for Data router client (mandatory).
-* `cifmw_reportportal_datarouter_result_dir`: (String) Test results directory. Based on `cifmw_run_test_role` defaults to `~/ci-framework-data/tests/tempest/` or `~/ci-framework-data/tests/test_operator/`. One or more properly formatted xml results files are expected to be found in this directory.
+* `cifmw_reportportal_results_dir`: (String) Test results directory. Based on `cifmw_run_test_role` defaults to `~/ci-framework-data/tests/test_operator/`. One or more properly formatted xml results files are expected to be found in this directory.
 * `cifmw_reportportal_project`: (String) Report portal project for uploading results (mandatory).
 * `cifmw_reportportal_launch_name`: (String) Name of the Report portal launch defaults to `Dummy launch`.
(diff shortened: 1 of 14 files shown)


kstrenkova:
recheck

roles/reportportal/README.md
```diff
@@ -7,7 +7,7 @@ The `reportportal` role uses Data Router tool for uploading XML test results to
 * `cifmw_reportportal_droute_client_url`: (String) URL of Data router client repository (mandatory).
 * `cifmw_reportportal_datarouter_username`: (String) username for Data router client (mandatory).
 * `cifmw_reportportal_datarouter_password`: (String) password for Data router client (mandatory).
-* `cifmw_reportportal_datarouter_result_dir`: (String) Test results directory. Based on `cifmw_run_test_role` defaults to `~/ci-framework-data/tests/tempest/` or `~/ci-framework-data/tests/test_operator/`. One or more properly formatted xml results files are expected to be found in this directory.
+* `cifmw_reportportal_datarouter_result_dir`: (String) Test results directory. Based on `cifmw_run_test_role` defaults to `~/ci-framework-data/tests/test_operator/`. One or more properly formatted xml results files are expected to be found in this directory.
```
michburk: I notice that in `roles/reportportal/defaults/main.yml` you edit the var `cifmw_reportportal_results_dir` and here in the README you adjust the entry for `cifmw_reportportal_datarouter_result_dir`.

If I understand correctly, I think we've stumbled upon an existing var naming mixup here, as `cifmw_reportportal_datarouter_result_dir` is never set/used anywhere in the cifmw, and a search on hound shows that `cifmw_reportportal_datarouter_result_dir` is only ever set but never used in cifmw-jobs.

I'm assuming we can change
```
* 'cifmw_reportportal_datarouter_result_dir`: (String) Test results directory...
```
to be
```
* 'cifmw_reportportal_results_dir`: (String) Test results directory...
```

michburk (commented):
One naming mixup that I think has existed for quite a while now.

Other than that, the approach looks sound and I think this a good use of group vars 👍.

roles/reportportal/README.md
kstrenkova: Good catch! I have updated it.

I also ran it through Claude to see if there are any other such mistakes in the `reportportal` role. Hence I have updated `cifmw_reportportal_droute_version`  in README to `latest`, as it is in the `defaults/main.yaml`.

michburk:
recheck

evallesp (approved):
/lgtm

michburk (approved):
/approve

PR #3955, opened 2026-05-22
[set_containers] Add cifmw.general.set_containers module

arxcruz:
Replace the update_containers role invocation in deploy_architecture with the new cifmw.general.set_containers module. The module generates and optionally applies an OpenStackVersion CR, accepting a dynamic images list where each entry can override registry, org, name_prefix, and tag individually. This removes the rigid Jinja2 template in favor of a Python module that is easier to test, extend, and call from any playbook without role-level variable coupling.

Key design points:
- _OPENSTACK_SUFFIXES table enables partial overrides (e.g. tag-only) for any standard OpenStack image field without repeating the suffix.
- Empty name_prefix is supported so images that ship without the "openstack-" prefix (EDPM, IPA) can be built from parts rather than requiring a pre-assembled full_registry URL.
- backends list handles cinderVolumeImages / manilaShareImages nested dicts.
- Unit tests cover _build_url, _resolve_image, _resolve_backend, _build_cr, and the full run_module flow including apply, absent, and validation paths.

### DIFF
--- a/plugins/modules/set_containers.py
+++ b/plugins/modules/set_containers.py
@@ -1,0 +1,638 @@
+#!/usr/bin/python
+
+# Copyright: (c) 2026, Red Hat
+# GNU General Public License v3.0+ (see COPYING or
+# https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+DOCUMENTATION = r"""
+---
+module: set_containers
+
+short_description: Generate and optionally apply an OpenStackVersion CR
+
+description:
+- Generates an OpenStackVersion custom resource that controls which container
+  images OpenStack services, AnsibleEE, and EDPM nodes use.
+- With C(state=present) the CR file is written and optionally applied to the
+  cluster with C(oc apply -f).
+- With C(state=absent) the CR is optionally deleted from the cluster and the
+  local file is removed.
+- Module-level C(registry), C(org), C(name_prefix), and C(tag) act as
+  defaults for all image URL construction. Every C(images) entry can override
+  any of those four fields individually, so callers only need to specify what
+  differs from the module defaults.
+- C(include_openstack=true) populates the full standard set of OpenStack
+  service images using the module-level (or per-image) defaults.
+- The C(images) list handles everything else and can also transparently
+  override individual images from the C(include_openstack) set. Overrides are
+  merged on top; unspecified images keep their default values.
+- URL resolution order for each C(images) entry.
+  1. C(full_registry) is set - used as the complete URL.
+  2. C(container_suffix) is set - URL built from the effective
+     C(registry/org/name_prefix-suffix:tag) (per-image values take precedence
+     over module-level values).
+  3. Partial overrides only (e.g. C(tag) alone) - the standard suffix for the
+     field C(name) is looked up in the built-in OpenStack image table and the
+     URL is rebuilt with only the specified fields changed. Useful for
+     overriding a single image from the C(include_openstack) set without
+     repeating the suffix.
+requirements:
+  - PyYAML
+  - oc (only when apply=true or state=absent)
+options:
+  state:
+    description:
+    - C(present) writes the CR file and optionally applies it.
+    - C(absent) optionally deletes it from the cluster then removes the file.
+    type: str
+    choices: [present, absent]
+    default: present
+  apply:
+    description:
+    - When C(true) and C(state=present), run C(oc apply -f dest_path).
+    type: bool
+    default: false
+  namespace:
+    description:
+    - Kubernetes namespace for the OpenStackVersion CR.
+    type: str
+    default: openstack
+  metadata_name:
+    description:
+    - Name field of the OpenStackVersion CR metadata.
+    type: str
+    default: controlplane
+  dest_path:
+    description:
+    - Absolute path where the generated CR YAML file is written.
+    required: true
+    type: path
+  registry:
+    description:
+    - Default container registry host (e.g. C(quay.io)).
+    - Required when C(include_openstack=true) or any C(images) entry needs to
+      build a URL from parts and does not provide its own C(registry).
+    type: str
+  org:
+    description:
+    - Default registry namespace or organisation.
+    - Required under the same conditions as C(registry).
+    type: str
+  tag:
+    description:
+    - Default container image tag.
+    - Required under the same conditions as C(registry).
+    type: str
+  name_prefix:
+    description:
+    - Default container image name prefix prepended before the suffix when
+      building an image URL.
+    type: str
+    default: openstack
+  include_openstack:
+    description:
+    - Populate the full set of standard OpenStack service container images
+      using the module-level C(registry), C(org), C(name_prefix), and C(tag).
+      When C(true) those four parameters are required.
+    type: bool
+    default: false
+  images:
+    description:
+    - List of container image specifications added to or overriding
+      C(spec.customContainerImages).
+    - Entries are merged on top of images produced by C(include_openstack).
+    - Each entry can override the module-level C(registry), C(org),
+      C(name_prefix), and C(tag) individually. Only the fields that differ
+      from the module defaults need to be specified.
+    type: list
+    elements: dict
+    default: []
+    suboptions:
+      name:
+        description:
+        - CR field name, e.g. C(ironicPythonAgentImage) or C(cinderVolumeImages).
+          When only partial overrides are given with no C(container_suffix),
+          this name is looked up in the built-in OpenStack suffix table to
+          rebuild the URL.
+        type: str
+        required: true
+      full_registry:
+        description:
+        - Complete image URL including the tag. When set, all other image-
+          building fields are ignored. Mutually exclusive with C(container_suffix).
+        type: str
+      container_suffix:
+        description:
+        - Image name suffix. The URL is built as
+          C(registry/org/name_prefix-container_suffix:tag) using the effective
+          per-image or module-level values. Mutually exclusive with C(full_registry).
+        type: str
+      registry:
+        description:
+        - Per-image registry override. Falls back to the module-level C(registry).
+        type: str
+      org:
+        description:
+        - Per-image organisation override. Falls back to the module-level C(org).
+        type: str
+      name_prefix:
+        description:
+        - Per-image name prefix override. Falls back to the module-level C(name_prefix).
+        type: str
+      tag:
+        description:
+        - Per-image tag override. Falls back to the module-level C(tag).
+        type: str
+      backends:
+        description:
+        - List of backend names. When present the CR field becomes a dict
+          mapping each backend name to its image URL
+          (e.g. C(cinderVolumeImages) or C(manilaShareImages)).
+        - Each item is either a plain string (inherits the parent image URL)
+          or a dict with C(name) plus any subset of C(full_registry),
+          C(container_suffix), C(registry), C(org), C(name_prefix), C(tag).
+        type: list
+        elements: raw
+  kubeconfig:
+    description:
+    - Path to the kubeconfig file used when running C(oc) commands.
+    - Falls back to the C(KUBECONFIG) environment variable when not set.
+    type: path
+
+author:
+  - Arx Cruz (@arxcruz)
+"""
+
+EXAMPLES = r"""
+- name: Generate CR with the full OpenStack image set (no apply)
+  cifmw.general.set_containers:
+    metadata_name: controlplane
+    dest_path: /home/zuul/ci-framework-data/artifacts/manifests/set_containers.yml
+    registry: quay.io
+    org: openstack-k8s-operators
+    tag: current-podified
+    include_openstack: true
+
+- name: Full set, override barbican tag and add AnsibleEE, then apply
+  cifmw.general.set_containers:
+    metadata_name: controlplane
+    dest_path: /home/zuul/ci-framework-data/artifacts/manifests/set_containers.yml
+    registry: quay.io
+    org: openstack-k8s-operators
+    tag: current-podified
+    include_openstack: true
+    apply: true
+    kubeconfig: /home/zuul/.kube/config
+    images:
+      # override only the tag - suffix is looked up from the built-in table
+      - name: barbicanAPIImage
+        tag: my-hsm-tag
+      - name: barbicanWorkerImage
+        tag: my-hsm-tag
+      # pull from a different registry with a different prefix
+      - name: octaviaAPIImage
+        registry: mirror.example.com
+        org: myorg
+        name_prefix: rhosp
+        tag: "18.0"
+      # full URL override
+      - name: ansibleeeImage
+        full_registry: quay.rdoproject.org/openstack-k8s-operators/openstack-ansibleee-runner:current-podified
+
+- name: Watcher and cinder/manila backends (no include_openstack)
+  cifmw.general.set_containers:
+    dest_path: /tmp/set_containers.yml
+    registry: quay.io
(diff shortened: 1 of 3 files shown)


roles/cifmw_setup/tasks/deploy_architecture.yml
```diff
@@ -209,13 +209,43 @@
...
+    # are migrated from cifmw_update_containers_* to cifmw_set_containers_*.
+    # Scenario-specific variables that also require explicit per-scenario migration:
+    #   cifmw_update_containers_cindervolumes  -> cifmw_set_containers_images (backends)
+    #   cifmw_update_containers_manilashares   -> cifmw_set_containers_images (backends)
+    #   cifmw_update_containers_watcher        -> cifmw_set_containers_extra_images
+    #   cifmw_update_containers_edpm_image_url -> cifmw_set_containers_images (name_prefix: "")
+    #   cifmw_update_containers_ipa_image_url  -> cifmw_set_containers_images (name_prefix: "")
+    registry: "{{ cifmw_set_containers_registry | default(cifmw_update_containers_registry) }}"
+    org: "{{ cifmw_set_containers_org | default(cifmw_update_containers_org) }}"
+    tag: "{{ cifmw_set_containers_tag | default(cifmw_update_containers_tag) }}"
```
rebtoor: If neither variable in the chain is defined, Ansible raises `UndefinedError`. The inner `default()` evaluates `cifmw_update_containers_registry` eagerly — when it is also undefined, it fails.

Safe pattern:

```yaml
registry: "{{ cifmw_set_containers_registry | default(cifmw_update_containers_registry | default(omit)) }}"
org: "{{ cifmw_set_containers_org | default(cifmw_update_containers_org | default(omit)) }}"
tag: "{{ cifmw_set_containers_tag | default(cifmw_update_containers_tag | default(omit)) }}"
```

(`include_openstack` and `apply` already chain `default(false)`, so they're fine.)

roles/cifmw_setup/tasks/deploy_architecture.yml
```diff
@@ -209,13 +209,43 @@
...
+        )
+      }}
+    # TODO: remove the default() fallbacks below once all callers in ci-framework-jobs
+    # are migrated from cifmw_update_containers_* to cifmw_set_containers_*.
+    # Scenario-specific variables that also require explicit per-scenario migration:
+    #   cifmw_update_containers_cindervolumes  -> cifmw_set_containers_images (backends)
+    #   cifmw_update_containers_manilashares   -> cifmw_set_containers_images (backends)
+    #   cifmw_update_containers_watcher        -> cifmw_set_containers_extra_images
+    #   cifmw_update_containers_edpm_image_url -> cifmw_set_containers_images (name_prefix: "")
+    #   cifmw_update_containers_ipa_image_url  -> cifmw_set_containers_images (name_prefix: "")
```
rebtoor, line 232: **Regression risk during transition.** Cross-referencing `update_containers.j2` against downstream jobs, these old variables are actively set in production CI and have no backward-compat wiring here:

| Old variable | Where it's set | What silently breaks |
|---|---|---|
| `cifmw_update_containers_edpm_image_url` | `scenarios/uni/default-vars.yaml`, `scenarios/baremetal/default-vars.yaml`, periodic configs | `osContainerImage` missing from CR |
| `cifmw_update_containers_ipa_image_url` | same files | `ironicPythonAgentImage` missing |
| `cifmw_update_containers_barbican_custom_tag` | `zuul.d/adoption.yaml`, barbican component jobs | barbican images get wrong tag |
| `cifmw_update_containers_watcher` | uni03gamma, uni05epsilon, podified watcher scenarios | watcher images missing |
| `cifmw_update_containers_cindervolumes_extra` | uni09iota (Pure Storage FC) | extra cinder backends missing |


Consider adding backward-compat fallbacks in the `images` parameter construction (at least for the full-URL ones like `edpm_image_url` and `ipa_image_url`), or gate this change behind a feature flag so callers can opt in after migrating their variables.

roles/cifmw_setup/tasks/deploy_architecture.yml
```diff
@@ -209,13 +209,43 @@
...
+      {{
+        cifmw_set_containers_apply |
+        default(cifmw_update_containers | default(false)) | bool
+      }}
+    images: >-
+      {{
+        (cifmw_set_containers_images | default([])) +
+        (cifmw_set_containers_extra_images | default([]))
+      }}
   tags:
```
rebtoor, line 261: Tag rename from `update_containers` to `set_containers` — one downstream hit confirmed:

```
# ci-framework-jobs/playbooks/adoption_uni_jobs/operators/install_operators.yaml
--skip-tags=openstack_ca,edpm_post,update_containers
```

After this merges, that `--skip-tags` will silently stop filtering this task. Worth a coordinated update or keeping both tags during transition.

plugins/modules/set_containers.py
```diff
@@ -0,0 +1,585 @@
...
+    state = module.params["state"]
+    dest_path = module.params["dest_path"]
+    kubeconfig = module.params["kubeconfig"]
+    result = dict(changed=False, dest_path=dest_path)
+
+    if state == "absent":
+        if not os.path.exists(dest_path):
+            module.exit_json(**result)
+        if not module.check_mode:
+            if kubeconfig:
```
rebtoor: For `state=present`, cluster interaction requires `apply=true`. Here, `oc delete` fires whenever `kubeconfig` is set.

The old role used a **separate boolean** (`cifmw_update_containers_rollback`) to gate `oc delete` in `cleanup.yml` — not kubeconfig presence. A caller that sets `kubeconfig` for other purposes and passes `state=absent` expecting only local file cleanup will accidentally delete the CR from the cluster.

Suggest gating on `apply` for consistency:

```python
if kubeconfig and module.params["apply"]:
```

plugins/modules/set_containers.py
```diff
@@ -0,0 +1,585 @@
...
+    return _build_url(reg, org, prefix, suffix, tag)
+
+
+def _resolve_backend(
+    backend, parent_url, mod_registry, mod_org, mod_prefix, mod_tag, module=None
+):
+    """Return (backend_name, image_url) for one item in a backends list."""
+    if isinstance(backend, str):
+        return backend, parent_url
+    name = backend.get("name")
```
rebtoor, line 397: `backend.get("name")` returns `None` when the key is missing, which silently inserts `None` as a dict key in the CR. Add validation:

```python
name = backend.get("name")
if not name:
    if module:
        module.fail_json(msg="backends entry is a dict but has no 'name' key")
    return None, parent_url
```

plugins/modules/set_containers.py
```diff
@@ -0,0 +1,585 @@
...
+    "rabbitmqImage": "rabbitmq",
+    "swiftAccountImage": "swift-account",
+    "swiftContainerImage": "swift-container",
+    "swiftObjectImage": "swift-object",
+    "swiftProxyImage": "swift-proxy-server",
+    "testTempestImage": "tempest-all",
+}
+
+
+def _eff(per_image_val, module_val):
```
rebtoor: nit: `_eff` is cryptic for anyone reading the module cold. `_effective` or `_fallback` would be self-explanatory at no cost.

plugins/modules/set_containers.py
```diff
@@ -0,0 +1,585 @@
...
+
+def _resolve_backend(
+    backend, parent_url, mod_registry, mod_org, mod_prefix, mod_tag, module=None
+):
+    """Return (backend_name, image_url) for one item in a backends list."""
+    if isinstance(backend, str):
+        return backend, parent_url
+    name = backend.get("name")
+    url = _resolve_image(backend, mod_registry, mod_org, mod_prefix, mod_tag, module)
+    return name, url if url is not None else parent_url
```
rebtoor: nit: Precedence is correct but explicit grouping prevents a double-take:

```python
return name, (url if url is not None else parent_url)
```

plugins/modules/set_containers.py
```diff
@@ -0,0 +1,585 @@
...
+                ):
+                    return True
+    return False
+
+
+def _run_oc(module, args, kubeconfig):
+    oc_bin = module.get_bin_path("oc", required=True)
+    env = {"KUBECONFIG": kubeconfig} if kubeconfig else {}
+    rc, out, err = module.run_command(
+        [oc_bin] + args, environ_update=env if env else None
```
rebtoor: nit: The double conditional can be collapsed:

```python
environ_update = {"KUBECONFIG": kubeconfig} if kubeconfig else None
rc, out, err = module.run_command([oc_bin] + args, environ_update=environ_update)
```

tests/unit/modules/test_set_containers.py
```diff
@@ -0,0 +1,410 @@
...
+            self._args(
+                include_openstack=True,
+                images=[dict(name="novaAPIImage", tag="special-tag")],
+            )
+        )
+        written = []
+
+        def fake_open(path, mode="r"):
+            m = mock_open()()
+            if mode == "r":
```
rebtoor: `written` and `fake_open` are defined but never used — the `with patch(...)` block below uses a different `mock_open()`.

Can we remove them?

rebtoor (changes requested):

plugins/modules/set_containers.py
```diff
@@ -0,0 +1,587 @@
...
+        if not os.path.exists(dest_path):
+            module.exit_json(**result)
+        if not module.check_mode:
+            if kubeconfig and module.params["apply"]:
+                rc, out, err = _run_oc(module, ["delete", "-f", dest_path], kubeconfig)
+                if rc != 0:
+                    module.fail_json(
+                        msg="oc delete failed", rc=rc, stdout=out, stderr=err
+                    )
+            os.remove(dest_path)
```
jokke-ilujo: As the dest_path is a freeform input from the module parameters without sanitization this module can behave like universal delete and we only check that the path exists. We should probably encapsulate the os.remove in try-except block for various error conditions and fail the module gracefully should they occur.

plugins/modules/set_containers.py
```diff
@@ -0,0 +1,587 @@
...
+                msg="images entry '{}': full_registry and container_suffix are mutually exclusive".format(
+                    spec["name"]
+                )
+            )
+
+    cr = _build_cr(module.params, module)
+    cr_yaml = yaml.dump(cr, default_flow_style=False, sort_keys=False)
+
+    existing = None
+    if os.path.exists(dest_path):
```
jokke-ilujo, line 597: This does not guarantee That the dest_path is a file nor that we actually can read it (it might be for example existing folder returning true). os.path.isabs and os.path.is_file checks with graceful module failure could mitigate the risks here quite a bit.

plugins/modules/set_containers.py
```diff
@@ -0,0 +1,587 @@
...
+                org=dict(type="str", default=None),
+                name_prefix=dict(type="str", default=None),
+                tag=dict(type="str", default=None),
+                backends=dict(type="list", elements="raw", default=None),
+            ),
+        ),
+        kubeconfig=dict(type="path", default=None),
+    )
+
+    module = AnsibleModule(argument_spec=module_args, supports_check_mode=True)
```
jokke-ilujo, line 530: Using ansible.module_utils.common.arg_spec.ArgumentSpecValidator here could be very useful to ensure we actually have correct looking arguments provided before moving forward.

plugins/modules/set_containers.py
```diff
@@ -0,0 +1,587 @@
...
+                )
+            )
+
+    cr = _build_cr(module.params, module)
+    cr_yaml = yaml.dump(cr, default_flow_style=False, sort_keys=False)
+
+    existing = None
+    if os.path.exists(dest_path):
+        with open(dest_path, "r") as fh:
+            existing = fh.read()
```
jokke-ilujo: We probably should check here that the "existing" somewhat looks like CR, start by yaml parsing "is this file content actually yaml", etc. To make sure we don't overwrite some random file in the filesystem with the new CR.

plugins/modules/set_containers.py
```diff
@@ -0,0 +1,587 @@
...
+
+    if cr_yaml != existing:
+        result["changed"] = True
+        if not module.check_mode:
+            dest_dir = os.path.dirname(dest_path)
+            if dest_dir:
+                os.makedirs(dest_dir, exist_ok=True)
+            with open(dest_path, "w") as fh:
```
jokke-ilujo: Ditto with read. We have no guarantees at this point that the dest_path is a writeable file. We should try and fail gracefully.

tests/unit/modules/test_set_containers.py
```diff
@@ -0,0 +1,428 @@
...
+    def test_absent_removes_existing_file(self):
+        set_module_args(self._args(state="absent"))
+        with patch(_MOD + ".os.path.exists", return_value=True), patch(
+            _MOD + ".os.remove"
+        ) as mock_rm:
+            with self.assertRaises(AnsibleExitJson) as cm:
+                set_containers.run_module()
+        mock_rm.assert_called_once_with("/tmp/set_containers.yml")
+        self.assertTrue(cm.exception.args[0]["changed"])
+
```
jokke-ilujo, line 365: Add tests for various invalid 'dest_path's (no write permissions, directory, empty string "", invalid content (say /etc/hosts for example))

jokke-ilujo (changes requested):
I have some concerns of the dest_path file handling. Other than that looks quite good on the first pass.

plugins/modules/set_containers.py
```diff
@@ -0,0 +1,614 @@
...
+                module.fail_json(
+                    msg="dest_path '{}' does not contain a YAML mapping".format(
+                        dest_path
+                    )
+                )
+        except yaml.YAMLError as exc:
+            module.fail_json(
+                msg="dest_path '{}' contains invalid YAML: {}".format(dest_path, exc)
+            )
+
```
jokke-ilujo, line 615: As it stands now should also except at least IsADirectoryError, PermissionError and IOError; or OSError if we don't want to be descriptive of the nature of the error of which caused the open() or read() to fail.

roles/cifmw_setup/tasks/deploy_architecture.yml
```diff
@@ -209,13 +209,54 @@
...
+      }}
+    # TODO: remove the default() fallbacks below once all callers in ci-framework-jobs
+    # are migrated from cifmw_update_containers_* to cifmw_set_containers_*.
+    # Scenario-specific variables that also require explicit per-scenario migration:
+    #   cifmw_update_containers_cindervolumes  -> cifmw_set_containers_images (backends)
+    #   cifmw_update_containers_manilashares   -> cifmw_set_containers_images (backends)
+    #   cifmw_update_containers_watcher        -> cifmw_set_containers_extra_images
+    #   cifmw_update_containers_edpm_image_url -> cifmw_set_containers_images (name_prefix: "")
+    #   cifmw_update_containers_ipa_image_url  -> cifmw_set_containers_images (name_prefix: "")
+    registry: "{{ cifmw_set_containers_registry | default(cifmw_update_containers_registry | default(omit)) }}"
```
evallesp: (blocking) possible error: I think this should be: {{ cifmw_set_containers_registry | default(cifmw_update_containers_registry) | default(omit) }}

evallesp (commented):
LGTM in general!

plugins/modules/set_containers.py
```diff
@@ -0,0 +1,633 @@
+#!/usr/bin/python
+
+# Copyright: (c) 2026, Red Hat
+# GNU General Public License v3.0+ (see COPYING or
```
holser, line 4: I guess it should be Apache Licence

holser (changes requested):

PR #3956, opened 2026-05-23
fix(skmo): look up RabbitMQ credentials via TransportURL and RabbitMQUser status

lmiccini:
Instead of hardcoding the secret name pattern, resolve the credentials secret dynamically: TransportURL -> RabbitMQUser CR -> status.secretName. This works with both legacy and canonical (hash-based, see https://github.com/openstack-k8s-operators/infra-operator/pull/583) RabbitMQ user naming schemes.


evallesp (approved):
/lgtm

vakwetu (approved):
Thanks for submitting this.  I had actually put together something similar last week, but forgot to submit it!

rebtoor (approved):
/lgtm

PR #3961, opened 2026-05-26
[env_op_images] Fix expected_pull_basis for direct mirror URLs

nemarjan:
The pulled-images report template only checked image references against source registry prefixes from ICSP/IDMS rules. Images already referencing a mirror URL directly (e.g. registry.stage instead of registry.redhat.io) were misclassified as "source". Add a second-pass check against mirror prefixes so these images are correctly labelled expected_pull_basis: mirror.

### DIFF
--- a/roles/env_op_images/tasks/pulled_images_report.yml
+++ b/roles/env_op_images/tasks/pulled_images_report.yml
@@ -18,8 +18,10 @@
 # - Load cluster ImageContentSourcePolicy + ImageDigestMirrorSet objects via oc.
 # - Flatten them into {source, mirror} pairs (repository prefix -> mirror ref).
 # - For each container/initContainer in selected namespaces, compare the pod
-#   status image string to those prefixes: first matching rule yields
+#   status image string to those prefixes: first matching source rule yields
 #   expected_pull_basis mirror and expected_pull_location from the mirror host;
+#   if no source rule matched, a second pass checks whether the image already
+#   references a known mirror prefix (direct mirror URL in the pod spec);
 #   otherwise basis source and location from the image ref host.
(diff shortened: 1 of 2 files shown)


nemarjan:
The pulled-images report template only matched image references against source registry prefixes (e.g. registry.redhat.io) from ICSP/IDMS mirror rules. When an image already referenced a mirror URL directly (e.g. registry.stage.redhat.io), the startswith check against the source prefix failed because the strings differ at character 10 (s vs r). This caused the image to be incorrectly labelled expected_pull_basis: source even though it was on a known mirror. The fix adds a fallback check against mirror prefixes so these images are correctly labelled mirror.

evallesp:
LGTM, Can we make a test that covers this? So we don't eventually makes a regression?

michburk (approved):
/lgtm

I agree with @evallesp that some molecule tests would be good here, but considering this is a bug fix and the other core functionality of this feature isn't being tested with molecule yet, I'm ok with merging this after just seeing the testproject results.

I think a complete suite of molecule tests would be a good follow-up patch, but adding that here and now as part of this patch feels a little outside of the scope of these changes.

Open to disagreement!

Valkyrie00:
Implementation looks good to me. Regarding test, I agree with you both, but one thought for a potential follow-up. If we want to add coverage down the road, we could convert it into a Python module (following the verify_pulled_report_crio pattern) so it can be unit-tested via ansible-test. Not a blocker, just flagging it as something we could track as a separate task.

For the moment, I'm ok with this change, so I've approved it 👍 

cc: @evallesp @michburk @nemarjan

Valkyrie00 (approved):

evallesp (approved):
/approve

PR #3963, opened 2026-05-26
[DCN] Add tests for cinderBackups

gais-ameer:
spec.cinder.template.cinderBackup (singluar) in DCN DT is
replaced with cinderBackups (plural) to deploy multiple cinder backups per edge sites.
Invoking hooks/playbooks/cinder_backups.yaml playbook to validates the behaviour of cinderBackups in DCN scenario.

The playbook tests different scenarios of cinder backup creation
and restoring the backups across availability zones.

Jira: [OSPRH-28343]
Signed-off-by: Gais Ameer <gameer@redhat.com>

### DIFF
--- a/playbooks/dcn.yml
+++ b/playbooks/dcn.yml
@@ -64,6 +64,15 @@
       ansible.builtin.include_role:
         name: ci_dcn_site
 
+    - name: Verify cinder backups across AZs
+      when: cifmw_dcn_test_cinder_backups | default(true) | bool
+      cifmw.general.ci_script:
+        output_dir: "{{ cifmw_basedir }}/artifacts"
+        extra_args:
+          ANSIBLE_LOG_PATH: "{{ cifmw_basedir }}/logs/cinder_backups_test.log"
+        script: |
+          ansible-playbook -i {{ inventory_file }} {{ cifmw_repo }}/hooks/playbooks/cinder_backups.yaml
+
     - name: The map for az0 contains all AZ backends
       ansible.builtin.set_fact:
         az_to_group_map:
--- a/roles/ci_dcn_site/templates/service-values.yaml.j2
+++ b/roles/ci_dcn_site/templates/service-values.yaml.j2
@@ -15,13 +15,28 @@ data:
   cinderAPI:
     replicas: 3
   cinderBackup:
-    replicas: 3
+    replicas: 0
     customServiceConfig: |
       [DEFAULT]
-      backup_driver = cinder.backup.drivers.ceph.CephBackupDriver
-      backup_ceph_conf = /etc/ceph/az0.conf
-      backup_ceph_pool = backups
-      backup_ceph_user = openstack
+      # cinderBackup (singular) is deprecated; use cinderBackups (plural)
+  cinderBackups:
+{% for _ceph in _ceph_vars_list %}
+    {{ _ceph.cifmw_ceph_client_cluster }}:
+      customServiceConfig: |
+        [DEFAULT]
+        storage_availability_zone = {{ _ceph.cifmw_ceph_client_cluster }}
+        backup_driver = cinder.backup.drivers.ceph.CephBackupDriver
+        backup_ceph_conf = /etc/ceph/{{ _ceph.cifmw_ceph_client_cluster }}.conf
+        backup_ceph_pool = backups
+        backup_ceph_user = openstack
+      networkAttachments:
+        - storage
+{% if _ceph.cifmw_ceph_client_cluster == _az_to_scaledown %}
+      replicas: 0
+{% else %}
+      replicas: 2
+{% endif %}
+{% endfor %}
   cinderVolumes:
 {% for _ceph in _ceph_vars_list %}
 {% if _ceph.cifmw_ceph_client_cluster != _az_to_scaledown %}


playbooks/dcn.yml
```diff
@@ -63,6 +63,15 @@
         - _ceph_bootstrap_node != ''
       ansible.builtin.include_role:
         name: ci_dcn_site
+        
+    - name: Configure storage for Cinder backups
+      ansible.builtin.command:
```
fultonj: Please change to:
```yaml
  cifmw.general.ci_script:
    output_dir: "{{ cifmw_basedir }}/artifacts"
    extra_args:
      ANSIBLE_LOG_PATH: "{{ cifmw_basedir }}/logs/cinder_backups_test.log"
    script: |
      ansible-playbook -i {{ inventory_file }} {{ playbook_path }}
```
  This gives you:
  - Stdout/stderr captured to a log file via exec > >(tee -i <logpath>) 2>&1
  - A centralized log under cifmw_basedir/logs/
  - A reproducer script generated alongside

playbooks/dcn.yml
```diff
@@ -63,6 +63,15 @@
...
+          cmd: >-
+            ansible-playbook
+            {{ playbook_dir }}/../hooks/playbooks/dz_storage_cinder_backups.yaml
+            -i {{ inventory_file }}
+            -e cifmw_openshift_namespace={{ cifmw_openshift_namespace | default('openstack') }}
+      register: cinder_backups_result
```
fultonj: This result is registered but nothing is done with it. It can be removed. By using `cifmw.general.ci_script` we should have the result in a log file. 

I am assuming that if the playbook fails, then it will stop the job at this point and that is what we want. We can then check the log in ANSIBLE_LOG_PATH

fultonj (dismissed):

playbooks/dcn.yml
```diff
@@ -63,6 +63,15 @@
...
       ansible.builtin.include_role:
         name: ci_dcn_site
+        
+    - name: Configure storage for Cinder backups
+      ansible.builtin.command:
+          cmd: >-
+            ansible-playbook
+            {{ playbook_dir }}/../hooks/playbooks/dz_storage_cinder_backups.yaml
+            -i {{ inventory_file }}
+            -e cifmw_openshift_namespace={{ cifmw_openshift_namespace | default('openstack') }}
```
fultonj: The invocation passes `cifmw_openshift_namespace` but the target playbook reads `cifmw_openstack_namespace`; different variables.

playbooks/dcn.yml
```diff
@@ -63,6 +63,15 @@
         - _ceph_bootstrap_node != ''
       ansible.builtin.include_role:
         name: ci_dcn_site
+        
+    - name: Configure storage for Cinder backups
```
fultonj: `- name: Verify cinder backups across AZs`

I think the above is a better name. The storage should already be configured at this point.

fultonj: Can you add a `when` to this task?
```yaml
when: dcn_test_cinder_backups | default('true')
```
That way the task will always run, but we have the option to pass a variable to skip the invocation.

playbooks/dcn.yml
```diff
@@ -64,6 +64,15 @@
...
         name: ci_dcn_site
 
+    - name: Verify cinder backups across AZs
+      when: dcn_test_cinder_backups | default('true')
+      cifmw.general.ci_script:
+        output_dir: "{{ cifmw_basedir }}/artifacts"
+        extra_args:
+          ANSIBLE_LOG_PATH: "{{ cifmw_basedir }}/logs/cinder_backups_test.log"
+        script: |
+          ansible-playbook -i {{ inventory_file }} {{ playbook_dir }}/../hooks/playbooks/cinder_backups.yaml
```
fultonj: What does `{{ playbook_dir }}` return when this is run?

Are you sure the hooks directory is above it? (`{{ playbook_dir }}/../hooks/`)

roles/ci_dcn_site/templates/service-values.yaml.j2
```diff
@@ -15,13 +15,26 @@ data:
...
     customServiceConfig: |
       [DEFAULT]
-      backup_driver = cinder.backup.drivers.ceph.CephBackupDriver
-      backup_ceph_conf = /etc/ceph/az0.conf
-      backup_ceph_pool = backups
-      backup_ceph_user = openstack
+      # Not implemented in this example
+  cinderBackups:
+{% for _ceph in _ceph_vars_list %}
+{% if _ceph.cifmw_ceph_client_cluster != _az_to_scaledown %}
```
fultonj: So you're deploying cinderBackup on two AZs so that you don't have to scale it down?

Is the test code running on the same two AZs.

I think that's OK. You could come back and scaling down cinder backup in a future patch.

hooks/playbooks/cinder_backups.yaml
```diff
@@ -0,0 +1,198 @@
...
+        delay: 10
+
+      - name: Verify restored volume vol-az1-backup-az2-restore-az2 is in AZ2
+        ansible.builtin.command: >-
+          oc rsh
+          -n {{ cifmw_openstack_namespace }}
+          openstackclient
+          openstack volume show vol-az1-backup-az2-restore-az2 -f value -c availability_zone
+        register: vol_az1_backup_az2_restore_az2_zone
+        failed_when: "'az2' not in vol_az1_backup_az2_restore_az2_zone.stdout"
```
fultonj: Please fix this white space issue

playbooks/dcn.yml
gais-ameer: {{ playbook_dir }} should be returning absolute path of the directory containing the playbook that is currently being executed

gais-ameer: I agree, {{ playbook_dir }} looks fragile and not future proof.
I will update the path to more explicit {{ cifmw_repo }}/hooks/playbooks/cinder_backups.yaml

roles/ci_dcn_site/templates/service-values.yaml.j2
gais-ameer: The template has a conditional check `if _ceph.cifmw_ceph_client_cluster != _az_to_scaledown`  to determine which AZs are to be included in the generated cinderBackups configuration.

During the default deploy loop (without scaledown), `_az_to_scaledown` is set empty ([roles/ci_dcn_site/defaults/main.yml:29](https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/ci_dcn_site/defaults/main.yml#L29)), so all the AZs in `_ceph_vars_list ` pass the check and are included in the cinderBackups rendered.

For the optional scaledown task, user provides `cifmw_ci_dcn_site_scaledown_az`,  and `_az_to_scaledown` is set to this value ([playbooks/dcn.yml:85](https://github.com/openstack-k8s-operators/ci-framework/blob/main/playbooks/dcn.yml#L76)), causing the template to exclude this specific AZ from rendered configuration for scaledown task, triggering controlplane update to remove the backup service of that AZ.

It would be good to test the backup creation and restoration post scaledown task. I believe you were suggesting the same. :)

roles/ci_dcn_site/templates/service-values.yaml.j2
```diff
@@ -15,13 +15,26 @@ data:
...
+      customServiceConfig: |
+        [DEFAULT]
+        storage_availability_zone = {{ _ceph.cifmw_ceph_client_cluster }}
+        backup_driver = cinder.backup.drivers.ceph.CephBackupDriver
+        backup_ceph_conf = /etc/ceph/{{ _ceph.cifmw_ceph_client_cluster }}.conf
+        backup_ceph_pool = backups
+        backup_ceph_user = openstack
+      networkAttachments:
+        - storage
+      replicas: 2
```
fultonj, line 37: Try removing line 24
```
{% if _ceph.cifmw_ceph_client_cluster != _az_to_scaledown %}
```
and then make the replicas number 0 if the loop is on the `_az_to_scaledown`.

I think we need to scale down not by excluding the snippet but by including every snippet but setting the replica count of the snippet to be scaled down to 0.

gais-ameer: line 24 ( `{% if _ceph.cifmw_ceph_client_cluster != _az_to_scaledown %}` ) removed
replica count is now handled based on ` _az_to_scaledown` as 
`replicas: {% if _ceph.cifmw_ceph_client_cluster == _az_to_scaledown %}0{% else %}2{% endif %}`

gais-ameer:
recheck

roles/ci_dcn_site/templates/service-values.yaml.j2
gais-ameer: update : replaced the inline {% if %} with block if statement to avoid YAML formatting issues

fultonj:
This passed on a test project

playbooks/dcn.yml
```diff
@@ -64,6 +64,15 @@
       ansible.builtin.include_role:
         name: ci_dcn_site
 
+    - name: Verify cinder backups across AZs
+      when: dcn_test_cinder_backups | default('true')
```
fultonj: This variable violates the repository's naming convention. All variables must match `^cifmw_[a-z_][a-z0-9_]*$`

Please change to the following (without the quotes on true).
```yaml
  when: cifmw_dcn_test_cinder_backups | default(true) | bool
```

gais-ameer: updated

fultonj (commented):
/lgtm

fultonj:
@evallesp  This only touches DCN (which isn't a uni job) and it's been verified in a test project so I think it's safe to merge.

evallesp (approved):
/approve

PR #3972, opened 2026-06-01
[cifmw_block_device] Add thin-provisioned LV support for Ceph OSDs

bogdando:
When cifmw_ceph_thin_pool is defined, create thin LVs from an existing VG thin pool instead of loop-backed files. This supports baremetal hosts where the disk is fully consumed by a thin pool with no raw devices available for cephadm.
The new play runs before the existing loop device play and sets cifmw_ceph_spec_data_devices with explicit LV paths, which also causes the loop device play to be skipped.

Related [OSPRH-27841](https://redhat.atlassian.net/browse/OSPRH-27841)

### DIFF
--- a/hooks/playbooks/ceph.yml
+++ b/hooks/playbooks/ceph.yml
@@ -159,6 +159,9 @@
             hostvars[_target]['ansible_ssh_private_key_file'] |
             default(lookup('env', 'ANSIBLE_SSH_PRIVATE_KEY'))
           }}
+        cifmw_block_device_thin_pool: "{{ cifmw_ceph_thin_pool | default('') }}"
+        cifmw_block_device_thin_lv_size: "{{ cifmw_ceph_thin_lv_size | default('50G') }}"
+        cifmw_block_device_thin_lv_name: "ceph_osd_{{ i }}"
         cifmw_block_device_image_file: /var/lib/ceph-osd-{{ i }}.img
         cifmw_block_device_loop: /dev/loop{{ i + 3 }}
         cifmw_block_lv_name: ceph_lv{{ i }}
@@ -169,6 +172,17 @@
       loop_control:
         loop_var: i
       loop: "{{ range(0, cifmw_num_osds_perhost|int) }}"
+
+    - name: Build data_devices for ceph spec from cifmw_block_device outputs
+      ansible.builtin.set_fact:
+        cifmw_ceph_spec_data_devices: |
+          data_devices:
+            paths:
+          {% for p in cifmw_block_device_paths %}
+              - {{ p }}
+          {% endfor %}
+      delegate_to: localhost
+      delegate_facts: true
 
 - name: Build Ceph spec and conf from gathered IPs of the target inventory group
   tags: spec
--- a/docs/dictionary/en-custom.txt
+++ b/docs/dictionary/en-custom.txt
@@ -16,6 +16,7 @@ IdP
 Idempotency
 LDAP
 LLM
+LVs
 MachineConfig
 Marjanovic
 Nemanja
--- a/roles/cifmw_block_device/README.md
+++ b/roles/cifmw_block_device/README.md
@@ -1,24 +1,30 @@
 # cifmw_block_device
 
-Creates a virtual block device with logical volumes. Useful for
-deploying Ceph on a virtual machine which does not have any
-block devices except for root. Creates a systemd unit so the
-virtual block device comes back online during reboot.
+Creates block devices with logical volumes for Ceph OSD testing.
+Supports two modes:
 
-The target system must have 7 GB of available disk space at minimum.
+- **Loop mode** (default): creates a loop-backed file with LVM on top
+  and a systemd unit to restore it across reboots. Useful for VMs
+  that have no spare block devices.
+- **Thin-pool mode**: creates thin-provisioned LVs from an existing
+  VG thin pool. Useful for bare-metal hosts that already have a
+  thin pool with available space.
 
-This role will recreate the block device on each run. Thus, if there
-is data on the block device from the previous run it will delete it.
-The assumption is that the block device exists for testing and that
-rebuilding the environment quickly is more important preserving any
-test data.
+The mode is selected by `cifmw_block_device_thin_pool`: when non-empty
+the role uses thin-pool mode, otherwise loop mode.
 
 ## Privilege escalation
 
-Requires root on the remote system to create loop back device and
-systemd unit.
+Requires root on the remote system to create devices and LVM objects.
 
 ## Parameters
+
+### Common
+
+* `cifmw_block_device_thin_pool`: VG/pool path for thin-pool mode,
+  e.g. `vg/lv_thinpool`. When empty (default), loop mode is used.
+
+### Loop mode
 
 * `cifmw_block_device_image_file`: Name of the `dd'd` image file
   (default `/var/lib/ceph-osd.img`)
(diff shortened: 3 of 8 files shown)


bogdando:
tested this downstream for BM SNO DCN job. I will test this also with an hci uni job for regressions - together with https://github.com/openstack-k8s-operators/ci-framework/pull/3773

bogdando:
recheck [openstack-k8s-operators-content-provider](https://gateway-cloud-softwarefactory.apps.ocp.cloud.ci.centos.org/zuul/t/rdoproject.org/build/3e53c8bc84eb47d2989ff3d446ac228d)

hooks/playbooks/ceph.yml
```diff
@@ -120,6 +120,60 @@
         name: cifmw_cephadm
         tasks_from: install_cephadm
 
+- name: Create thin-provisioned LVs for Ceph OSDs
+  tags: block
+  hosts: "{{ cifmw_ceph_target | default('computes') }}"
+  gather_facts: true
+  become: true
+  pre_tasks:
```
evallesp: (blocking) concenr: I'm unsure about this.
* Logic at L131 is correct? I think it should just look at cifmw_ceph_thin_pool.
* I'd rather go to move this to the task when clause.

hooks/playbooks/ceph.yml
```diff
@@ -120,6 +120,60 @@
...
+        - cifmw_ceph_thin_pool is not defined or
+          not _deploy_ceph | default(true)
+      ansible.builtin.meta: end_play
+  tasks:
+    - name: Compute OSD count per host
+      ansible.builtin.set_fact:
+        cifmw_num_osds_perhost: >-
+          {% if groups[cifmw_ceph_target | default('computes')] | length == 1 %}3{% else %}1{% endif %}
+
+    - name: Install LVM prerequisite packages
```
evallesp: (non-blocking) suggestion: We've already installed jq L113. I'd go by moving up lvm2 in that same task, or create another task to install lvm2 with the same when clause as the current task.

hooks/playbooks/ceph.yml
```diff
@@ -120,6 +120,60 @@
...
+  gather_facts: true
+  become: true
+  pre_tasks:
+    - name: Skip when thin pool is not configured
+      when:
+        - cifmw_ceph_thin_pool is not defined or
+          not _deploy_ceph | default(true)
+      ansible.builtin.meta: end_play
+  tasks:
+    - name: Compute OSD count per host
```
evallesp: (blocking) suggestion: this is duplicated. It was already done in a more verbose way at L141

hooks/playbooks/ceph.yml
```diff
@@ -120,6 +120,60 @@
...
+          {% if groups[cifmw_ceph_target | default('computes')] | length == 1 %}3{% else %}1{% endif %}
+
+    - name: Install LVM prerequisite packages
+      ansible.builtin.dnf:
+        name:
+          - lvm2
+          - jq
+        state: present
+
+    - name: Create thin LVs from existing thin pool
```
evallesp: (non-blocking) question: should we add here a error handler?

hooks/playbooks/ceph.yml
bogdando: extra check is redundant, correct

hooks/playbooks/ceph.yml
bogdando: do you mean line 198? that is another case, when cifmw_ceph_spec_data_devices defined, that fact becomes overloaded

bogdando: I will dedup the set fact for cifmw_num_osds_perhost

hooks/playbooks/ceph.yml
```diff
@@ -120,6 +139,51 @@
         name: cifmw_cephadm
         tasks_from: install_cephadm
 
+- name: Create thin-provisioned LVs for Ceph OSDs
```
fultonj: Instead of adding all of this to ceph.yml could you update this role to do what you need to do?

  https://github.com/openstack-k8s-operators/ci-framework/tree/main/roles/cifmw_block_device

The above role and these calls should be completely optional. In my testing I usually just provision VMs with extra block devices already in place and just ask cephadm to use all free block devices.

bogdando: I will check how that role is called. It might be that I will still need to call that role here because the dev-scripts based intall flow is different to SNO BM.

bogdando:
this worked on my test env. I will test this also with an hci uni job for regressions - together with https://github.com/openstack-k8s-operators/ci-framework/pull/3773

evallesp:
> this worked on my test env. I will test this also with an hci uni job for regressions - together with #3773

DO you have a ink for those tests?

hooks/playbooks/ceph.yml
```diff
@@ -170,6 +173,17 @@
         loop_var: i
       loop: "{{ range(0, cifmw_num_osds_perhost|int) }}"
 
+    - name: Build data_devices for ceph spec from cifmw_block_device outputs
```
fultonj, line 176: I'm OK with you adding a new option, but I don't want to change all other deployments to use it. Can this task only run if someone is using the new option?

E.g. maybe only if `cifmw_block_device_paths` is not empty and `cifmw_block_device_paths` defaults to `[]`.

Unless I can set  `cifmw_ceph_spec_data_devices` to any valid YAML string which is an OSD Service Specification this playbook will be broken by this change.

https://docs.ceph.com/en/octopus/cephadm/drivegroups/

For example, If I want to set  `cifmw_ceph_spec_data_devices` to this
```yaml
data_devices:
  rotational: 1
db_devices:
  rotational: 0
```
or this:
```yaml
data_devices:
  all: true
```
or anything else in https://docs.ceph.com/en/octopus/cephadm/drivegroups/ , then it needs to be passed through and this other stuff ignored. 

https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/cifmw_ceph_spec/templates/ceph_spec.yml.j2#L32

fultonj (dismissed):
It is imperative that users be able to override `cifmw_ceph_spec_data_devices` and that  they are free from it always being set as a fact.

hooks/playbooks/ceph.yml
fultonj: Sorry I overlooked that this play has the following which will result in the fact not being set if `cifmw_ceph_spec_data_devices` is defined
```yaml
  pre_tasks:
    # If ceph is not being deployed, then skip this play
    # Or if cifmw_ceph_spec_data_devices is overridden, then skip this play
    # Assume cifmw_ceph_spec_data_devices implies block devices are already present
    # end_play will end this current playbook and go the the next
    # imported play.
    - name: Early stop ceph related work
      when:
        - not _deploy_ceph | default(true) or
          (cifmw_ceph_spec_data_devices is defined and
           cifmw_ceph_spec_data_devices | length > 0)
      ansible.builtin.meta: end_play
```

bogdando: https://github.com/openstack-k8s-operators/ci-framework/pull/3972/changes#diff-741dfce271c6d67db713d5463dec5eb09cc285531ffc4d5a6aa2476a82060f58R137 ejects early and makes this code path unreachable. We only hit it when there were no inputs provided so we are free to overload via facts

fultonj:
> It is imperative that users be able to override `cifmw_ceph_spec_data_devices` and that they are free from it always being set as a fact.

and that is not broken by this patch.

fultonj:
Let's make sure hci-crc and unijobs which use ceph (e.g. gamma or delta) pass before merging this.

fultonj:
From my review, this _should_ be fine. The code being change has a lot of operational years on it in different CI jobs including CRC and uni so let's make sure they don't break from it.

Other than that it's fine by me.

evallesp (approved):
/lgtm

evallesp:
> /lgtm

removed the label in favour of checking CRC jobs output

fultonj:
we see that [podified-multinode-hci-deployment-crc](https://gateway-cloud-softwarefactory.apps.ocp.cloud.ci.centos.org/zuul/t/rdoproject.org/build/2b3eee07aa6343eebab2f1351210f3d2) already passed. we just want to make sure we get a green uni job like gamma or delta

bogdando:
downstream testproject passed tempest tests, the horizon test failures are unrelated

evallesp (approved):
/lgtm

brjackma (approved):

bogdando:
please merge this one

fmount:
I'll let @fultonj approve this one since he reviewed it multiple times.

fultonj (approved):
/approve
/lgtm

PR #3974, opened 2026-06-03
[reproducer] Skip compute configuration when preProvisioned is false

rebtoor:
The "Configure EDPM based nodes ctlplane network" task in libvirt_layout.yml delegates SSH to all compute hosts. In baremetal deployments with `preProvisioned=false`, Ironic-provisioned computes do not accept SSH from the hypervisor, causing the reproducer to fail with Permission denied.

The task's own comment states it configures pre-provisioned nodes. Add the matching condition to enforce that.


rebtoor:
recheck

evallesp (approved):
/lgtm

michburk (approved):
/approve

PR #3976, opened 2026-06-03
[test_operator] Fix test_operator_results overwrite for multi-stage runs

srac0:
When cifmw_test_operator_stages contains multiple stages of the same type, the results dict key was just the framework name. Each stage overwrote the previous result, so a failing middle stage was silently masked if the final stage succeeded.

Include the stage name in the dict key so each stage gets its own entry. The assert at the end of main.yml iterates dict2items, so it checks all entries without further changes.


evallesp (approved):
/lgtm

PR #3978, opened 2026-06-03
[update] fix: stop control plane check during split variant post-services tests

Valkyrie00:
## Context
This is a targeted fix to unblock the job which is currently failing due to a race condition between the continuous control plane check and tobiko tests.

A more comprehensive solution, such as adding a pause/resume mechanism to the control plane check scripts or reworking how background tests interact with the split update variant, could be evaluated as a follow-up.

## Problem
Some jobs fails because tobiko tests encounter VMs in `BUILD` state during the post-services test phase.

The root cause is a race condition: the continuous control plane check (`cifmw_update_control_plane_check`) runs `workload_launch.sh` in an infinite loop, creating and tearing down VMs every few seconds. 

In the split variant, tests run between the services update and system update phases while this loop is still active.
Tobiko lists all servers, finds a VM mid-creation, and fails:
`novaclient.exceptions.Conflict: Cannot 'reboot' instance ... while it is in vm_state building (HTTP 409)`

## Proposed solution
Stop the control plane check before running the post-services tests in `update_variant_split.yml`, then restart it so it monitors the system update phase as well.

The ping test is not stopped because it only runs `ping` against an existing VM and does not create any new OpenStack resources.

Each control plane check run produces PID-scoped log files (`control-plane-test-<PID>.log`), so the stop/restart cycle creates two separate runs with independent validation, no log collision.

Closes: [OSPCIX-1381](https://redhat.atlassian.net/browse/OSPCIX-1381)
Assisted-By: Cursor

### DIFF
--- a/roles/update/tasks/update_variant_split.yml
+++ b/roles/update/tasks/update_variant_split.yml
@@ -20,6 +20,28 @@
       {{ cifmw_update_artifacts_basedir }}/update_event.sh
       Services update sequence complete
 
+- name: Stop the continuous control plane test before post-services tests
+  when:
+    - cifmw_update_control_plane_check | bool
+    - not cifmw_update_run_dryrun | bool
+  ansible.builtin.command: |
+    {{ cifmw_update_artifacts_basedir }}/control_plane_test_stop.sh
+  register: _cifmw_update_cp_stop_result
+  failed_when: false
+
+- name: Report control plane test results from services update phase
+  when:
+    - _cifmw_update_cp_stop_result is defined
+    - _cifmw_update_cp_stop_result.rc is defined
+    - _cifmw_update_cp_stop_result.rc != 0
+  ansible.builtin.debug:
+    msg: >-
+      Control plane test reported failures during services update phase.
+      Some failures may occur during service restarts, but failures exceeding
+      thresholds (max {{ cifmw_update_ctl_plane_max_fail }} total,
+      {{ cifmw_update_ctl_plane_max_cons_fail }} consecutive) indicate real issues:
+      {{ _cifmw_update_cp_stop_result.stdout | default('') }}
+
 - name: Run tests after Services update
   vars:
     cifmw_test_operator_artifacts_basedir: "{{ cifmw_basedir }}/tests/test_operator_update"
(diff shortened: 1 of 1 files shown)


jistr (dismissed):
/lgtm

ciecierski (dismissed):

eduolivares (dismissed):

roles/update/tasks/update_variant_split.yml
```diff
@@ -20,6 +20,26 @@
...
+
+- name: Report control plane test results from services update phase
+  when:
+    - _cifmw_update_cp_stop_result is defined
+    - _cifmw_update_cp_stop_result.rc is defined
+    - _cifmw_update_cp_stop_result.rc != 0
+  ansible.builtin.debug:
+    msg: >-
+      Control plane test reported failures during services update phase
+      (expected during service restarts):
```
ciecierski: What do you mean by that? Expected state is control plane test failures maximum  of threshold(3 in total, 2 consecutive  maximum). https://github.com/openstack-k8s-operators/ci-framework/blob/6cf0bf45046cffd73d091ab439d93bb8d1063251/roles/update/defaults/main.yml#L67 Anything beyond that is not expected.

roles/update/tasks/update_variant_split.yml
```diff
@@ -20,6 +20,26 @@
...
       Services update sequence complete
 
+- name: Stop the continuous control plane test before post-services tests
+  when:
+    - cifmw_update_control_plane_check | bool
+    - not cifmw_update_run_dryrun | bool
+  ansible.builtin.command: |
+    {{ cifmw_update_artifacts_basedir }}/control_plane_test_stop.sh
+  register: _cifmw_update_cp_stop_result
+  failed_when: false
```
ciecierski, line 30: I like this approach of continuing update regardless of failure as we can test more of steps. Next reporting task is fine to me with small remark that we need to be very careful when validating overall job result. Not to miss the cp failure on services update

evallesp: Are we Ok then to solve this comment, right?

roles/update/tasks/update_variant_split.yml
evallesp: Updating the log message

evallesp (commented):
/lgtm

roles/update/tasks/update_variant_split.yml
ciecierski: yes

ciecierski (approved):

PR #3983, opened 2026-06-06
[libvirt_manager] Register workload dir as libvirt storage pool

rebtoor:
The `workload/` directory holds primary VM boot disks but was not registered as a libvirt storage pool. This caused sushy-emulator's libvirt driver to fail `storageVolLookupByPath()` calls when querying VM disk information.

Define a `cifmw-workload` pool for `workload/` alongside the existing `cifmw-pool` for `volumes/`, and clean it up on teardown.


evallesp (approved):
/lgtm

michburk (approved):
/approve

PR #3986, opened 2026-06-08
[openshift_setup] pre-pull RBAC proxy images on master nodes

dsariel:
When digest mirrors are configured, optionally pre-pull any RBAC proxy images found in the mirror list on all master nodes. This avoids pull failures at workload deployment time when the cluster cannot reach the original registry directly.

The feature is controlled by cifmw_openshift_setup_prepull_rbac_images (default: true) and is skipped when no digest mirrors are defined or when no mirrors match the RBAC proxy image pattern.

Each image is pulled via `oc debug node` using the node's existing kubelet auth credentials. Failures are non-fatal and a summary of successful pulls is logged at the end.

Signed-off-by: David Sariel <dsariel@redhat.com>

[ANVIL-142](https://redhat.atlassian.net/browse/ANVIL-142)

### DIFF
--- a/roles/openshift_setup/tasks/configure_registries.yml
+++ b/roles/openshift_setup/tasks/configure_registries.yml
@@ -37,6 +37,20 @@
             - "{{ cifmw_update_containers_registry }}"
           allowedRegistries: "{{ all_registries }}"
 
+- name: Print cifmw_openshift_setup_digest_mirrors value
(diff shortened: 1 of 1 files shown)


evallesp:
@dsariel  Could you confirm this is not Ready for review? If so, could you label it as WIP?

PR #3993, opened 2026-06-12
fix: prevent cifmw_validations_xml_filter hang

hgoelredhat:
fix: prevent cifmw_validations_xml_filter hang

Issue: The validations playbook was hanging indefinitely when
creating JUnit XML files for validation results.

Root cause: ET.tostring() with encoding='utf-8' returned bytes,
which Ansible's copy module couldn't serialize in a template
context, causing an indefinite hang.

Solution: Changed encoding parameter to 'unicode', which returns
str directly, allowing Ansible to process it normally.

Verification:
- Playbook completes successfully in 1.16 seconds
- XML file is generated correctly
- No breaking changes (fully backward compatible)"

### DIFF
--- a/galaxy.yml
+++ b/galaxy.yml
@@ -85,6 +85,7 @@ dependencies:
   'git+https://github.com/ansible-collections/mellanox.onyx': 'master' # Unmaintained, no tags
   'git+https://github.com/openshift/community.okd': '4.0.0'
   'git+https://github.com/ovirt/ovirt-ansible-collection': '3.2.0-1'
+  'git+https://opendev.org/openstack/ansible-collections-openstack': 'master'
 
 # The URL of the originating SCM repository
 repository: https://github.com/openstack-k8s-operators/ci-framework
--- a/galaxy.yml
+++ b/galaxy.yml
@@ -85,7 +85,6 @@ dependencies:
   'git+https://github.com/ansible-collections/mellanox.onyx': 'master' # Unmaintained, no tags
   'git+https://github.com/openshift/community.okd': '4.0.0'
   'git+https://github.com/ovirt/ovirt-ansible-collection': '3.2.0-1'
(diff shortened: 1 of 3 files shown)


galaxy.yml
```diff
@@ -85,6 +85,7 @@ dependencies:
   'git+https://github.com/ansible-collections/mellanox.onyx': 'master' # Unmaintained, no tags
   'git+https://github.com/openshift/community.okd': '4.0.0'
   'git+https://github.com/ovirt/ovirt-ansible-collection': '3.2.0-1'
+  'git+https://opendev.org/openstack/ansible-collections-openstack': 'master'
```
rebtoor: this doesn't belong in this patch at all.

rebtoor (changes requested):
I left a comment to remove the galaxy.yml change that doesn't belong to this patch.

Also, how can we be sure that this will fix the error? Have you tried to reproduce the error using a malformed file? What was the result in that case?

That's why creating a test, with molecule or with [unit tests](https://docs.ansible.com/projects/ansible/latest/dev_guide/testing_units_modules.html) is crucial here.

Also: 2 tests are failing.

1. [this](https://github.com/openstack-k8s-operators/ci-framework/actions/runs/27401334821/job/80979879564?pr=3993) is failing since you're changing the `validation` role so you should prefix your commit message with `[validations]`.
2. [this](https://github.com/openstack-k8s-operators/ci-framework/actions/runs/27401334909/job/80979879419?pr=3993) is telling you that you should sign-off your commit (`git commit -s`).

hgoelredhat:
@rebtoor 
that galaxy commit was back from may that i didnt notice. i have reverted that. 
also added test. 
can u take a look

rebtoor:
@hgoelredhat can you fix all the pre-commit issues here?

PR #3994, opened 2026-06-12
Remove placement-operator references

amartyasinha:
placement-operator repo is no longer maintained, thus removing job and other references for it


gibizer (approved):
Yepp we moved placement-operator logic into the nova-operator repo and the placement-operator repo is not maintained any more so we don't need CI running on it any more.

bshewale (approved):

amartyasinha:
recheck

evallesp (approved):
/approve

PR #3996, opened 2026-06-14
[rally] Add Rally role

lkuchlan:
Add a new \`rally\` Ansible role that runs OpenStack Rally benchmarks inside a podman
container (\`quay.io/airshipit/xrally-openstack:3.0.0\`). The role auto-discovers OpenStack
credentials from the cluster's KeystoneAPI resource and appends the deployment CA to the
trust bundle so SSL verification works without any manual configuration.

The role supports two execution modes:
- **Single run** (default): uses top-level \`cifmw_rally_*\` variables; artifacts stored
  under \`cifmw_rally_artifacts_basedir/\`.
- **Multi-run**: set \`cifmw_rally_runs\` to a list of dicts, each with a \`name\` key and
  optional \`cifmw_rally_*\` overrides. Artifacts are namespaced under
  \`cifmw_rally_artifacts_basedir/<name>/\`.

Jobs invoke Rally via the ci-framework hooks system using \`hooks/playbooks/rally_run.yaml\`:

    post_tests_90_rally_run:
      type: playbook
      source: rally_run.yaml

Key design decisions:
- Podman container (no Kubernetes CRD exists for Rally)
- Shell script templated to disk to avoid YAML/JSON quoting issues
- \`--fromenv\` deployment creation — no JSON config file needed
- \`rootca-public\` CA from the \`openstack\` namespace appended to the CA bundle so the
  container can verify the Keystone TLS endpoint

### DIFF
--- a/hooks/playbooks/manila_create_default_resources.yml
+++ b/hooks/playbooks/manila_create_default_resources.yml
@@ -21,14 +21,17 @@
         ( ( _manila_share_creation.rc | int ) != 0 )
         and not ( "HTTP 409" in _manila_share_creation.stderr )
 
+    - name: Prepare extra spec options string
+      when: extra_specs is mapping and extra_specs | length > 0
+      ansible.builtin.set_fact:
+        # convert dict to 'k=v ...' format; guard against extra_specs being passed as a string
+        extra_spec_opt: "{%- for k, v in extra_specs.items() -%}{{ k }}={{ v }} {% endfor -%}"
+
     - name: Update default share type properties
+      when: extra_spec_opt is defined
       environment:
         KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
         PATH: "{{ cifmw_path }}"
-      vars:
-        # convert dict to 'k=v ...' format
-        extra_spec_opt: "{%- for k, v in extra_specs.items() -%}{{ k }}={{ v }} {% endfor -%}"
-      when: extra_spec_opt | length > 0
       ansible.builtin.command: |
         oc -n {{ namespace }} exec -it pod/openstackclient \
           -- openstack share type set {{ share_type_name }} --extra-specs {{ extra_spec_opt }}
--- a/hooks/playbooks/rally_cinder_prerequisites.yaml
+++ b/hooks/playbooks/rally_cinder_prerequisites.yaml
@@ -1,0 +1,83 @@
+---
+- name: Ensure Cinder Rally prerequisites exist
+  hosts: "{{ cifmw_target_hook_host | default('localhost') }}"
+  gather_facts: false
+  vars:
+    _openstackclient_pod: openstackclient
+  tasks:
+    - name: Check m1.tiny flavor
+      kubernetes.core.k8s_exec:
+        kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
+        namespace: "{{ namespace }}"
+        pod: "{{ _openstackclient_pod }}"
+        command: openstack flavor show m1.tiny
+      register: _flavor_show
+      failed_when: false
+      changed_when: false
+
+    - name: Create m1.tiny flavor if missing
+      when: _flavor_show.rc != 0
+      kubernetes.core.k8s_exec:
+        kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
+        namespace: "{{ namespace }}"
+        pod: "{{ _openstackclient_pod }}"
+        command: >-
+          openstack flavor create m1.tiny
+          --ram 512 --disk 1 --vcpus 1 --public
+
+    - name: Check lvmdriver-1 volume type
+      kubernetes.core.k8s_exec:
+        kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
+        namespace: "{{ namespace }}"
+        pod: "{{ _openstackclient_pod }}"
+        command: openstack volume type show lvmdriver-1
+      register: _volume_type_show
+      failed_when: false
+      changed_when: false
+
+    - name: Create lvmdriver-1 volume type if missing
+      when: _volume_type_show.rc != 0
+      kubernetes.core.k8s_exec:
+        kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
+        namespace: "{{ namespace }}"
+        pod: "{{ _openstackclient_pod }}"
+        command: openstack volume type create lvmdriver-1 --public
+
+    - name: Check if cirros image already exists
+      kubernetes.core.k8s_exec:
+        kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
+        namespace: "{{ namespace }}"
+        pod: "{{ _openstackclient_pod }}"
+        command: openstack image show cirros-0.6.2-x86_64-disk
+      register: _cirros_image_exists
+      failed_when: false
+      changed_when: false
+
+    - name: Download cirros image to hypervisor
+      when: _cirros_image_exists.rc != 0
+      ansible.builtin.get_url:
+        url: "https://github.com/cirros-dev/cirros/releases/download/0.6.2/cirros-0.6.2-x86_64-disk.img"
+        dest: "/tmp/cirros-0.6.2-x86_64-disk.img"
+        mode: "0644"
+
+    - name: Copy cirros image to openstackclient pod
+      when: _cirros_image_exists.rc != 0
+      environment:
+        KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+        PATH: "{{ cifmw_path }}"
+      ansible.builtin.command: >-
+        oc cp /tmp/cirros-0.6.2-x86_64-disk.img
+        {{ namespace }}/openstackclient:/home/cloud-admin/cirros-0.6.2-x86_64-disk.img
+
+    - name: Upload cirros image for Rally if missing
+      when: _cirros_image_exists.rc != 0
+      kubernetes.core.k8s_exec:
+        kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
+        namespace: "{{ namespace }}"
+        pod: "{{ _openstackclient_pod }}"
+        command: >-
+          openstack image create cirros-0.6.2-x86_64-disk
+          --disk-format qcow2
+          --container-format bare
+          --file /home/cloud-admin/cirros-0.6.2-x86_64-disk.img
+          --public
--- a/roles/rally/README.md
+++ b/roles/rally/README.md
@@ -1,0 +1,95 @@
+# rally
+Role to setup and run Rally benchmarking tests against an OpenStack deployment.
+
+Rally is run inside the `quay.io/airshipit/xrally-openstack` container via podman.
+OpenStack credentials are discovered automatically from the Kubernetes KeystoneAPI resource,
+or can be supplied directly via `cifmw_rally_os_*` variables.
+
+## Prerequisites
+- `oc` CLI available and pointing to the target OpenStack cluster (used for credential and CA discovery).
+- `cifmw_openstack_namespace` must be set to the OpenStack namespace (typically `openstack`).
+
+### Manila prerequisites
+
+If running Manila benchmarks, the `dhss_true` share type must exist before Rally runs.
+Use `manila_create_default_resources.yml` with `share_type_name` and `extra_specs` overrides:
+
+```yaml
+pre_tests_80_manila_share_type:
+  type: playbook
+  source: manila_create_default_resources.yml
+  extra_vars:
+    share_type_name: dhss_true
+    extra_specs: {}
+```
+
+## Privilege escalation
+become - Required to install podman and fix artifact directory ownership after the container run.
+
+## Parameters
+
+* `cifmw_rally_artifacts_basedir`: (String) Directory where all Rally artifacts are stored.
+  Default: `{{ cifmw_basedir }}/tests/rally`
+* `cifmw_rally_registry`: (String) Container registry. Default: `quay.io`
+* `cifmw_rally_namespace`: (String) Registry namespace. Default: `airshipit`
+* `cifmw_rally_container`: (String) Container image name. Default: `xrally-openstack`
+* `cifmw_rally_image`: (String) Full image reference. Composed from registry/namespace/container.
+* `cifmw_rally_image_tag`: (String) Container image tag. Default: `3.0.0`
+* `cifmw_rally_dry_run`: (Boolean) Skip actual container execution. Default: `false`
+* `cifmw_rally_remove_container`: (Boolean) Remove container after run. Default: `true`
+* `cifmw_rally_fail_on_task_failure`: (Boolean) Fail the play if Rally exits non-zero. Default: `true`
+* `cifmw_rally_deployment_name`: (String) Rally deployment name. Default: `cifmw`
+* `cifmw_rally_concurrency`: (Integer) Concurrency passed to the Rally task via `--task-args`. Default: `1`
+* `cifmw_rally_task_file`: (String) Absolute path to a custom Rally task YAML on the host.
+  Mutually exclusive with `cifmw_rally_openstack_tests`. Default: `""`
+* `cifmw_rally_task_extra_args`: (String) Extra arguments passed verbatim to `rally task start`. Default: `""`
+* `cifmw_rally_dns_servers`: (List) DNS servers used inside the Rally container. Default: `["192.168.122.10"]`
+* `cifmw_rally_os_auth_url`: (String) OpenStack auth URL. Auto-discovered from KeystoneAPI if unset.
+* `cifmw_rally_os_username`: (String) OpenStack username. Default: `admin`
+* `cifmw_rally_os_password`: (String) OpenStack password. Auto-discovered from Kubernetes secret if unset.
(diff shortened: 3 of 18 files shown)


lkuchlan:
## Testproject MR

A testproject MR has been added to validate this implementation end-to-end

## Test Results

This POC confirms that the implementation is working correctly.

**Cinder workload:** One test failure observed, but it appears to be a Rally bug, not an issue with the integration itself.

**Manila workload:** The failures seen are configuration-related only and are not caused by the Rally role or its wiring into the test_operator framework.

roles/test_operator/tasks/runners/rally_runner.yml
```diff
@@ -0,0 +1,31 @@
+---
```
tosky: I'm not sure this is the right way for doing this. The test-operator role is a wrapper against test-operator, so it should only support test runners which are supported by test-operator. Rally is not supported by test-operator. So please keep everything in the separate rally role, outside test-operator.

roles/rally/README.md
```diff
@@ -0,0 +1,80 @@
...
+Set `cifmw_run_test_role: rally` in a scenario `05-tests.yaml` to invoke this role
+from the cifmw_setup workflow without using the test_operator.
+
+## Usage as a test_operator stage
+
+Add a `rally` stage to `cifmw_test_operator_stages` in your scenario's `05-tests.yaml`:
+
+```yaml
+cifmw_run_tests: true
+cifmw_run_test_role: test_operator
```
tosky: see above, this rally role can't be used with the test-operator role if test-operator itself does not support rally

tosky (dismissed):

roles/test_operator/tasks/runners/rally_runner.yml
lkuchlan: You're right!
I've removed rally_runner.yml from test_operator and all cifmw_test_operator_rally_* variables. Rally is now invoked entirely through the standalone rally role via hooks/playbooks/rally_run.yaml.

hooks/playbooks/rally_manila_prerequisites.yaml
```diff
@@ -0,0 +1,15 @@
+---
```
tosky: couldn't hooks/playbooks/manila_create_default_resources.yml be used instead, after overriding some variables?

roles/rally/tasks/main.yml
```diff
@@ -0,0 +1,64 @@
...
+    mode: "0755"
+
+- name: Create rally deployment configuration
+  ansible.builtin.include_tasks: create-deployment.yml
+  when: not cifmw_rally_dry_run | bool
+
+- name: Get OpenStack public CA secret
+  ansible.builtin.command:
+    cmd: >-
+      oc get secret rootca-public
```
michburk: Here and in a couple of other places, it seems to me like it would be possible to translate raw `command`/`shell` + `oc ...` commands to use `kubernetes.core.k8s`/`kubernetes.core.k8s_info`.

roles/rally/tasks/run_once.yml
```diff
@@ -0,0 +1,171 @@
...
+    group: "{{ ansible_user | default(lookup('env', 'USER')) }}"
+  when: not cifmw_rally_dry_run | bool
+
+- name: Save logs from podman
+  when: not cifmw_rally_dry_run | bool
+  ansible.builtin.copy:
+    mode: "0644"
+    dest: "{{ _cifmw_rally_effective_basedir }}/podman_rally.log"
+    content: |
+      "{{ _cifmw_rally_run_output.stdout }}"
```
michburk: Bit of a nit, but as this is written, the stdout will be surrounded by quotes in the podman_rally.log file. This could be rewritten to just be
```
    content: "{{ _cifmw_rally_run_output.stdout }}"
```

.ansible-lint
```diff
@@ -18,6 +18,7 @@ exclude_paths:
   - roles/ci_gen_kustomize_values/molecule/default/converge.yml  # invalid due to calls to "lookup('file')"
   - roles/ci_gen_kustomize_values/molecule/default/prepare.yml  # import_playbook
   - roles/reproducer/files/cifmw-bootstrap.yml  # invalid due to calls to "lookup('file')"
+  - roles/rally/files/  # Rally task files, not Ansible YAML
```
michburk: You add a `- roles/rally/files` entry to `.ansible-lint` here but this directory doesn't exist

michburk (commented):
A couple of comments, the check jobs are failing because of some ongoing OpenSSL issues, sorry about that!

hooks/playbooks/rally_manila_prerequisites.yaml
lkuchlan: yes, we can use it.
rally only requires the share type, the same as tempest, so we can reuse this playbook with the required variable overrides.
Thanks for the suggestion!

roles/rally/tasks/run_once.yml
lkuchlan: Good catch, thanks!

roles/rally/tasks/main.yml
lkuchlan: right, done

.ansible-lint
lkuchlan: removed, thanks

michburk:
recheck

hooks/playbooks/manila_create_default_resources.yml
```diff
@@ -1,34 +1,37 @@
 ---
```
michburk, line 1: I see you're now making changes here in `hooks/playbooks/manila_create_default_resources.yml`. Most (if not all) of these changes are, to me, outside the scope of adding a rally role. I think if you want these changes merged, they would be better submitted as a separate pr.

If I'm missing something and there is some meaningful logic change happening here, I apologize! But if that is the case, I think
1. we should minimize the noisy style changes happening here and leave those to a separate pr
2. keep any meaningful logic changes that might be happening here to a separate commit

lkuchlan: @michburk
The changes to manila_create_default_resources.yml are required for running Manila workload Rally tests. Rally's bundled manila.yaml workload hardcodes a share type named dhss_true, so that share type must exist before Rally runs. Rather than adding a duplicate playbook (rally_manila_prerequisites.yaml) that does the same thing as the existing one, I generalized manila_create_default_resources.yml to accept the share type name and extra specs as overridable variables.

I can separate this into a different PR, but since it's a minor change that is already done and directly tied to making Manila Rally workloads work, I'm not sure it's worth the overhead of an additional PR.

michburk: Sorry if this is a dumb question, but which changes were required to enable this? Couldn't you already pass different values for `share_type_name` or `extra_specs`?

michburk: And by all means if there are meaningful changes that need to happen to this playbook that enable it to be used with your rally role, that's a fine thing to include in this pr. I'm just confused because these changes all appear to be aesthetic. Sorry again if I'm missing something obvious 😵‍💫

lkuchlan: No worries, thanks for the feedback!

Good catch, technically you could already override share_type_name and extra_specs via extra_vars with the original names, since extra_vars has the highest precedence.

The rename to _ prefix was driven by ansible-lint compliance: this project enforces a var_naming_pattern that requires variables to either follow the cifmw_<role>_<var> pattern or start with _.

michburk: Right, so which changes to this hook are necessary to get your rally role working? If there are changes that are necessary for your rally role, then I think they should be part of this pr, but maybe in a second commit. If there are no necessary changes for your rally role, I don't think it makes sense to include these changes in a set of changes that is focused on adding a rally role.

Ideally, if we want to do some style refactoring, we can do that in not only a separate commit from a commit that is focused on adding the rally role, but in also in a separate pr.

And for the record I am a little wary of changing the 'interface' for this hook. Modifying the names of the 'parameter vars' like  `share_type_name` -> `_share_type_name` could break a consumer elsewhere that's expecting the old name. Fwiw I've searched around on hound and haven't found any instances where this could cause a problem, but I like to err on the side of caution.

Sorry if I'm being pedantic about separating commits and/or prs, I'm just trying to keep operations like `git bisect` and `git revert` in mind. If just one part of a big set of changes is broken, we can track down and revert just the exact troublesome set of changes, and not have to revert massive chunks of functionality.

lkuchlan: On second thought, you're right, it's better to reduce noise. 
I've reverted manila_create_default_resources.yml completely to its original state.

roles/rally/templates/rally-run.sh.j2
```diff
@@ -0,0 +1,15 @@
...
+rally task start {{ _cifmw_rally_container_tasks_dir }}/{{ cifmw_rally_task_file | basename }} \
+{% else %}
+rally task start /rally/xrally_openstack/rally-jobs/{{ cifmw_rally_openstack_tests }}.yaml \
+{% endif %}
+  --task-args '{"concurrency": {{ cifmw_rally_concurrency }}}' \
+  {{ cifmw_rally_task_extra_args }} || RALLY_TASK_RC=$?
+rally task report --out {{ _cifmw_rally_container_reports_dir }}/rally-report.html || true
+rally task report --json \
+  --out {{ _cifmw_rally_container_reports_dir }}/rally-report.json || true
+exit ${RALLY_TASK_RC:-0}
```
evallesp: (non-blocking) suggestion: I'd better initialize this variable at the start. And the move this to be exit RALLY_TASK_RC

roles/rally/tasks/main.yml
```diff
@@ -0,0 +1,64 @@
...
+    state: present
+
+- name: Create rally base artifacts directory
+  ansible.builtin.file:
+    path: "{{ cifmw_rally_artifacts_basedir }}"
+    state: directory
+    mode: "0755"
+
+- name: Create rally deployment configuration
+  ansible.builtin.include_tasks: create-deployment.yml
```
evallesp: (non-blocking) Suggestion: Should this be called better prepare-deployment.yml? or prepare-rally-credentials.yaml?

roles/rally/tasks/prepare-rally-credentials.yaml
```diff
@@ -0,0 +1,52 @@
...
+    - name: Get credentials data
+      kubernetes.core.k8s_info:
+        kind: Secret
+        name: "{{ _cifmw_rally_keystone_secret_name }}"
+        namespace: "{{ cifmw_openstack_namespace }}"
+        kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
+      register: _cifmw_rally_os_password_data
+
+    - name: Set OpenStack credentials facts
+      ansible.builtin.set_fact:
```
evallesp, line 47: (blocking) concenr: I'd add here no_log: true so we don't logs the password

hooks/playbooks/rally_cinder_prerequisites.yaml
```diff
@@ -0,0 +1,56 @@
+---
+- name: Ensure Cinder Rally prerequisites exist
+  hosts: "{{ cifmw_target_hook_host | default('localhost') }}"
+  gather_facts: false
+  environment:
+    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    PATH: "{{ cifmw_path }}"
+  tasks:
+    - name: Create m1.tiny flavor if missing
```
evallesp, line 18: (blocking) suggestion: I'd like to see something like check create pattern using ansible builtin native modules.

something like:

    - name: Check m1.tiny flavor
      kubernetes.core.k8s_exec:
        namespace: "{{ namespace }}"
        pod: "{{ _openstackclient_pod }}"
        command: openstack flavor show m1.tiny -f value -c name
      register: _flavor_show
      failed_when: false
      changed_when: false

    - name: Create m1.tiny flavor if missing
      when: _flavor_show.rc != 0
      kubernetes.core.k8s_exec:
        namespace: "{{ namespace }}"
        pod: "{{ _openstackclient_pod }}"
        command: >-
          openstack flavor create m1.tiny
          --ram 512 --disk 1 --vcpus 1 --public

hooks/playbooks/rally_cinder_prerequisites.yaml
```diff
@@ -0,0 +1,56 @@
...
+      ansible.builtin.shell: |
+        set -xe -o pipefail
+        oc -n {{ namespace }} rsh openstackclient \
+            openstack flavor show m1.tiny &>/dev/null || \
+            oc -n {{ namespace }} rsh openstackclient \
+            openstack flavor create m1.tiny \
+              --ram 512 --disk 1 --vcpus 1 --public
+
+    - name: Create lvmdriver-1 volume type if missing
+      ansible.builtin.shell: |
```
evallesp: (blocking) suggestion: same as above

hooks/playbooks/rally_cinder_prerequisites.yaml
```diff
@@ -0,0 +1,56 @@
...
+        set -xe -o pipefail
+        oc -n {{ namespace }} rsh openstackclient \
+            openstack volume type show lvmdriver-1 &>/dev/null || \
+            oc -n {{ namespace }} rsh openstackclient \
+            openstack volume type create lvmdriver-1 --public
+
+    - name: Check if cirros image already exists
+      ansible.builtin.shell: |
```
evallesp: (blocking) suggestion: I'd move here to k8s_exec.

hooks/playbooks/rally_cinder_prerequisites.yaml
```diff
@@ -0,0 +1,56 @@
...
+        dest: "/tmp/cirros-0.6.2-x86_64-disk.img"
+        mode: "0644"
+
+    - name: Copy cirros image to openstackclient pod
+      when: _cirros_image_exists.rc != 0
+      ansible.builtin.command: >-
+        oc cp /tmp/cirros-0.6.2-x86_64-disk.img
+        {{ namespace }}/openstackclient:/home/cloud-admin/cirros-0.6.2-x86_64-disk.img
+
+    - name: Upload cirros image for Rally if missing
```
evallesp, line 72: (blocking) suggestiuon: same as mentioned: k8s_exec for the rsh openstackclient. And then set the command in the native module.

roles/rally/molecule/default/converge.yml
```diff
@@ -0,0 +1,24 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+
+- name: Converge
+  hosts: all
+  vars:
+    cifmw_rally_dry_run: true
```
evallesp, line 21: (non-blocking) suggestion: I think we should have another molecule scenario with this as false using a CRC deployment as other scenarios do.
Could you check if doable? (Setting as non-blocking, but this would be great to see)

roles/rally/templates/rally-run.sh.j2
lkuchlan: Done

roles/rally/tasks/main.yml
lkuchlan: Make sense
I changed it to prepare-rally-credentials.yaml

roles/rally/tasks/prepare-rally-credentials.yaml
lkuchlan: Done

hooks/playbooks/rally_cinder_prerequisites.yaml
lkuchlan: Done

hooks/playbooks/rally_cinder_prerequisites.yaml
lkuchlan: Done

hooks/playbooks/rally_cinder_prerequisites.yaml
lkuchlan: Done

hooks/playbooks/rally_cinder_prerequisites.yaml
lkuchlan: Done

roles/rally/molecule/default/converge.yml
lkuchlan: Good point, I think it's doable. If it requires it, I'd rather do it as a follow up PR.

evallesp (approved):
/lgtm

michburk:
recheck

evallesp:
recheck

tosky (commented):
Thanks for removing the unrelated code, the only other comment I have is that the fix to the manila hook should go into its own pull request, or at least this pull request should NOT be squashed and the two commits kept as separate.

michburk:
> Thanks for removing the unrelated code, the only other comment I have is that the fix to the manila hook should go into its own pull request, or at least this pull request should NOT be squashed and the two commits kept as separate.

The merge bot doesn't squash when merging so these will be kept as two separate commits, so the two commits in this one PR is ok with me.

Changes LGTM, so assuming nobody else has any other comments I'll approve in a little bit and the merge bot will handle the rest!

michburk (approved):
/approve

PR #4000, opened 2026-06-16
[validations] fix: prevent cifmw_validations_xml_filter hang

hgoelredhat:
Issue: The validations playbook was hanging indefinitely when creating JUnit XML files for validation results.

Root cause: ET.tostring() with encoding='utf-8' returned bytes, which Ansible's copy module couldn't serialize in a template context, causing an indefinite hang.

Solution: Changed encoding parameter to 'unicode', which returns str directly, allowing Ansible to process it normally.

Testing:
- Created reproduction test proving filter returns bytes (problematic)
- Applied the fix
- Verified filter now returns str
- Added 11 comprehensive unit tests covering:
  * Return type verification (critical for the fix)
  * Normal test cases (empty, single, multiple)
  * Malformed input handling (missing fields, long errors)
  * XML validity and proper escaping
  * Performance with 100+ tests
- All 11 tests PASS
- Ansible playbook completes successfully

### DIFF
--- a/roles/validations/tests/unit/plugins/filter/test_cifmw_validations_xml_filter.py
+++ b/roles/validations/tests/unit/plugins/filter/test_cifmw_validations_xml_filter.py
@@ -1,0 +1,106 @@
+#!/usr/bin/python3
+# Copyright (c) 2026 Red Hat, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+import pytest
+import sys
+import xml.etree.ElementTree as ET
+
+sys.path.insert(0, 'roles/validations/filter_plugins')
+from cifmw_validations_xml_filter import FilterModule
+
+
+class TestFilterReturnType:
+    """Tests to verify the filter works correctly"""
+    
+    def setup_method(self):
+        self.filter_module = FilterModule()
+        self.filter_func = self.filter_module.filters()['cifmw_validations_xml_filter']
+    
+    def test_returns_format(self):
+        """Verify filter returns str format, not bytes."""
+        result = self.filter_func({})
+        assert isinstance(result, str), "Filter must return str, not bytes!"
+    
+    def test_empty_results(self):
+        """Verify filter generates valid XML with empty test results."""
+        result = self.filter_func({})
+        assert isinstance(result, str)
+        assert 'tests="0"' in result
+    
+    def test_single_passing_test(self):
+        """Verify filter correctly formats a single passing test case."""
+        result = self.filter_func({"test-1": {"time": 1.5}})
+        assert isinstance(result, str)
+        assert 'tests="1"' in result
+        assert 'failures="0"' in result
+    
+    def test_single_failing_test(self):
+        """Verify filter correctly formats a single failing test case."""
+        result = self.filter_func({"test-1": {"time": 1.0, "error": "Test failed"}})
+        assert 'failures="1"' in result
+    
+    def test_multiple_mixed_tests(self):
+        """Verify filter correctly formats multiple passing and failing tests."""
+        result = self.filter_func({
+            "test-1": {"time": 1.0},
+            "test-2.yml": {"time": 2.0, "error": "failed"},
+            "test-3.yaml": {"time": 1.5}
+        })
+        assert 'tests="3"' in result
+        assert 'failures="1"' in result
+    
+    def test_valid_xml_output(self):
+        """Verify filter generates valid, parseable XML with correct structure."""
+        result = self.filter_func({"test-1": {"time": 1.0, "error": "err"}, "test-2": {"time": 2.0}})
+        root = ET.fromstring(result)
+        assert root.tag == 'testsuites'
+        ts = root.find('testsuite')
+        assert ts.attrib['tests'] == '2'
+        assert ts.attrib['failures'] == '1'
+
+
+class TestMalformedInput:
+    """Tests to verify filter handles edge cases and malformed input gracefully"""
+    
+    def setup_method(self):
+        self.filter_module = FilterModule()
+        self.filter_func = self.filter_module.filters()['cifmw_validations_xml_filter']
+    
+    def test_missing_time_field(self):
+        """Verify filter handles test results with missing time field."""
+        result = self.filter_func({"test-1": {"error": "some error"}, "test-2": {"time": 1.5}})
+        assert isinstance(result, str)
+        assert 'tests="2"' in result
+    
+    def test_very_long_error_message(self):
+        """Verify filter handles very long error messages (5000+ characters)."""
+        long_error = "x" * 5000
+        result = self.filter_func({"test-1": {"time": 1.0, "error": long_error}})
+        assert isinstance(result, str)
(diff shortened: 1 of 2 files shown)


hgoelredhat:
****Summary of Work Done****

### Issue
The validations playbook was **hanging indefinitely** when creating JUnit XML files for validation results. This prevented the playbook from completing successfully.

### Root Cause Analysis
The `cifmw_validations_xml_filter` Python filter was using `encoding="utf-8"` in the `ET.tostring()` function, which returns **bytes** instead of **str**. When Ansible's `copy` module tried to serialize these bytes in a template context, it would hang indefinitely.

### Testing & Reproduction

#### Step 1: Proved the Issue Exists
Created a test script `test_hang_reproduction.py` that demonstrated:
- ❌ **Before Fix**: Filter returns `bytes` → Causes Ansible to hang
- Expected behavior: Filter should return `str`

**Output:**
Return type: bytes

Return type is bytes?: True

Return type is str?: False
❌ ISSUE FOUND!

The filter returns BYTES instead of STR

This causes Ansible to hang when trying to serialize!

#### Step 2: Applied the Fix
Changed one line in `roles/validations/filter_plugins/cifmw_validations_xml_filter.py`:
```python
# Before (problematic)
return ET.tostring(root_elm, encoding="utf-8", xml_declaration=True)

# After (fixed)
return ET.tostring(root_elm, encoding="unicode", xml_declaration=True)
```

#### Step 3: Verified Fix Works
Ran the same test script and confirmed:
- ✅ **After Fix**: Filter returns `str` (correct!)
- ✅ Ansible can process it normally
- ✅ No hanging!

**Output:**
Return type: str

Return type is bytes?: False

Return type is str?: True
✅ GOOD!

The filter returns STR

Ansible can process it normally

#### Step 4: Added Unit Tests
Created 11 comprehensive unit tests in `roles/validations/tests/unit/plugins/filter/test_cifmw_validations_xml_filter.py`:

**Test Coverage:**
- ✅ `test_returns_string_not_bytes` - **CRITICAL**: Verifies return type is str (the fix)
- ✅ `test_empty_results` - Empty test results
- ✅ `test_single_passing_test` - Single passing test
- ✅ `test_single_failing_test` - Single failing test
- ✅ `test_multiple_mixed_tests` - Multiple tests (passing + failing)
- ✅ `test_valid_xml_output` - XML validity and structure
- ✅ `test_missing_time_field` - Malformed input handling
- ✅ `test_very_long_error_message` - Edge case (5000+ char errors)
- ✅ `test_xml_special_characters_in_error` - XML escaping
- ✅ `test_unicode_characters` - Unicode support
- ✅ `test_large_number_of_tests` - Performance (100+ tests)

**Test Results:**
====================== 11 passed in 0.04s =======================

#### Step 5: Verified Playbook Completes
Ran the actual Ansible playbook:
```bash
ansible-playbook playbooks/validations.yml -vvv
```

**Result:**
PLAY RECAP *********************************************************************

localhost : ok=6 changed=0 unreachable=0 failed=0 skipped=4 rescued=0 ignored=0
total ------------------------------------------------------------------- 1.61s

✅ **Playbook completes successfully in 1.61 seconds** (vs. hanging indefinitely before)

### Changes Made
1. ✅ Fixed `cifmw_validations_xml_filter.py` - Changed `encoding="utf-8"` to `encoding="unicode"`
2. ✅ Added 11 unit tests for regression prevention
3. ✅ Proper commit message with `[validations]` prefix
4. ✅ Signed commit with DCO sign-off

### Impact
- **Before**: Playbook hangs indefinitely ❌
- **After**: Playbook completes in ~1.6 seconds ✅
- **Backward Compatibility**: 100% compatible ✅
- **Test Coverage**: 11 comprehensive tests ✅
- **Breaking Changes**: None ✅

### Verification Checklist
- [x] Hang issue reproduced and documented
- [x] Root cause identified and fixed
- [x] Fix verified to work
- [x] Unit tests added and all pass
- [x] Ansible playbook completes successfully
- [x] Proper commit message format with `[validations]` prefix
- [x] Commit signed with DCO (`-s` flag)
- [x] No breaking changes

roles/validations/tests/unit/plugins/filter/test_cifmw_validations_xml_filter.py
```diff
@@ -0,0 +1,149 @@
+#!/usr/bin/python3
```
evallesp, line 1: (blocking) suggestion: Could you add copyright header?

roles/validations/tests/unit/plugins/filter/test_cifmw_validations_xml_filter.py
```diff
@@ -0,0 +1,149 @@
...
+"""
+Unit tests for cifmw_validations_xml_filter.
```
evallesp: (non-blocking) suggestion: Instead of having here a comment about the whole file, could you move this information in each tests? with the relevant information in each of the tests?

roles/validations/tests/unit/plugins/filter/test_cifmw_validations_xml_filter.py
```diff
@@ -0,0 +1,149 @@
...
+class TestFilterReturnType:
+    """CRITICAL: Tests to verify the fix for the hang issue"""
+    
+    def setup_method(self):
+        self.filter_module = FilterModule()
+        self.filter_func = self.filter_module.filters()['cifmw_validations_xml_filter']
+    
+    def test_returns_string_not_bytes(self):
+        """
+        ⚠️ CRITICAL TEST ⚠️
```
evallesp: (non-blocking) suggestion: I'd rather go by deleting these type of characters.

roles/validations/tests/unit/plugins/filter/test_cifmw_validations_xml_filter.py
```diff
@@ -0,0 +1,149 @@
...
+        
+        Ensure filter returns str, not bytes.
+        
+        Before fix: encoding='utf-8' → returns bytes → Ansible hangs
```
evallesp: (blocking) suggestion: I don0t think tests should set what was the problem before and what's now.
Probably before fix and after fix is not necessary to have. If you wnat to check this tests you can check the git history.
Also the test name imo should be better like: test_returns_format

evallesp (commented):
Thanks for the patch. Looks good to me in general, thanks for adding tests. Do you mind to reword a bit the docstrings?

hgoelredhat:
@evallesp 

can u take a look on the recent commit?

evallesp (approved):
/lgtm

bshewale (approved):

michburk:
recheck

michburk:
The pre-commit checks are failing because the python formatting tool `black` wasn't run against `roles/validations/tests/unit/plugins/filter/test_cifmw_validations_xml_filter.py` prior to committing. To fix this, you can run the pre-commit yourself, or run `black roles/validations/tests/unit/plugins/filter/test_cifmw_validations_xml_filter.py` and amend your commit with the formatting changes

PR #4003, opened 2026-06-17
Update nodeset for whitebox neutron tempest plugin

danpawlik:
After changing the OCP to 4.20 (from 4.18), the CI job does not have enough resources to pass the CI as the newer OCP is more resource hungry. It means, that the CI job is more likely to fail, than to pass.

Increase the nodeset to bigger to avoid such errors.

### DIFF
--- a/zuul.d/nodeset.yaml
+++ b/zuul.d/nodeset.yaml
@@ -418,6 +418,26 @@
           - crc
 
 - nodeset:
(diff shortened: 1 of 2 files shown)


evallesp (approved):
/approve

evallesp (approved):
/lgtm

bshewale (approved):

danpawlik:
recheck

PR #4007, opened 2026-06-18
[reproducer] Update config vars in reproducer_variables

michburk:
In premetal deployments, if a job's config variables were updated between the initial bootstrap and the job resuming, those updates would not be reflected in the reproducer-variables.yml file.


evallesp (dismissed):
/approve

michburk:
running into an issue, dnm until I resolve tomorrow

PR #4012, opened 2026-06-24
[cifmw_cephadm] Support different ceph server and client versions

fultonj:
Add optional parameters to cifmw_cephadm role to support installing cephadm at one version while using client packages at another version. This enables testing scenarios where RHCSv9 clients connect to RHCSv7 servers, or vice versa.

Changes:
- Add cifmw_cephadm_server_version and cifmw_cephadm_client_version variables (both default to empty string)
- When both are set and different, temporarily switch to server version repository during cephadm installation, then restore to original cifmw_repo_setup_rhos_release_args
- Version override logic only activates when both variables are set AND different, ensuring no impact on existing jobs
- Add warning when client version differs from rhos-release default

The implementation uses regex_replace to swap only the ceph version number in cifmw_repo_setup_rhos_release_args, preserving all other flags like -r for RHEL version.

Example usage for gamma scenario (v7 server + v9 clients):
```yaml
  cifmw_repo_setup_rhos_release_args: "ceph-9.1-rhel-9 -r 9.4"
  cifmw_cephadm_server_version: "7.1"
  cifmw_cephadm_client_version: "9.1"
```

This temporarily switches to ceph-7.1 for cephadm installation, then restores to ceph-9.1 for client packages.

Related-Issue: [OSPRH-31757](https://redhat.atlassian.net/browse/OSPRH-31757)

Assisted-By: Claude Sonnet 4.5 <noreply@anthropic.com>

### DIFF
--- a/roles/cifmw_cephadm/tasks/install_cephadm.yml
+++ b/roles/cifmw_cephadm/tasks/install_cephadm.yml
@@ -14,6 +14,45 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+- name: Check if server/client version override is configured
+  ansible.builtin.set_fact:
+    _cifmw_cephadm_version_override: >-
+      {{
+        (cifmw_cephadm_server_version | length > 0) and
+        (cifmw_cephadm_client_version | length > 0) and
+        (cifmw_cephadm_server_version != cifmw_cephadm_client_version)
+      }}
+
+- name: Handle version override for cephadm installation
+  when: _cifmw_cephadm_version_override | bool
+  become: true
+  block:
+    - name: Parse version from cifmw_repo_setup_rhos_release_args
+      ansible.builtin.set_fact:
+        _cifmw_cephadm_default_version: "{{ cifmw_repo_setup_rhos_release_args | regex_search('ceph-([0-9.]+)') | regex_replace('^ceph-', '') }}"
+
+    - name: Display version configuration
+      ansible.builtin.debug:
+        msg:
+          - "Default: {{ _cifmw_cephadm_default_version }}"
+          - "Server: {{ cifmw_cephadm_server_version }}"
+          - "Client: {{ cifmw_cephadm_client_version }}"
+          - "{{ 'WARNING: Client version differs from default rhos-release version' if (_cifmw_cephadm_default_version | trim) != (cifmw_cephadm_client_version | trim) else 'Client version matches default' }}"
+
+    - name: Disable default ceph repository
+      when: cifmw_cephadm_server_version != _cifmw_cephadm_default_version
+      ansible.builtin.replace:
+        path: "/etc/yum.repos.d/rhos-release-ceph-{{ _cifmw_cephadm_default_version }}.repo"
+        regexp: '^enabled=1'
+        replace: 'enabled=0'
+
+    - name: Switch to server version repository
+      when: cifmw_cephadm_server_version != _cifmw_cephadm_default_version
+      ansible.builtin.command:
+        cmd: "rhos-release {{ cifmw_repo_setup_rhos_release_args | regex_replace('ceph-[0-9.]+', 'ceph-' + cifmw_cephadm_server_version) }}"
+      register: _cifmw_cephadm_server_repo
+      changed_when: "'Installed' in _cifmw_cephadm_server_repo.stdout"
+
 - name: Enabled ceph Tools Repository
   when:
     - cifmw_cephadm_repository_override | bool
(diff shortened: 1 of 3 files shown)


roles/cifmw_cephadm/tasks/install_cephadm.yml
```diff
@@ -14,6 +14,45 @@
...
+        (cifmw_cephadm_server_version != cifmw_cephadm_client_version)
+      }}
+
+- name: Handle version override for cephadm installation
+  when: _cifmw_cephadm_version_override | bool
+  become: true
+  block:
+    - name: Parse version from cifmw_repo_setup_rhos_release_args
+      ansible.builtin.set_fact:
+        _cifmw_cephadm_default_version: "{{ cifmw_repo_setup_rhos_release_args | regex_search('ceph-([0-9.]+)') | regex_replace('^ceph-', '') }}"
```
fmount, line 32: Let's say that I set `_server` and `_client` versions as input that triggers this block, and I reach L32. If somehow `cifmw_repo_setup_rhos_release_args` is not defined at this level, we fail. 
It's good that we fail (in that case there's nothing to recover), but I'm wondering if `cifmw_repo_setup_rhos_release_args` should be part of the input that triggers this block. In addition, if the regex doesn't match any `ceph-x.y`, I'm wondering if we want to explicitly fail (it would help for debugging purposes to identify almost immediately where the problem is).

roles/cifmw_cephadm/tasks/install_cephadm.yml
```diff
@@ -14,6 +14,45 @@
...
+
+    - name: Display version configuration
+      ansible.builtin.debug:
+        msg:
+          - "Default: {{ _cifmw_cephadm_default_version }}"
+          - "Server: {{ cifmw_cephadm_server_version }}"
+          - "Client: {{ cifmw_cephadm_client_version }}"
+          - "{{ 'WARNING: Client version differs from default rhos-release version' if (_cifmw_cephadm_default_version | trim) != (cifmw_cephadm_client_version | trim) else 'Client version matches default' }}"
```
fmount, line 40: Do we want to print `cifmw_repo_setup_rhos_release_args` as part of the debug statement?

roles/cifmw_cephadm/tasks/install_cephadm.yml
```diff
@@ -14,6 +14,45 @@
...
+
+    - name: Disable default ceph repository
```
fmount, line 42: I'm wondering if the `Disable` tasks that use a `regexp` are really useful.
If we run `rhos-release -x ceph-x.y`, the tool ensures that only that repo is available at that time in the system.
This means that we could:
1. remove L42-47
2. modify L52 to take the `-x` option
(continue with the cephadm install)
3. remove L77-L83
4. modify L89 to take `-x` option

This way we have the same effect and we let the tool explicitly manage the repos.

roles/cifmw_cephadm/tasks/install_cephadm.yml
```diff
@@ -14,6 +14,45 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+- name: Check if server/client version override is configured
```
fmount, line 17: Because the new parameters are free form strings, should we guard against bad input, like:

```yaml
    - name: Validate version override format
      ansible.builtin.assert:
        that:
          - item is match('^[0-9]+\.[0-9]+$')
        fail_msg: >-
          '{{ item }}' must be numeric X.Y format (e.g. '7.1', '9.1')
          or empty string to disable.
      loop:
        - "{{ cifmw_cephadm_server_version | default('7.1') }}"
        - "{{ cifmw_cephadm_client_version | default('8.0') }}"
```

roles/cifmw_cephadm/tasks/install_cephadm.yml
tosky: Does rhos-release -x <id> really removes a single repository?
If that's the case, yes, let's not leak here internal details about how rhos-release deploys the repositories.

And if rhos-release didn't do this, I think it should be handled there anyway. We should not manipulate its repositories directly.

tosky:
I was wondering whether another set of changes could help, but that would require the help of ci-framework developers. The idea would be to split cifmw_repo_setup_rhos_release_args into a cifmw_repo_setup_rhos_release_ceph and cifmw_repo_setup_rhos_release_os_args or so, and have ci-framework combine them. This way we would have to deal with just a single variable. Moreover, I've seen this pattern in use, and it would be beneficial for every users.

roles/cifmw_cephadm/tasks/install_cephadm.yml
evallesp: Agree on this, we can check if 'ceph-([0-9.]+)' is there at L20. We stop early and we don't let the playbook running with unexpected results!

evallesp:
> I was wondering whether another set of changes could help, but that would require the help of ci-framework developers. The idea would be to split cifmw_repo_setup_rhos_release_args into a cifmw_repo_setup_rhos_release_ceph and cifmw_repo_setup_rhos_release_os_args or so, and have ci-framework combine them. This way we would have to deal with just a single variable. Moreover, I've seen this pattern in use, and it would be beneficial for every users.

This is a really good idea.

fultonj:
Thanks for the reviews. I'll prototype another approach which avoids touching rhos-release. If it works, I might close this. I'll be back in touch.

roles/cifmw_cephadm/tasks/install_cephadm.yml
fultonj: No `-x` purges everything and then adds what's in the arguments.

fultonj:
I'm going to test with https://github.com/openstack-k8s-operators/ci-framework/pull/4015 and see if I can acheive the same thing. If so I'll close this.

fultonj:
Obsoleted by https://github.com/openstack-k8s-operators/ci-framework/pull/4015

PR #4013, opened 2026-06-24
[multiple] Remove unnecessary registries from defaults

Valkyrie00:
## Summary
Remove registries that are not needed in upstream defaults from the `openshift_setup` role.
A mechanism already exists to supply additional registries at the job level via `cifmw_openshift_setup_allowed_extra_registries`, which is consumed in `roles/openshift_setup/tasks/configure_registries.yml` and merged into the final `allowedRegistries` list.

## Note
A hardcoded registry reference was found in `hooks/playbooks/config_cluster_for_disconnected_deployment.yml`. This needs to be investigated and removed in a dedicated follow-up task.


evallesp:
recheck

evallesp (approved):
/lgtm

michburk (approved):
/lgtm

bshewale (approved):

PR #4017, opened 2026-06-25
Strenghten ready condition on set openstack containers

averdagu:
set_openstack_containers wait for the openstack-operator-init
pod to be reinstalled, problem is that openstack-operator
takes a minute to be restarted, which in some scenarios can
cause problems, improving the wait logic to avoid that.

Resolves: [OSPRH-31605](https://redhat.atlassian.net/browse/OSPRH-31605)

### DIFF
--- a/roles/set_openstack_containers/tasks/main.yml
+++ b/roles/set_openstack_containers/tasks/main.yml
@@ -119,6 +119,33 @@
     - operator_name != 'openstack'
     - not cifmw_architecture_scenario is defined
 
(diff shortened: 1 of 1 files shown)


averdagu:
Was tested on a testproject and update job passed the failing point

karelyatin:
LGTM, do fix github workflows

michburk (approved):
/lgtm

evallesp (approved):
/approve

PR #4021, opened 2026-06-26
[reproducer] Refresh premetal config vars from resolved hostvars

michburk:
In premetal deployments, config variables updated between bootstrap and pickup were not reflected in reproducer-variables.yml. Overwrite it at pickup time by dumping the hypervisor's filtered hostvars using the same filter chain as configure_bm_ocp_controller.yml.

### DIFF
--- a/roles/reproducer/tasks/overwrite_zuul_vars.yml
+++ b/roles/reproducer/tasks/overwrite_zuul_vars.yml
@@ -1,75 +1,115 @@
 ---
-# NOTE(dpawlik): Earlier, our idea was to dump vars as
-# it is roles/cifmw_setup/tasks/bootstrap.yml and read it,
-# then dump it and replace reproducer-variables.yml.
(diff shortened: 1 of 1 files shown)


roles/reproducer/tasks/overwrite_zuul_vars.yml
```diff
@@ -1,71 +1,74 @@
...
         dest: "{{ cifmw_basedir }}/parameters/reproducer-variables.yml"
         mode: "0664"
         backup: true
       no_log: "{{ cifmw_nolog | default(true) | bool }}"
       delegate_to: controller-0
 
-    - name: Overwrite custom-params.yml
+    - name: Write merged custom-params.yml to controller-0
+      vars:
+        _merged: >-
```
evallesp: (non-blocking) suggestion: I think we can move this to the block vars level, as _merged is the same used at L52. So we don't need to calculate it twice.

PR #4023, opened 2026-06-29
[update] Add optional cleanup of OpenStack resources

mnietoji:
The update role creates test workloads (ping test VM, ports) on OpenStack during the update procedure but never cleans them up. This leaves ports allocated on the external network, which blocks subsequent test jobs from recreating subnets.

Add a cleanup step at the end of the update role that runs the workload_launch.sh script in cleanup mode to delete the test VM and associated ports. The cleanup runs by default and can be disabled with cifmw_update_dont_cleanup for debugging.

Based on the work from #3335 by Sofer Athlan-Guyot which was closed by the stale bot before merging.

Related-Issue: #[OSPRH-20307](https://redhat.atlassian.net/browse/OSPRH-20307)

### DIFF
--- a/roles/update/defaults/main.yml
+++ b/roles/update/defaults/main.yml
@@ -48,6 +48,7 @@ cifmw_update_openstack_cmd: >-
   oc rsh -n {{ cifmw_update_namespace }} openstackclient openstack
 
 ## User facing
+cifmw_update_cleanup: true
 cifmw_update_openstack_update_kpatch: false
 
 cifmw_update_reboot_test: false
--- a/roles/update/tasks/cleanup.yml
+++ b/roles/update/tasks/cleanup.yml
@@ -14,6 +14,20 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
-- name: Cleaning the World
-  ansible.builtin.debug:
-    msg: "So here update should clean things up!"
+- name: Cleanup ping test resources
+  when:
+    - cifmw_update_ping_test | bool
+  block:
+    - name: Cleaning the ping vm if needed
+      ansible.builtin.shell: |
+        set -e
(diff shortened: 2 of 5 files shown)


roles/update/defaults/main.yml
```diff
@@ -48,6 +48,7 @@ cifmw_update_openstack_cmd: >-
   oc rsh -n {{ cifmw_update_namespace }} openstackclient openstack
 
 ## User facing
+cifmw_update_dont_cleanup: false
```
michburk: maybe a nit and if you disagree and want to keep it this way, that's fine, but I think `cifmw_update_cleanup: true` would be clearer than the double negative of `...dont_cleanup: false`

roles/update/tasks/cleanup.yml
```diff
@@ -14,6 +14,19 @@
...
+        set -e
+        set -o pipefail
+        cat {{ cifmw_update_workload_launch_script }} | \
+        podman exec -i lopenstackclient env WKL_MODE=cleanup bash -i 2>&1 \
+        {{ cifmw_update_timestamper_cmd }} | tee {{ cifmw_update_artifacts_basedir }}/workload_cleanup.log
+
+    - name: Get logs from update instance cleanup
+      ansible.builtin.shell: >
+        podman cp lopenstackclient:{{ cifmw_update_artifacts_basedir_suffix }}/.
+        {{ cifmw_update_artifacts_basedir }}
```
michburk: could be `ansible.builtin.comand`

michburk (commented):
two small things to consider, largely lgtm

michburk (approved):
/lgtm

evallesp (approved):
/approve

PR #4024, opened 2026-06-30
[test_operator] Support merging skiplist with exclude list

kstrenkova:
Adding cifmw_test_operator_tempest_exclude_list_merge variable to enable combining list_skipped.yml with
cifmw_test_operator_tempest_exclude_list. When set to true, both sources are merged into a single exclude list, which helps to later remove workarounds that merged these lists manually.

Signed-off-by: Katarina Strenkova kstrenko@redhat.com

### DIFF
--- a/roles/test_operator/tasks/tempest-tests.yml
+++ b/roles/test_operator/tasks/tempest-tests.yml
@@ -51,6 +51,7 @@
 - name: Configuring tests to be skipped via skiplist
   when: >
     stage_vars_dict.cifmw_test_operator_tempest_exclude_list is not defined or
+    stage_vars_dict.cifmw_test_operator_tempest_exclude_list_merge | bool or
     stage_vars_dict.cifmw_test_operator_tempest_tests_exclude_override_scenario | bool
   block:
     - name: Copy list_skipped to artifacts dir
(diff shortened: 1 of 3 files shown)


kstrenkova:
I have tested this change locally (ran the test_operator role, once with the new parameter set to false and once set with true) and it seems to be working as expected :D 

The workaround we are trying to remove is present in the ci-framework-jobs repository (search for `TODO (anbanerj): Remove this task when tempest supports it`), if that would help you understand the bigger picture a bit better.

roles/test_operator/tasks/tempest-tests.yml
```diff
@@ -66,10 +67,18 @@
       register:
         list_skipped
 
-    - name: Set variable
+    - name: Merge tests from exclude list when merge is enabled
+      when:
```
evallesp, line 71: (non-blocking) concern: Unsure but might be good if we check cifmw_test_operator_tempest_exclude_list > 0 ? So L76 canoot loop over empty string.

evallesp (dismissed):
LGTM in general!

roles/test_operator/tasks/tempest-tests.yml
kstrenkova: You're right, I have added the condition :D

kstrenkova:
recheck

michburk:
recheck

evallesp (approved):
/lgtm

michburk (approved):
/approve

PR #4028, opened 2026-07-02
[federation] Fix RHSSO OperatorGroup conflict with RHOSO operators

srac0:
The federation role was deploying the RHSSO OperatorGroup into openstack-operators, which already has an OperatorGroup from the OpenStack operator. OLM enforces one OperatorGroup per namespace, so the second one triggered TooManyOperatorGroups and blocked InstallPlan creation for the RHSSO subscription.

Change the default cifmw_federation_operator_namespace from openstack-operators to keycloak-operators, and add a task to create that namespace before applying the OLM manifest.

### DIFF
--- a/docs/dictionary/en-custom.txt
+++ b/docs/dictionary/en-custom.txt
@@ -583,6 +583,7 @@ rsa
 rsync
 runtime
 RHCOS
+RHSSO
 scansettingbinding
 scap
 scp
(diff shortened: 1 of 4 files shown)


evallesp (approved):
/lgtm

Valkyrie00 (approved):
/approve

PR #4029, opened 2026-07-02
Fix bmo01 scenario template generation key

abhibongale:
bmo01 scneario template generates duplicate YAML keys when processing multi-ppart inventory hostnames, causing Nodeset deployment timeout.

This bug casues data loss (first host) when adding a misnamed node (duplicate key survivor) This PR tries to fix.

fixes: [OSPRH-32215](https://redhat.atlassian.net/browse/OSPRH-32215)


michburk (approved):
/approve

brjackma (approved):

PR #4030, opened 2026-07-02
[cifmw_cephadm] Add fallback when cifmw_cephadm_rpm_url 404s

fultonj:
If `cifmw_cephadm_rpm_url` returns a 404, strip the filename from the URL to get the parent directory, fetch the index page, and search it for a cephadm RPM. If one is found, warn and update `cifmw_cephadm_rpm_url` to the discovered URL. If none is found, fail with a descriptive error.

This handles cases where the exact RPM version in the URL has changed (e.g. a compose rebuild with a new NVR) but a valid cephadm package still exists in the same directory.

Related-Issue: [OSPRH-31757](https://redhat.atlassian.net/browse/OSPRH-31757)


fultonj:
Abandoning since we should have a stable URL.

PR #4032, opened 2026-07-03
[libvirt_manager] Create DNS/DHCP records when mac_map is pre-defined

rebtoor:
When `cifmw_libvirt_manager_mac_map` is defined (baremetal deploy-bm
flow), `generate_networking_data.yml` is skipped because MAC addresses
are already known. However, this also skips the DNS/DHCP record
creation (`create_dns_records.yml`) which is called at the end of
that file.

Without DHCP reservations, compute VMs provisioned by Ironic never
get a ctlplane IP, leaving them unreachable for EDPM bootstrap.

This adds a separate path that loads `cifmw_networking_env_definition`
from the infra environment file (`/etc/ci/env/networking-environment-definition.yml`)
and creates DNS/DHCP records when `mac_map` is pre-defined.

### DIFF
--- a/roles/libvirt_manager/tasks/deploy_layout.yml
+++ b/roles/libvirt_manager/tasks/deploy_layout.yml
@@ -118,6 +118,30 @@
     - cifmw_libvirt_manager_mac_map is undefined
   ansible.builtin.include_tasks: generate_networking_data.yml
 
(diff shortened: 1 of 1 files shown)


rebtoor:
recheck

PR #4045, opened 2026-07-09
[cifmw_backup_restore] Add Octavia OVN sync after Neutron sync

stuggi:
Sync Octavia load balancer entries to OVN NB database when Octavia with OVN provider is deployed. Checks spec.octavia.enabled on the OpenStackControlPlane CR, skips if Octavia is not enabled.

Placed right after the Neutron sync and before EDPM deployment so OVN NB DB is fully populated when data plane nodes reconnect.

### DIFF
--- a/roles/cifmw_backup_restore/tasks/restore.yml
+++ b/roles/cifmw_backup_restore/tasks/restore.yml
@@ -597,7 +597,36 @@
   changed_when: true
 
 # ========================================
-# Step 11: Restore DataPlane (Order 60)
+# Step 11: Sync Octavia to OVN (optional)
+# ========================================
+# Only when Octavia with OVN provider is deployed. Repopulates OVN NB DB
+# with load balancer entries from Octavia's database.
+# Ref: https://docs.openstack.org/ovn-octavia-provider/latest/admin/driver.html#octavia-db-to-ovn-database-population
+- name: Check if Octavia is enabled
+  ansible.builtin.shell: |
+    oc get openstackcontrolplane {{ _ctlplane_name.stdout }} \
+      -n {{ cifmw_backup_restore_namespace }} \
+      -o jsonpath='{.spec.octavia.enabled}'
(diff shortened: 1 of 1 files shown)


abays (dismissed):
/lgtm

roles/cifmw_backup_restore/tasks/restore.yml
```diff
@@ -620,7 +650,7 @@
     _step_name: "Step 11 (DataPlane restore)"
```
michburk: looks like the number in this `_step_name` is out of sync now,
`Step 11` -> `Step 12`

roles/cifmw_backup_restore/tasks/restore.yml
```diff
@@ -597,7 +597,37 @@
...
+  register: _octavia_ovn_sync
+  when: _octavia_enabled.stdout == 'true'
+  changed_when: true
+
+- name: Report Octavia OVN sync result
+  ansible.builtin.debug:
+    msg: >-
+      Octavia OVN sync:
+      {{ 'completed' if _octavia_enabled.stdout == 'true' else 'skipped (Octavia not enabled)' }}.
+  when: _octavia_enabled.stdout == 'true' or not (_octavia_enabled.stdout == 'true')
```
michburk: when condition is always true, could be removed

roles/cifmw_backup_restore/tasks/restore.yml
stuggi: done

roles/cifmw_backup_restore/tasks/restore.yml
stuggi: done

michburk (approved):
/lgtm

abays (approved):
/lgtm
/approve

abays:
@michburk Would you mind approving if you're good with this?

PR #4051, opened 2026-07-10
[env_op_images] Reduce POST-RUN console output

sdatko:
This commit suppress verbose console output from Kubernetes API
queries and CRI-O log collection tasks to reduce POST-RUN phase
output by ~80% in Zuul jobs.

The suppressed data that were previously visible in `job-output.txt`
were actually redundant and already collected in ci-framework-data/
artifacts anyway as:
- `operator_images.yaml`
- `crio_logs/*`

Keeping `job-output.txt` smaller reduces impact on the web browser
when opening Zuul build logs.

For debugging purposes logging can be re-enabled by setting
the `cifmw_nolog: false` variable.

Changes:
- Add `no_log` to 4 tasks `env_op_images` role.
- Reworked the CRI-O log collection tasks.

Impact:
- 75% reduction in console output (~60k → ~15k lines)
- 90% smaller file size (~20 MB → ~2 MB)
- All data still preserved for subsequent tasks + dumped as artifacts

Assisted-by: Claude Sonnet 4.5

### DIFF
--- a/roles/env_op_images/tasks/main.yml
+++ b/roles/env_op_images/tasks/main.yml
@@ -87,6 +87,7 @@
         field_selectors:
           - status.phase=Running
       register: all_pods_list
+      no_log: "{{ cifmw_nolog | default(true) | bool }}"
 
     - name: Retrieve openstack-operator-index pod
       vars:
@@ -111,6 +112,7 @@
         field_selectors:
           - status.phase=Running
       register: pod_list
+      no_log: "{{ cifmw_nolog | default(true) | bool }}"
 
     - name: Get operator images and pods
       when: not cifmw_env_op_images_dryrun | bool
@@ -144,6 +146,7 @@
       loop: "{{ selected_pods }}"
       loop_control:
         label: "{{ item.metadata.name }}"
+      no_log: "{{ cifmw_nolog | default(true) | bool }}"
 
     - name: Write images to file
       vars:
--- a/roles/env_op_images/tasks/verify_pulled_report_crio.yml
+++ b/roles/env_op_images/tasks/verify_pulled_report_crio.yml
@@ -68,26 +68,18 @@
                     | map(attribute='metadata.name') | list
                   }}
 
-            - name: Fetch CRI-O unit logs per node
-              ansible.builtin.command:
+            - name: Fetch and write CRI-O logs per node
+              ansible.builtin.shell:
                 cmd: >-
-                  oc adm node-logs "{{ item }}" -u crio
-                  --since=-24h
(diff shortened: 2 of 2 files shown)


danpawlik (dismissed):

danpawlik:
nice improvement @sdatko !

evallesp (dismissed):
/approve

sdatko:
FINALLY READY FOR REVIEW

roles/env_op_images/tasks/main.yml
```diff
@@ -87,6 +87,7 @@
         field_selectors:
           - status.phase=Running
       register: all_pods_list
+      no_log: true
```
nemarjan: (suggestion) Nice work on the log reduction -- the 80% drop in job-output.txt is a big win. I see the underlying data is still preserved as artifacts, so nothing is lost on  successful runs.

One thought on the no_log: true approach: if any of these tasks ever fail, the error details (API response, traceback, etc.) would be hidden in job-output.txt too, which could make debugging trickier. The artifacts won't help there since they either won't exist or will be incomplete on failure.

What do you think about using no_log: "{{ cifmw_nolog | default(true) | bool }}" instead? It gives you the same quiet output by default, but lets someone toggle it off when they're investigating a broken job.

nemarjan (commented):
LGTM overall -- the 80% reduction in job-output.txt is a big win, and I like that all the data is still preserved as artifacts. Just left a small suggestions inline, nothing blocking.

roles/env_op_images/tasks/main.yml
sdatko: Ach, good point ~ I was originally thinking about creating new role variable, but I forgot about the `cifmw_nolog` present already in other places – I will update the PR 🙏

Valkyrie00: I agree with @nemarjan's suggestion. it would be great to control it using a variable

sdatko: I already wrote I will do it :stuck_out_tongue: I was just waiting for current CI jobs to ensure they do not break any parts of adoption jobs by chance (functionality-wise the change from `true` to variable is the same so I will skip 2nd run).

sdatko: Pushed

rebtoor:
/lgtm

thanks @sdatko

danpawlik (approved):

evallesp (approved):

nemarjan (approved):
/lgtm
/approve

brjackma (approved):

PR #4055, opened 2026-07-14
DT05 with custom ASN

mtomaska:
- Adds ability to set same ASN for spines and leafs in preparation for mixed iBGP and eBGP deployment.
- conditional add `route-reflector-client` for spines when when remote_as is set to `internal` as it is incase of iBGP (same asn)
Continuation on top of https://github.com/openstack-k8s-operators/ci-framework/pull/3900 

Related: [OSPRH-32310](https://redhat.atlassian.net/browse/OSPRH-32310)                                                    
Assisted-By: Claude Code

### DIFF
--- a/playbooks/bgp/prepare-bgp-spines-leaves.yaml
+++ b/playbooks/bgp/prepare-bgp-spines-leaves.yaml
@@ -272,9 +272,25 @@
         state: present
       when: _ip_version == 6
 
+    - name: Assign loopback IPv4 on router loopback interface
+      when:
+        - _ip_version == 4
+        - (router_loopback_ip | default('') | length) > 0
+      become: true
+      ansible.builtin.command:
+        cmd: >-
+          ip address replace {{ router_loopback_ip }}/32 dev lo
+      changed_when: true
+
     - name: Configure FRR
       vars:
-        _router_id: "{{ '' if _ip_version == 4 else '1.1.1.1' }}"
+        # Prefer loopback IP as BGP router-id when set; else 1.1.1.1 (IPv4 and IPv6).
+        _router_id: >-
+          {{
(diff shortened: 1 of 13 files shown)


eduolivares:
Hi Miro,

I tested the all-iBGP topology (all nodes in AS 65000, `remote-as internal` everywhere) on titan145 and found that **router-0 receives no BGP routes from the spines**. The root cause is the **iBGP split-horizon rule**: spine nodes learn routes from leaves via iBGP but cannot re-advertise them to router-0 (also iBGP) unless the spine is configured as a route reflector.

The fix is to add `neighbor downlink route-reflector-client` in the IPv4 and IPv6 address families on the spines when the downlink peering is iBGP. I verified this on titan145 — after adding the directive and restarting FRR, router-0 immediately started receiving all expected BGP routes.

I think `playbooks/bgp/templates/spine-frr.conf.j2` needs a conditional block like:

```jinja2
  address-family ipv4 unicast
    redistribute connected
    neighbor downlink default-originate
    neighbor downlink prefix-list only-host-prefixes in
{% if cifmw_bgp_spine_downlink_remote_as | default('external') == 'internal' %}
    neighbor downlink route-reflector-client
{% endif %}
```

And the same for `address-family ipv6 unicast`. Without this, anyone who overrides to the all-iBGP topology will hit the split-horizon issue — spines silently drop leaf-learned routes instead of forwarding them to router-0.

mtomaska:
recheck

mtomaska:
recheck

eduolivares:
recheck

eduolivares (approved):

Valkyrie00 (approved):
/lgtm
/approve

PR #4060, opened 2026-07-15
[cinder_multiattach] Wait for Glance reconciliation after config patch

fmount:
The `oc wait --for=condition=Ready` on the `controlplane` returns immediately when the condition is already true, before the `StatefulSet` rolling update triggered by the config change completes. 
This causes a race with subsequent hooks (e.g. amphora image upload) that depend on `Glance` being stable.

This patch replaces the inline `oc wait` with two explicit steps:
1. Poll the `Glance CR` until `customServiceConfig` reflects the change (`oc apply` happened and the `rollout` started)
2. Wait for the `controlplane` `Ready` condition after reconciliation starts (which is what we had previously)

Closes: [OSPCIX-1454](https://redhat.atlassian.net/browse/OSPCIX-1454)

### DIFF
--- a/hooks/playbooks/cinder_multiattach_volume_type.yml
+++ b/hooks/playbooks/cinder_multiattach_volume_type.yml
@@ -28,6 +28,19 @@
     - name: Block to configure cinder_volume_type when needed
       when: configure_cinder_volume_type | default(false) | bool
       block:
+        - name: Get OpenStackControlPlane CR name
+          kubernetes.core.k8s_info:
+            kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
+            api_version: core.openstack.org/v1beta1
+            kind: OpenStackControlPlane
+            namespace: "{{ namespace }}"
+          register: _ctlplane_cr
+
+        - name: Set OpenStackControlPlane CR name
+          ansible.builtin.set_fact:
+            _glance_cinder_volume_type_ctlplane_name: >-
+              openstackcontrolplane.core.openstack.org/{{ _ctlplane_cr.resources[0].metadata.name }}
+
         - name: Create tempfile
           ansible.builtin.tempfile:
             state: file
@@ -37,8 +50,7 @@
         - name: Write current glance customServiceConfig to tempfile
           ansible.builtin.shell: |
             set -xe -o pipefail
-            crname=$(oc get openstackcontrolplane -o name -n {{ namespace }})
-            oc -n {{ namespace }} get ${crname} -o jsonpath={.spec.glance.template.customServiceConfig} > {{ _glance_custom_service_config_file.path }}
+            oc -n {{ namespace }} get {{ _glance_cinder_volume_type_ctlplane_name }} -o jsonpath={.spec.glance.template.customServiceConfig} >{{ _glance_custom_service_config_file.path }}
           changed_when: false
 
         - name: Ensure cinder_volume_type is configured with proper value in tempfile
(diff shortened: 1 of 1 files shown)


hooks/playbooks/cinder_multiattach_volume_type.yml
```diff
@@ -65,8 +74,29 @@
...
           ansible.builtin.shell: |
             set -xe -o pipefail
-            crname=$(oc get openstackcontrolplane -o name -n {{ namespace }})
-            oc -n {{ namespace }} patch ${crname} --type=merge --patch "{{ _yaml_patch | to_nice_yaml }}"
-            oc -n {{ namespace }} wait ${crname} --for condition=Ready --timeout=10m
+            oc -n {{ namespace }} patch {{ _ctlplane_name.stdout }} --type=merge --patch "{{ _yaml_patch | to_nice_yaml }}"
           changed_when: _glance_ini_file.changed
           when: _glance_ini_file.changed
+
+        - name: Verify Glance CR reflects the cinder_volume_type change
```
tosky: maybe I would add here a comment to explain why this is needed

fmount: Good idea, I added it.

tosky:
The failure in the precommit job looks legit, but please wait before updating this as I'm testing it

fmount:
recheck

hooks/playbooks/cinder_multiattach_volume_type.yml
```diff
@@ -65,8 +74,50 @@
...
+          ansible.builtin.shell: |
+            set -xe -o pipefail
+            oc -n {{ namespace }} get {{ _glance_name.stdout }} -o jsonpath='{.spec.customServiceConfig}'
+          register: _glance_cr_config
+          retries: 30
+          delay: 5
+          until: "'cinder_volume_type' in _glance_cr_config.stdout"
+          when: _glance_ini_file.changed
+
+        - name: Wait for reconciliation to start
```
fmount: @tosky I'm making a test with a pause. This is not the final version but something I'd like to test to prove a theory. I will update this patch with the final version after I look at the next set of logs.

fmount: @tosky ok the pause worked, I just verified with the uni07eta testproject. I might wait for the current execution to finish, then I'm going to update the patch and test the final version on it.

hooks/playbooks/cinder_multiattach_volume_type.yml
```diff
@@ -28,6 +28,16 @@
     - name: Block to configure cinder_volume_type when needed
       when: configure_cinder_volume_type | default(false) | bool
       block:
+        - name: Get OpenStackControlPlane CR name
```
evallesp, line 31: (blocking) suggestion: Could you use ansible k8s_info?

(non-blocking) Do we want to add retry logic?

hooks/playbooks/cinder_multiattach_volume_type.yml
```diff
@@ -37,8 +47,7 @@
         - name: Write current glance customServiceConfig to tempfile
           ansible.builtin.shell: |
             set -xe -o pipefail
-            crname=$(oc get openstackcontrolplane -o name -n {{ namespace }})
-            oc -n {{ namespace }} get ${crname} -o jsonpath={.spec.glance.template.customServiceConfig} > {{ _glance_custom_service_config_file.path }}
+            oc -n {{ namespace }} get {{ _ctlplane_name.stdout }} -o jsonpath={.spec.glance.template.customServiceConfig} > {{ _glance_custom_service_config_file.path }}
```
evallesp: praise: this is much better!!

hooks/playbooks/cinder_multiattach_volume_type.yml
```diff
@@ -65,8 +74,44 @@
...
+            oc -n {{ namespace }} patch {{ _ctlplane_name.stdout }} --type=merge --patch "{{ _yaml_patch | to_nice_yaml }}"
           changed_when: _glance_ini_file.changed
           when: _glance_ini_file.changed
+
+        # After patching the controlplane, there is a window where both
+        # Glance and the controlplane still report Ready because the
+        # reconciliation has not propagated yet. Wait for Glance to
+        # enter reconciliation (Ready=False) before waiting for it
+        # and the controlplane to become Ready again.
+        - name: Get Glance CR name # noqa: no-handler
```
evallesp, line 89: (blocking) suggestion: can we use k8s_info?

hooks/playbooks/cinder_multiattach_volume_type.yml
```diff
@@ -37,8 +47,7 @@
         - name: Write current glance customServiceConfig to tempfile
           ansible.builtin.shell: |
```
evallesp, line 51: (non-blocking) thought would be awesome if we refactor this eventually! to follow a more ansible idiomantic approach

tosky:
recheck

hooks/playbooks/cinder_multiattach_volume_type.yml
tosky: Addressed (hopefully) the blocking suggestion, while Francesco is not around.

As for the retry, IMHO if this simple get call fails, we may have bigger issues, but it can be added?

hooks/playbooks/cinder_multiattach_volume_type.yml
tosky: That would require probably a bigger effort, I guess

hooks/playbooks/cinder_multiattach_volume_type.yml
tosky: Addressed(2) (hopefully) the blocking suggestion, while Francesco is not around.

tosky:
I've tested the new version and it seems to be working (contact me for more details)

michburk (approved):
/lgtm

evallesp (approved):
/approve

PR #4070, opened 2026-07-23
[adoption_osp_deploy] Recognize dcn-scenario-defined central stack name

gais-ameer:
`edpm_nodes:` generates only special-cased stack name `overcloud`, but
 `dcn_storage.yaml` names its central stack `central`, so `edpm_nodes`
rendered empty for that scenario. Widening the check to 
`stack in ['overcloud', 'central']` to cover the `dcn` scenario.


fultonj (approved):
/lgtm
/approve

PR #4077, opened 2026-07-27
[adoption] Enable network service on OSP17

xbezdick:
OSP17 requires network service running as ifcfg is used as backend for os-net-config. Without it nodes loose network connection on reboot or ifcfg to nmstate migration.

Resolves: [OSPRH-31307](https://redhat.atlassian.net/browse/OSPRH-31307)

### DIFF
--- a/roles/adoption_osp_deploy/tasks/prepare_overcloud.yml
+++ b/roles/adoption_osp_deploy/tasks/prepare_overcloud.yml
@@ -340,6 +340,17 @@
       loop_control:
         loop_var: overcloud_vm
 
+    - name: Enable and start network service for overcloud nodes
+      become: true
+      delegate_to: "{{ overcloud_vm }}"
+      ansible.builtin.systemd:
+        name: network
+        enabled: true
+        state: started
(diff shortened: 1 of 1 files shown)


roles/adoption_osp_deploy/tasks/prepare_overcloud.yml
```diff
@@ -340,6 +340,17 @@
       loop_control:
         loop_var: overcloud_vm
 
+    - name: Enable and start network service for overcloud nodes
```
ciecierski, line 343: Do you need it before configuring network interfaces, which is one task before?

xbezdick: No, we just need to make sure it's running before/after does not matter that much.

jistr (approved):

holser (approved):

holser (approved):
/LGTM
/Approved

holser:
recheck

holser:
recheck

PR #4084, opened 2026-07-30
[hooks] Wait for cinder-backup services before running backup/restore checks

gais-ameer:
The backup/restore validation in distributed architectures 
could start hitting the Cinder API before all per-AZ cinder-backup 
services had finished initializing, causing intermittent failures.

Adding a wait task that polls `openstack volume service list` 
and blocks until no cinder-backup service reports state as "down" 
in any availability zone, with a rescue block that dumps the 
last observed Host/State values before failing.

### DIFF
--- a/hooks/playbooks/cinder_backups.yaml
+++ b/hooks/playbooks/cinder_backups.yaml
@@ -1,5 +1,5 @@
 ---
-# Test Cinder backup and restore capabilities across availability zones in dz-storage DT
+# Test Cinder backup and restore capabilities across availability zones in dz-storage/DCN DT
 # Tests three scenarios:
 # 1. AZ1 backs up to AZ1 and restores to AZ1
 # 2. AZ1 backs up to AZ2 and restores to AZ1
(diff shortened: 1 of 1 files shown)


hooks/playbooks/cinder_backups.yaml
```diff
@@ -14,6 +14,36 @@
...
+              openstack volume service list
+              --service cinder-backup
+              -c Host -c State
+              -f json
+            register: cinder_backup_service_status
+            until: >-
+              (cinder_backup_service_status.stdout | from_json)
+              | selectattr('State', 'equalto', 'down')
+              | list
+              | length == 0
```
abays: Should we also check here to make sure there are Cinder backup services present?  This could technically pass as-is before any service appears, right?

gais-ameer: you are right , I will modify that

gais-ameer: @abays I have updated the task to cover the case when no cinder-backup service is present at all.

gais-ameer: `(cinder_backup_service_status.stdout | from_json | length > 0)` would fail when `openstack volume service list` return empty for cinder-backup

gais-ameer:
recheck

gais-ameer:
recheck

hooks/playbooks/cinder_backups.yaml
```diff
@@ -14,6 +14,39 @@
...
+                | list
+                | length == 0)
+              retries: 30
+              delay: 10
+        rescue:
+            - name: Show cinder-backup service status
+              ansible.builtin.debug:
+                  msg: >-
+                    Here is the output of openstack volume service list:
+                    {{ cinder_backup_service_status.stdout | from_json }}
```
abays: It might be better to use the `default` filter here instead, because if `oc rsh` itself fails (pod not found, network issue) or returns non-JSON output on the final retry, `cinder_backup_service_status.stdout` won't be valid JSON. The rescue debug task does `{{ cinder_backup_service_status.stdout | from_json
  }}`, which would raise a Jinja2 error inside the rescue block, masking the actual failure.

gais-ameer: How about I remove the `from_json` so that the rescue will print the non-json errors as well ?

abays: Yes, that was my thinking too, but just with a `default` in case the output is empty, to make it obvious in that case.  i.e. `{{ cinder_backup_service_status.stdout | default('(no output)') }}`

abays (approved):
/lgtm
/approve

michburk (approved):
/approve

PR #4086, opened 2026-07-31
[reproducer] Add remove top-level keys from a YAML file

evallesp:
Plugin for removing top-level key that appears in a provided
key-source yamls files.

Used at overwrite_zuul_vars file for removing variables
added at: exclude_discovered_files as in premetal path
these variables were already loaded in previous premetal step.

### DIFF
--- a/plugins/modules/strip_dict_keys.py
+++ b/plugins/modules/strip_dict_keys.py
@@ -1,0 +1,160 @@
+#!/usr/bin/python
+
+# Copyright: (c) 2026, Red Hat
+# GNU General Public License v3.0+ (see COPYING or
+# https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+DOCUMENTATION = r"""
+---
+module: strip_dict_keys
+short_description: Remove top-level keys from a YAML file
+
+description:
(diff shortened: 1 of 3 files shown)


plugins/modules/strip_dict_keys.py
```diff
@@ -0,0 +1,160 @@
...
+        except Exception as e:
+            module.fail_json(msg=f"Failed to read keys-from file {kf}: {e}")
+        if isinstance(keys_data, dict):
+            keys_to_remove.update(keys_data.keys())
+
+    removed = sorted(k for k in keys_to_remove if k in base)
+    filtered = {k: v for k, v in base.items() if k not in keys_to_remove}
+    changed = len(removed) > 0
+
+    if not module.check_mode:
```
Valkyrie00: Hey @evallesp 👋, I noticed the module writes to the destination file unconditionally, even when no keys were actually removed. Since `yaml.dump` doesn't preserve comments or original formatting, this means the file could get silently reformatted in CI without any logical change.

What do you think about guarding the write with:  `if changed and not module.check_mode:`
so we only touch the file when there's actually something to strip?

evallesp: Oh sure, good catch! Thanks for the review.

Valkyrie00 (approved):
/lgtm

bshewale (approved):

PR #4089, opened 2026-08-05
[test_operator] Add WA for HorizonTest admin credentials

kstrenkova:
This WA sets the HorizonTest admin credentials from OpenStackConfigMap and OpenStackConfigSecret respectively. This will be done automatically once TCIB and test-operator fixes reach downstream. Until then, this WA is needed for running horizon tests with correct credentials.

### DIFF
--- a/roles/test_operator/tasks/runners/horizontest_runner.yml
+++ b/roles/test_operator/tasks/runners/horizontest_runner.yml
@@ -1,28 +1,64 @@
 ---
-- name: Fetch admin password from osp-secret
-  no_log: true
-  ansible.builtin.command:
-    cmd: >-
-      oc get secret osp-secret
-      -n {{ cifmw_test_operator_namespace }}
-      -o jsonpath='{.data.AdminPassword}'
-  register: _horizontest_admin_password_b64
-  failed_when: false
-  changed_when: false
-
-- name: Override admin password from osp-secret
-  no_log: true
-  when:
-    - _horizontest_admin_password_b64.rc == 0
-    - _horizontest_admin_password_b64.stdout | length > 0
+# Note(kstrenko): Remove WA once the TCIB and test-operator fix reaches downstream
+- name: Check if admin credentials need fetching
   ansible.builtin.set_fact:
-    stage_vars_dict: >-
+    _horizontest_fetch_creds: >-
       {{
-        stage_vars_dict | combine({
-          'cifmw_test_operator_horizontest_admin_password':
-            _horizontest_admin_password_b64.stdout | b64decode
-        })
+        stage_vars_dict.cifmw_test_operator_horizontest_admin_username | default('') | length == 0 or
+        stage_vars_dict.cifmw_test_operator_horizontest_admin_password | default('') | length == 0
       }}
+
+- name: Fetch admin credentials from OpenStack config
+  when: _horizontest_fetch_creds | bool
+  block:
+    - name: Fetch OpenStackConfigMap
+      kubernetes.core.k8s_info:
+        kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
+        api_key: "{{ cifmw_openshift_token | default(omit) }}"
+        context: "{{ cifmw_openshift_context | default(omit) }}"
+        kind: ConfigMap
+        name: openstack-config
+        namespace: "{{ stage_vars_dict.cifmw_test_operator_namespace }}"
+      register: _horizontest_configmap
+
+    - name: Fetch OpenStackConfigSecret
+      kubernetes.core.k8s_info:
+        kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
+        api_key: "{{ cifmw_openshift_token | default(omit) }}"
+        context: "{{ cifmw_openshift_context | default(omit) }}"
+        kind: Secret
+        name: openstack-config-secret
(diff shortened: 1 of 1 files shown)


kstrenkova:
recheck

roles/test_operator/tasks/runners/horizontest_runner.yml
```diff
@@ -1,29 +1,65 @@
...
+      kubernetes.core.k8s_info:
+        kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
+        api_key: "{{ cifmw_openshift_token | default(omit) }}"
+        context: "{{ cifmw_openshift_context | default(omit) }}"
+        kind: ConfigMap
+        name: openstack-config
+        namespace: "{{ stage_vars_dict.cifmw_test_operator_namespace }}"
+      register: _horizontest_configmap
+
+    - name: Fetch OpenStackConfigSecret
```
Valkyrie00, line 24: Hey @kstrenkova 👋, the previous implementation had no_log: true to keep the admin password out of CI logs, could you add `no_log: true `(or `no_log: "{{ cifmw_nolog | default(true) | bool }}"`) here, in to "Override admin credentials" and "Patch admin credentials" tasks? What do you think?

kstrenkova: Hey @Valkyrie00 ! The password will be visible in the task `Print CR` anyway :sweat_smile: I do agree it would be better with `no_log: true`, but if you look into the job logs, the password is not visible in these tasks either way. If you don't mind, I would like to get it merged even if it's not perfect due to the CIX (and also because it will be removed soon).

roles/test_operator/tasks/runners/horizontest_runner.yml
```diff
@@ -1,29 +1,65 @@
...
+        namespace: "{{ stage_vars_dict.cifmw_test_operator_namespace }}"
+      register: _horizontest_secret
+
+    - name: Override admin credentials in stage_vars_dict
+      ansible.builtin.set_fact:
+        stage_vars_dict: >-
+          {{
+            stage_vars_dict | combine({
+              'cifmw_test_operator_horizontest_admin_username':
+                (_horizontest_configmap.resources[0].data['clouds.yaml'] | from_yaml).clouds.default.auth.username,
```
Valkyrie00, line 40: (nit) just a heads-up, the expressions access resources[0] and then chain several nested keys (.data['clouds.yaml'] | from_yaml).clouds.default.auth.username) without guards. If the ConfigMap/Secret doesn't exist or has an unexpected structure, the error will be pretty opaque. Not a blocker, but worth keeping in mind if you see weird failures in CI.

roles/test_operator/tasks/runners/horizontest_runner.yml
```diff
@@ -1,29 +1,65 @@
...
-          'cifmw_test_operator_horizontest_admin_password':
-            _horizontest_admin_password_b64.stdout | b64decode
-        })
+        stage_vars_dict.cifmw_test_operator_horizontest_admin_username | default('') | length == 0 or
+        stage_vars_dict.cifmw_test_operator_horizontest_admin_password | default('') | length == 0
       }}
 
+- name: Fetch admin credentials from OpenStack config
+  when: _horizontest_fetch_creds | bool
+  block:
```
Valkyrie00, line 13: Hey @kstrenkova, small note, the `block`: doesn't have a `rescue`: section, so if any of the credential-fetching steps fail, you won't get much context in the logs about what went wrong. Not blocking, just something that could save debugging time later.

roles/test_operator/tasks/runners/horizontest_runner.yml
kstrenkova: You are right. I am counting on the fact that for a successful setup the ConfigMap and Secret should be there. If this would permanent code I would change it, but because of the time constraint I would like to keep it as is.

roles/test_operator/tasks/runners/horizontest_runner.yml
kstrenkova: Ack. I agree here as well, but same as before, I am expecting the WA to be around 2 weeks, so hopefully we will not need a debugging for it.

roles/test_operator/tasks/runners/horizontest_runner.yml
Valkyrie00: Sounds good, let's get it in! Just a heads-up, the password also leaks through the "Print CR before applying" debug task, so it's two places now. I know it's temporary, but these things tend to pop up in audits later :( and then we're the ones chasing the fix. No rush, but maybe a small follow-up once the dust settles?

Valkyrie00 (approved):
/lgtm

roles/test_operator/tasks/runners/horizontest_runner.yml
kstrenkova: I can take a look at making it safe (no showing of the `AdminPassword`) as a follow-up. Although, spoiler, I am planning to remove `AdminPassword` all together in the future :smile: Sadly, since we ship test-operator to customers it will need to stay with a deprecation warning for some time :(

evallesp (approved):
/approve

evallesp (approved):

PR #4090, opened 2026-08-05
Read whitebox DB password from osp-secret

ashu-011:
install_yamls PR openstack-k8s-operators/install_yamls#1158 replaced hardcoded passwords with dynamically generated per-service secrets. Whitebox Tempest needs Galera root credentials via whitebox-database.password, which must match DbRootPassword in osp-secret.
Fetch DbRootPassword from osp-secret at runtime and inject it into the top-level tempestconfRun.overrides before applying the Tempest CR. Workflow steps inherit this value from the global spec.

### DIFF
--- a/roles/test_operator/tasks/tempest-tests.yml
+++ b/roles/test_operator/tasks/tempest-tests.yml
@@ -191,6 +191,109 @@
               combine({'cifmw_test_operator_tempest_workflow': overriden_workflow})
           }}
 
+# Prefer k8s_info over bare `oc get` so kubeconfig/token match the rest of this role.
+- name: Fetch database root password from osp-secret
+  no_log: true
+  when: not cifmw_test_operator_dry_run | bool
+  kubernetes.core.k8s_info:
+    kubeconfig: "{{ cifmw_openshift_kubeconfig }}"
+    api_key: "{{ cifmw_openshift_token | default(omit) }}"
+    context: "{{ cifmw_openshift_context | default(omit) }}"
+    kind: Secret
+    name: osp-secret
+    namespace: "{{ stage_vars_dict.cifmw_test_operator_namespace }}"
+  register: _tempest_osp_secret
+  failed_when: false
+
+- name: Set whitebox database password fact from osp-secret
+  no_log: true
+  when:
+    - not cifmw_test_operator_dry_run | bool
+    - _tempest_osp_secret is defined
+    - _tempest_osp_secret.resources | default([]) | length > 0
+    - _tempest_osp_secret.resources[0].data.DbRootPassword is defined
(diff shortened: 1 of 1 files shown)


evallesp:
To fix CI you need to set "[test_operator]" at the start of the commit message tittle.

roles/test_operator/tasks/tempest-tests.yml
```diff
@@ -191,6 +191,47 @@
...
+
+    - name: Add whitebox database password to the overrides section in Tempest CR
+      no_log: true
+      vars:
+        _tempest_db_root_password: >-
+          {{ _tempest_osp_secret.resources[0].data.DbRootPassword | b64decode }}
+        _overrides_cleaned: >-
+          {{
+            (test_operator_cr.spec.tempestconfRun.overrides | default('')) |
+            regex_replace('whitebox-database\\.password\\s+\\S+\\s*', '')
```
evallesp: (non-blocking) suggestion: I think it might be better:
| regex_replace("whitebox-database\\.password\\s+.*?(\\s+|$)", "") | trim }}

Double quote allows us: \. to become literal.
Also \S+ to .*? allows us having multi word password.

ashu-011: Done

ashu-011:
test project attached in CIX https://redhat.atlassian.net/browse/OSPCIX-1483

evallesp (approved):
/lgtm

roles/test_operator/tasks/tempest-tests.yml
```diff
@@ -191,6 +191,47 @@
...
+        _overrides_cleaned: >-
+          {{
+            (test_operator_cr.spec.tempestconfRun.overrides | default('')) |
+            regex_replace("whitebox-database\.password\s+.*?(\s+|$)", "") | trim
+          }}
+      ansible.builtin.set_fact:
+        test_operator_cr: >-
+          {{
+              test_operator_cr |
+              combine({'spec': {'tempestconfRun': {'overrides':
```
evallesp: (blocking) question: Do we need to inject the new password to also workflow steps? (cifmw_test_operator_tempest_workflow), as each of them carries (IIUC) their own tempestconfRun.overrides with have no update about whitebox DB password.

evallesp:
(blocking) suggestion: Could you update the commit: "Fetch DbRootPassword from osp-secret at runtime and inject it into tempestconfRun.overrides (including workflow steps) before applying the Tempest CR."
IIUC we're not setting this in the workflow steps, we're doing this in the global level, so workflow steps also inherits the value.

evallesp (approved):
/lgtm
/approve

ashu-011:
retest

PR #4091, opened 2026-08-05
[federation] Recreate Keycloak Route with reencrypt TLS

Deydra71:
RHSSO only exposes keycloak.local via Ingress, so readiness against keycloak-<ns>.<domain> returns 503 after the manual Route was removed (in https://github.com/openstack-k8s-operators/ci-framework/pull/4074).

Restore an explicit reencrypt Route with the expected host and service CA `destinationCACertificate` to avoid the old passthrough SSL race.

### DIFF
--- a/roles/federation/tasks/run_keycloak_setup.yml
+++ b/roles/federation/tasks/run_keycloak_setup.yml
@@ -180,6 +180,51 @@
   retries: 30
   delay: 10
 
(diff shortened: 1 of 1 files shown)


Deydra71:
Tested in testproject, and it resolved the early keycloak failure

evallesp (approved):
/lgtm

afaranha (approved):

PR #4092, opened 2026-08-05
[scenarios] Pin DB root password for whitebox tempest compat

sbauza:
install_yamls PR openstack-k8s-operators/install_yamls#1158 replaced
hardcoded passwords with dynamically generated per-service secrets.
The `whitebox_tempest_plugin` tempestconf overrides still reference the
old hardcoded `whitebox-database.password` value (`12345678`), causing
`Access Denied` errors against Galera.

As a short-term fix, pin `DB_ROOT_PASSWORD` in the `multinode-ci` and
`hci_ceph_backends` scenarios so that the deployed Galera root password
matches what whitebox tempest expects. A proper long-term fix (reading
`DbRootPassword` from `osp-secret` at runtime) is tracked in #4090.

### DIFF
--- a/scenarios/centos-9/hci_ceph_backends.yml
+++ b/scenarios/centos-9/hci_ceph_backends.yml
@@ -15,6 +15,7 @@ cifmw_rgw_ssl_backward_compatibility: false
 cifmw_install_yamls_vars:
   BMO_SETUP: false
   INSTALL_CERT_MANAGER: false
+  DB_ROOT_PASSWORD: "12345678"
 
 cifmw_edpm_prepare_skip_crc_storage_creation: true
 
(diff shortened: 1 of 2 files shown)


gibizer (dismissed):
That sounds like a reasonable short term solution to unblock the work.

sbauza:
Closing in favor of the downstream ci-framework-jobs fix: https://gitlab.cee.redhat.com/ci-framework/ci-framework-jobs/-/merge_requests/3516

That MR pins DB_ROOT_PASSWORD in the edpm_base job definition where the whitebox tempest overrides live, which is a more targeted fix than changing the upstream scenario files.

PR #4093, opened 2026-08-07
Fix missing is succeeded guard on Wait for NHC CR task

antonioromito:
- The previous fix (f9b046b3) added `is succeeded` guards to three of the four `NodeHealthCheck` tasks but missed the "Wait for Node Health Check CR to be created" task
- Without this guard, a transient API failure on that task causes an unhandled error instead of being skipped gracefully — inconsistent with the other three tasks in the same flow


antonioromito:
recheck

evallesp (approved):
/lgtm

michburk (approved):
/approve

PR #4095, opened 2026-08-10
[adoption_osp_deploy] Skip nic1 configuration for bgp overcloud

eduolivares:
nic1/eth0 is configured during the VM creation and it does not need to be included in the os-net-config configuration file. When it is included, starting network.service fails because dhclient is already running on that NIC. That is the reason why the default template (for non-BGP environments) does not include nic1.


evallesp (approved):
/lgtm

michburk (approved):
/lgtm

michburk (approved):
/approve

PR #4096, opened 2026-08-10
[cleanup_openstack] fix reuse_ocp workflow

bogdando:
Fix BMH detach wait when no status exist

The until condition in detach_bmh.yaml accesses
bmh_status.resources[0].status.operationalStatus without
guarding against the case where the BareMetalHost CR exists
but has no .status subresource. This happens when the BMH
was created manually and Metal3 has not reconciled it (e.g.
pre-existing bare metal hosts in reuse_ocp workflows).
In that state, operationalStatus will never become
'detached' because Metal3 never provisioned the host.

Treat a missing .status as success: the detached annotation
is already applied by the preceding patch task, and a BMH
with no status was never Metal3-provisioned, so there is
nothing to deprovision.

Additionally chown stale root-owned log files for zuul
to access those without issue in reuse_ocp.

### DIFF
--- a/deploy-edpm-reuse.yaml
+++ b/deploy-edpm-reuse.yaml
@@ -108,6 +108,25 @@
         name: reproducer
         tasks_from: reuse_main
 
+    - name: Find stale root-owned log files
+      delegate_to: controller-0
+      ansible.builtin.find:
+        paths: /home/zuul/ci-framework-data/logs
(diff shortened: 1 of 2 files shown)


evallesp (approved):
/lgtm

nemarjan (approved):
/approve

PR #4098, opened 2026-08-12
[edpm_build_images] Fix bootc output directory permissions

rabi:
Run the bootc output directory creation task with privilege escalation so bootc builds can write under the RPM-installed edpm-image-builder path in /usr/share.


evallesp (approved):
/lgtm
/approve

PR #4100, opened 2026-08-12
[update_containers] Discover cinder and manila backend keys

rabi:
Resolve cinder volume and manila share image map keys from the live OpenStackControlPlane CR when callers do not provide them explicitly. This avoids emitting misleading default backend names and keeps backend-keyed customContainerImages aligned with the deployed control plane.

Depends-On: https://github.com/openstack-k8s-operators/ci-framework/pull/4104


rabi:
recheck

rabi:
recheck

rabi:
recheck

rabi:
recheck

evallesp:
Hi @rabi Do you think it might good to check if we can move this patch to use the python module: https://github.com/openstack-k8s-operators/ci-framework/pull/4117/changes#diff-169cd4c0f3c2f8df434e1ce8d7f495f29cc04c10bce40d212e199fafe18ed201R225

So we can move edpm_prepare to the new way while maintaining: deploy_architecture.yml edpm_baremetal_update callers in the previous way until. So modifying: update_containers/defaults/main.yml or update_containers.j2 wouldn't affect them. As with this new approach we don't need to update them.

What I'm unsure about the current state of the patch:
* cifmw_setup at deploy_architecture and edpm_baremetal_update where not updated while the role defaults were modified affecting also them. This might affect ast the new [] default (IIUC) would stop emiting the cindeVBolume and ManilaShare images overrides.
* Template emits a bare cinderVolumeImages: (null) when no cinder backends exist and .get(..., {}) | list runtimeError when value is null as it's not iterable.

rabi:
> Do you think it might good to check if we can move this patch to use the python module: https://github.com/openstack-k8s-operators/ci-framework/pull/4117

Thanks @evallesp. I had proposed this PR much before 4117 was proposed (though it merged quickly). At that time update_containers was the only way to update the openstackversion images without a big refactoring. Anyway, I'll try and change to use set_container and fix other issues.

roles/edpm_prepare/README.md
```diff
@@ -21,3 +21,5 @@ This role doesn't need privilege escalation.
 openstack control plane deployment status. Useful when using the role to only update the control plane resource, scenario where it may be in a `ready` status. Defaults to `30`.
 * `cifmw_edpm_prepare_extra_kustomizations`: (List) Extra Kustomizations to apply on top of the controlplane CRs. Defaults to `[]`.
 * `cifmw_edpm_prepare_bootc_os_image_url`: (String) Default bootc OS container image used when content-provider did not build one. Defaults to `quay.io/openstack-k8s-operators/edpm-bootc:latest-qcow2`.
+* `cifmw_edpm_prepare_cindervolumes`: (List) cinder volume backend names used as keys under `customContainerImages.cinderVolumeImages`. When empty, discovered from the kustomized `OpenStackControlPlane` CR. Defaults to `[]`.
```
tosky, line 24: but why cifmw_edpm_* for these variables? Aren't those containers mainly running on the control plane?

rabi: Though the role name is `edpm_prepare` it does  everything, deploying meta operator and the controlplane CR.

https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/edpm_prepare/tasks/kustomize_and_deploy.yml#L121

evallesp: Should we also update edpm_update? As now default value is [] instead of [default].
cinderVolumeImages/manilaShareImages would not emit anything from edpm_update.
As there we don't have: _ctlplane_spec.get

rabi: Not sure I understand what you mean by edpm_update? These vars are not used by update workflow (ex. edpm_baremetal_update). If we want to change them that should be separate PR I guess as update workflow roles still use update_containers.

evallesp:
@rabi its pending a rebase.

roles/edpm_prepare/README.md
evallesp: I think it might work!

rabi:
> @rabi its pending a rebase.

yeah, rebased for the nth time :smile:

evallesp (dismissed):
/lgtm

roles/edpm_prepare/tasks/kustomize_and_deploy.yml
```diff
@@ -120,6 +79,137 @@
...
+          }}
+        include_openstack: >-
+          {{
+            cifmw_set_containers_include_openstack |
+            default(cifmw_update_containers_openstack | default(false)) | bool
+          }}
+        apply: >-
+          {{
+            cifmw_set_containers_apply |
+            default(cifmw_update_containers | default(false)) | bool
```
nemarjan, line 160: (question) Hey @rabi, one small thing -- the old "Apply custom OpenStackVersion container overrides" task had apply: default(true) (old line 67), while the new consolidated task at line 160 uses apply: default(cifmw_update_containers | default(false)). For callers that set cifmw_set_containers_images without also setting cifmw_update_containers, this changes the apply behavior from true to false. Is that intentional, or should the fallback preserve the old default for that path?

rabi: Yeah, that default did flip from true to false. However, in practice cifmw_set_containers_images is currently unused (no job/scenario in ci-framework or ci-framework-jobs sets it), so there won't be any issues. But if we plan to use that interface in the future, I can change it though.

rabi: On second thought I think it's consistent with cifmw_update_containers (false), the set_containers module's own apply default (false), and how the same module is called in cifmw_setup/deploy_architecture.yml  (also false-default). So as it's not used we can change it to default(false).

nemarjan: Makes sense. Since no one currently sets cifmw_set_containers_images, there's no regression. And default(false) is the better default anyway -- applying to a live cluster should be explicit. Let's leave it as-is. LGTM.

michburk (approved):
/approve

PR #242, opened 2023-06-02
Collect meta operator related logs

raukadah:
As a pull request owner and reviewers, I checked that:
- [X] Appropriate testing is done and actually running

PR #4102, opened 2026-08-12
(multiple) Fix RHEL 10 compatibility issues

imatza-rh:
## Summary

Four fixes for RHEL 10 hypervisor support, all gated by `ansible_distribution_major_version >= 10` (no RHEL 9 regression).

### 1. dnsmasq: Version-conditional PID file path

RHEL 10 [inverted the SELinux file context equivalency](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/10/html/10.0_release_notes/new-features-and-enhancements) from `/run = /var/run` to `/var/run = /run`. Setting fcontext on `/var/run/` paths now fails:

```
ValueError: File spec /var/run/cifmw-dnsmasq.pid conflicts with
equivalency rule '/var/run /run'; Try adding '/run/cifmw-dnsmasq.pid' instead
```

Verified in Zuul build [2c922ff6](https://sf.apps.int.gpc.ocp-hub.prod.psi.redhat.com/zuul/t/components-integration/build/2c922ff69b684d528d7f51a72f2f14b9) (line 2992). The systemd unit already uses `/run/cifmw-dnsmasq.pid`.

Fix uses Jinja2 conditional: `/run/` on RHEL 10+, `/var/run/` on RHEL 9 (original behavior preserved).

### 2. repo_setup: Exclude python3 from rhos-release repos

`rhos-release` configures RHEL 9 repos on RHEL 10 hypervisors. Installing `nmstate` (el9) pulls `python3-libnmstate` which requires `python3.9dist(pyyaml)` (verified via `rpm -qR`), causing dnf to **downgrade** `python3` from 3.12 to 3.9 (confirmed in `/var/log/dnf.log` on serval64).

This breaks **9 RHEL 10 system tools** whose shebangs use `#!/usr/bin/python3 -sP` — the [`-P` flag](https://docs.python.org/3.11/using/cmdline.html#cmdoption-P) requires Python 3.11+:
- `firewalld`, `firewall-cmd`, `firewall-offline-cmd`
- `jsonschema`, `normalizer`, `distro`
- `activate-global-python-argcomplete`, `register-python-argcomplete`, `python-argcomplete-check-easy-install-script`

Fix uses `dnf config-manager --setopt=REPO.exclude=python3*,python3-libs*` on all rhos-release repos.

### 3. repo_setup: Disable rhos-release-extras on RHEL 10+

The `rhos-release-extras` repo uses `$releasever` (=10 on RHEL 10) but the server has no RHEL 10 content ([HTTP 404](https://sf.apps.int.gpc.ocp-hub.prod.psi.redhat.com/zuul/t/components-integration/build/0f7a8ea1286a4570a13aa69795c78837) at line 7561). This causes `devscripts make requirements` to fail.

### 4. repo_setup: Install python3-setuptools on RHEL 10+

RHEL 10 does not ship `python3-setuptools` by default (RHEL 9 did). Ansible's built-in `pip` module (`ansible/modules/pip.py:281`) imports `from pkg_resources import Requirement` — `pkg_resources` is part of `setuptools`. When the inner reproducer runs `ansible.builtin.pip` on the hypervisor to install dev-scripts Python dependencies, it fails with `ModuleNotFoundError: No module named 'pkg_resources'`.

SSH-verified on serval64 (RHEL 10.2): `rpm -qa | grep setuptools` returns empty, `python3 -c "import pkg_resources"` fails. After `dnf install python3-setuptools` (69.0.3 from rhel-10-baseos): import succeeds.

## Verification

SSH-verified on serval64.lab (RHEL 10.2):
- `python3 --version` → 3.9.18 (broken state after rhos-release)
- `python3.12 /usr/sbin/firewalld --debug` → starts normally
- `/var/log/dnf.log`: `Downgraded: python3-3.9.18-3.el9_4.13`
- `rpm -qR python3-libnmstate`: `python3.9dist(pyyaml)`
- `head -1 /usr/sbin/firewalld`: `#!/usr/bin/python3 -sP`

Build [36923d2e](https://sf.apps.int.gpc.ocp-hub.prod.psi.redhat.com/zuul/t/components-integration/build/36923d2ec94a476eb9e52e9a5d6d0477) confirmed dnsmasq SELinux fix works (line 3019).
Build [0f7a8ea1](https://sf.apps.int.gpc.ocp-hub.prod.psi.redhat.com/zuul/t/components-integration/build/0f7a8ea1286a4570a13aa69795c78837) confirmed python3 exclude works (521 ok tasks, firewalld started).

## References

- [RHEL 10.0 Release Notes — SELinux equivalency inversion](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/10/html/10.0_release_notes/new-features-and-enhancements)
- [Python 3.11 -P flag documentation](https://docs.python.org/3.11/using/cmdline.html#cmdoption-P)
- [Fedora SELinux policy issue #242 — /run /var/run equivalency](https://github.com/fedora-selinux/selinux-policy/issues/242)
- [Red Hat KB: Adding a SELinux context fails with ValueError](https://access.redhat.com/solutions/3294811)

### DIFF
--- a/roles/repo_setup/tasks/rhos_release.yml
+++ b/roles/repo_setup/tasks/rhos_release.yml
@@ -47,3 +47,51 @@
     cmd: >-
       rhos-release {{ cifmw_repo_setup_rhos_release_args }} \
         -t {{ cifmw_repo_setup_output }}
+
+- name: Make generated repos RHEL 10 safe
+  when:
+    - ansible_distribution_major_version | int >= 10
+  become: "{{ not cifmw_repo_setup_output.startswith(ansible_user_dir) }}"
+  vars:
+    _rhel_major: "{{ ansible_distribution_major_version }}"
+    _rhel_ver: "{{ ansible_distribution_version }}"
+    _base_url: >-
+      http://download.devel.redhat.com/released/rhel-{{ _rhel_major }}/RHEL-{{ _rhel_major }}/{{ _rhel_ver }}
+  block:
+    - name: Disable all rhos-release generated repos
+      ansible.builtin.shell:
+        cmd: >-
+          set -o pipefail;
+          for f in {{ cifmw_repo_setup_output }}/rhos-release*.repo
+          {{ cifmw_repo_setup_output }}/osp-trunk*.repo; do
+            [ -f "$f" ] || continue;
+            sed -i 's/^enabled=1/enabled=0/' "$f";
+          done
+      changed_when: true
+
+    - name: Add RHEL 10 base repos to generated directory
+      ansible.builtin.copy:
+        dest: "{{ cifmw_repo_setup_output }}/rhel-{{ _rhel_major }}-{{ item.name }}.repo"
+        content: |
+          [rhel-{{ _rhel_major }}-{{ item.name }}]
+          name=RHEL {{ _rhel_major }} {{ item.path }}
+          baseurl={{ _base_url }}/{{ item.path }}/x86_64/os
+          gpgcheck=0
+          enabled=1
+        mode: "0644"
+      loop:
+        - name: baseos
+          path: BaseOS
+        - name: appstream
+          path: AppStream
+      loop_control:
+        label: "{{ item.name }}"
+
+    - name: Ensure python3 build deps are available on RHEL 10+
+      become: true
+      ansible.builtin.dnf:
+        name:
+          - python3-setuptools
+          - gcc
+          - python3-devel
+        state: present
--- a/roles/reproducer/tasks/rhos_release.yml
+++ b/roles/reproducer/tasks/rhos_release.yml
@@ -9,13 +9,39 @@
   ansible.builtin.command:
     cmd: "rhos-release {{ _rhos_release_args_override | default(cifmw_repo_setup_rhos_release_args) | default('rhel') }}"
 
+# ansible_* under delegate_to is evaluated from the inventory host
+# (hypervisor). RHEL 9 computes on a RHEL 10 hypervisor must not get this
+# block (uni03gamma HCI). Read VERSION_ID from the connection target.
+# Do not use setup: without delegate_facts it assigns facts to the
+# hypervisor (Ansible "Delegating facts").
+- name: Detect OS on the rhos-release target
+  ansible.builtin.command:
+    argv:
+      - awk
+      - -F=
+      - /^VERSION_ID=/ { gsub(/"/, "", $2); print $2 }
+      - /etc/os-release
+  register: _rhos_release_version_id
+  changed_when: false
+  failed_when: >-
+    _rhos_release_version_id.rc != 0 or
+    _rhos_release_version_id.stdout | trim | length == 0
+
+- name: Show detected rhos-release target OS
+  ansible.builtin.debug:
+    msg: "rhos-release target VERSION_ID={{ _rhos_release_version_id.stdout }}"
+
+# Disable-all is intentional on RHEL 10 (not extras-only). Selective
+# python3 exclude is overwritten by repo_setup sync_repos.yml, and
+# osp-trunk-candidate.repo is not a rhos-release-* file. See KB
+# "Python3 Contamination - Structural Problem (2026-08-18)".
 - name: Make RHEL 10+ safe from rhos-release RHEL 9 repo contamination
   when:
-    - ansible_distribution_major_version | int >= 10
+    - ((_rhos_release_version_id.stdout | trim).split('.') | first | int) >= 10
   become: true
   vars:
-    _rhel_major: "{{ ansible_distribution_major_version }}"
-    _rhel_ver: "{{ ansible_distribution_version }}"
+    _rhel_ver: "{{ _rhos_release_version_id.stdout | trim }}"
+    _rhel_major: "{{ (_rhos_release_version_id.stdout | trim).split('.') | first }}"
     _base_url: >-
       http://download.devel.redhat.com/released/rhel-{{ _rhel_major }}/RHEL-{{ _rhel_major }}/{{ _rhel_ver }}
   block:
--- a/roles/reproducer/tasks/rhos_release.yml
+++ b/roles/reproducer/tasks/rhos_release.yml
@@ -9,6 +9,50 @@
   ansible.builtin.command:
     cmd: "rhos-release {{ _rhos_release_args_override | default(cifmw_repo_setup_rhos_release_args) | default('rhel') }}"
 
+- name: Make RHEL 10+ safe from rhos-release RHEL 9 repo contamination
+  when:
+    - ansible_distribution_major_version | int >= 10
+  become: true
+  vars:
+    _rhel_major: "{{ ansible_distribution_major_version }}"
+    _rhel_ver: "{{ ansible_distribution_version }}"
+    _base_url: >-
+      http://download.devel.redhat.com/released/rhel-{{ _rhel_major }}/RHEL-{{ _rhel_major }}/{{ _rhel_ver }}
+  block:
+    - name: Disable all rhos-release generated repos
+      ansible.builtin.shell:
+        cmd: >-
+          set -o pipefail;
+          for f in /etc/yum.repos.d/rhos-release*.repo
+          /etc/yum.repos.d/osp-trunk*.repo; do
+            [ -f "$f" ] || continue;
+            sed -i 's/^enabled=1/enabled=0/' "$f";
+          done
+      changed_when: true
+
+    - name: Add RHEL 10 base repos
+      ansible.builtin.yum_repository:
+        name: "rhel-{{ _rhel_major }}-{{ item.name }}"
+        description: "RHEL {{ _rhel_major }} {{ item.path }}"
+        baseurl: "{{ _base_url }}/{{ item.path }}/x86_64/os"
+        gpgcheck: false
+        enabled: true
+      loop:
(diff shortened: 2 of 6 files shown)


roles/repo_setup/tasks/rhos_release.yml
```diff
@@ -47,3 +47,42 @@
...
+  when:
+    - ansible_distribution_major_version | int >= 10
+  become: true
+  block:
+    - name: Get rhos-release repo IDs
+      ansible.builtin.shell:
+        cmd: >-
+          set -o pipefail;
+          dnf repolist --all 2>/dev/null |
+          awk '/rhos/ {print $1}' |
```
davidsaOpenu: pattern matches "rhos" as a substring anywhere in the repo ID, not just as a prefix. Anchoring with awk '$1 ~ /^rhos/' would be tighter.

roles/reproducer/tasks/rhos_release.yml
```diff
@@ -9,6 +9,46 @@
...
+  when:
+    - ansible_distribution_major_version | int >= 10
+  become: true
+  block:
+    - name: Get rhos-release repo IDs
+      ansible.builtin.shell:
+        cmd: >-
+          set -o pipefail;
+          dnf repolist --all 2>/dev/null |
+          awk '/rhos/ {print $1}' |
```
davidsaOpenu: pattern matches "rhos" as a substring anywhere in the repo ID, not just as a prefix. Anchoring with awk '$1 ~ /^rhos/' would be tighter.

roles/reproducer/tasks/rhos_release.yml
```diff
@@ -9,6 +9,46 @@
   ansible.builtin.command:
     cmd: "rhos-release {{ _rhos_release_args_override | default(cifmw_repo_setup_rhos_release_args) | default('rhel') }}"
 
+- name: Protect RHEL 10+ from rhos-release RHEL 9 repo contamination
```
davidsaOpenu: These two blocks roles/repo_setup/tasks/rhos_release.yml and roles/reproducer/tasks/rhos_release.yml are nearly identical (same repo-discovery awk, same exclude logic).

The suggested fix is to pull the shared logic out into its own small, focused role — following the pattern already used throughout ci-framework.

New role: roles/cifmw_rhel10_repo_guard/
```
# roles/cifmw_rhel10_repo_guard/defaults/main.yml
cifmw_rhel10_repo_guard_repo_pattern: '^rhos'
cifmw_rhel10_repo_guard_exclude_globs: "python3*,python3-libs*"
cifmw_rhel10_repo_guard_extra_packages: []   # e.g. [gcc, python3-devel] when a caller needs build deps too
```

```
# roles/cifmw_rhel10_repo_guard/tasks/main.yml
---
- name: Protect system Python from RHEL 9 repo contamination
  when:
    - ansible_distribution_major_version | int >= 10
  become: true
  block:
    - name: Get rhos-release repo IDs
      ansible.builtin.shell:
        cmd: >-
          set -o pipefail;
          dnf repolist --all 2>/dev/null |
          awk '$1 ~ /{{ cifmw_rhel10_repo_guard_repo_pattern }}/ {print $1}' |
          tr '\n' ' '
      register: _rhel10_repo_guard_repo_ids
      changed_when: false

    - name: Exclude python3 packages from matched repos
      ansible.builtin.command:
        cmd: >-
          dnf config-manager --save
          --setopt={{ item }}.exclude={{ cifmw_rhel10_repo_guard_exclude_globs }}
      loop: "{{ _rhel10_repo_guard_repo_ids.stdout.split() }}"
      register: _rhel10_repo_guard_exclude
      changed_when: "'changed' in _rhel10_repo_guard_exclude.stdout | default('changed')"
      # or, more simply, just check dnf's actual exit/no-op behavior if it
      # reports one consistently — the point is: not an unconditional `true`

    - name: Disable rhos-release-extras repo
      ansible.builtin.command:
        cmd: dnf config-manager --disable rhos-release-extras
      register: _rhel10_repo_guard_disable_extras
      failed_when: false
      changed_when: _rhel10_repo_guard_disable_extras.rc == 0

    - name: Ensure required packages are installed
      ansible.builtin.package:
        name: "{{ ['python3-setuptools'] + cifmw_rhel10_repo_guard_extra_packages }}"
        state: present

```

And then the calls become:

```
# roles/repo_setup/tasks/rhos_release.yml
- name: Protect system Python on RHEL 10+ from RHEL 9 repo contamination
  ansible.builtin.include_role:
    name: cifmw_rhel10_repo_guard
  vars:
    cifmw_rhel10_repo_guard_extra_packages:
      - gcc
      - python3-devel
```


```
# roles/reproducer/tasks/rhos_release.yml
- name: Protect RHEL 10+ from rhos-release RHEL 9 repo contamination
  ansible.builtin.include_role:
    name: cifmw_rhel10_repo_guard
```

roles/repo_setup/tasks/rhos_release.yml
imatza-rh: Fixed — anchored to `$1 ~ /^rhos/` in c20f30c. Verified on live RHEL 10.2 node: both patterns match the same 22 repos (all rhos-release and rhosp-* IDs), `rhelosp-ceph-*` correctly excluded by both.

roles/reproducer/tasks/rhos_release.yml
imatza-rh: Fixed — same anchoring applied in c20f30c.

roles/reproducer/tasks/rhos_release.yml
imatza-rh: Agreed on the DRY principle — the shared awk+exclude logic is a good extraction candidate. However the two blocks have meaningful differences that make a direct extraction tricky:

1. **Disable scope**: repo_setup disables only `rhos-release-extras`; reproducer disables 6 additional repos (`rhosp-rhel-9.4-{appstream,baseos,fdp-cdn,ha,image-build-override}`)
2. **Base repo setup**: reproducer creates RHEL 10 base repos via `yum_repository` (compute VMs don't have them); repo_setup doesn't need this (hypervisor already has base repos)
3. **Package module**: The proposed role uses `ansible.builtin.package`, which breaks on compute VMs where `ansible_facts['pkg_mgr']` isn't gathered — the `pkg_mgr` bug this push fixes (see build [d69d8ad4](https://sf.apps.int.gpc.ocp-hub.prod.psi.redhat.com/logs/d69/components-integration/d69d8ad4231d4109b87b36b89134abb3/)). Must use `ansible.builtin.dnf` for hosts without gathered facts.

I'd suggest merging the current fixes (awk anchoring + dnf) to unblock the pipeline, then extracting the shared role as a follow-up with proper parameterization for these differences. WDYT?

imatza-rh:
recheck

roles/reproducer/tasks/rhos_release.yml
evallesp: I think it's a really good idea.
Unsure if handling now tbh, but it'd be really good to have this noted to work on this soon.

evallesp (dismissed):
/lgtm

roles/reproducer/tasks/rhos_release.yml
```diff
@@ -9,6 +9,50 @@
...
+    - ansible_distribution_major_version | int >= 10
+  become: true
+  vars:
+    _rhel_major: "{{ ansible_distribution_major_version }}"
+    _rhel_ver: "{{ ansible_distribution_version }}"
+    _base_url: >-
+      http://download.devel.redhat.com/released/rhel-{{ _rhel_major }}/RHEL-{{ _rhel_major }}/{{ _rhel_ver }}
+  block:
+    - name: Disable all rhos-release generated repos
+      ansible.builtin.shell:
```
Valkyrie00, line 23: Hey @imatza-rh 👋 ,I have the same doubt as in the message above from coderabbit,I noticed the disable loop turns off every rhos-release and osp-trunk repo, while the PR description talks about excluding python3 and disabling only rhos-release-extras. Was the pivot to disable-all intentional?

imatza-rh: Yes - disable-all is intentional. The PR description is stale from the first iteration (python3 exclude + extras-only).

Selective `python3*` exclude does not stick: `roles/repo_setup/tasks/sync_repos.yml` wipes `/etc/yum.repos.d` and copies the staging dir, which drops those `exclude=` settings. A narrow `rhos-release*` disable also misses `osp-trunk-candidate.repo`, which is how el9 python3 came back. Disabling every `rhos-release*.repo` and `osp-trunk*.repo`, then adding RHEL 10 BaseOS/AppStream, is the structural fix.

This follow-up does not change that. It only stops the block from running on the wrong OS. `ansible_distribution_major_version` is evaluated from the inventory host (the hypervisor), so on uni03gamma a RHEL 10 hypervisor applied `dnf install python3-devel` to RHEL 9.6 HCI computes:

https://sf.apps.int.gpc.ocp-hub.prod.psi.redhat.com/zuul/t/components-integration/build/039feeddebcc4580a1e70e41c3215d0a

(`python3-devel-3.12.12-5.el10_2` vs `python3-3.9.21-2.el9_6.6`). The gate now reads `VERSION_ID` from the delegated target's `/etc/os-release`. RHEL 9 computes skip the block; RHEL 10 controller/hypervisor still get disable-all.

evallesp (approved):
/lgtm

nemarjan (approved):
/approve

PR #4105, opened 2026-08-13
Bgp adoption fix nic1 2ndtry

eduolivares:
fix-on-fix because [1] did not work as expected.

nic1/eth0 obtains a default route via DHCP, conflicting with the default
route obtained via BGP, which is the one that should be applied.
For that reason, nic1 needs to be included in os-net-config file, but
disabling DHCP on it, in order to avoid that conflict.

Then, the temporary default routes for the overcloud nodes need to be
added right before the overcloud deployment starts.

[1] https://github.com/openstack-k8s-operators/ci-framework/pull/4095

### DIFF
--- a/hooks/playbooks/adoption_bgp_pre_overcloud.yaml
+++ b/hooks/playbooks/adoption_bgp_pre_overcloud.yaml
@@ -14,37 +14,24 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
-- name: BGP pre_undercloud hook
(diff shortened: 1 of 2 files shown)


eduolivares:
recheck

lmiccini (approved):

jistr (approved):

evallesp (approved):
/lgtm

evallesp (approved):
/approve

PR #4108, opened 2026-08-13
Adapt install-openstack-lightspeed playbook to current operator

malingatembo:
Updated the playbook to work with the current OpenStack Lightspeed operator after org migration from openstack-lightspeed to openstack-k8s-operators.

Changes:
1. Namespace: Changed default from openshift-lightspeed to openstack-lightspeed to match operator's suggested namespace in csv
2. CR creation: Added STEP 3 to create OpenStackLightspeed CR
   - Deploys the actual service (not just operator)
   - Configurable LLM parameters (endpoint, credentials, model)
   - Graceful timeout with ignore_errors for test environments
3. Kubeconfig fix: Resolved variable recursion bug on line 29
   - Renamed cifmw_openshift_kubeconfig to kubeconfig_path
   - Prevents 'recursive loop detected' error

Jira ticket: [OSPRH-33342](https://redhat.atlassian.net/browse/OSPRH-33342)

### DIFF
--- a/hooks/playbooks/install-openstack-lightspeed.yml
+++ b/hooks/playbooks/install-openstack-lightspeed.yml
@@ -135,6 +135,7 @@
           stringData:
             apitoken: "{{ cifmw_openstack_lightspeed_api_token }}"
       when: cifmw_openstack_lightspeed_api_token is defined
+      no_log: true
 
     - name: Verify LLM credentials secret exists
       kubernetes.core.k8s_info:
@@ -145,7 +146,27 @@
       failed_when: secret_check.resources | length == 0
       when: cifmw_openstack_lightspeed_api_token is not defined
 
-    - name: Create TLS CA certificate bundle ConfigMap
+    - name: Download TLS CA certificate from URL
+      ansible.builtin.uri:
+        url: "{{ cifmw_openstack_lightspeed_ca_cert_url }}"
+        return_content: true
+      register: ca_cert_download
+      when: cifmw_openstack_lightspeed_ca_cert_url is defined
+
+    - name: Create TLS CA certificate bundle ConfigMap (from URL)
+      kubernetes.core.k8s:
+        state: present
+        definition:
+          apiVersion: v1
+          kind: ConfigMap
+          metadata:
+            name: "{{ cifmw_openstack_lightspeed_tls_ca_cert_bundle | default('openstack-lightspeed-certs') }}"
+            namespace: "{{ openstack_lightspeed_namespace }}"
+          data:
+            ca-bundle.crt: "{{ ca_cert_download.content }}"
+      when: cifmw_openstack_lightspeed_ca_cert_url is defined
+
+    - name: Create TLS CA certificate bundle ConfigMap (from content)
       kubernetes.core.k8s:
         state: present
         definition:
@@ -156,7 +177,9 @@
             namespace: "{{ openstack_lightspeed_namespace }}"
           data:
             ca-bundle.crt: "{{ cifmw_openstack_lightspeed_ca_cert }}"
-      when: cifmw_openstack_lightspeed_ca_cert is defined
+      when:
+        - cifmw_openstack_lightspeed_ca_cert is defined
+        - cifmw_openstack_lightspeed_ca_cert_url is not defined
 
     - name: Verify TLS certificate bundle ConfigMap exists
       kubernetes.core.k8s_info:
@@ -165,7 +188,9 @@
         namespace: "{{ openstack_lightspeed_namespace }}"
       register: cert_check
       failed_when: cert_check.resources | length == 0
-      when: cifmw_openstack_lightspeed_ca_cert is not defined
+      when:
+        - cifmw_openstack_lightspeed_ca_cert is not defined
+        - cifmw_openstack_lightspeed_ca_cert_url is not defined
 
     # STEP 4: Create OpenStackLightspeed CR to deploy the service
     # Note: Uses configurable parameters for LLM integration
@@ -189,23 +214,6 @@
             modelName: "{{ cifmw_openstack_lightspeed_model_name }}"
             llmCredentials: "{{ cifmw_openstack_lightspeed_llm_credentials | default('openstack-lightspeed-apitoken') }}"
             tlsCACertBundle: "{{ cifmw_openstack_lightspeed_tls_ca_cert_bundle | default('openstack-lightspeed-certs') }}"
-
-    - name: Wait for OpenStackLightspeed CR to be ready
-      kubernetes.core.k8s_info:
-        api_version: lightspeed.openstack.org/v1beta1
-        kind: OpenStackLightspeed
-        name: "{{ cifmw_openstack_lightspeed_cr_name | default('openstacklightspeed-sample') }}"
-        namespace: "{{ openstack_lightspeed_namespace }}"
-      register: osls_cr
-      until:
-        - osls_cr.resources is defined
-        - osls_cr.resources | length > 0
-        - osls_cr.resources[0].status is defined
-        - osls_cr.resources[0].status.conditions is defined
-        - osls_cr.resources[0].status.conditions | selectattr('type', 'equalto', 'Ready') | selectattr('status', 'equalto', 'True') | list | length > 0
-      retries: 60
-      delay: 10
-      ignore_errors: true
 
     - name: Display deployment summary
       ansible.builtin.debug:
--- a/hooks/playbooks/install-openstack-lightspeed.yml
+++ b/hooks/playbooks/install-openstack-lightspeed.yml
@@ -145,6 +145,7 @@
       register: secret_check
       failed_when: secret_check.resources | length == 0
       when: cifmw_openstack_lightspeed_api_token is not defined
+      no_log: true 
 
     - name: Download TLS CA certificate from URL
       ansible.builtin.uri:
@@ -153,7 +154,7 @@
       register: ca_cert_download
       when: cifmw_openstack_lightspeed_ca_cert_url is defined
 
-    - name: Create TLS CA certificate bundle ConfigMap (from URL)
+    - name: Create TLS CA certificate bundle ConfigMap
       kubernetes.core.k8s:
         state: present
         definition:
@@ -163,23 +164,9 @@
             name: "{{ cifmw_openstack_lightspeed_tls_ca_cert_bundle | default('openstack-lightspeed-certs') }}"
             namespace: "{{ openstack_lightspeed_namespace }}"
           data:
-            ca-bundle.crt: "{{ ca_cert_download.content }}"
-      when: cifmw_openstack_lightspeed_ca_cert_url is defined
-
-    - name: Create TLS CA certificate bundle ConfigMap (from content)
-      kubernetes.core.k8s:
-        state: present
-        definition:
-          apiVersion: v1
-          kind: ConfigMap
-          metadata:
-            name: "{{ cifmw_openstack_lightspeed_tls_ca_cert_bundle | default('openstack-lightspeed-certs') }}"
-            namespace: "{{ openstack_lightspeed_namespace }}"
-          data:
-            ca-bundle.crt: "{{ cifmw_openstack_lightspeed_ca_cert }}"
-      when:
-        - cifmw_openstack_lightspeed_ca_cert is defined
-        - cifmw_openstack_lightspeed_ca_cert_url is not defined
+            ca-bundle.crt: >-
+              {{ ca_cert_download.content if cifmw_openstack_lightspeed_ca_cert_url is defined else cifmw_openstack_lightspeed_ca_cert }}
+      when: cifmw_openstack_lightspeed_ca_cert_url is defined or cifmw_openstack_lightspeed_ca_cert is defined
 
     - name: Verify TLS certificate bundle ConfigMap exists
       kubernetes.core.k8s_info:
--- a/hooks/playbooks/install-openstack-lightspeed.yml
+++ b/hooks/playbooks/install-openstack-lightspeed.yml
@@ -19,16 +19,16 @@
   connection: local
   vars:
     # OpenStack Lightspeed configuration
-    # Note: Installing in openshift-lightspeed namespace to ensure compatibility
-    openstack_lightspeed_namespace: "{{ cifmw_openstack_lightspeed_namespace | default('openshift-lightspeed') }}"
+    # Note: Using openstack-lightspeed namespace per operator's suggested namespace
+    openstack_lightspeed_namespace: "{{ cifmw_openstack_lightspeed_namespace | default('openstack-lightspeed') }}"
     openstack_lightspeed_operator_group: "{{ cifmw_openstack_lightspeed_operator_group | default('openstack-lightspeed-operator-group') }}"
     openstack_lightspeed_catalog_image: "{{ cifmw_openstack_lightspeed_catalog_image | default('quay.io/openstack-lightspeed/operator-catalog:latest') }}"
     openstack_lightspeed_catalog_name: "{{ cifmw_openstack_lightspeed_catalog_name | default('openstack-lightspeed-catalog') }}"
 
     # Kubeconfig path - use user-provided or default to CRC location
-    cifmw_openshift_kubeconfig: "{{ cifmw_openshift_kubeconfig | default(ansible_env.HOME ~ '/.crc/machines/crc/kubeconfig') }}"
+    kubeconfig_path: "{{ cifmw_openshift_kubeconfig | default(ansible_env.HOME ~ '/.crc/machines/crc/kubeconfig') }}"
   environment:
-    KUBECONFIG: "{{ cifmw_openshift_kubeconfig }}"
+    KUBECONFIG: "{{ kubeconfig_path }}"
 
   tasks:
     # STEP 1: Deploy OpenStack Lightspeed catalog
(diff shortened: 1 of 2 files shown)


malingatembo:
recheck

hooks/playbooks/install-openstack-lightspeed.yml
```diff
@@ -121,8 +121,45 @@
...
+        state: present
+        definition:
+          apiVersion: lightspeed.openstack.org/v1beta1
+          kind: OpenStackLightspeed
+          metadata:
+            name: "{{ cifmw_openstack_lightspeed_cr_name | default('openstacklightspeed-sample') }}"
+            namespace: "{{ openstack_lightspeed_namespace }}"
+          spec:
+            llmCredentials: "{{ cifmw_openstack_lightspeed_llm_credentials | default('openstack-lightspeed-apitoken') }}"
+            llmEndpoint: "{{ cifmw_openstack_lightspeed_llm_endpoint | default('https://api.models.corp/v1/chat/completions') }}"
```
lpiwowar: **issue (blocking):** ⚠️ This link should not be part of the repo (downstream).

hooks/playbooks/install-openstack-lightspeed.yml
```diff
@@ -121,8 +121,45 @@
...
+      kubernetes.core.k8s:
+        state: present
+        definition:
+          apiVersion: lightspeed.openstack.org/v1beta1
+          kind: OpenStackLightspeed
+          metadata:
+            name: "{{ cifmw_openstack_lightspeed_cr_name | default('openstacklightspeed-sample') }}"
+            namespace: "{{ openstack_lightspeed_namespace }}"
+          spec:
+            llmCredentials: "{{ cifmw_openstack_lightspeed_llm_credentials | default('openstack-lightspeed-apitoken') }}"
```
lpiwowar: **issue (blocking):** Where is the secret created? IMO it should be created as part of this playbook.

---

Btw, we have to find a mechanism how to pass securely the token into the job and handle it securely. It must not appear in the logs. There should be somewhere a repo with Zuul secrets somewhere if I remember correctly. The secrets are stored there encrypted using public key.

---

**Update:** For upstream we might need to add the token here [1]. I'm not 💯 sure about it. Further investigation would be needed. Also, we need to agree what provider we are going to use upstream (OpenAI?).

[1] https://review.rdoproject.org/r/plugins/gitiles/config (`zuul.d/secrets.yaml`)

hooks/playbooks/install-openstack-lightspeed.yml
```diff
@@ -121,8 +121,45 @@
...
+            llmEndpoint: "{{ cifmw_openstack_lightspeed_llm_endpoint | default('https://api.models.corp/v1/chat/completions') }}"
+            llmEndpointType: "{{ cifmw_openstack_lightspeed_llm_endpoint_type | default('openai') }}"
+            modelName: "{{ cifmw_openstack_lightspeed_model_name | default('openai/gpt-oss-20b') }}"
+            tlsCACertBundle: "{{ cifmw_openstack_lightspeed_tls_ca_cert_bundle | default('openstack-lightspeed-certs') }}"
```
lpiwowar, line 199: **issue (blocking):** Where is the ConfigMap created? IMO it should be created as part of this playbook.

lpiwowar (changes requested):
Good start!:) Couple of comments.

hooks/playbooks/install-openstack-lightspeed.yml
malingatembo: Fixed, please check commit c76fd9cf0

- Removed the hardcoded endpoints 
- llmEndpoint and modelName are now required variables and have no set defaults. This requirement has also been added to the README

hooks/playbooks/install-openstack-lightspeed.yml
malingatembo: Fixed in commit c76fd9c. Added secret creation as STEP 3
The playbook now 
1. creates the secret from the `cifmw_openstack_lightspeed_api_token` when the variable is provided when the varible is provided 
2. validate that the secret exists when the variable isnot provided
3. otherwise fails when the secret is missing 

Thank you for the thought ahead. I will take a look :), and also I think Zuul's encrypted secrets can be parsed as Ansible vars, configured in the CI job setup ( I'll have to check it )

I will also link the link in the OSPRH-33343, it's exactly the ticket i'm starting work on to deal with this.

hooks/playbooks/install-openstack-lightspeed.yml
malingatembo: In commit  c76fd9c. Playbook now creates the ConfigMap from `cifmw_openstack_lightspeed_ca_cert` when the var is provided. 
 Please check it. Thank you

malingatembo:
@lpiwowar, please take a look at commit c76fd9c where the three blocking issues are addressed. Thanks

hooks/playbooks/install-openstack-lightspeed.yml
```diff
@@ -121,8 +121,95 @@
...
+        state: present
+        definition:
+          apiVersion: lightspeed.openstack.org/v1beta1
+          kind: OpenStackLightspeed
+          metadata:
+            name: "{{ cifmw_openstack_lightspeed_cr_name | default('openstacklightspeed-sample') }}"
+            namespace: "{{ openstack_lightspeed_namespace }}"
+          spec:
+            # REQUIRED: LLM endpoint must be provided via cifmw_openstack_lightspeed_llm_endpoint
+            # Example: https://your-llm-service/v1/chat/completions
```
lpiwowar: **suggestion (non-blocking):** Comment like this is not necessarey. We are already documenting the interface for this playbook in the README.md

hooks/playbooks/install-openstack-lightspeed.yml
```diff
@@ -121,8 +121,95 @@
...
+            llmEndpoint: "{{ cifmw_openstack_lightspeed_llm_endpoint }}"
+            llmEndpointType: "{{ cifmw_openstack_lightspeed_llm_endpoint_type | default('openai') }}"
+            # REQUIRED: Model name must be provided via cifmw_openstack_lightspeed_model_name
+            # Example: gpt-4, llama3.1:8b, granite-3.1-8b
```
lpiwowar: ditto

hooks/playbooks/install-openstack-lightspeed.yml
```diff
@@ -121,8 +121,95 @@
...
+      when: cifmw_openstack_lightspeed_api_token is defined
+
+    - name: Verify LLM credentials secret exists
+      kubernetes.core.k8s_info:
+        kind: Secret
+        name: "{{ cifmw_openstack_lightspeed_llm_credentials | default('openstack-lightspeed-apitoken') }}"
+        namespace: "{{ openstack_lightspeed_namespace }}"
+      register: secret_check
+      failed_when: secret_check.resources | length == 0
+      when: cifmw_openstack_lightspeed_api_token is not defined
```
lpiwowar, line 147: **issue (blocking):** This should have 💯 `no_log: true` to prevent accidental leakage of the token through the logs.

hooks/playbooks/install-openstack-lightspeed.yml
```diff
@@ -121,8 +121,95 @@
...
+      kubernetes.core.k8s:
+        state: present
+        definition:
+          apiVersion: v1
+          kind: ConfigMap
+          metadata:
+            name: "{{ cifmw_openstack_lightspeed_tls_ca_cert_bundle | default('openstack-lightspeed-certs') }}"
+            namespace: "{{ openstack_lightspeed_namespace }}"
+          data:
+            ca-bundle.crt: "{{ cifmw_openstack_lightspeed_ca_cert }}"
```
lpiwowar: **suggection (probably blocking):** I think we might want to be able to download the custom CA cert from an URL the same way as here [1]. Except that we are not creating a `Secret` but `ConfigMap`. This is downstream related (let's talk about it on Slack).


[1] https://github.com/openstack-k8s-operators/ci-framework/blob/main/hooks/playbooks/install_custom_ca_certs.yaml#L2

hooks/playbooks/install-openstack-lightspeed.yml
```diff
@@ -121,8 +121,95 @@
...
+      register: osls_cr
+      until:
+        - osls_cr.resources is defined
+        - osls_cr.resources | length > 0
+        - osls_cr.resources[0].status is defined
+        - osls_cr.resources[0].status.conditions is defined
+        - osls_cr.resources[0].status.conditions | selectattr('type', 'equalto', 'Ready') | selectattr('status', 'equalto', 'True') | list | length > 0
+      retries: 60
+      delay: 10
+      ignore_errors: true
```
lpiwowar: **question (blocking):** Is it possible for this step to ever fail with `ignore_errors: true`?

lpiwowar (changes requested):
Nice update! Thank you!:) Couple of follow up comments.

hooks/playbooks/install-openstack-lightspeed.yml
malingatembo: Thanks for catching the error in this logic. I've removed this wait task entirely in commit 97d20cc5. The playbook now installs the operator and creates the CR.

I was thinking that we can add the validation logic instead to the ticket: Task
[OSPRH-33343]; Create a job upstream that deploys openstack lightspeed. what do you think?

hooks/playbooks/install-openstack-lightspeed.yml
malingatembo: Fixed in commit 97d20cc5

hooks/playbooks/install-openstack-lightspeed.yml
malingatembo: good point, removed in 4c0c896e76cb95f31374e233e00f7b1074ca27b8

hooks/playbooks/install-openstack-lightspeed.yml
malingatembo: thanks, removed in 4c0c896e76cb95f31374e233e00f7b1074ca27b8

hooks/playbooks/install-openstack-lightspeed.yml
malingatembo: i've implemented url download support in 97d20cc5 (using ansible.builtin.uri). In `hooks/playbooks/install_custom_ca_certs.yaml,` they use get_url + slurp pattern. Happy to discuss ours vs their implemtnaion.

hooks/playbooks/install-openstack-lightspeed.yml
```diff
@@ -121,8 +121,103 @@
...
+        state: present
+        definition:
+          apiVersion: v1
+          kind: ConfigMap
+          metadata:
+            name: "{{ cifmw_openstack_lightspeed_tls_ca_cert_bundle | default('openstack-lightspeed-certs') }}"
+            namespace: "{{ openstack_lightspeed_namespace }}"
+          data:
+            ca-bundle.crt: "{{ ca_cert_download.content }}"
+      when: cifmw_openstack_lightspeed_ca_cert_url is defined
```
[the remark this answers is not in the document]
malingatembo: The HTTP in the example was an error. The actual Red Hat cert server (https://certs.corp.redhat.com) uses HTTPS. Example has been updated in 7b0eb9a679b31f83b62d9392c551400e4e6df3f2
Runtime implementation is not necessary

hooks/playbooks/install-openstack-lightspeed.yml
lpiwowar: @malingatembo, we want to add `no_log: true` to the second task that is handling the Secret as well. Since it is handling the Secret in some way, there is a slight chance that with `-vvvv` it would expose the value. It is better to be safe than sorry here.

hooks/playbooks/install-openstack-lightspeed.yml
```diff
@@ -121,8 +121,99 @@
...
+          apiVersion: v1
+          kind: ConfigMap
+          metadata:
+            name: "{{ cifmw_openstack_lightspeed_tls_ca_cert_bundle | default('openstack-lightspeed-certs') }}"
+            namespace: "{{ openstack_lightspeed_namespace }}"
+          data:
+            ca-bundle.crt: "{{ cifmw_openstack_lightspeed_ca_cert }}"
+      when:
+        - cifmw_openstack_lightspeed_ca_cert is defined
+        - cifmw_openstack_lightspeed_ca_cert_url is not defined
```
lpiwowar: **suggestion (non-blocking):** These two can be probably merged into one. Something like:

```
ca-bundle.crt: >-
  {{
    ca_cert_download.content
    if cifmw_openstack_lightspeed_ca_cert_url is defined
    else cifmw_openstack_lightspeed_ca_cert
  }}
```

hooks/playbooks/install-openstack-lightspeed.yml
```diff
@@ -121,8 +121,99 @@
       retries: 30
       delay: 10
 
+    # STEP 3: Create prerequisites (secrets and certificates)
+
+    - name: Create LLM API token secret
```
lpiwowar, line 126: **Blocking:** Before we start chasing reviews, it would be good to see it being executed in a job somewhere. It is going to save us from a lot of trouble later. :) That way, we chase the reviews ideally once.

hooks/playbooks/install-openstack-lightspeed.yml
lpiwowar: Thanks!:)

hooks/playbooks/install-openstack-lightspeed.yml
lpiwowar: Thanks!:)

hooks/playbooks/install-openstack-lightspeed.yml
lpiwowar: Thanks!:)

hooks/playbooks/install-openstack-lightspeed.yml
lpiwowar: Thanks!:)

hooks/playbooks/install-openstack-lightspeed.yml
lpiwowar: @malingatembo, the validation that OpenStack Lightspeed has been deployed successfully must be part of this playbook :see_no_evil:. The` ignore_errors: true` setting was causing the job not to fail if there was an issue during the OpenStack Lightspeed deployment. What we want:

- Deploy OpenStack Lightspeed -> validate that OpenStack Lightspeed has been deployed successfully -> if OpenStack Lightspeed deployment failed -> fail the job.

lpiwowar (changes requested):
Thanks for the updates!:) 🎉 Couple of follow up comments.

hooks/playbooks/install-openstack-lightspeed.yml
malingatembo: thank you, fixed in 7a8915dd39c3adcdc841d8d0bba01665c7a8a4c3

hooks/playbooks/install-openstack-lightspeed.yml
malingatembo: Fair point. I have since tested the playbook on CRC with Downstream resources. I've attached the execution output in JIRA OSPRH-33342. I'll start OSPRH-33343 (Create upstream job that deploys OpenStack Lightspeed) to set up the Zuul CI job that calls this playbook. when this goes successfully, we can call in the  reviews :) 
Thanks

hooks/playbooks/install-openstack-lightspeed.yml
malingatembo: @lpiwowar I have now added the validation to the playbook; fails the job if the operator deployment doesn't succeed. commit 8ac5287d

hooks/playbooks/install-openstack-lightspeed.yml
```diff
@@ -121,8 +121,122 @@
...
+    - name: Assert OpenStackLightspeed deployment successful
+      ansible.builtin.assert:
+        that:
+          - cr_status.resources is defined
+          - cr_status.resources | length > 0
+          - cr_status.resources[0].status is defined
+          - cr_status.resources[0].status.conditions is defined
+          - cr_status.resources[0].status.conditions | selectattr('type', 'equalto', 'Ready') | map(attribute='status') | select('equalto', 'True') | list | length > 0
+        fail_msg: "OpenStackLightspeed CR did not reach Ready state. Current status: {{ cr_status.resources[0].status | default('No status available') }}"
+        success_msg: "OpenStackLightspeed CR is Ready"
```
lpiwowar: **question (non-blocking):** Why do we need the `cr_status` check here again when we have it in the task above as well?

hooks/playbooks/install-openstack-lightspeed.yml
```diff
@@ -121,8 +121,122 @@
...
+        openstack_lightspeed_cr_spec:
+          llmEndpoint: "{{ cifmw_openstack_lightspeed_llm_endpoint }}"
+          llmEndpointType: "{{ cifmw_openstack_lightspeed_llm_endpoint_type | default('openai') }}"
+          modelName: "{{ cifmw_openstack_lightspeed_model_name }}"
+          llmCredentials: "{{ cifmw_openstack_lightspeed_llm_credentials | default('openstack-lightspeed-apitoken') }}"
+
+    - name: Add tlsCACertBundle to CR spec if TLS cert vars provided
+      ansible.builtin.set_fact:
+        openstack_lightspeed_cr_spec: "{{ openstack_lightspeed_cr_spec | combine({'tlsCACertBundle': cifmw_openstack_lightspeed_tls_ca_cert_bundle | default('openstack-lightspeed-certs')}) }}"
+      when: cifmw_openstack_lightspeed_ca_cert is defined or cifmw_openstack_lightspeed_ca_cert_url is defined
```
lpiwowar: **question (non-blocking):** Have you tried to experiment with `omit` keyword? It would simplify things quite a lot if it works.

```
tlsCACertBundle: >-
  {{ cifmw_openstack_lightspeed_tls_ca_cert_bundle | default('openstack-lightspeed-certs')
     if cifmw_openstack_lightspeed_ca_cert is defined or cifmw_openstack_lightspeed_ca_cert_url is defined else omit }}
```

lpiwowar (commented):
Just two comments. But at this point if it works, it works:). Let's focus on getting the job done and we can polish all of these things at the end.

hooks/playbooks/install-openstack-lightspeed.yml
lpiwowar: @malingatembo what about having only single task for the OpenStackLightspeed CR? Like this (please double-check and test first):

```
- name: Create OpenStackLightspeed custom resource
  kubernetes.core.k8s:
    state: present
    definition:
      apiVersion: lightspeed.openstack.org/v1beta1
      kind: OpenStackLightspeed
      metadata:
        name: "{{ cifmw_openstack_lightspeed_cr_name | default('openstacklightspeed-sample') }}"
        namespace: "{{ openstack_lightspeed_namespace }}"
      spec:
        llmEndpoint: "{{ cifmw_openstack_lightspeed_llm_endpoint }}"
        llmEndpointType: "{{ cifmw_openstack_lightspeed_llm_endpoint_type | default('openai') }}"
        modelName: "{{ cifmw_openstack_lightspeed_model_name }}"
        llmCredentials: "{{ cifmw_openstack_lightspeed_llm_credentials | default('openstack-lightspeed-apitoken') }}"
        tlsCACertBundle: "{{ cifmw_openstack_lightspeed_tls_ca_cert_bundle | default('openstack-lightspeed-certs') if (cifmw_openstack_lightspeed_ca_cert is defined or cifmw_openstack_lightspeed_ca_cert_url is defined) else omit }}"
```

lpiwowar:
@malingatembo you probably want to:

- squash the commits
- take a look at [1] and
- request a review from ci-framework folks:)

Once [1] is resolved I'll take a look as well.

[1] https://github.com/openstack-k8s-operators/ci-framework/pull/4108#discussion_r4002927959

lpiwowar:
recheck

```
2026-09-15 10:53:12.611808 | controller |     changed: false
2026-09-15 10:53:12.611817 | controller |     msg: 'Failed to download metadata for repo ''epel-low-priority'': Yum repo downloading
2026-09-15 10:53:12.611826 | controller |       error: Downloading error(s): repodata/1c1b6dab08445ef99da9a54c7364764e2a0d6d49185f2d2b04dd8564135754d3-primary.xml.xz
2026-09-15 10:53:12.611835 | controller |       - Cannot download, all mirrors were already tried without success'
2026-09-15 10:53:12.611843 | controller |     rc: 1
```

lpiwowar (dismissed):
LGTM! 👍 Let's merge if everything is alright and iterate on this:)

lpiwowar (approved):

brjackma (approved):
/lgtm
/approve

PR #4112, opened 2026-08-17
[cifmw_cephadm] add generic ceph-config override passthrough

tosky:
The role applies a fixed set of `ceph config set` commands in tasks/cephadm_config_set.yml, one task per option. A scenario that needs to tune any other cluster option currently has to patch the role. Add a generic, opt-in `cifmw_cephadm_extra_config` list so scenarios can apply arbitrary `ceph config set <who> <key> <value>` entries without a new task per option. It defaults to empty (no-op) and is applied after the built-in configuration, so it can also override earlier values.

For example, a deployment can bound the MDS cap-revoke stall that wedges a single-worker CephFS/NFS-Ganesha consumer by evicting an unresponsive cap-holder before it hits session_autoclose:

    cifmw_cephadm_extra_config:
      - who: mds
        key: mds_cap_revoke_eviction_timeout
        value: "25"

Assisted-By: Claude Opus 4.8 <noreply@anthropic.com>
Signed-off-by: Luigi Toscano <ltoscano@redhat.com>

### DIFF
--- a/roles/cifmw_cephadm/defaults/main.yml
+++ b/roles/cifmw_cephadm/defaults/main.yml
@@ -54,6 +54,18 @@ cifmw_cephadm_certs: /etc/pki/tls
 cifmw_cephadm_debug: false
 cifmw_cephadm_min_compat_client: "mimic"
 cifmw_cephadm_auth_allowed_ciphers: ""
+# Extra `ceph config set` entries, applied after the built-in config so they
+# can also override earlier values. Each item maps to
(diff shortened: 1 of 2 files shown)


fultonj:
+1 to this generic solution. Thanks @tosky

tosky:
Internally tested, ready for review.

fultonj (commented):
/lgtm

fmount (approved):
/lgtm

sdatko (approved):
/lgtm /approve

sdatko (commented):
/approve

PR #4117, opened 2026-08-18
[cifmw_setup][set_containers] add cifmw.general.set_containers module

arxcruz:
Replace the update_containers role invocation in deploy_architecture with the new cifmw.general.set_containers module. The module generates and optionally applies an OpenStackVersion CR, accepting a dynamic images list where each entry can override registry, org, name_prefix, and tag individually. This removes the rigid Jinja2 template in favor of a Python module that is easier to test, extend, and call from any playbook without role-level variable coupling.

Key design points:
- _OPENSTACK_SUFFIXES table enables partial overrides (e.g. tag-only) for any standard OpenStack image field without repeating the suffix.
- Empty name_prefix is supported so images that ship without the "openstack-" prefix (EDPM, IPA) can be built from parts rather than requiring a pre-assembled full_registry URL.
- backends list handles cinderVolumeImages / manilaShareImages nested dicts.
- Unit tests cover _build_url, _resolve_image, _resolve_backend, _build_cr, and the full run_module flow including apply, absent, and validation paths.

### DIFF
--- a/roles/cifmw_setup/tasks/deploy_architecture.yml
+++ b/roles/cifmw_setup/tasks/deploy_architecture.yml
@@ -213,13 +213,55 @@
     - operator
     - edpm_bootstrap
 
-- name: Update containers in deployed OSP operators
-  vars:
-    cifmw_update_containers_metadata: controlplane
-  ansible.builtin.include_role:
-    name: update_containers
+- name: Set containers for deployed OSP operators
+  cifmw.general.set_containers:
+    dest_path: >-
+      {{
+        cifmw_set_containers_dest_path | default(
+          (cifmw_basedir, 'artifacts', 'manifests', 'set_containers.yml') |
+          ansible.builtin.path_join
+        )
+      }}
+    # TODO: remove the default() fallbacks below once all callers in ci-framework-jobs
+    # are migrated from cifmw_update_containers_* to cifmw_set_containers_*.
+    # Scenario-specific variables that also require explicit per-scenario migration:
+    #   cifmw_update_containers_cindervolumes  -> cifmw_set_containers_images (backends)
+    #   cifmw_update_containers_manilashares   -> cifmw_set_containers_images (backends)
+    #   cifmw_update_containers_watcher        -> cifmw_set_containers_extra_images
+    #   cifmw_update_containers_edpm_image_url -> cifmw_set_containers_images (name_prefix: "")
+    #   cifmw_update_containers_ipa_image_url  -> cifmw_set_containers_images (name_prefix: "")
+    registry: "{{ cifmw_set_containers_registry | default(cifmw_update_containers_registry) | default(omit) }}"
+    org: "{{ cifmw_set_containers_org | default(cifmw_update_containers_org) | default(omit) }}"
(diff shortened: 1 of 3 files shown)


Valkyrie00 (dismissed):
/lgtm

Valkyrie00:
Old ref: #3955

danpawlik (dismissed):

roles/cifmw_setup/tasks/deploy_architecture.yml
```diff
@@ -213,13 +213,54 @@
...
+      }}
+    # TODO: remove the default() fallbacks below once all callers in ci-framework-jobs
+    # are migrated from cifmw_update_containers_* to cifmw_set_containers_*.
+    # Scenario-specific variables that also require explicit per-scenario migration:
+    #   cifmw_update_containers_cindervolumes  -> cifmw_set_containers_images (backends)
+    #   cifmw_update_containers_manilashares   -> cifmw_set_containers_images (backends)
+    #   cifmw_update_containers_watcher        -> cifmw_set_containers_extra_images
+    #   cifmw_update_containers_edpm_image_url -> cifmw_set_containers_images (name_prefix: "")
+    #   cifmw_update_containers_ipa_image_url  -> cifmw_set_containers_images (name_prefix: "")
+    registry: "{{ cifmw_set_containers_registry | default(cifmw_update_containers_registry) | default(omit) }}"
```
evallesp: (blocking) suggestion: I think these three vars should be:
registry: >-
  {{ cifmw_set_containers_registry |
    default(cifmw_update_containers_registry | default(cifmw_default_registry)) }}
org: >-
  {{ cifmw_set_containers_org |
     default(cifmw_update_containers_org | default(cifmw_default_container_image_namespace)) }}
tag: >-
  {{ cifmw_set_containers_tag |
     default(cifmw_update_containers_tag | default(cifmw_default_container_image_tag)) }}

I don't think now that the last omit is necessary, the thing is that, after the change of how this is consumed, no anymore like include_role: update_containers, so default vars at: group_vars/all.yml are not defined, which are:

cifmw_update_containers_registry: "{{ cifmw_default_registry }}"
cifmw_update_containers_org: "{{ cifmw_default_container_image_namespace }}"
cifmw_update_containers_tag: "{{ cifmw_default_container_image_tag }}"

roles/cifmw_setup/tasks/deploy_architecture.yml
```diff
@@ -213,13 +213,54 @@
     - operator
     - edpm_bootstrap
 
-- name: Update containers in deployed OSP operators
-  vars:
-    cifmw_update_containers_metadata: controlplane
-  ansible.builtin.include_role:
-    name: update_containers
+- name: Set containers for deployed OSP operators
+  cifmw.general.set_containers:
```
evallesp, line 217: (blocking) suggestion: There's some fallback loginc in the python code, but I think makes sense here adding explicit  value: kubeconfig: "{{ cifmw_openshift_kubeconfig }}"

evallesp (approved):
/lgtm

evallesp (commented):
/approve

Valkyrie00 (approved):

PR #4125, opened 2026-08-20
[bgp] Optionally configure router-0 as a BGP-EVPN VTEP

eduolivares:
The BGP-EVPN dataplane test needs the DC router (router-0) to reach tenant VM IPs advertised via BGP-EVPN Type-5 routes. Until now the router only ran `advertise-all-vni`, so the Type-5 routes propagated through the fabric but router-0 was not a VTEP and could not decapsulate or forward tenant traffic, leaving the private IPs unreachable.

Make router-0 a real EVPN VTEP, gated behind `router_evpn_vtep` (default `false`, so existing plain-BGP and non-EVPN jobs are unchanged):

- `router-frr.conf.j2`: add the L3VNI VRF (`vrf vrf-l3vniN`/`vni N`), `import vrf` into the default instance so ingress traffic on the default VRF can resolve tenant routes, and a second `router bgp <asn> vrf vrf-l3vniN` instance that imports the VTEP IP from the default VRF and advertises only the VTEP `/32` into EVPN. The VTEP address is router-0's loopback IP (`router_loopback_ip`), which is already assigned to `lo`, so there is no separate VTEP-IP knob to configure. The `route-target` matches the EDPM/OVN side (`<edpm_asn>:<vni>`).
- `prepare-bgp-spines-leaves.yaml`: create the VTEP netdevs (`vrf-l3vniN` -> `br-l3vniN` -> `vx-0-N`) before FRR starts so zebra can program the L3VNI. The VXLAN `dstport` is `4789` to match the OVN dataplane on EDPM nodes; the kernel FRR default port is advertisement-only and silently blackholes decapsulated packets.
- Add an idempotent SNAT rule that rewrites tenant-bound traffic to the VTEP IP (router-0's loopback), so the VM replies to an address router-0 advertises into EVPN and no worker subnet needs to be leaked into the tenant VRF. The destination is the whole tempest `project_network_cidr` (`router_tenant_cidr`, default `10.100.0.0/16`) because tempest carves `/28` subnets sequentially out of it -- a single `/24` route flakes once the allocator advances past `10.100.0.255`.

The worker-3 node route and the `bgpnet-worker-3` NAD route (`10.100.0.0/16 -> 100.64.10.1`) live in the `architecture` repo and are handled by a separate change.

Related-ticket: [OSPRH-34785](https://redhat.atlassian.net/browse/OSPRH-34785)

### DIFF
--- a/playbooks/bgp/prepare-bgp-spines-leaves.yaml
+++ b/playbooks/bgp/prepare-bgp-spines-leaves.yaml
@@ -282,6 +282,86 @@
           ip address replace {{ router_loopback_ip }}/32 dev lo
       changed_when: true
 
+    # EVPN VTEP is IPv4-only: the Linux dataplane (VRF/bridge/VXLAN + SNAT) is
+    # built only for IPv4, so fail early instead of rendering FRR EVPN config
+    # with no matching dataplane on an IPv6 run. It also reuses router_loopback_ip
+    # as the VTEP address (already assigned to lo above), so require it to be set.
+    - name: Assert EVPN VTEP prerequisites
+      when: router_evpn_vtep | default(false)
+      ansible.builtin.assert:
+        that:
+          - _ip_version == 4
+          - (router_loopback_ip | default('') | length) > 0
+        fail_msg: >-
+          router_evpn_vtep requires IPv4 (_ip_version == 4, got {{ _ip_version }})
+          and a non-empty router_loopback_ip (got '{{ router_loopback_ip | default('') }}').
+
+    # Make router-0 an EVPN VTEP so it can reach tenant IPs advertised via
+    # BGP-EVPN Type-5 routes. Netdevs are created before FRR starts so zebra can
+    # map "vni N" to the VRF and program the L3VNI. They are not persistent, but
+    # the router is never rebooted (same rationale as the IPv6 route below).
+    # NOTE: dstport must be 4789 to match the OVN dataplane VXLAN port on EDPM
+    # nodes; the kernel FRR device default (8472/60010) is advertisement-only.
+    - name: Configure EVPN VTEP netdevs on router
+      when:
+        - _ip_version == 4
+        - router_evpn_vtep | default(false)
+      become: true
+      vars:
+        _vni: "{{ router_evpn_l3vni | default(1) }}"
+        _table: "{{ 1000 + (router_evpn_l3vni | default(1) | int) }}"
+      block:
+        - name: Create the L3VNI VRF device
+          ansible.builtin.shell:
+            cmd: |
+              set -o errexit
+              ip link show vrf-l3vni{{ _vni }} \
+                || ip link add vrf-l3vni{{ _vni }} type vrf table {{ _table }}
+              ip link set vrf-l3vni{{ _vni }} up
+            executable: /bin/bash
+          changed_when: true
+
+        - name: Create the L3VNI bridge and enslave it to the VRF
+          ansible.builtin.shell:
+            cmd: |
+              set -o errexit
+              ip link show br-l3vni{{ _vni }} \
+                || ip link add br-l3vni{{ _vni }} type bridge
+              ip link set br-l3vni{{ _vni }} master vrf-l3vni{{ _vni }}
+              ip link set br-l3vni{{ _vni }} up
+            executable: /bin/bash
+          changed_when: true
+
+        - name: Create the L3VNI VXLAN device and enslave it to the bridge
+          ansible.builtin.shell:
+            cmd: |
+              set -o errexit
+              ip link show vx-0-{{ _vni }} \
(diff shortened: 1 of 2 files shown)


playbooks/bgp/prepare-bgp-spines-leaves.yaml
```diff
@@ -282,6 +282,86 @@
...
+      block:
+        - name: Create the L3VNI VRF device
+          ansible.builtin.shell:
+            cmd: |
+              set -o errexit
+              ip link show vrf-l3vni{{ _vni }} \
+                || ip link add vrf-l3vni{{ _vni }} type vrf table {{ _table }}
+              ip link set vrf-l3vni{{ _vni }} up
+            executable: /bin/bash
+          changed_when: true
```
evallesp, line 327: (non-blocking) question: Unsure why we want changed_when as false here in L333 L345 and L432
Might be good checking the rc, but as this shouldn't be run twice (AFAIK), I'm markign this as non-blocking.

playbooks/bgp/prepare-bgp-spines-leaves.yaml
```diff
@@ -282,6 +282,86 @@
...
+
+    # Make router-0 an EVPN VTEP so it can reach tenant IPs advertised via
+    # BGP-EVPN Type-5 routes. Netdevs are created before FRR starts so zebra can
+    # map "vni N" to the VRF and program the L3VNI. They are not persistent, but
+    # the router is never rebooted (same rationale as the IPv6 route below).
+    # NOTE: dstport must be 4789 to match the OVN dataplane VXLAN port on EDPM
+    # nodes; the kernel FRR device default (8472/60010) is advertisement-only.
+    - name: Configure EVPN VTEP netdevs on router
+      when:
+        - _ip_version == 4
```
evallesp: (blocking) suggestion: I think it's safe to remove this clause as it's already check at L293

playbooks/bgp/prepare-bgp-spines-leaves.yaml
```diff
@@ -328,6 +408,29 @@
...
+        # SNAT tenant-bound traffic to the loopback/VTEP IP so the return path
+        # works without leaking the worker-3 subnet into the tenant VRF: the VM
+        # sees the VTEP IP as source and replies to it, and router-0 advertises
+        # only that /32 into EVPN. The destination is the whole tempest
+        # project_network_cidr (default 10.100.0.0/16), not a single subnet,
+        # because tempest carves /28 subnets sequentially out of it.
+        # -C || -A keeps the rule idempotent across playbook re-runs.
+        - name: SNAT tenant-bound traffic to the EVPN VTEP IP
+          when:
+            - _ip_version == 4
```
evallesp: (blocking) suggestion: I think it's safe to remove this clause as it's already check at L293

evallesp (commented):
in general lgtm. Some comments set.

playbooks/bgp/prepare-bgp-spines-leaves.yaml
eduolivares: I think checking rc would not help. Each of these tasks is a compound command.
The task's rc is the rc of the last command (`ip link set ... up`), which returns 0 whether or not the device was actually created.
Hence, `changed_when: <rc-based>` wouldn't distinguish create-vs-already-exists — it'd report "changed" on every run, same as the current hardcoded value.

To get genuinely idempotent reporting I'd have to split each task into a show check (`failed_when: false`, `changed_when: false`) + a `when: rc != 0` create — doubling 3 tasks to 6 for a purely cosmetic gain on a router that's provisioned exactly once and never re-run in practice.

I'd prefer to keep it as-is.

@evallesp, what do you think?

evallesp: Indeed!

eduolivares:
recheck

eduolivares:
recheck

eduolivares:
@nemarjan @evallesp 
Could you please review this PR again? I simply resolved one comment from Enrique since last time you reviewed it.

Could you also please override `rdoproject.org/github-check`? The jobs that fail don't test the changes from this PR and they fail because of issues with repos:
`Failed to download metadata for repo ''epel-low-priority''`

evallesp:
> @nemarjan @evallesp Could you please review this PR again? I simply resolved one comment from Enrique since last time you reviewed it.
> 
> Could you also please override `rdoproject.org/github-check`? The jobs that fail don't test the changes from this PR and they fail because of issues with repos: `Failed to download metadata for repo ''epel-low-priority''`

Trigger the ci job again. Let's see this time if it gets better as I've checked in the zuul job build history now they're green. If not I'd say it's safe to override them.

eduolivares:
recheck

evallesp (commented):
/lgtm
/approve

evallesp (approved):

PR #4127, opened 2026-08-20
[update] Fix floating IP leak in DPDK workload cleanup

mnietoji:
When `workload_dpdk` is true, the `sanity_teardown` function only deletes the DPDK port but does not clean up the floating IP created by `set_vm_ip`. The VM is deleted (which disassociates the FIP) but the floating IP remains allocated, leaking one IP per update run.

Over multiple runs this exhausts the access network's IP pool, causing subsequent tempest tests to fail with
`IpAddressGenerationFailure: No more IP addresses available`.

Add floating IP cleanup to the DPDK branch of `sanity_teardown`, matching the existing cleanup logic in the default (non-SRIOV, non-DPDK) branch.


evallesp (approved):
/lgtm

nemarjan (approved):
/approved

PR #4130, opened 2026-08-20
[cifmw_cephadm] Allow legacy CephX while daemons use aes256k

rebtoor:
Ceph 20.2.4 (CVE-2025-30156) can fail the CephX handshake for OpenStack Glance, Cinder, and Manila. Those Antelope service images still ship librados that only understands key type `aes`, while Tentacle daemons may use `aes256k`.

This is a fresh cephadm cluster every CI run, so the CVE upgrade path is:

1. Keep `auth_allowed_ciphers=aes,aes256k` so daemons can use the new type and legacy `aes` clients can still authenticate.
2. Do **not** set `auth_preferred_cipher`. New keys would otherwise default to `aes256k` before OpenStack images understand that type.
3. Mint `client.openstack` as type `aes` until Glance/Cinder/Manila images ship librados that understands `aes256k`.
4. Keep the RBD vs CephFS cap split (`profile rbd` on RBD pools, `allow rw` on CephFS pools, plus `mon: allow r`). That is a separate Manila/RBD fix, not the errno 5 connect path.

HCI scenario `hci_ceph_backends.yml` sets the allow-list and `cifmw_ceph_key_cipher: aes`. The `cifmw_cephadm` role does not set `auth_preferred_cipher`.

### DIFF
--- a/hooks/playbooks/ceph.yml
+++ b/hooks/playbooks/ceph.yml
@@ -372,6 +372,10 @@
         all_addresses: ansible_all_ipv6_addresses
         cidr: 64
 
+    # OpenStack service images still ship librados that only understands
+    # CephX type aes. CVE-2025-30156: keep daemons on aes256k via
+    # auth_allowed_ciphers, but mint client.openstack as aes until those
+    # images are upgraded. Do not set auth_preferred_cipher.
     - name: Generate a cephx key
       cephx_key:
         cipher: "{{ cifmw_ceph_key_cipher | default('aes') }}"
@@ -386,11 +390,36 @@
             mode: '0600'
             caps:
               mgr: allow *
-              mon: profile rbd
-              osd: "{{ pools | map('regex_replace', '^(.*)$',
-                                   'profile rbd pool=\\1') | join(', ') }}"
+              mon: allow r, profile rbd
+              osd: "{{ _cifmw_cephadm_osd_caps }}"
       vars:
-        pools: "{{ cifmw_cephadm_pools | map(attribute='name') | list }}"
+        _cifmw_cephadm_rbd_pools: >-
+          {{
+            cifmw_cephadm_pools
+            | selectattr('application', 'equalto', 'rbd')
+            | map(attribute='name')
+            | list
+          }}
+        _cifmw_cephadm_cephfs_pools: >-
+          {{
+            cifmw_cephadm_pools
+            | selectattr('application', 'equalto', 'cephfs')
+            | map(attribute='name')
+            | list
+          }}
+        _cifmw_cephadm_osd_caps: >-
+          {{
+            (
+              _cifmw_cephadm_rbd_pools
+              | map('regex_replace', '^(.*)$', 'profile rbd pool=\\1')
+              | list
+              + (
+                _cifmw_cephadm_cephfs_pools
+                | map('regex_replace', '^(.*)$', 'allow rw pool=\\1')
+                | list
+              )
+            ) | join(', ')
+          }}
       no_log: "{{ cifmw_nolog | default(true) | bool }}"
 
     # for deploying external ceph for 17.1 using cifmw, we need this playbook to create keyring
--- a/roles/cifmw_cephadm/README.md
+++ b/roles/cifmw_cephadm/README.md
@@ -136,6 +136,16 @@ that they do not need to be changed for a typical EDPM deployment.
   `ceph mon set auth_allowed_ciphers <value>` during cluster configuration.
    Example values are `"aes,aes256k"` or `"aes256k"` or `"aes"`.
    Defaults to `""` (unset, no command is run).
+   For Tentacle (v20) HCI with older OpenStack librados, set
+   `"aes,aes256k"` so daemons can use `aes256k` while `client.openstack`
+   keys of type `aes` still authenticate (CVE-2025-30156). Do not set
+   `auth_preferred_cipher` until Glance/Cinder/Manila images understand
+   `aes256k`.
+
+* `cifmw_ceph_key_cipher`: (String) Cipher passed to the `cephx_key` module
+  when generating `client.openstack`. Defaults to `aes`. Keep `aes` for
+  OpenStack clients until service images ship librados that understands
+  `aes256k`.
 
 Use the `cifmw_cephadm_pools` list of dictionaries to define pools for
 Nova (vms), Cinder (volumes), Cinder-backups (backups), and Glance (images).
@@ -169,7 +179,7 @@ cifmw_cephadm_keys:
     mode: '0600'
     caps:
       mgr: allow *
-      mon: profile rbd
+      mon: allow r, profile rbd
       osd: profile rbd pool=vms, profile rbd pool=volumes, profile rbd pool=backups, profile rbd pool=images
 ```
 
--- a/scenarios/centos-9/hci_ceph_backends.yml
+++ b/scenarios/centos-9/hci_ceph_backends.yml
@@ -11,6 +11,12 @@ cifmw_cephadm_version: "tentacle"
 cifmw_cephadm_prepare_host: true
 # Apply the new spec version for RGW/TLS
 cifmw_rgw_ssl_backward_compatibility: false
+# CVE-2025-30156: Tentacle daemons may use aes256k, but Antelope
+# Glance/Cinder/Manila librados only understands aes. Allow both
+# types and mint client.openstack as aes. Do not set
+# auth_preferred_cipher until those service images are upgraded.
+cifmw_ceph_key_cipher: aes
+cifmw_cephadm_auth_allowed_ciphers: "aes,aes256k"
 
 cifmw_install_yamls_vars:
   BMO_SETUP: false


hooks/playbooks/ceph.yml
```diff
@@ -386,11 +396,36 @@
...
       vars:
-        pools: "{{ cifmw_cephadm_pools | map(attribute='name') | list }}"
+        _cifmw_cephadm_rbd_pools: >-
+          {{
+            cifmw_cephadm_pools
+            | selectattr('application', 'equalto', 'rbd')
+            | map(attribute='name')
+            | list
+          }}
+        _cifmw_cephadm_cephfs_pools: >-
```
fmount, line 403: ok this worked in the past and probably changed with the new release.  To give more context, this applied `profile rbd pool=X` to every pool, including the `CephFS` pools (`cephfs.cephfs.meta` and `cephfs.cephfs.data`) and while working, it was wrong.
I assume that on older Ceph releases this happened to work because cap validation was more flexible, but Ceph Tentacle (v20) enforces stricter cephx cap validation, causing RADOS permission denied errors when `Manila` try to access CephFS-backed pools.
Is that the case? @fultonj to double check.

hooks/playbooks/ceph.yml
```diff
@@ -372,9 +372,19 @@
         all_addresses: ansible_all_ipv6_addresses
         cidr: 64
 
+    - name: Select cephx cipher default for Ceph release
+      ansible.builtin.set_fact:
+        _cifmw_ceph_key_cipher_default: >-
+          {{
+            'aes256k' if (
+              cifmw_cephadm_version | default('') == 'tentacle'
+              or (cifmw_cephadm_container_tag | default('v18') is match('^v(20|[2-9][0-9])'))
```
fmount: should we swiftch the default here to `v20` at this point for the main branch? 19 will be based on v20+ and I'm not sure we currently have upstream jobs based on previous releases.
We might need to patch `github.com/openstack-k8s-operators/ci-framework/blob/1256a1cd4c0d52707f14c975f510cca324dabbe2/zuul.d/edpm_multinode.yaml#L347` and make sure it is reconciled we new defaults, but it might allow to simplify the logic for the main branch and start testing from `v20` onward.

roles/cifmw_cephadm/tasks/cephadm_config_set.yml
```diff
@@ -62,6 +62,16 @@
     {{ cifmw_cephadm_auth_allowed_ciphers }}
   changed_when: false
 
+- name: Set Ceph auth_preferred_cipher
+  become: true
+  when:
+    - cifmw_cephadm_auth_preferred_cipher is defined
+    - cifmw_cephadm_auth_preferred_cipher | length > 0
+  ansible.builtin.command: |
+    {{ cifmw_cephadm_ceph_cli }} mon set auth_preferred_cipher \
```
fmount: I wasn't aware about `auth_preferred_cipher` and I thought `auth_allowed_ciphers` [1] was sufficient to give to Ceph a list of allowed ciphers to accept. 
What's the difference between these two commands? @fultonj might know this already.

[1] https://github.com/openstack-k8s-operators/ci-framework/commit/c7a2c0e47275581991ed5cb1fbeabcea852c3e79

hooks/playbooks/ceph.yml
```diff
@@ -372,9 +372,19 @@
...
+            ) else 'aes'
+          }}
+
     - name: Generate a cephx key
       cephx_key:
-        cipher: "{{ cifmw_ceph_key_cipher | default('aes') }}"
+        cipher: "{{ cifmw_ceph_key_cipher | default(_cifmw_ceph_key_cipher_default) }}"
```
fmount: Ok I recall we added support to customize ciphers via [1] and I'm wondering if we should simply update our defaults based on `v20` instead of guessing the `ceph tag` (that can't be misleading unless you run `podman run --rm <container> ceph versions` and `grep` for the release).
So the real question is: should we update the defaults to tentacle (v20) by default and update the cipher to `aes256k` ? v20:aes256.
I assume that in the next future (weeks?) the assumption provided by this default will no longer be valid and we need to modify this logic again to meet new backports to the ceph clients.

[1] https://github.com/openstack-k8s-operators/ci-framework/commit/03fdc8f3fdcfb279244745076f95825b604f4216

fmount (commented):
We might need to iterate over this patch due to the recent ceph release, but I have a few questions:
1. should we default to v20:aes256k?
2. Instead of guessing the release via regex, do we want to keep a mapping of the current status (e.g. {v20:aes256k, v19:aes, etc ...} ) to make easy follow up based on the Ceph backports?
3. is auth_preferred_cipher something we really need to enforce if we create the client.openstack key with the right cipher that we map to a struct or we guess via regex?
4. should we update jobs like: `https://github.com/openstack-k8s-operators/ci-framework/blob/main/zuul.d/edpm_multinode.yaml#L348` to finally switch to v20:aes256k as main is our 19 base (and 19 will be based on v20+).

I'd like also @fultonj to review the above and we can follow up in case CI is blocked and we need to land this patch asap.

fultonj:
This patch changes the generated cephx key like this:

```diff
  cifmw_cephadm_keys:
    - name: client.openstack
      caps:
        mgr: allow *
-       mon: profile rbd
+       mon: allow r, profile rbd
-       osd: profile rbd pool=vms, profile rbd pool=volumes, profile rbd pool=backups, profile rbd pool=images, profile rbd pool=manila_data, profile rbd pool=manila_metadata
+       osd: profile rbd pool=vms, profile rbd pool=volumes, profile rbd pool=backups, profile rbd pool=images, allow rw pool=manila_data, allow rw pool=manila_metadata
```
So:
1. mon: keeps `profile rbd` but also gets `allow r`

This is not necessary since `profile rbd` includes `allow r` https://github.com/ceph/ceph/blob/main/src/mon/MonCap.cc#L348

2. osd: changes from `profile rbd` to `allow rw` for manila pools

`profile rbd` effectively gave us RWX and this would change it to just RW https://github.com/ceph/ceph/blob/main/src/osd/OSDCap.cc#L354

Before we make this change, can I know which CI Job failed and how?

I only want to CI the permissions we document and recommend for users. If we're going to change it, then that's something people are going to have to change on their systems so we need to be deliberate and know exactly why.

I can see the argument that X is not necessary, but this patch is about making an update to deal with aes256k and I don't yet see the connection to the new cipher and the caps change.

Let's split out the aes256k adjustments for our jobs with the caps change. I'll propose the change to fix the jobs in a separate PR.

fultonj:
@rebtoor 

Can we split this into two PRs? 

One to fix broken CI jobs and another to explore changing the caps? 

Here's what I'm proposing for the first: https://github.com/openstack-k8s-operators/ci-framework/pull/4144

If the caps change and CI breakages are actually intertwined, then please share the logs so I can debug. 

Thanks

fmount (approved):
We need to investigate more and see if this is solved by #4144. I think @fultonj is going to review this and see if we can land it.

Valkyrie00 (approved):
Okay, I've approved #4144 it's about to be merged. I'm waiting for it to be reviewed, anyway this one in general /lgtm

fultonj:
> We might need to iterate over this patch due to the recent ceph release, but I have a few questions:
> 
> 1. should we default to v20:aes256k?

https://github.com/openstack-k8s-operators/ci-framework/pull/4144 has merged so jobs which use scenarios/centos-9/hci_ceph_backends.yml will now get AES not AES256k. However, anyone can override these values in their job.

For example uni04delta does not use `scenarios/centos-9/hci_ceph_backends.yml` and sets `cifmw_ceph_key_cipher: aes256k`. uni04delta uses new enough ceph clients so it works wit that new cipher.

OpenStack servers need Ceph clients 20.2.4 or 19.2.6 or newer in order to use AES256k. I expect all containers will eventually get updated and then we can revert  https://github.com/openstack-k8s-operators/ci-framework/pull/4144 

> 2. Instead of guessing the release via regex, do we want to keep a mapping of the current status (e.g. {v20:aes256k, v19:aes, etc ...} ) to make easy follow up based on the Ceph backports?

Yes.

> 3. is auth_preferred_cipher something we really need to enforce if we create the client.openstack key with the right cipher that we map to a struct or we guess via regex?

Setting `auth_preferred_cipher` is not necessary to make a CI job pass with the new cipher. It's effectively the default cipher for new keyrings if you create keys without passing the cipher flag or without passing the key itself. We should be able to get by without an override for it.

> 4. should we update jobs like: `https://github.com/openstack-k8s-operators/ci-framework/blob/main/zuul.d/edpm_multinode.yaml#L348` to finally switch to v20:aes256k as main is our 19 base (and 19 will be based on v20+).

If we're talking about the basis of 19, then yes, it should use aes256k. 

For jobs which use the following file, are they using Ceph clients 20.2.4 or 19.2.6 or newer? Or is it a mix? If it's a mix and it's going to stay a mix, then we need to fork out our scenarios. That additional complexity shouldn't be needed though. Better to just revert https://github.com/openstack-k8s-operators/ci-framework/pull/4144 when containers are built with new ceph libraries and use aes256k in all RHOSO based CI. 

https://github.com/openstack-k8s-operators/ci-framework/blob/main/scenarios/centos-9/multinode-ci.yml

> I'd like also @fultonj to review the above and we can follow up in case CI is blocked and we need to land this patch asap.

I think the ASAP issue has been addressed by this patch. If I'm wrong please point me at the failing job.

https://github.com/openstack-k8s-operators/ci-framework/pull/4144

Now that the above has merged this patch is should now be about changing the caps in our standard openstack keyring but I only want to change caps if it's really necessary. I have not yet seen proof that it's needed and the caps are essentially the same in the end as per my last comment.

fultonj (changes requested):
This patch changed the caps and overrode the cipher params in `scenarios/centos-9/hci_ceph_backends.yml`. The cipher params matter has now been merged in https://github.com/openstack-k8s-operators/ci-framework/pull/4144. 

If we still need to override the caps, then please show me which job requires that change and we'll refine this further.

PR #4131, opened 2026-08-21
[edpm_prepare] Apply set_containers overrides from job vars

rebtoor:
## Summary

- Add one small task to `edpm_prepare` that calls `cifmw.general.set_containers` when jobs provide `cifmw_set_containers_images`.
- Leave the legacy `update_containers` role untouched.
- Add `preserve_unlisted` to `set_containers` so partial override lists keep unspecified images from the current OpenStackVersion CR before apply.

### Problem

Consumer jobs (for example the s2i content provider) return a *partial*
`customContainerImages` map: only the services that were built, not the full
OpenStack image set. Those overrides must be applied during `edpm_prepare`,
before `OpenStackControlPlane` is deployed. Applying them later (for example
via a `pre_tests` hook) is too late — the control plane and dataplane may
already be running with operator-default images.

### Approach

- Jobs pass `cifmw_set_containers_images` as a list of `{name, full_registry}` entries.
- `edpm_prepare` runs one generic `set_containers` task when that list is non-empty.
- `preserve_unlisted` reads the current OpenStackVersion CR, keeps entries not
  present in the override list, overlays the new entries, writes the manifest,
  and applies it via the module's existing `oc apply -f` path.
- No s2i-specific logic lives in ci-framework; job YAML owns the translation
  from provider output to `cifmw_set_containers_images`.

### Out of scope (follow-up)

- Migrating the legacy `update_containers` role to `set_containers` for all
  EDPM and DLRN flows. The architecture deploy path already uses
  `set_containers`; this change adds a narrow override hook for EDPM jobs that
  supply explicit image lists.
- Replacing `update_containers` wholesale in this PR would duplicate the large
  parameter surface already present in `deploy_architecture.yml` and expand
  review scope beyond what consumer jobs need today.

### Alternatives considered

- **`pre_tests` hook + `oc patch`:** wrong point in the deploy timeline.
- **Inlining a full `set_containers` call in `edpm_prepare`:** rejected; job
  vars plus a thin task keeps the framework generic.
- **Migrating `update_containers` to `set_containers` in this PR:** larger
  change, separate from the partial-override use case.
- **Applying partial manifests without `preserve_unlisted`:** would replace the
  entire `customContainerImages` map and drop unspecified services.

## Test plan

- [ ] `make ansible_test` (set_containers unit tests)
- [ ] Cross-repo Zuul check with `Depends-On` from s2i-openstack-containers PR #91

Depends-on: https://github.com/openstack-k8s-operators/ci-framework/pull/4136

### DIFF
--- a/plugins/modules/set_containers.py
+++ b/plugins/modules/set_containers.py
@@ -56,6 +56,16 @@ options:
     - When C(true) and C(state=present), run C(oc apply -f dest_path).
     type: bool
     default: false
+  preserve_unlisted:
+    description:
+    - When C(true), read the current OpenStackVersion CR from the cluster
+      and keep C(spec.customContainerImages) entries that are not overridden
+      before writing/applying the manifest.
+    - Requires C(kubeconfig) when the cluster lookup is needed.
+    - A missing OpenStackVersion is treated as an empty map. Any other
+      C(oc get) failure is an error.
+    type: bool
+    default: false
   namespace:
     description:
     - Kubernetes namespace for the OpenStackVersion CR.
@@ -250,6 +260,7 @@ changed:
     returned: always
 """
 
+import json
 import os
 
 from ansible.module_utils.basic import AnsibleModule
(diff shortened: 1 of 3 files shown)


rebtoor:
recheck

Valkyrie00 (dismissed):
/lgtm

rebtoor:
Addressed the Qodo comments that still applied after the `preserve_unlisted` rewrite:

- Fetch existing images with `oc get -o json` and extract `spec.customContainerImages` instead of jsonpath + `json.loads`.
- `oc get` failures are now hard errors, except NotFound (treated as an empty map).
- Unit tests now assert the merged map, NotFound fallback, and non-NotFound get failure.

Left as-is: check-mode still does a read-only `oc get` so dry-run can report an accurate `changed`. The older comments about `_edpm_s2i_set_containers_images`, `patch_merge`, and apply-with-empty-images no longer apply to this patch.

evallesp (dismissed):
/approve

Valkyrie00 (dismissed):

rebtoor:
recheck

arxcruz:
recheck

Valkyrie00 (approved):

nemarjan (approved):

rebtoor:
recheck

rebtoor:
recheck

rebtoor:
recheck

PR #4140, opened 2026-08-25
Add resources required to run lightspeed-tests

jancervenka:
test-operator expects a ConfigMap and two secrets to exists before it will start executing a test suite. We create the expected dummy resources with no data to satisfy the requirement.

test-operator is using the default ServiceAccount to run the tests. We need to create a ClusterRole and ClusterRoleBinding to grant the default SA access to the Lightspeed service.

### DIFF
--- a/hooks/playbooks/install-openstack-lightspeed.yml
+++ b/hooks/playbooks/install-openstack-lightspeed.yml
@@ -135,6 +135,7 @@
           stringData:
             apitoken: "{{ cifmw_openstack_lightspeed_api_token }}"
       when: cifmw_openstack_lightspeed_api_token is defined
+      no_log: true
 
     - name: Verify LLM credentials secret exists
       kubernetes.core.k8s_info:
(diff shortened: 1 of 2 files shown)


jancervenka:
Hi @malingatembo! we were discussing with @lpiwowar (see: https://github.com/openstack-k8s-operators/lightspeed-tests/pull/12) create some additional resources in your playbook. The resources are required by the test-operator to successfully run the lightspeed-test suite.

See the commit message:

> test-operator expects a ConfigMap and two secrets to exists before it will start executing a test suite. We create the expected dummy resources with no data to satisfy the requirement.
> 
> test-operator is using the default ServiceAccount to run the tests. We need to create a ClusterRole and ClusterRoleBinding to grant the default SA access to the Lightspeed service.

lpiwowar:
@jancervenka thanks for the PR! :tada: The initial idea was for this PR to build on top of @malingatembo's PR. We can then use `Depends-On:` and test your changes alongside the changes from @malingatembo's PR -> #4108 (once the job Malinga Tembo is working on is ready -> openstack-k8s-operators/lightspeed-operator#65)

If it turns out that the changes you introduced work, then we can merge both PRs into one (into @malingatembo's PR #4108 with a proper commit attributing your work, of course :).

I suggest:
- Marking this as a draft and building this PR on top of the one from @malingatembo. I wonder whether it is possible to use the stacked PRs GitHub recently introduced for this (I have not experimented with them yet). But what you can definitely do is check out the branch from @malingatembo's PR and check out a new branch out of it with your changes.


Just want to say that we are getting a little bit ahead ourselves here. But if everything clicks then the PR merging is going to be easy. Also, I'm personally ok just waiting it out until the job @malingatembo is working on gets ready. One way or the other I would mark this as a draft for now :see_no_evil: .

jancervenka:
the PR is checked out of @malingatembo's work and marked as a draft.

brjackma:
@jancervenka Would you mind amending the commit title with the conventional prefix. I would suggest: 

[lightspeed] Add resources required to run lightspeed-tests

Additional Context: We usually prefix the commit title with the name of the role/ hook being changed in the ci-framework

hooks/playbooks/install-openstack-lightspeed.yml
```diff
@@ -76,6 +76,35 @@
           metadata:
             name: "{{ openstack_lightspeed_namespace }}"
 
+    - name: Create ClusterRole for Lightspeed tests
+      kubernetes.core.k8s:
```
brjackma, line 80: I see the five new kubernetes.core.k8s tasks don't have retries/delay/until. I know the existing create-tasks in this file don't have them either, but best practices say any k8s API call should be retried ijn case CI networks are flaky.

What do you think about adding something like register: _result / retries: 5 / delay: 10 / until: _result is not failed to these tasks? It'd make the playbook more resilient

PR #4141, opened 2026-08-25
[disconnected] Add option to perform image signature verification

drosenfe:
Add option to the existing configure openshift cluster for disconnected deployment hook to perform image signature verification. This may be used when oc mirror v2 has mirrored both container images and their cryptographic signatures.

jira: https://redhat.atlassian.net/browse/OSPRH-35167

Signed-off-by: David Rosenfeld drosenfe@redhat.com


drosenfe:
recheck

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
```diff
@@ -8,10 +8,12 @@
   vars:
     oc_mirror_download_url: "{{ cifmw_disconnected_mirror_url | default('https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/latest/oc-mirror.rhel9.tar.gz') }}"
     mirror_registry_url: "{{ cifmw_disconnected_registry_url | default('https://mirror.openshift.com/pub/cgw/mirror-registry/latest/mirror-registry-amd64.tar.gz') }}"
+    generate_cluster_image_pol_url: "{{ cifmw_generate_cluster_image_pol_url | default('https://raw.githubusercontent.com/momoah/snolibvirt/refs/heads/main/files/generate_cluster_image_policies.py') }}"
```
evallesp: (blocking) question: Unsure if this is a typo, could we use a non-personal fork?

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
```diff
@@ -295,6 +297,42 @@
...
+
+        - name: Update output file in generate cluster image policy script
+          ansible.builtin.replace:
+            path: "{{ disconnect_working_dir }}/generate_cluster_image_policies.py"
+            regexp: ClusterImagePolicies.yaml
+            replace: '{{ disconnect_working_dir }}/ClusterImagePolicies.yaml'
+
+        - name: Execute generate cluster image policy script
+          ansible.builtin.command:
+            cmd: python {{ disconnect_working_dir }}/generate_cluster_image_policies.py
```
evallesp: (non-blocking) question: should this be python3?

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
```diff
@@ -295,6 +297,42 @@
...
+
+        - name: Apply yaml files from results directory to cluster
```
evallesp: (non-blocking) suggestion: WDYT about changing this task name as it's the same as L286

evallesp (commented):
lgtm in general.

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
rabi: Yeah, we should not depend on any arbitray uncontrolled script. At best we can create a module like plugins/modules/generate_cluster_image_policies.py with the code and parameterize some of the hardcoded stuff. Module can be called like

 ```yaml                                                                                                                                                                                                            
   - name: Generate ClusterImagePolicies from mirror data                                                                                                                                                           
     cifmw.general.generate_cluster_image_policies:                                                                                                                                                                 
       input_dir: "{{ mirror_location }}/working-dir/cluster-resources"                                                                                                                                             
       output_file: "{{ disconnect_working_dir }}/ClusterImagePolicies.yaml"                                                                                                                                        
     when: verify_image_signatures                                                                                                                                                                                  
                                                                                                                                                                                                                    
   - name: Apply ClusterImagePolicies                                                                                                                                                                               
     kubernetes.core.k8s:                                                                                                                                                                                           
       state: present                                                                                                                                                                                               
       src: "{{ disconnect_working_dir }}/ClusterImagePolicies.yaml"                                                                                                                                                
     when: verify_image_signatures                                                                                                                                                                                  
 ```

drosenfe: Rewritten as suggested by adding a new module and not using non-personal fork.

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
drosenfe: python statement totally removed from latest version.

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
drosenfe: Task name has been changed.

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
```diff
@@ -295,6 +296,18 @@
       retries: 10
       delay: 30
 
+    - name: Generate ClusterImagePolicies from mirror data
+      cifmw.general.generate_cluster_image_policies:
```
evallesp, line 300: praise: Great! Thanks !

plugins/modules/generate_cluster_image_policies.py
```diff
@@ -0,0 +1,139 @@
...
+    sigstore_key = """LS0tLS1CRUdJTiBQVUJMSUMgS0VZLS0tLS0KTUlJQ0lqQU5CZ2txaGtpRzl3MEJBUUVGQUFPQ0FnOEFNSUlDQ2dLQ0FnRUEwQVN5dUgyVExXdkJVcVBIWjRJcAo3NWc3RW5jQmtnUUhkSm5qenhBVzVLUVRNaC9zaUJvQi9Cb1NydGlQTXduQ2hiVENuUU9JUWVadURpRm5odUo3Ck0vRDNiN0pvWDBtMTIzTmNDU242N21BZGpCYTZCZzZrdWtaZ0NQNFpVWmVFU2FqV1gvRWp5bEZjUkZPWFc1N3AKUkRDRU40MkovallsVnF0K2c5K0dya2VyOFN6ODZIM2wwdGJxT2RqYnovVnhIWWh3RjBjdFVNSHN5VlJEcTJRUAp0cXpOWGxtbE1oUy9Qb0ZyNlI0dS83SENuL0srTGVnY08yZkFGT2I0MEt2S1NLS1ZENmxld1VaRXJob3AxQ2dKClhqRHRHbW1POWRHTUY3MW1mNkhFZmFLU2R5K0VFNmlTRjJBMlZ2OVFoQmF3TWlxMmtPekVpTGc0bkFkSlQ4d2cKWnJNQW1QQ3FHSXNYTkdaNC9RK1lUd3dsY2UzZ2xxYjVMOXRmTm96RWRTUjlOODVERVNmUUxRRWRZM0NhbHdLTQpCVDFPRWhFWDF3SFJDVTRkck1PZWo2Qk5XMFZ0c2NHdEhtQ3JzNzRqUGV6aHdOVDh5cGt5UytUMHpUNFRzeTZmClZYa0o4WVNIeWVuU3pNQjJPcDJidnNFM2dyWStzNzRXaEc5VUlBNkRCeGNUaWUxNU5Tekt3Znphb05XT0RjTEYKcDdCWThhYUhFMk1xRnhZRlgrSWJqcGtRUmZhZVFRc291REZkQ2tYRUZWZlBwYkQyZGs2RmxlYU1UUHV5eHRJVApnalZFdEdRSzJxR0NGR2lRSEZkNGhmVitlQ0E2M0pybzF6MHpvQk01QmJJSVEzK2VWRnd0M0FsWnA1VVZ3cjZkCnNlY3FraS95cm12M1kwZHFaOVZPbjNVQ0F3RUFBUT09Ci0tLS0tRU5EIFBVQkxJQyBLRVktLS0tLQ=="""
```
evallesp, line 70: (non-blocking) suggestion: Let's move this below NO_POLICIES in upper case as a constant.

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
```diff
@@ -12,6 +12,7 @@
     mirror_location: "{{ disconnect_working_dir }}/mirror_location"
     local_registry: "{{ disconnect_working_dir }}/local_registry"
     oc_mirror_cert_manager_catalog_url: "{{ cifmw_cert_manager_catalog_url | default('registry.redhat.io/redhat/redhat-operator-index:v4.18') }}"
+    verify_image_signatures: "{{ cifmw_disconnected_verify_image_signatures | default(false) }}"
```
rabi, line 15: Image signature verification is true by default downstream right? Otherwise one has to use insecureAcceptAnything: true in /etc/containers/policy.json for default/registry.

rabi (commented):
A question on the default otherwise good for me.

plugins/modules/generate_cluster_image_policies.py
```diff
@@ -0,0 +1,139 @@
...
+
+    # Deduplicate
+    unique_pairs = sorted(set(pairs))
+
+    # Generate policy YAMLs
+    policies = []
+    for mirror, source in unique_pairs:
+        name = mirror.split("/")[-1].replace(".", "-").replace(":", "-")
+        policy = {
+            "apiVersion": "config.openshift.io/v1",
```
evallesp, line 100: (non-blocking) question: DO we need to check openshift deployed version? I think in > 4.18 this is correct. If not this is not applied and we're silently not running this.

We can comment this that requires specific ocp version to run, or retrieving the info by "oc explain clusterimagepolicy --recursive | head -1" to parametrice this line, or maybe just not running the python code at all.

plugins/modules/generate_cluster_image_policies.py
```diff
@@ -0,0 +1,139 @@
...
+        }
+        policies.append(policy)
+
+    # Output to file
+    with open(output_file, "w") as f:
+        f.write(
+            "---\n" + "\n---\n".join(yaml.dump(p, sort_keys=False) for p in policies)
+        )
+
+    print(f"Generated {len(policies)} ClusterImagePolicy objects in {output_file}")
```
evallesp, line 127: (blocking) let's move this to module.log or put in a result field. This might bring runtime error of not able to parse JSON object.

plugins/modules/generate_cluster_image_policies.py
```diff
@@ -0,0 +1,139 @@
...
+author:
+  - David Rosenfeld (@drosenfe)
+"""
+
+EXAMPLES = r"""
+- name: Generate ClusterImagePolicies from mirror data
+  cifmw.general.generate_cluster_image_policies:
+    input_dir: "{{ mirror_location }}/working-dir/cluster-resources"
+    output_file: "{{ disconnect_working_dir }}/ClusterImagePolicies.yaml"
+"""
```
evallesp, line 40: (non-blocking) suggestion: Let's add the RETURN explanation.

plugins/modules/generate_cluster_image_policies.py
```diff
@@ -0,0 +1,139 @@
...
+    sigstore_key = """LS0tLS1CRUdJTiBQVUJMSUMgS0VZLS0tLS0KTUlJQ0lqQU5CZ2txaGtpRzl3MEJBUUVGQUFPQ0FnOEFNSUlDQ2dLQ0FnRUEwQVN5dUgyVExXdkJVcVBIWjRJcAo3NWc3RW5jQmtnUUhkSm5qenhBVzVLUVRNaC9zaUJvQi9Cb1NydGlQTXduQ2hiVENuUU9JUWVadURpRm5odUo3Ck0vRDNiN0pvWDBtMTIzTmNDU242N21BZGpCYTZCZzZrdWtaZ0NQNFpVWmVFU2FqV1gvRWp5bEZjUkZPWFc1N3AKUkRDRU40MkovallsVnF0K2c5K0dya2VyOFN6ODZIM2wwdGJxT2RqYnovVnhIWWh3RjBjdFVNSHN5VlJEcTJRUAp0cXpOWGxtbE1oUy9Qb0ZyNlI0dS83SENuL0srTGVnY08yZkFGT2I0MEt2S1NLS1ZENmxld1VaRXJob3AxQ2dKClhqRHRHbW1POWRHTUY3MW1mNkhFZmFLU2R5K0VFNmlTRjJBMlZ2OVFoQmF3TWlxMmtPekVpTGc0bkFkSlQ4d2cKWnJNQW1QQ3FHSXNYTkdaNC9RK1lUd3dsY2UzZ2xxYjVMOXRmTm96RWRTUjlOODVERVNmUUxRRWRZM0NhbHdLTQpCVDFPRWhFWDF3SFJDVTRkck1PZWo2Qk5XMFZ0c2NHdEhtQ3JzNzRqUGV6aHdOVDh5cGt5UytUMHpUNFRzeTZmClZYa0o4WVNIeWVuU3pNQjJPcDJidnNFM2dyWStzNzRXaEc5VUlBNkRCeGNUaWUxNU5Tekt3Znphb05XT0RjTEYKcDdCWThhYUhFMk1xRnhZRlgrSWJqcGtRUmZhZVFRc291REZkQ2tYRUZWZlBwYkQyZGs2RmxlYU1UUHV5eHRJVApnalZFdEdRSzJxR0NGR2lRSEZkNGhmVitlQ0E2M0pybzF6MHpvQk01QmJJSVEzK2VWRnd0M0FsWnA1VVZ3cjZkCnNlY3FraS95cm12M1kwZHFaOVZPbjNVQ0F3RUFBUT09Ci0tLS0tRU5EIFBVQkxJQyBLRVktLS0tLQ=="""
+
+    # Collect mirror-source pairs
+    pairs = []
+
+    for fname in os.listdir(input_dir):
```
evallesp, line 75: (blocking) suggestion: let's check this exists first so we can fail_json in case of not existing.

plugins/modules/generate_cluster_image_policies.py
```diff
@@ -0,0 +1,139 @@
...
+                for doc in docs:
+                    if not doc or "spec" not in doc:
+                        continue
+                    key = (
+                        "imageDigestMirrors"
+                        if "imageDigestMirrors" in doc["spec"]
+                        else "imageTagMirrors"
+                    )
+                    for entry in doc["spec"].get(key, []):
+                        source = entry["source"]
```
evallesp, line 88: (non-blocking) suggestion: I'd check if both keys exists before trying to access them.

plugins/modules/generate_cluster_image_policies.py
```diff
@@ -0,0 +1,139 @@
...
+    mod_args = {
+        "input_dir": {"required": True, "type": "str"},
+        "output_file": {"required": True, "type": "str"},
+    }
+
+    module = AnsibleModule(argument_spec=mod_args, supports_check_mode=False)
+
+    result = {
+        "success": False,
+        "changed": False,
```
evallesp, line 61: (blocking) suggestion: We need to change this as True somewhere. Might be good L134?

plugins/modules/generate_cluster_image_policies.py
```diff
@@ -0,0 +1,139 @@
...
+
+options:
+  input_dir:
+    description:
+    - Absolute path to cluster-resources mirror directory
+    required: True
+    type: str
+  output_file:
+    description:
+    - Absolute path to directory when ClusterImagePolicy file is created
```
evallesp, line 27: (blocking) typo: I think this should be "to file" instead "to directory"

plugins/modules/generate_cluster_image_policies.py
```diff
@@ -0,0 +1,139 @@
...
+
+    # Output to file
+    with open(output_file, "w") as f:
+        f.write(
+            "---\n" + "\n---\n".join(yaml.dump(p, sort_keys=False) for p in policies)
+        )
+
+    print(f"Generated {len(policies)} ClusterImagePolicy objects in {output_file}")
+
+    # Ensure some cluster image policies were created
```
evallesp, line 129: (blocking) suggestion: let's move this to L 121, so we don't create a file the header.

evallesp (commented):
LGTM in general!

evallesp:
(blocking) suggestion: Also I think this new python code suitable to have some tests located at: tests/unit/modules/test_generate_cluster_image_policies.py.
This would be taken automatically by molecule.

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
drosenfe: Image signature verification is not true by default downstream. This is from /etc/containers/policy.json:

`"default": [
        {
            "type": "insecureAcceptAnything"
        }
    ]`

What I've found is that the test operator that runs tempest tests doesn't support image signature verification. I left the default at false to be consistent with current downstream jobs.

hooks/playbooks/config_cluster_for_disconnected_deployment.yml
drosenfe: Thank you.

plugins/modules/generate_cluster_image_policies.py
drosenfe: Moved and capitalized in latest version.

plugins/modules/generate_cluster_image_policies.py
drosenfe: I added that openshift 4.19 or later is required to the Documentation description.

plugins/modules/generate_cluster_image_policies.py
drosenfe: Changed to use logging module in latest version.

plugins/modules/generate_cluster_image_policies.py
drosenfe: RETURN explanation has been added.

plugins/modules/generate_cluster_image_policies.py
drosenfe: Checking for existence of input_dir has been added.

plugins/modules/generate_cluster_image_policies.py
drosenfe: Check for existence of source and mirrors has been added.

plugins/modules/generate_cluster_image_policies.py
drosenfe: Setting changed to True has been added to suggested line.

plugins/modules/generate_cluster_image_policies.py
drosenfe: Change to file instead of directory has been made.

plugins/modules/generate_cluster_image_policies.py
drosenfe: Suggested move has been made.

drosenfe:
> (blocking) suggestion: Also I think this new python code suitable to have some tests located at: tests/unit/modules/test_generate_cluster_image_policies.py. This would be taken automatically by molecule.

Not sure how to make molecule tests. Can you re-review the rest while I try to figure out molecule tests?

Anyway this can be done without unit tests? This is way out of my area of expertise.

I was able to use unittest to make some trivial units tests like verifying inputs were present. Is this module a good candidate for unit tests since everything is in main and there are only two inputs - input_dir and output_file?

michburk:
Hey @drosenfe Enrique is on PTO for a little while, sorry about that!
If you look in https://github.com/openstack-k8s-operators/ci-framework/tree/main/tests/unit/modules
You'll see some example unit test files. Creating a file here will include that test suite in the `cifmw-pod-ansible-test` check job, and the unit tests will automatically get run as a part of that job. The scripts in the link above could serve as inspiration/a starting point.

~I'll take a look at the code and leave a review shortly.~ Generally lgtm, just a couple of questions and a note on logging!

plugins/modules/generate_cluster_image_policies.py
michburk: question: Beyond documenting, would it also make sense to explicitly check and fail if the openshift version is too old and `cifmw_disconnected_verify_image_signatures` is true? Or would some other failure occur that would make it obvious what the exact issue is? I'm missing a bit of context about what happens when trying to do this on an older ocp version, so I'm just asking for the sake of info, not necessarily trying to suggest any changes.

plugins/modules/generate_cluster_image_policies.py
michburk: If I'm not mistaken the logging as it is written now won't be recorded anywhere. If this message is useful and you'd like to return a message you could include a field like `result['message']` in the result dict.
Additionally I notice you add but don't use a `debug` field in the result dict:
```
    result = {
        "success": False,
        "changed": False,
        "debug": {},
    }
```
This can be removed or you could include some debugging info here.

plugins/modules/generate_cluster_image_policies.py
michburk: question: Do you foresee this key being rotated / using this module with a different key? Could it be better to parameterize this, then fetch this and pass it to the module from the calling site? If the answer to those questions is no then this is fine, just curious about usage/chance of this changing.

plugins/modules/generate_cluster_image_policies.py
drosenfe: Image signature verification was added to openshift 4.19. This was the only documentation I could find about how to perform it: https://developers.redhat.com/articles/2025/08/27/how-verify-container-signatures-disconnected-openshift#    The author checked in the script here: https://github.com/momoah/snolibvirt/blob/main/files/generate_cluster_image_policies.py    The first version of this review just used that script with modification for input_dir and output_file to match our environment. The review comments were to make it a module. I rewrote the script as a module. Its turning out to be more than expected since I wasn't the original author. Haven't tried on old openshift versions, but image signature verification would not be supported.

plugins/modules/generate_cluster_image_policies.py
drosenfe: I don't know the answer. It seems like whoever maintains Red Hat's public key would determine when/if it gets rotated.

drosenfe:
recheck

plugins/modules/generate_cluster_image_policies.py
drosenfe: Logging and the debug statement entirely removed from latest version.

drosenfe:
recheck

plugins/modules/generate_cluster_image_policies.py
michburk: Ah looking back at Enrique's comment today I think I initially misunderstood what he was saying. If I understand correctly, the issue is about `"apiVersion": "config.openshift.io/v1"`, If I understand correctly (and do correct me if I'm wrong!) using `config.openshift.io/v1` will only work on ocp verison 4.20 (or later), as the `ClusterImagePolicy` api had yet to be promoted to v1 in versions 4.19 and prior. Note that the code example you cite uses `"apiVersion": "config.openshift.io/v1alpha1"`.

Hence the suggestion to use `oc explain clusterimagepolicy`.

If I'm not mistaken though, we expect to use this on ocp 4.20 or later so this shouldn't be an issue, but maybe it's worth explicitly checking the current cluster's version and making sure we're not on an earlier version than 4.20 and printing an error message if we are?
If you don't think it's too an important a point though, I'm ok with the documentation alone.

drosenfe: @michburk I added a check to ensure cluster version is 4.20 or larger. The latest version passes unit tests when executed from a command line. However, the same unit tests fail when executed by cifmw-pod-ansible-test with error: raise TypeError('path-like args is not allowed when 'shell is true'). I specifically have shell=False. The pod tests must be overriding what my module has set. Unless you see anything I am doing wrong I'm inclined to leave out the version check.

michburk:
Looks like `cifmw-pod-ansible-test` is failing on some of your new tests.
```
=========================== short test summary info ============================
FAILED tests/unit/modules/test_generate_cluster_image_policies.py::TestGenerateClusterImagePolicies::test_no_input_files
FAILED tests/unit/modules/test_generate_cluster_image_policies.py::TestGenerateClusterImagePolicies::test_invalid_input_file
FAILED tests/unit/modules/test_generate_cluster_image_policies.py::TestGenerateClusterImagePolicies::test_valid_input_file
```

See the full log at https://gateway-cloud-softwarefactory.apps.ocp.cloud.ci.centos.org/logs//a9c/rdoproject.org/a9c7195423a4408fa882f25b16e9de7d/ansible_test.log

plugins/modules/generate_cluster_image_policies.py
```diff
@@ -0,0 +1,182 @@
...
+    match = re.search(r"(?i)version\s+([0-9]+\.[0-9]+)", clusterver_out.stdout)
+
+    # Verify that cluster version was found
+    if not match:
+        result["error"] = NO_CLUSTERVERSION
+        module.fail_json(msg=result["error"])
+
+    # Verify that cluster version is not less than 4.20
+    clusterver = match.group(1)
+    if float(clusterver) < 4.20:
```
michburk: it would be safer to split the major and minor version and compare them as ints. While it's unlikely that anyone would ever try it, clusterver 4.2 would actually pass this check. (`float(4.2)` is not less than `float(4.20)`)

maybe

```
   match = re.search(r"(?i)version\s+([0-9]+\.[0-9]+)", clusterver_out.stdout)
   if not match:
       module.fail_json(msg=NO_CLUSTERVERSION)

   try:
       cluster_version = tuple(int(part) for part in match.group(1).split("."))
   except (TypeError, ValueError):
       module.fail_json(msg=NO_CLUSTERVERSION)

   if cluster_version < (4, 20):
       module.fail_json(msg=UNSUPPORTED_CLUSTERVERSION)
```

would work?

tests/unit/modules/test_generate_cluster_image_policies.py
```diff
@@ -0,0 +1,90 @@
+# Copyright: (c) 2026, Red Hat
+
+# GNU General Public License v3.0+ (see COPYING or
+# https://www.gnu.org/licenses/gpl-3.0.txt)
+
+import json
+import os
+import tempfile
+
+from unittest.mock import patch
```
michburk: I think the ansible tests check job is failing because the module is trying to run
```
    clusterver_out = subprocess.run(
        ["oc", "get", "clusterversion"],
        capture_output=True,
        text=True,
        check=True,
        shell=False,
    )
```

Without an `oc` binary, this fails.

try mocking the subprocess.run output with some output that matches the shape you're expecting for each test

michburk:
@drosenfe very sorry I didn't catch your comment/updates earlier!

FWIW in the log I linked, I see the tests failing with
```
>               raise child_exception_type(errno_num, err_msg, err_filename)
E               FileNotFoundError: [Errno 2] No such file or directory: 'oc'

/usr/lib64/python3.9/subprocess.py:1837: FileNotFoundError
```
The test environment doesnt expect to actually have a cluster set up, so it won't have `oc` installed.

I opened a new pr with your changes plus a couple more in a second commit: https://github.com/openstack-k8s-operators/ci-framework/pull/4203

Much of the ci is still running, but I got the ansible test passing:
[cifmw-pod-ansible-test](https://gateway-cloud-softwarefactory.apps.ocp.cloud.ci.centos.org/zuul/t/rdoproject.org/build/72d1c0a379d54c32a858b65912ea938d)
(full transparency: the suggested changes in the test script are 100% ai-generated, sorry if the mocked output doesn't perfectly match the expected output shape, if the idea seems worthwhile feel free to take it or modify it as you see fit.)

PR #4144, opened 2026-08-25
Use AES Ceph cipher for CentOS-9 HCI scenario

fultonj:
Update scenarios/centos-9/hci_ceph_backends.yml so that jobs which depend on that file use the Ceph old AES cipher and not the new AES256k cipher.

### DIFF
--- a/scenarios/centos-9/hci_ceph_backends.yml
+++ b/scenarios/centos-9/hci_ceph_backends.yml
@@ -11,6 +11,13 @@ cifmw_cephadm_version: "tentacle"
 cifmw_cephadm_prepare_host: true
 # Apply the new spec version for RGW/TLS
 cifmw_rgw_ssl_backward_compatibility: false
+# If your OpenStack service does not have Ceph client version
+# 20.2.4+ or 19.2.6+, then it cannot speak the AES256k cipher
+# which guards against CVE-2025-30156. These overrides result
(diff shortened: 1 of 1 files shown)


fmount (approved):
I'm ok with this change because I think we are not going to rebuild Antelope clients to include a new ceph-common (if available) from centos storage sig. iirc Antelope goes with squid clients, and we need backports for it before switching to the new defaults, so this makes sense to me.
I was under impression that this was already covered by #4130, but I might be wrong and I might need to investigate more.
Being that said, the change of this patch makes sense and /lgtm

rebtoor:
> I'm ok with this change because I think we are not going to rebuild Antelope clients to include a new ceph-common (if available) from centos storage sig. iirc Antelope goes with squid clients, and we need backports for it before switching to the new defaults, so this makes sense to me. I was under impression that this was already covered by #4130, but I might be wrong and I might need to investigate more. Being that said, the change of this patch makes sense and /lgtm

https://github.com/rebtoor/ci-framework/blob/c07d8d4f7727303ceccc711beb282ccb51374fe2/scenarios/centos-9/hci_ceph_backends.yml#L18-L19

I changed those values as well but it's not important. I'm fine with the gradual approach. lgtm from me as well.

Valkyrie00 (approved):
/lgtm

PR #4151, opened 2026-08-31
Fix BGP setup in reproducer deployments

fultonj:
The BGP preparation hook delegated libvirt tasks to the literal hypervisor host. Generated reproducer inventories identify that host through controller-0.cifmw_hypervisor_host, typically as hypervisor-1. This caused the hook to fail with:

  Could not resolve hostname hypervisor

Use the inventory's recorded hypervisor name for both delegated tasks.

Fabric nodes also tried to download rhos-release before trusting the Red Hat internal CA, causing:

 CERTIFICATE_VERIFY_FAILED: self-signed certificate in chain

Install the internal CA before downloading rhos-release so FRR can be installed on the leaf, spine, and router nodes.

Co-authored-by: Codex (GPT-5) <codex@openai.com>

### DIFF
--- a/playbooks/bgp/prepare-bgp-spines-leaves.yaml
+++ b/playbooks/bgp/prepare-bgp-spines-leaves.yaml
@@ -17,7 +17,7 @@
     fabric_list: "{{ leafs_list + spines_list + routers_list }}"
   tasks:
     - name: Start spine and leaf VMs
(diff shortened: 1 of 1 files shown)


fultonj:
recheck

evallesp (approved):
/lgtm
/approve

PR #4154, opened 2026-09-02
[cifmw_ceph_spec] Avoid ini lookup on Python 3.12

imatza-rh:
## Summary

uni03gamma-rhel10 fails in the inner Ceph hook when reading `public_network` from `initial_ceph.conf`. Some ansible-core builds still call `ConfigParser.readfp`, which Python 3.12 removed (ansible/ansible#81656). Confirmed on build [3fbf2665](https://sf.apps.int.gpc.ocp-hub.prod.psi.redhat.com/zuul/t/components-integration/build/3fbf26650038474ba636aba475401657) (`ansible-core-2.16.16-2.el10` RPM + Python 3.12).

Upgrading ansible-core is not the fix: pip 2.16.16 works on 3.12, the RHEL RPM does not. Regex-parsing the file, or substituting `storage_network_range`, would also be wrong. uni03gamma extra-vars override the role param with ctlplane ([OSPRH-6675](https://redhat.atlassian.net/browse/OSPRH-6675)); extra-vars beat `set_fact` of the same name (ansible/ansible#19934), so a snapshot named `cifmw_ceph_spec_public_network` cannot be consumed later.

## Approach

- `cifmw_ceph_spec` exports the CIDR it actually rendered as `cifmw_ceph_spec_rendered_public_network` / `_private_network` (distinct names).
- `hooks/playbooks/ceph.yml` reads that host fact from `localhost` instead of `ansible.builtin.ini`.
- Callers that must parse INI text (`update.yml` fsid, molecule verify) use a new `cifmw.general.from_ini` filter based on `ConfigParser.read_file`.

Do not fold this into #4102. RHEL 9 jobs are unchanged: they never hit the `readfp` crash, and uni03gamma still honors [OSPRH-6675](https://redhat.atlassian.net/browse/OSPRH-6675).

## Test plan

- [x] Unit tests for `from_ini` (14 cases, including `%s` interpolation disabled and [OSPRH-6675](https://redhat.atlassian.net/browse/OSPRH-6675) ctlplane)
- [x] Extra-vars vs role-param vs default on Python 3.12 with ansible-core 2.15.5 (`readfp` crash) and 2.16.16
- [ ] Zuul: point testproject !2625 at this PR (`Depends-On`) and confirm the Ceph hook passes `Get Storage network range`

Related-Issue: #[OSPNW-1691](https://redhat.atlassian.net/browse/OSPNW-1691)

### DIFF
--- a/ci/playbooks/collections/ansible_collections/cifmw/general/plugins/filter/from_ini.py
+++ b/ci/playbooks/collections/ansible_collections/cifmw/general/plugins/filter/from_ini.py
@@ -1,0 +1,1 @@
+../../../../../../../../plugins/filter/from_ini.py
--- a/hooks/playbooks/ceph.yml
+++ b/hooks/playbooks/ceph.yml
@@ -416,10 +416,12 @@
       when: adoption_deploy_ceph_for_tripleo | default (false)
       no_log: "{{ cifmw_nolog | default(true) | bool }}"
 
-    # public network always exist because is provided by the ceph_spec role
+    # Use the snapshot exported by cifmw_ceph_spec, not storage_network_range
+    # and not the extra-var name: extra-vars beat set_fact of the same name,
+    # and uni03gamma (OSPRH-6675) overrides the role param with ctlplane.
     - name: Get Storage network range
       ansible.builtin.set_fact:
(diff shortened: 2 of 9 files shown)


evallesp (approved):
Great! THanks a lot
/lgtm
/approve

PR #4155, opened 2026-09-02
[reproducer] Fix IPv6 NDP stale cache causing AnsibleEE SSH timeouts

renjingxiao:
macvlan CNI assigns a new random MAC per AnsibleEE pod while Whereabouts IPAM reuses the same IPv6 address from a small pool. When pods cycle rapidly during Job backoff, EDPM nodes' NDP cache still maps the reused address to the previous pod's MAC. Return traffic is sent to the stale MAC and silently dropped on the linux bridge, causing SSH connection timeouts.

Reduce base_reachable_time_ms from 30000 to 5000 on the actual ctlplane interface of each EDPM node, so NDP entries go STALE within 5s and the kernel re-probes on the next packet. TCP retransmits within 1s, by which time the cache is updated.

Also set the default for any interfaces created later.

Related-to: [OSPCIX-1486](https://redhat.atlassian.net/browse/OSPCIX-1486)
Assisted-by: Claude Code Opus 4.6

### DIFF
--- a/roles/reproducer/tasks/libvirt_layout.yml
+++ b/roles/reproducer/tasks/libvirt_layout.yml
@@ -122,3 +122,46 @@
   loop_control:
     loop_var: compute
     label: "{{ compute.key }}"
+
+# macvlan CNI assigns a random MAC per AnsibleEE pod, but Whereabouts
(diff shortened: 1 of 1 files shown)


evallesp (approved):
/approve

nemarjan (approved):
/approve

PR #4157, opened 2026-09-03
[test_operator] Remove HorizonTest credentials WA

kstrenkova:
There is a HorizonTest WA for credentials due to a change from static to dynamic passwords. This PR aims to remove this WA since the actual fix has reached downstream and we do not have to use WA anymore.


kstrenkova:
I have created this PR to test whether we can actually remove the WA. Once I get a passing test job, I will mark it here. Until then this is a work in progress.

Edit: After running the job, it seems the test-operator change is already downstream, but the tcib change has yet to propagate. I am keeping this ticket as a draft and will try to run it again next week.

kstrenkova:
I am still waiting for the new image in tcib.

PR #4159, opened 2026-09-03
[docs] Clarify public vs internal variable naming for Qodo

nemarjan:
Qodo was flagging underscore-prefixed register and set_fact variables as cifmw_ naming violations. The cifmw_ prefix applies only to the role's public interface; internal runtime variables use a leading underscore by convention.


nemarjan:
recheck

michburk (approved):
/lgtm

evallesp (approved):
/approve

PR #4162, opened 2026-09-03
[edpm_prepare] Skip manila/cinder override if explicitly set

karelyatin:
If cinder volume and manila share images are explicitly set as part of cifmw_set_containers_images, we shouldn't override those.

Follow up of https://github.com/openstack-k8s-operators/ci-framework/pull/4100


fmount (approved):
/lgtm

tosky:
I can't vote, but big +1 - in the long run, we probably don't need the two new variables at all.

danpawlik (approved):

michburk (approved):
/approve

PR #4165, opened 2026-09-07
[test_operator] Add test_router_rescheduling to skip list

karelyatin:
This test is only relevant for ml2/ovs setups while we only support ml2/ovn. This was not visible until we started switching to neutron master S2I images and that requires [1] to be available in tempest containers. But since we have jobs not fully switched to S2I containers yet adding this explicitly to skip list.
This can be cleaned from the skip list once we fully switch to S2I containers including the tempest ones in master branch.

[1] https://review.opendev.org/c/openstack/tempest/+/1003527

Related-Issue: #[OSPRH-36303](https://redhat.atlassian.net/browse/OSPRH-36303)


karelyatin:
Validate in https://gateway-cloud-softwarefactory.apps.ocp.cloud.ci.centos.org/zuul/t/rdoproject.org/buildset/a9a469674375403b8aa372b3d6f5bd3e with master content set

evallesp (approved):
/lgtm

nemarjan (approved):
/approve

PR #4167, opened 2026-09-08
[s2i_content_set] Resolve promoted S2I images for OpenStackVersion

rebtoor:
## Summary
- Add a `s2i_content_set` role that resolves a hash-stable S2I content set (`s2iCommit` + image digests) from a pin file, a post-merge catalog, or Quay labels, then emits `cifmw_set_containers_images` for `edpm_prepare`.
- Wire it into `deploy-edpm.yml` behind `cifmw_s2i_content_set` so operator jobs can consume `quay.io/openstack-s2i-containers` without racing `:master-latest`.
- Preserve unlisted `OpenStackVersion` keys so unmapped services stay on payload defaults.

Related-Issue: [ANVIL-275](https://redhat.atlassian.net/browse/ANVIL-275)

## Test plan
- [x] Molecule `s2i_content_set` (catalog + inspect plays)
- [x] `cifmw-molecule-s2i_content_set` in Zuul
- [x] Exercise with the openstack-operator `openstack-operator-tempest-multinode` change (Depends-On)


roles/s2i_content_set/handlers/main.yml
```diff
@@ -0,0 +1,15 @@
+---
```
danpawlik: is it needed as there is no handlers?

roles/s2i_content_set/defaults/main.yml
```diff
@@ -0,0 +1,102 @@
...
+# License for the specific language governing permissions and limitations
+# under the License.
+
+
+# All variables intended for modification should be placed in this file.
+# All variables within this role should have a prefix of "cifmw_s2i_content_set"
+
+# Explicit s2i-openstack-containers git SHA. Wins over the pin file and
+# over inspecting :master-latest.
+cifmw_s2i_content_set_commit: ""
```
danpawlik, line 23: is it needed as it is required? Then in https://github.com/openstack-k8s-operators/ci-framework/pull/4167/changes#diff-b6a992b74941f91b646e1077775f0b481a7ecf09c71c0e3ad5d27250b9781f0dR21 you also set default which is here with same value

roles/s2i_content_set/tasks/emit.yml
```diff
@@ -0,0 +1,58 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+- name: Drop skipped OpenStackVersion fields
+  ansible.builtin.set_fact:
+    cifmw_s2i_content_set_resolved_images: >-
+      {{
+        cifmw_s2i_content_set_resolved_images |
```
danpawlik: that might be confusing as you are overwriting the var. If someone would debug that, this part might be confusing. Suggest to use new var.

roles/s2i_content_set/tasks/inspect_set.yml
```diff
@@ -0,0 +1,137 @@
...
+
+- name: Require image mappings for live inspect
+  when: not cifmw_s2i_content_set_mappings_stat.stat.exists
+  ansible.builtin.fail:
+    msg: >-
+      S2I image mappings not found at {{ cifmw_s2i_content_set_mappings_file }}.
+      Add openstack-k8s-operators/s2i-openstack-containers to required-projects.
+
+- name: Load image mappings
+  ansible.builtin.include_vars:
```
danpawlik, line 34: nit: suggest to use helpers: https://github.com/openstack-k8s-operators/ci-framework/blob/main/roles/cifmw_helpers/tasks/include_file.yml as include_vars are taking vars on the host where ansible was executed

danpawlik (dismissed):

roles/s2i_content_set/molecule/default/files/skopeo
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
```
evallesp, line 1: (non-blocking) suggestion: Super small nit, do you think we can add ".sh" suffix to this file?

roles/s2i_content_set/tasks/write_artifact.yml
```diff
@@ -0,0 +1,50 @@
...
+- name: Check for Zuul log directory
+  ansible.builtin.stat:
+    path: "{{ ansible_user_dir }}/zuul-output/logs"
+  register: cifmw_s2i_content_set_zuul_log_dir
+
+- name: Copy artifact into Zuul logs
+  when: cifmw_s2i_content_set_zuul_log_dir.stat.exists
+  ansible.builtin.copy:
+    src: "{{ cifmw_s2i_content_set_dest_path }}"
+    dest: "{{ ansible_user_dir }}/zuul-output/logs/s2i-content-set.yaml"
```
evallesp, line 48: (non-blocking) suggestion: Might be better to store in an artifacts folder?

roles/s2i_content_set/defaults/main.yml
```diff
@@ -0,0 +1,102 @@
...
+# OpenStackVersion field names to omit so preserve_unlisted keeps payload
+# defaults. Empty here; jobs that must match s2i-openstack-deploy-validation
+# pass the unready keys.
+cifmw_s2i_content_set_skip_images: []
+
+# Continue when an individual image inspect fails (partial set).
+cifmw_s2i_content_set_ignore_missing: false
+
+# Backend maps are not expressible in image-mappings.yaml scalars.
+cifmw_s2i_content_set_backend_images:
```
evallesp, line 94: (non-blocking) question: Should we check if the deployment uses cinder and/or manila before populating this? This might break with default value of cifmw_s2i_content_set_ignore_missing

roles/s2i_content_set/molecule/default/files/content-set.yaml
```diff
@@ -0,0 +1,19 @@
...
+  keystone/keystone: quay.io/openstack-s2i-containers/openstack-keystone@sha256:1111111111111111111111111111111111111111111111111111111111111111
+  glance/glance-api: quay.io/openstack-s2i-containers/openstack-glance-api@sha256:2222222222222222222222222222222222222222222222222222222222222222
+  neutron/neutron-server: quay.io/openstack-s2i-containers/openstack-neutron-server@sha256:3333333333333333333333333333333333333333333333333333333333333333
+openstackVersion:
+  customContainerImages:
+    keystoneAPIImage: quay.io/openstack-s2i-containers/openstack-keystone@sha256:1111111111111111111111111111111111111111111111111111111111111111
+    glanceAPIImage: quay.io/openstack-s2i-containers/openstack-glance-api@sha256:2222222222222222222222222222222222222222222222222222222222222222
+    neutronAPIImage: quay.io/openstack-s2i-containers/openstack-neutron-server@sha256:3333333333333333333333333333333333333333333333333333333333333333
+    cinderVolumeImages:
+      volume1: quay.io/openstack-s2i-containers/openstack-cinder-volume@sha256:4444444444444444444444444444444444444444444444444444444444444444
```
evallesp, line 19: (non-blocking) question: Should we have here volume2 with another digest? If so, https://github.com/openstack-k8s-operators/ci-framework/pull/4167/changes#diff-71aeb849814f576a32c1960dbb9e047183fd21e15cd8df3e55433dd1ee301712 

We might end up by having the same digest for all of the backends items in the array.

roles/s2i_content_set/defaults/main.yml
```diff
@@ -0,0 +1,102 @@
...
+# All variables within this role should have a prefix of "cifmw_s2i_content_set"
+
+# Explicit s2i-openstack-containers git SHA. Wins over the pin file and
+# over inspecting :master-latest.
+cifmw_s2i_content_set_commit: ""
+
+cifmw_s2i_content_set_registry: quay.io
+cifmw_s2i_content_set_namespace: openstack-s2i-containers
+cifmw_s2i_content_set_stream: master
+cifmw_s2i_content_set_prefix: openstack
```
evallesp, line 28: (non-blocking) question: I think this and cifmw_s2i_content_set_src should be docummeted?

roles/s2i_content_set/tasks/inspect_set.yml
```diff
@@ -0,0 +1,137 @@
...
+      image: "{{ item.image }}"
+      backends: "{{ item.backends }}"
+  loop: "{{ cifmw_s2i_content_set_backend_images }}"
+  loop_control:
+    label: "{{ item.name }}"
+
+- name: Inspect probe image at master-latest for s2iCommit
+  when: cifmw_s2i_content_set_resolved_commit | length == 0
+  vars:
+    cifmw_s2i_content_set_inspect_name: "{{ cifmw_s2i_content_set_probe_image }}"
```
evallesp, line 68: (non-blocking) question: I think openstack-keystone might get done two times. So two skopeo calls against same image might happen?

roles/s2i_content_set/handlers/main.yml
rebtoor: Dropped. It was the role skeleton and nothing notifies it.

roles/s2i_content_set/defaults/main.yml
rebtoor: It is optional, not required. Empty means “discover the commit” (pin, then catalog, then probe labels). The `| default('')` in `tasks/main.yml` is only a guard if the var is unset; same empty default as `defaults/main.yml`.

roles/s2i_content_set/tasks/emit.yml
rebtoor: Agreed. Skip + overlay now land in `cifmw_s2i_content_set_merged` so we do not clobber `cifmw_s2i_content_set_resolved_images`.

roles/s2i_content_set/tasks/inspect_set.yml
rebtoor: `include_file.yml` is for remote files that contain Jinja, and it does not support `name:`, so it would dump `openstack_version` into the play vars. `image-mappings.yaml` is static YAML and we load it namespaced as `cifmw_s2i_content_set_mappings`. Same pattern as the pin/catalog loads. This role runs on the job node where the s2i checkout already lives, so `include_vars` reads the right path.

roles/s2i_content_set/molecule/default/files/skopeo
rebtoor: The file has to be named `skopeo` so it shadows the real binary on `PATH` in the inspect scenario. A `.sh` suffix would not be found.

roles/s2i_content_set/tasks/write_artifact.yml
rebtoor: The role already writes `{{ cifmw_basedir }}/artifacts/s2i-content-set.yaml`. The `zuul-output/logs/` copy is only so it shows up at the log root; `collect-logs` already picks up `ci-framework-data/artifacts/`.

roles/s2i_content_set/defaults/main.yml
rebtoor: We always resolve those `OpenStackVersion` map fields; we do not look at whether Cinder/Manila are enabled on the CR. Unused image fields on the version CR are fine. If the image is missing from the promoted set, the default `ignore_missing: false` should fail. Jobs that want payload defaults for those keys can put them in `cifmw_s2i_content_set_skip_images`.

roles/s2i_content_set/molecule/default/files/content-set.yaml
rebtoor: `volume1` / `volume2` are backend names for the same `openstack-cinder-volume` image, so they share one digest. A second digest in the fixture would be fake. The catalog path also takes the first digest for all keys in that map.

roles/s2i_content_set/defaults/main.yml
rebtoor: Documented `cifmw_s2i_content_set_src` and `_prefix`. Registry / namespace / stream were already in the README; they are listed together with prefix now.

roles/s2i_content_set/tasks/inspect_set.yml
rebtoor: Same image name, two refs. The probe is `openstack-keystone:master-latest` to read `org.opencontainers.image.revision`. The loop then inspects `openstack-keystone:master-<sha>`. If commit is already set (pin / explicit SHA), the probe is skipped and keystone is inspected once.

evallesp (approved):
/lgtm

rebtoor:
recheck

michburk (approved):
/approve

rebtoor:
Recheck

brjackma:
recheck

brjackma:
recheck

rebtoor:
@brjackma care to read a lgtm?

We can override the zuul check...

brjackma:
recheck

Valkyrie00 (approved):
/lgtm

PR #4169, opened 2026-09-09
[reproducer] Revert "Fix IPv6 NDP stale cache causing AnsibleEE SSH timeouts"

renjingxiao:
This reverts commit a92ed0a1859fd17f5f5bbf9651a534ae90c69ff0.

The retries patch addresses the same symptom in the nodeset itself and should be sufficient on its own. Reverting the CI workaround to avoid hiding the issue.


brjackma (approved):

michburk (approved):
/lgtm

PR #4171, opened 2026-09-10
[operator_build] Make tasks fails on command error

karelyatin:
These task succeeds always as the last command will never fail. Make the task fail on any command error by using set -eo pipefail.


abays (dismissed):
/lgtm
/approve

michburk (approved):
/approve

evallesp (approved):
/lgtm

PR #4172, opened 2026-09-10
edpm_baremetal_update: use the split update procedure (services only)

jistr:
The monolithic update procedure is being removed. Switch the edpm_baremetal_update role to use the new split update procedure. As it looks like only the services update is relevant in the edpm_baremetal_update, we don't include the operating system update.

Assisted-by: goose+gpt
Resolves: https://redhat.atlassian.net/browse/OSPRH-36564

### DIFF
--- a/ci/playbooks/edpm_baremetal_update/update-edpm.yml
+++ b/ci/playbooks/edpm_baremetal_update/update-edpm.yml
@@ -168,22 +168,22 @@
         (        cifmw_update_containers_watcher is defined and
         cifmw_update_containers_watcher | bool)
 
-    - name: Set update step to Starting the CI update sequence
+    - name: Set update step to Starting the services update sequence
       ansible.builtin.command:
         cmd: >
           {{ cifmw_update_artifacts_basedir }}/update_event.sh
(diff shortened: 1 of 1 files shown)


jistr:
recheck

jistr:
This PR is to unblock https://github.com/openstack-k8s-operators/openstack-operator/pull/2060

jistr:
recheck

jistr:
recheck

jistr:
All jobs passed now but not in the same run

cjeanner (approved):

ciecierski:
looks good
As a next step we need to think where we want to plug system update. This can be a follow-up change if we want to keep the change simple.

ciecierski (approved):

jistr:
Thanks, @ciecierski re system upgrade, based on how i understood the baremetal update job, it actually doesn't reboot so the system update isn't utilized fully. I could do a follow up change and include the system update to just test the pre-reboot part though, i guess that wouldn't hurt.

brjackma (approved):

PR #4174, opened 2026-09-10
[kustomize_deploy] Randomize libvirt-secret password at deploy time

abays:
The architecture repo's lib/dataplane/nodeset component generates a libvirt-secret Secret from a hardcoded libvirt-secret.env file (LibvirtPassword=12345678). Extend the post-kustomize manifest injection pattern introduced for osp-secret so this well-known default password is also randomized before `oc apply`.

After each `kustomize build`, if the rendered dataplane manifest contains a libvirt-secret Secret, its data keys are replaced with a cryptographically random value via the shared `osp_secret_manifest.py` helper, unless a live cluster secret already exists (in which case its value is preserved across redeploys).

To support this, generalize `osp_secret_manifest.py` so its helpers and CLI commands (has, get, get-namespace, set, randomize) accept an optional secret name, defaulting to osp-secret for backward compatibility. Add a new inject_libvirt_secret_key.yml task file mirroring the osp-secret randomization block, wired into execute_step.yml so it runs for every stage but is a no-op except for the dataplane stage.


abays:
recheck

michburk (approved):
/lgtm

lmiccini (approved):

bshewale (approved):

PR #4177, opened 2026-09-14
[bgp] Default controller-0.cifmw_hypervisor_host to "hypervisor"

eduolivares:
controller-0.cifmw_hypervisor_host is successfully resolved by BGP jobs except with adoption jobs, when the playbook is executed during OSP deployment and from a node different than controller-0. For those cases, this patch sets the hypervisor_host variable default to "hypervisor".


eduolivares:
recheck

brjackma (approved):

Valkyrie00 (approved):
/approve

PR #4180, opened 2026-09-14
Mask OpenAI Secret

jancervenka:
Modify crawl_n_mask.py script to allow masking of OpenAI secret token strings sk-proj-XXX..., sk-svcacct-XXX..., sk-admin-XXX...

Ticket: https://redhat.atlassian.net/browse/OSPRH-35143


jancervenka:
recheck

lpiwowar (dismissed):
LGTM:), thank you! I guess we want to get a pair of eyes on this PR that belong to somebody from the ci-framework team.

Valkyrie00 (dismissed):
/lgtm

plugins/modules/crawl_n_mask.py
```diff
@@ -295,6 +295,9 @@
     "stack_domain_admin_password",
     "staticPasswords",
     "X-Auth-Token",
+    "sk-proj",
+    "sk-admin",
+    "sk-svcacct",
```
brjackma: (Suggestion) I'm a little concerned about adding these three to 'PROTECT_KEYS'. Since the PROTECT_KEYS has held only key names up to now. In this instance, we're adding prefixes to this list. It might make sense to separate those two things. 

We could add something like: 

`TOKEN_PREFIXES = frozenset(["sk-proj", "sk-admin", "sk-svcacct"])`

and then update the mask_log_line() function to use that when checking what to mask. 

WDYT?

jancervenka: Thank you for taking a look! @brjackma I separated the prefixes into the `TOKEN_PREFIXES` set.

brjackma (approved):

Valkyrie00 (approved):

PR #4150, opened 2026-08-29
Retry transient oc exec errors during shiftstack pod polling

tusharjadhav3302:
## Summary
- Treat known transient `oc exec` stderr patterns as retryable while polling long-running shiftstack tests in `shiftstackclient`
- Prevents false post-deployment failures when the API/admission webhook returns transient errors (for example `SCCExecRestrictions` timeouts) while the background command is still running
- Improves the unreachable-pod failure message with the polling attempt limit

## Context
`shiftstack-rhel9-rhoso18.0-ocp4.22` on serval70 failed in post-deployment after ~32 minutes even though:
- RHOSO deploy and `openstack-edpm` were healthy
- The nested OCP 4.22 install inside `shiftstackclient` continued and completed successfully afterward

Failure signature:
```
admission plugin "security.openshift.io/SCCExecRestrictions" failed to complete validation in 13s
```

Related Jira: https://redhat.atlassian.net/browse/OSPRH-36061

## Test plan
- [ ] Re-run `shiftstack-rhel9-rhoso18.0-ocp4.22` on serval70 and confirm post-deployment survives transient `oc exec` polling errors during `osp_verification_4.22_nightly.yaml`
- [ ] Confirm non-transient failures (for example pod `NotFound`) still fail promptly

Made with [Cursor](https://cursor.com)

PR #4192, opened 2026-09-18
[shiftstack] Bound soft-transient oc exec retries during pod polling

tusharjadhav3302:
## Summary
- Add a short consecutive-failure budget for soft-transient `oc exec` errors (`connection refused`, `i/o timeout`) while polling long-running shiftstack tests in `shiftstackclient`.
- Keeps the PR #4150 behavior for hard-transient errors (SCC / InternalError / stream reset): those still retry for the full poll window.
- Avoids failing on a brief management-API VIP blip (observed on serval70 ocp4.23 when `cifmw-dnsmasq` restarted mid-test) without spinning for ~8h when the API is truly unreachable.

## Why
On `shiftstack-rhel9-rhoso18.0-ocp4.23` (serval70, 2026-09-17) the outer poller failed immediately on:

```
The connection to the server api.ocp.openstack.lab:6443 was refused
```

The background `osp_verification_4.23_nightly` run continued and completed afterward. Re-adding `connection refused` to the unlimited hard-transient list was deliberately rejected in #4150 review; this change uses a bounded soft budget instead (default 12 × 5s ≈ 60s).

## Test plan
- [ ] ci-framework checks / molecule shiftstack
- [ ] Confirm a single `connection refused` during polling no longer fails the testconfig
- [ ] Confirm 12 consecutive soft-transient failures still fail promptly with the new message
- [ ] Confirm SCCExecRestrictions / Internal error still retry for the full window


brjackma:
recheck

brjackma (dismissed):

roles/shiftstack/tasks/exec_command_in_pod.yml
```diff
@@ -54,34 +54,130 @@
...
-      ansible.builtin.command:
-        cmd: >-
-          oc exec -n {{ namespace }} {{ pod_name }} --
-          bash -c 'if cat /tmp/cifmw_cmd_rc 2>/dev/null; then exit 0;
-          elif [ -f /tmp/cifmw_cmd_pid ] && kill -0 $(cat /tmp/cifmw_cmd_pid) 2>/dev/null; then exit 10;
-          else exit 2; fi'
+      vars:
+        _hard_transient_re: "{{ cifmw_shiftstack_exec_transient_patterns | join('|') }}"
+        _soft_transient_re: "{{ cifmw_shiftstack_exec_soft_transient_patterns | join('|') }}"
+      ansible.builtin.shell: |
```
michburk: Not blocking: Logic itself lgtm but I don't love having a big block of shell script in the middle of this ansible file, do you think it would make sense to pull this script out into a dedicated file? It could live somewhere like `roles/shiftstack/files/shiftstack_cmd_poll.sh`

tusharjadhav3302: Good call — extracted the poller to `roles/shiftstack/files/shiftstack_cmd_poll.sh` and invoke it from the wait task via `role_path`. Soft/hard transient handling and exit codes are unchanged; the Ansible task just sets the poll env vars and normalizes the attempts metadata from stderr.

brjackma:
recheck

roles/shiftstack/tasks/exec_command_in_pod.yml
```diff
@@ -54,34 +54,69 @@
...
+        CIFMW_SHIFTSTACK_POLL_SOFT_MAX: "{{ cifmw_shiftstack_exec_soft_transient_max }}"
+        CIFMW_SHIFTSTACK_POLL_HARD_RE: "{{ cifmw_shiftstack_exec_transient_patterns | join('|') }}"
+        CIFMW_SHIFTSTACK_POLL_SOFT_RE: "{{ cifmw_shiftstack_exec_soft_transient_patterns | join('|') }}"
       ansible.builtin.command:
-        cmd: >-
-          oc exec -n {{ namespace }} {{ pod_name }} --
-          bash -c 'if cat /tmp/cifmw_cmd_rc 2>/dev/null; then exit 0;
-          elif [ -f /tmp/cifmw_cmd_pid ] && kill -0 $(cat /tmp/cifmw_cmd_pid) 2>/dev/null; then exit 10;
-          else exit 2; fi'
+        cmd: "{{ role_path }}/files/shiftstack_cmd_poll.sh"
```
michburk, line 79: Just one slight concern over using `ansible.builtin.command` and `role_path` here.
Have you tested this and does this work as expected? Situations like this can be a little tricky, as `role_path` refers to the role path on the controlling host, but the command executes on the remote host, so the file might not be exactly where you expect.
Maybe `ansible.builtin.script` would be safer?

michburk (commented):
One concern, otherwise looks much cleaner. Thanks!

PR #4193, opened 2026-09-18
[bgp] Configure router-0 EVPN VTEP dynamically; keep only static SNAT

eduolivares:
The neutron-tempest-plugin EVPN VNI provisioner now configures router-0
as a BGP-EVPN VTEP dynamically during the tests: it creates the per-VNI
VRF/bridge/VXLAN netdevs (`evpnvrf-<vni>` / `br-<vni>` / `vxlan-<vni>`)
over SSH and applies the matching FRR config (per-VRF BGP instance plus
`import vrf` into the default instance, advertising the datapath `/32`
via `redistribute connected`). This makes the static VTEP setup added in
cc53bf5 redundant.

- **Revert** cc53bf5 ("[bgp] Optionally configure router-0 as a BGP-EVPN
  VTEP"), removing the static netdev creation and FRR EVPN blocks.
- **Re-add only the piece that must stay static**: the SNAT rule for
  tenant return traffic. The tempest client runs on a test pod that
  reaches the VMs via router-0, and its source IP lives on the fabric
  underlay, outside the EVPN advertised space, so the VM has no route
  back to it. SNATing the source to the VTEP IP (`router_loopback_ip`,
  which router-0 advertises into EVPN) makes the VM reply to an address
  that returns through the VXLAN tunnel, where conntrack un-NATs it. The
  rule now matches every `br-<vni>` bridge (`br-+`) instead of a single
  L3VNI, covering all VNIs the provisioner allocates without depending on
  the tenant CIDR or VNI.

The companion `ci-framework-jobs` change drops the now-unused scenario
variables (`router_evpn_l3vni`, `router_evpn_edpm_asn`,
`router_tenant_cidr`) and adds the tempest EVPN options.

Related-ticket: [OSPRH-34785](https://redhat.atlassian.net/browse/OSPRH-34785)


eduolivares:
recheck

michburk (approved):
/approve

Valkyrie00 (approved):
/lgtm

mtomaska:
LGTM

